PR #1, opened 2018-04-12
Kubeutil minor improvements

dkistner:
- Change Garden api group version
- Extend error handling in shoot watcher function
- Rename functions to according to Gardener component naming


dkistner:
@ccwienk Could you please review?

PR #7, opened 2018-04-16
Introduce base_definition

RaphaelVogel:
Group all non-variant definitions under base_definition attribute.
Doing this, users can more easily see inheritance of definitions


concourse/pipelines/factory.py
```diff
@@ -20,6 +20,14 @@ def __init__(self, raw_dict):
 
         if not 'variants' in raw_dict:
             raise ModelValidationError('at least one variant must be specified')
+        if not 'base_definition' in raw_dict:
+            raise ModelValidationError('at least one base_definition must be specified')
```
ccwienk: I do not think we should enfore a base definition

PR #9, opened 2018-04-18
Create image-pull-secret and tls-secret from config

AndreasBurger:
Depends on changes to config

### DIFF
--- a/concourse/setup.py
+++ b/concourse/setup.py
@@ -54,31 +54,46 @@ from kubernetes.client import (
     V1beta1IngressBackend, V1beta1IngressTLS, V1EnvVar,
 )
 
-
-IMAGE_PULL_SECRET_NAME = "ci-gcr-readonly"
-IMAGE_PULL_SECRET_FILE = "ci-gcr-readonly.yml"
-
-
 @ensure_annotations
-def create_image_pull_secret(namespace: str, config_path: str):
+def create_image_pull_secret(
+    config_factory: ConfigFactory,
+    config_name: str,
+    namespace: str,
+):
     """Create an image pull secret in the K8s cluster to allow pods to download images from gcr"""
-    core_api = kubeutil.Ctx().create_core_api()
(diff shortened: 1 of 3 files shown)


concourse/setup.py
```diff
@@ -98,30 +98,33 @@ def create_image_pull_secret(
...
-def create_tls_secret(namespace: str, config_path: str):
-    """Creates the TLS secret for concourse web component in the K8s cluster"""
-    # XXX: move those to cc-config
-    tls_key_file = "concourse-tls.key"
-    tls_cert_file = "concourse-tls.crt"
-    tls_secret_name = "concourse-web-tls"
-
-    secret_helper = KubernetesSecretHelper(kubeutil.Ctx().create_core_api())
+def create_tls_secret(
+    config_set: ConfigurationSet,
```
ccwienk: accept TlsSecret as input; while at it: introduce a dedicated model element type TlsConfig

concourse/setup.py
```diff
@@ -98,30 +98,33 @@ def create_image_pull_secret(
...
+    namespace: str,
+):
+    """Creates the configured TLS secret for the Concourse web-component in the K8s cluster"""
+    ensure_not_none(config_set)
+    ensure_not_empty(namespace)
+
+    # secret-name is predetermined by the Helm-chart
+    tls_secret_name = 'concourse-web-tls'
```
ccwienk: at least make this an argument of this function; even better: mv it to concourse-cfg (in cc-config). pass-in from there. if possible with reasonable efforts, also pass to helm-deploy; if this is not (easily) possible, document this in cc-config

concourse/setup.py
```diff
@@ -98,30 +98,33 @@ def create_image_pull_secret(
...
+
+    tls_config = config_set._cfg_element(cfg_type_name='tls_config')
```
ccwienk: rm

ccwienk: also: add tls_config name to concourse_cfg; rm from cfg_set

PR #10, opened 2018-04-18
kubeutil - remove annoying logs when catching http or key error while…

dkistner:
… waiting for shoot operation operation to complete


dkistner:
@RaphaelVogel, @ccwienk, @AndreasBurger Could you please have a look and merge?

This should reduce the amount of annoying and unnecessary logs during the execution of integration test

PR #13, opened 2018-04-25
Use kubeconfig passed in Configuration

AndreasBurger:


concourse/setup.py
```diff
@@ -166,16 +166,15 @@ def create_instance_specific_helm_values(concourse_cfg: ConcourseConfig):
 
 
 def deploy_concourse_landscape(
-        config_dir: str,
+        config_factory: ConfigFactory,
```
ccwienk: firstly: would it not suffice to pass-in the required cfg_set (in that case, the cfg_name would not be required)

concourse/setup.py
```diff
@@ -100,19 +103,27 @@ def create_image_pull_secret(
 
 @ensure_annotations
 def create_tls_secret(
-    tls_config: TlsConfig,
-    tls_secret_name: str,
+    config_factory: ConfigFactory,
```
ccwienk: please revert to passing tls_config + tls_secret_name; extend e.g. with initialised ctx;

kubeutil.py
```diff
@@ -69,6 +62,17 @@ def get_kubecfg(self):
             fail('KUBECONFIG env var must be set')
         return config.load_kube_config(ensure_file_exists(kubeconfig))
 
+    def _set_kubecfg(self, kubeconfig_dict: dict):
```
ccwienk: why make the setter private?

concourse/setup.py
```diff
@@ -209,7 +209,10 @@ def deploy_concourse_landscape(
         cfg_name = concourse_cfg.helm_chart_values()
     ).raw
 
+    # Set the global context to the cluster specified by the given config
     kubernetes_config = config_set.kubernetes()
+    kubeconfig = kubernetes_config.kubeconfig()
+    kubeutil.ctx._set_kubecfg(kubeconfig)
```
ccwienk: could be inlined

PR #14, opened 2018-04-30
Check cluster k8s-version before deploying concourse

AndreasBurger:


kubeutil.py
```diff
@@ -111,6 +111,10 @@ def create_extensions_v1beta1_api(self):
         cfg = self.get_kubecfg()
         return client.ExtensionsV1beta1Api(cfg)
 
+    def get_version_info(self):
```
ccwienk: rename: (get_?)cluster_version_info 
mv out of Ctx class. either introduce a new helper class, or make this a module function

concourse/setup.py
```diff
@@ -182,6 +182,27 @@ def deploy_concourse_landscape(
...
+
+    # Check whether the clusters k8s-version is identical to the configured one
+    cluster_version_info = kubeutil.ctx.get_version_info()
+    configured_version_info = kubernetes_config.cluster_version()
+
+    if (
+        cluster_version_info.major != configured_version_info['major'] or
+        cluster_version_info.minor != configured_version_info['minor']
+    ):
+        raise ValueError(
```
ccwienk: This is not the appropriate exception type

PR #18, opened 2018-05-16
Create functionality for automatic submodule-upgrade-proposals

AndreasBurger:
Creates the necessary functions and utilities to propose automated upgrades for submodules in a given repository based on the commit to which the submodule points at a tagged commit in another repository.

### DIFF
--- a/gitutil.py
+++ b/gitutil.py
@@ -1,0 +1,114 @@
+import os
+import subprocess
+
+import git
+
+from util import ensure_not_empty, ensure_directory_exists
+
+def update_submodule(
+    repo_path: str,
+    tree_ish: str,
+    submodule_path: str,
+    commit_hash: str,
+    author: str,
+    email: str,
+):
+    '''Update the submodule of a git-repository to a specific commit.
+
+    Create a new commit, with the passed tree-ish as parent, in the given repository.
+
+    Parameters
+    ------
+    repo_path : str
+        Path to a directory containing an intialised git-repo with a submodule to update.
+    tree_ish : str
+        Valid tree-ish to use as base for creating the new commit. This will be used as parent for the commit created by this function.
+        Example: 'master' for the head of the master-branch.
+    submodule_path : str
+        Path (relative to `repo_path`) to the submodule. Must be immediately below the root (i.e. `repo_path`) of the repository.
+    commit_hash : str
+        The hash the submodule should point to in the created commit. This should be a valid commit hash in the submodule's repository.
+    author : str,
+        Will be set as author of the created commit
+    email : str
+        Will be set for the author of the created commit
+
+    Returns
+    ------
+    str
+        The hexadecimal SHA-1 hash of the created commit
+    '''
+    ensure_directory_exists(repo_path)
+    repo_path = os.path.abspath(repo_path)
+
+    ensure_not_empty(tree_ish)
+    ensure_not_empty(submodule_path)
+    ensure_directory_exists(os.path.join(repo_path, submodule_path))
+    ensure_not_empty(commit_hash)
+    ensure_not_empty(author)
+    ensure_not_empty(email)
+
+    repo = git.Repo(repo_path)
+
+    # Fetch tree-ish from git.repo and create mktree-parseable string-representation of it.
+    # Since submodule-objects contain the SHA of the to-be-checked-out version of the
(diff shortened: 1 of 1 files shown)


lsutils/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+    # get information about submodule in the current repo
+    git_repo = git.Repo(repo_path)
+    submodule = find_submodule_by_path(git_repo, submodule_path)
+    current_submodule_sha = submodule.hexsha
+
+    p_username, p_repository = extract_owner_and_name(predecessor_repo_owner_and_name)
+    username, repository = extract_owner_and_name(repo_owner_and_name)
+
+    # get the relevant information about the submodule (i.e. the commit it's pointing to) from the predecessor
+    github = github3.github.GitHubEnterprise(
```
ccwienk: pass-in github_api object. also, there is already a ready-to-use function that creates such an object in githubutil.py.

lsutils/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+
+    info("Creating pull-request for upgrade.")
+    repo_helper.create_pull_request(
+        title=pr_title,
+        base=repo_branch,
+        head=rnd_string,
+    )
+
+
+def extract_owner_and_name(owner_and_name: str):
```
ccwienk: this is called parsing. also, this is definitely not landscape-specific and should go into the githubutil module (I think there already is a similar functionality there)

lsutils/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+        username, repository = owner_and_name.split('/')
+    except ValueError:
+      fail(
+          'Could not extract owner- and repository-name from String. Found: {actual}'.format(
+              actual=owner_and_name,
+          )
+      )
+    return username, repository
+
+def find_submodule_by_path(repo: git.Repo, path: str):
```
ccwienk: this is not ls-specific and should go into a git-util module

lsutils/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+    for submodule in repo.submodules:
+        if submodule.path == path:
+            return submodule
+    raise ValueError("No submodule found at path '{p}'".format(p=path))
+
+def update_submodule(predecessor_tree, sha, path):
```
ccwienk: this is not ls-specific and should to into a git-util module (also, consider introducing a helper class rather than defining top-level functions)

lsutils/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+    cache = predecessor_tree.cache
+    cache.add(sha, 0o160000, path, force=True)
```
ccwienk: please check if the octal literal 0o160000 is not already predefined as a constant in module 'git'. if so, please use the constant. if not, consider using one from 'os' module or define it yourself, using an expressive name and in particular state what this value means for git!

lsutils/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+    cache.set_done()
+    return predecessor_tree
+
+def get_submodule_pr_title(path: str, from_sha: str, to_sha: str):
```
ccwienk: this is not a getter! use a more appropriate verb such as 'calculate' (or leave out the verb).

also, a comment would be a good idea (state the intention of the PR title; in particular, that this is used to determine if a PR has already been created)

this function somewhat fits into this module finally (although I'd consider moving it to a class)

lsutils/__init__.py
```diff
@@ -0,0 +1,121 @@
+import git
+import github3
+import urllib3
+urllib3.disable_warnings()
+from githubutil import GitHubHelper
+import util
+from util import info, fail
+
+
+def create_submodule_update_pull_request(
```
ccwienk: consider moving to a class

lsutils/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+            path=submodule_path,
+            from_sha=current_submodule_sha,
+            to_sha=predecessor_submodule_sha,
+        )
+    if repo_helper.find_existing_pr(pr_title):
+        info("Found existing pull-request for upgrade. Skipping upgrade.")
+        return
+
+    # create new branch and checkout
+    rnd_string = util.random_str()
```
ccwienk: this should go into a separate function

PR #21, opened 2018-05-23
Create function to update submodules

AndreasBurger:
Create new gitutil module containing (for now) a function to create
commits updating submodules in repositories to a specified commit-hash and
an auxiliary function.

### DIFF
--- a/gitutil.py
+++ b/gitutil.py
@@ -1,0 +1,158 @@
+# Copyright (c) 2018 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+import subprocess
+
+import git
+
+from util import ensure_not_empty, ensure_directory_exists, fail
+
+def update_submodule(
+    repo_path: str,
+    tree_ish: str,
+    submodule_path: str,
+    commit_hash: str,
+    author: str,
+    email: str,
+):
+    '''Update the submodule of a git-repository to a specific commit.
+
+    Create a new commit, with the passed tree-ish as parent, in the given repository.
+
+    Note that this implementation only supports toplevel submodules. To be removed in a
+    future version.
+
+    Parameters
+    ------
+    repo_path : str
+        Path to a directory containing an intialised git-repo with a submodule to update.
+    tree_ish : str
+        Valid tree-ish to use as base for creating the new commit. This will be used as parent for the
+        commit created by this function.
+        Example: 'master' for the head of the master-branch.
+    submodule_path : str
+        Path (relative to the repository root) to the submodule. Must be immediately below the root
+        of the repository.
+    commit_hash : str
+        The hash the submodule should point to in the created commit. This should be a valid commit-
+        hash in the submodule's repository.
+    author : str,
+        Will be set as author of the created commit
+    email : str
+        Will be set for the author of the created commit
+
+    Returns
+    ------
+    str
+        The hexadecimal SHA-1 hash of the created commit
+    '''
+    repo_path = ensure_directory_exists(os.path.abspath(repo_path))
+
+    ensure_not_empty(submodule_path)
+    if '/' in submodule_path:
+        fail('This implementation only supports toplevel submodules: {s}'.format(s=submodule_path))
+
+    ensure_not_empty(tree_ish)
+    ensure_not_empty(commit_hash)
+    ensure_not_empty(author)
+    ensure_not_empty(email)
+
+    repo = git.Repo(repo_path)
+    _ensure_submodule_exists(repo, submodule_path)
+
+    # Create mk-tree-parseable string-representation from given tree-ish.
+    tree = repo.tree(tree_ish)
+    tree_representation = _serialise_and_update_submodule(tree, submodule_path, commit_hash)
+
+    # Pass the patched tree to git mk-tree using GitPython. We cannot do this in GitPython
+    # directly as it does not support arbitrary tree manipulation.
+    # We must keep a reference to auto_interrupt as it closes all streams to the subprocess
+    # on finalisation
+    auto_interrupt = repo.git.mktree(istream = subprocess.PIPE, as_process=True)
+    process = auto_interrupt.proc
+    stdout, _= process.communicate(input=tree_representation.encode())
+
+    # returned string is byte-encoded and newline-terminated
+    new_sha = stdout.decode('utf-8').strip()
+
+    # Create a new commit in the repo's object database from the newly created tree.
+    actor = git.Actor(author, email)
+    parent_commit = repo.commit(tree_ish)
+    commit = git.Commit.create_from_tree(
+      repo = repo,
+      tree = new_sha,
+      parent_commits = [parent_commit],
+      message='Upgrade submodule {s} to commit {c}'.format(
+          s=submodule_path,
+          c=commit_hash,
+      ),
+      author=actor,
+      committer=actor,
+    )
+
+    return commit.hexsha
+
+
+def _serialise_and_update_submodule(
+    tree: git.Tree,
+    submodule_path: str,
+    commit_hash: str,
+):
+    '''Return a modified, serialised tree-representation in which the given submodule's entry is
+    altered such that it points to the specified commit hash.
+    The returned serialisation  format is understood by git mk-tree.
+
+    Returns
+    ------
+        str
+            An updated serialised git-tree with the updated submodule entry
+    '''
+    # Since there is no direct way to get the ls-tree representation from GitPython we need to gather
+    # the necessary information for each element of the given tree ourselves.
+    return '\n'.join([
+        _serialise_object_replace_submodule(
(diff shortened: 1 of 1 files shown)


gitutil.py
```diff
@@ -0,0 +1,114 @@
...
+
+    repo = git.Repo(repo_path)
+
+    # Fetch tree-ish from git.repo and create mktree-parseable string-representation of it.
+    # Since submodule-objects contain the SHA of the to-be-checked-out version of the
+    # target repository, we replace that during the string building.
+    tree = repo.tree(tree_ish)
+    tree_representation = '\n'.join(_tree_string_generator(tree, submodule_path, commit_hash))
+
+    # Pass the serialised trees to git mk-tree using GitPython. We cannot do this in GitPython
```
ccwienk: isn't it just _one_ tree (which in turn contains a sequence of trees, blobs and submodule(s))?

gitutil.py
```diff
@@ -0,0 +1,114 @@
...
+    ensure_not_empty(author)
+    ensure_not_empty(email)
+
+    repo = git.Repo(repo_path)
+
+    # Fetch tree-ish from git.repo and create mktree-parseable string-representation of it.
+    # Since submodule-objects contain the SHA of the to-be-checked-out version of the
+    # target repository, we replace that during the string building.
+    tree = repo.tree(tree_ish)
+    tree_representation = '\n'.join(_tree_string_generator(tree, submodule_path, commit_hash))
```
ccwienk: structurally, this is much better; I'm still not happy with the name '_tree_string_generator'
how about 'serialise_and_patch_tree'?

gitutil.py
```diff
@@ -0,0 +1,114 @@
...
+
+    return commit.hexsha
+
+
+def _tree_string_generator(
+    tree: git.Tree,
+    submodule_path: str,
+    commit_hash: str,
+):
+    '''Create a generator that yields the ls-tree string-representation of the passed tree one line at a time.
```
ccwienk: what's an ls-tree?

gitutil.py
```diff
@@ -0,0 +1,114 @@
...
+        Will be set as author of the created commit
+    email : str
+        Will be set for the author of the created commit
+
+    Returns
+    ------
+    str
+        The hexadecimal SHA-1 hash of the created commit
+    '''
+    ensure_directory_exists(repo_path)
```
ccwienk: inline into one line, please

gitutil.py
```diff
@@ -0,0 +1,114 @@
...
+    repo_path = os.path.abspath(repo_path)
+
+    ensure_not_empty(tree_ish)
+    ensure_not_empty(submodule_path)
+    ensure_directory_exists(os.path.join(repo_path, submodule_path))
```
ccwienk: does not have to exit (in case of a bare repository); also: check for existence of path separator character (/) and fail if one is present, indicating that our implementation does not yet support this

gitutil.py
```diff
@@ -0,0 +1,137 @@
...
+        fail('Specified submodule-path is not in the top-level of the repository: {s}'.format(s=submodule_path))
+
+    ensure_not_empty(tree_ish)
+    ensure_not_empty(commit_hash)
+    ensure_not_empty(author)
+    ensure_not_empty(email)
+
+    repo = git.Repo(repo_path)
+
+    # Fetch tree-ish from git.repo and create mktree-parseable string-representation of it.
```
ccwienk: shorten; fetch is not the appropriate term. also, this is already expressed by the code (repo.tree)

gitutil.py
```diff
@@ -0,0 +1,137 @@
...
+    # Since submodule-objects contain the SHA of the to-be-checked-out version of the
```
ccwienk: this is an implementation detail of the '_serialise_and_path_tree'-function -> rm

gitutil.py
```diff
@@ -0,0 +1,137 @@
...
+    # target repository, we replace that during the string building.
+    tree = repo.tree(tree_ish)
+    tree_representation = '\n'.join(_serialise_and_patch_tree(tree, submodule_path, commit_hash))
+
+    # Pass the serialised tree to git mk-tree using GitPython. We cannot do this in GitPython
```
ccwienk: patched

gitutil.py
```diff
@@ -0,0 +1,137 @@
...
+    # directly as it does not support arbitrary tree manipulation.
+    auto_interrupt = repo.git.mktree(istream = subprocess.PIPE, as_process=True)
```
ccwienk: it might be worth-while mentioning that we must prevent the reference to "auto_interrupt" from being finalised

gitutil.py
```diff
@@ -0,0 +1,137 @@
...
+    str
+        The hexadecimal SHA-1 hash of the created commit
+    '''
+    ensure_directory_exists(repo_path)
+    repo_path = os.path.abspath(repo_path)
+
+    ensure_not_empty(submodule_path)
+    if '/' in submodule_path:
+        fail('Specified submodule-path is not in the top-level of the repository: {s}'.format(s=submodule_path))
+
```
ccwienk: please check "if path is a submodule", using:
git.util.assure_directory_exists(path, is_file=False)

gitutil.py
```diff
@@ -0,0 +1,137 @@
...
+    Parameters
+    ------
+    repo_path : str
+        Path to a directory containing an intialised git-repo with a submodule to update.
+    tree_ish : str
+        Valid tree-ish to use as base for creating the new commit. This will be used as parent for the commit created by this function.
+        Example: 'master' for the head of the master-branch.
+    submodule_path : str
+        Path (relative to `repo_path`) to the submodule. Must be immediately below the root (i.e. `repo_path`) of the repository.
```
ccwienk: relative to repo root. this needs not be the repo_path (altough in most cases it will be)

gitutil.py
```diff
@@ -0,0 +1,137 @@
...
+    Returns
+    ------
+    str
+        The hexadecimal SHA-1 hash of the created commit
+    '''
+    ensure_directory_exists(repo_path)
+    repo_path = os.path.abspath(repo_path)
+
+    ensure_not_empty(submodule_path)
+    if '/' in submodule_path:
```
ccwienk: please state _WHY_ this is a problem: because this implementation does not support it

gitutil.py
```diff
@@ -0,0 +1,137 @@
...
+
+    return commit.hexsha
+
+
+def _serialise_and_patch_tree(
+    tree: git.Tree,
+    submodule_path: str,
+    commit_hash: str,
+):
+    '''Create a generator that yields the string-representation of the passed tree as understood by
```
ccwienk: the fact that a generator is returned is correct; but this is not the main aspect.

-> this function accepts as input a tree and a submodule path with a commit and returns a modified serialised tree representation, where the entry pointing to the submodule is altered and all other entries are left unmodified.

it is also sufficient to state that the result is returned as an iterable (although it _might_ be more adequate to return the full string, as the single lines are probably of no use to anyone)

gitutil.py
```diff
@@ -0,0 +1,137 @@
...
+    # the necessary information for each element of the given tree ourselves.
+    submodule_found = False
+    for tree_element in tree:
+        # GitPython uses the special type 'submodule' for submodules whereas git uses 'commit'.
+        if tree_element.type == 'submodule':
+            element_type = 'commit'
+            # Replace the hash the of the 'commit'-tree with the passed value if the submodule is at the specified path
+            if tree_element.path == submodule_path:
+                element_sha = commit_hash
+                submodule_found = True
```
ccwienk: might be discarded again, as you'll already check for the existence of a sub-module at the given path earlier

gitutil.py
```diff
@@ -0,0 +1,137 @@
...
+    ------
+    str
+        The hexadecimal SHA-1 hash of the created commit
+    '''
+    ensure_directory_exists(repo_path)
+    repo_path = os.path.abspath(repo_path)
+
+    ensure_not_empty(submodule_path)
+    if '/' in submodule_path:
+        fail('Specified submodule-path is not in the top-level of the repository: {s}'.format(s=submodule_path))
```
ccwienk: use git.util.assure_directory_exists(path, is_file=False) to check if the path (directory) already exists. please try whether this function properly

ccwienk:
addendum: we should have a test for this function

gitutil.py
```diff
@@ -0,0 +1,143 @@
...
+    ------
+    repo_path : str
+        Path to a directory containing an intialised git-repo with a submodule to update.
+    tree_ish : str
+        Valid tree-ish to use as base for creating the new commit. This will be used as parent for the
+        commit created by this function.
+        Example: 'master' for the head of the master-branch.
+    submodule_path : str
+        Path (relative to the repository root) to the submodule. Must be immediately below the root
+        of the repository.
```
ccwienk: state - also in the docstring, that this restriction is an implementation limitation

gitutil.py
```diff
@@ -0,0 +1,143 @@
...
+    Returns
+    ------
+    str
+        The hexadecimal SHA-1 hash of the created commit
+    '''
+    repo_path = ensure_directory_exists(os.path.abspath(repo_path))
+
+    ensure_not_empty(submodule_path)
+    if '/' in submodule_path:
+        fail('Submodule-paths not directly below the repository root are not supported. Given path: {s}'.format(s=submodule_path))
```
ccwienk: singular (it is only one path)

gitutil.py
```diff
@@ -0,0 +1,143 @@
...
+            type=element_type,
+            # mode is a number in octal representation WITHOUT '0o' prefix
+            mode=format(tree_element.mode, 'o'),
+            path=tree_element.path,
+        )
+    return serialised_tree.strip('\n')
+
+
+def _ensure_submodule_exists(repo: git.Repo, path: str):
+    '''Use GitPython to verify that a submodule with the given path exists in the repository.'''
```
ccwienk: it is irrelevant which library we are using; in particluar, this can be seen in the code

gitutil.py
ccwienk: since you are no longer evaluating it, this should be removed (as in: must)

PR #24, opened 2018-05-28
add method to obtain iterator for config-names to ConfigurationSet

AndreasBurger:


model/__init__.py
```diff
@@ -298,6 +298,14 @@ def secrets_server(self, cfg_name=None):
     def protecode(self, cfg_name=None):
         return self.cfg_factory.protecode(self._default_name('protecode', cfg_name))
 
+    def iter_configs(self, cfg_type_name):
```
ccwienk: add a docstring stating the contract:
-> returns a generator yielding all config_elements for the given cfg_type_name
add type-hint for cfg_type_name
state that ValueError be risen in case the cfg_type_name does not exist

model/__init__.py
```diff
@@ -298,6 +298,14 @@ def secrets_server(self, cfg_name=None):
...
+        if not cfg_type_name in self.raw:
+            raise ValueError("No config of type '{c}' in ConfigSet.".format(c=cfg_type_name))
```
ccwienk: you could be nicer and print the known cfg_type_names

model/__init__.py
```diff
@@ -298,6 +298,14 @@ def secrets_server(self, cfg_name=None):
...
+        if 'config_names' in self.raw[cfg_type_name]:
+            yield from self.raw[cfg_type_name]['config_names']
```
ccwienk: return the actual cfg_elements rather than just their names (this is what the method name suggests)

model/__init__.py
ccwienk: this function should be mv'ed to the cfg_factory. it may be re-used from cfg_factory here.

model/__init__.py
```diff
@@ -163,6 +169,23 @@ def secrets_server(self, cfg_name):
     def protecode(self, cfg_name):
         return self._cfg_element(cfg_type_name='protecode', cfg_name=cfg_name)
 
+    def iter_config_names(self, cfg_set_name: str, cfg_type_name: str):
```
ccwienk: rm cfg_set from signature

ccwienk: _cfg_element_names
optionally with iter_prefix

test/model_factory_test.py
```diff
@@ -168,3 +168,77 @@ def test_from_dict_fails_on_missing_cfg_types(self):
...
+
+    def test_iterate_config_names_handles_lists(self):
+        factory = ConfigFactory.from_dict(self.raw_dict)
+        cfg_list = list(
+            factory.iter_config_names(
+                cfg_set_name='second_set',
+                cfg_type_name='a_type',
+            )
+        )
+        self.assertEqual(len(cfg_list), 2)
```
ccwienk: self.assertEqual(cfg_list, [<first>, <second>]
better yet: use sets, as we do not guarantee the order

test/model_factory_test.py
```diff
@@ -168,3 +168,77 @@ def test_from_dict_fails_on_missing_cfg_types(self):
...
+        self.assertIn('second_value_of_a', cfg_list)
+        self.assertIn('third_value_of_a', cfg_list)
+
+    def test_iterate_config_names_handles_single_entries(self):
```
ccwienk: please rename to 'works_with' + add a short comment referring to the normalisation.

test/model_factory_test.py
```diff
@@ -168,3 +168,77 @@ def test_from_dict_fails_on_missing_cfg_types(self):
...
+        factory = ConfigFactory.from_dict(self.raw_dict)
+        cfg_list = list(
+            factory.iter_config_names(
+                cfg_set_name='first_set',
+                cfg_type_name='a_type',
+            )
+        )
+        self.assertEqual(len(cfg_list), 1)
```
ccwienk: as above - compare with should-be-set

test/model_factory_test.py
```diff
@@ -168,3 +168,77 @@ def test_from_dict_fails_on_missing_cfg_types(self):
...
+        self.assertIn('first_value_of_a', cfg_list)
+
+    def test_iterate_config_names_fails_on_unknown_config_type(self):
+        factory = ConfigFactory.from_dict(self.raw_dict)
+        with self.assertRaises(ValueError):
+            test_list = list(factory.iter_configs(
+                cfg_set_name='first_set',
+                cfg_type_name='c_type',
```
ccwienk: rather use a more expressive type name; such as "no_such_type" or sth. similar. this makes it:
- less likely that a c_type is later introduced when extending the test
- easier to understand that the type should be absent

test/model_factory_test.py
```diff
@@ -168,3 +168,77 @@ def test_from_dict_fails_on_missing_cfg_types(self):
...
+            ))
+
+    def test_iterate_config_names_fails_on_unknown_config_set_name(self):
```
ccwienk: remove

model/__init__.py
```diff
@@ -328,6 +328,29 @@ def secrets_server(self, cfg_name=None):
...
         return self.cfg_factory.protecode(self._default_name('protecode', cfg_name))
 
+    def _cfg_element_names(self, cfg_type_name: str):
+        ''' Return an iterator yielding the names of all config-elements for a given ConfigType
+        in this ConfigurationSet
+        '''
+        ensure_not_empty(cfg_type_name)
+
+        types = self._cfg_types()
+        if not cfg_type_name in types:
```
ccwienk: remove this

instead: collect all config_names for the given cfg_type_name or the empty list / set / iterator

in addition: check if cfg_type is known by our factory; fail in this case

model/__init__.py
```diff
@@ -328,6 +328,29 @@ def secrets_server(self, cfg_name=None):
...
+            raise ValueError("No entry of type '{c}' in ConfigSet '{n}'. Stored types: {k}".format(
+                c = cfg_type_name,
+                n = self.name(),
+                k = ', '.join(types),
+            ))
+        yield from self.raw[cfg_type_name]['config_names']
```
ccwienk: replace with ....get(cfg_type_name, {})

ccwienk: replace with ....get(cfg_type_name, {})

ccwienk: replace with ....get(cfg_type_name, {})

test/model_validation_test.py
```diff
@@ -225,3 +225,65 @@ def test_validation_fails_on_missing_key(self):
...
+    def test_cfg_element_names_works_with_single_entry(self):
+        # We specifically test the single-entry-case here because they are normalised
+        # internally the ConfigSet
+        cfg_set = self.cfg_factory.cfg_set('first_set')
+        cfg_names_set = set(cfg_set._cfg_element_names(
+                cfg_type_name='a_type',
+            ))
+        self.assertEqual(cfg_names_set, {'first_value_of_a'})
+
+    def test_cfg_element_names_works_with_lists(self):
```
ccwienk: multiple_elements

model/__init__.py
```diff
@@ -136,6 +142,29 @@ def _cfg_element(self, cfg_type_name: str, cfg_name: str):
         element_instance = element_type(**kwargs)
         return element_instance
 
+    def _cfg_elements(self, cfg_type_name: str):
+        ''' Return an iterator yielding all config_elements of a given config type.
```
ccwienk: actually, it's a generator..

model/__init__.py
```diff
@@ -136,6 +142,29 @@ def _cfg_element(self, cfg_type_name: str, cfg_name: str):
...
+        Raise ValueError if the config type is unknown.
+        '''
+        ensure_not_empty(cfg_type_name)
+
+        for element_name in self._cfg_element_names(cfg_type_name):
+            yield self._cfg_element(cfg_type_name, element_name)
+
+    def _cfg_element_names(self, cfg_type_name: str):
+        ''' Return an iterator yielding the configuration-names of the given config type specified by its name.
```
ccwienk: same

test/model_factory_test.py
```diff
@@ -73,6 +85,12 @@ def setUp(self):
           model:
             cfg_type_name: a_type
             type: NamedModelElement
+        unused_type:
```
ccwienk: name proposal: defined_but_unused_type - OR: add a comment stating that we define the type but decidedly do not define any elements

test/model_factory_test.py
```diff
@@ -73,6 +85,12 @@ def setUp(self):
...
+          src:
+          - file: unused_type_values.xxx
```
ccwienk: is this attribute actually required if there are no type instances (I did not test)?
if not, this could be omitted (same goes for the file)

test/model_factory_test.py
```diff
@@ -73,6 +85,12 @@ def setUp(self):
...
+          model:
+            cfg_type_name: unused_type
+            type: NamedModelElement
```
ccwienk: same here: is the type actually required? Could still make sense to leave it int, even if it is not required

model/__init__.py
```diff
@@ -294,6 +294,33 @@ def _cfg_element(self, cfg_type_name: str, cfg_name=None):
             cfg_name=cfg_name,
         )
 
+    def _cfg_elements(self, cfg_type_name: str):
+        ''' Return an iterator yielding all config_elements for a given ConfigType in this ConfigurationSet.
```
ccwienk: iterable

ccwienk: use consistent wording with cfg_factory._cfg_elements

model/__init__.py
```diff
@@ -294,6 +294,33 @@ def _cfg_element(self, cfg_type_name: str, cfg_name=None):
...
+        for element_name in self._cfg_element_names(cfg_type_name):
+            yield self._cfg_element(cfg_type_name, element_name)
+
+    def _cfg_element_names(self, cfg_type_name: str):
+        ''' Return an iterator yielding the names of all config-elements for a given ConfigType
+        in this ConfigurationSet.
+        Raise a ValueError if the specified cfg_type is unknown.
+        '''
+        ensure_not_empty(cfg_type_name)
+
```
ccwienk: all_cfg_elem_names = self._cfg_factory.cfg_element_names(cfg_type_name)
"return only names from our cfg-set"

test/model_factory_test.py
```diff
@@ -55,6 +55,8 @@ def test_cfg_set_parsing(self):
 
         self.assertEqual(first_elem_from_fac.raw, first_element.raw)
 
+    ### Tests for _cfg_element_names and _cfg_elements in ConfigFactory
+
     def test_cfg_element_names(self):
```
ccwienk: this may be the method name you test; it would however be nice to be more expressive; e.g. "test_enumerate_element_names" or sth. similar

ccwienk: other naming proposal: test_cfg_element_names_should_return_all_element_names

test/model_factory_test.py
```diff
@@ -67,6 +69,49 @@ def test_cfg_element_names_fails_on_unknown_config_type(self):
                 cfg_type_name='made_up_config_type',
             ))
 
+    ### Tests for _cfg_element_names and _cfg_elements in ConfigurationSet
+
+    def test_cfg_element_names_in_config_set_returns_empty_iterator_for_unused_type(self):
```
ccwienk: iterable

model/__init__.py
```diff
@@ -294,6 +294,33 @@ def _cfg_element(self, cfg_type_name: str, cfg_name=None):
...
+        known_types = self.cfg_factory._cfg_types()
+        if not cfg_type_name in known_types:
+            raise ValueError("Unknown config type '{c}'. Known types: {k}".format(
+                c = cfg_type_name,
+                k = ', '.join(known_types.keys()),
+            ))
+        if cfg_type_name in self.raw:
+            yield from self.raw[cfg_type_name]['config_names']
```
ccwienk: please just return a set

PR #28, opened 2018-05-30
Use cluster-config configured in ConcourseConfig when deploying Concourse

AndreasBurger:
Use the cluster-config configured explicitly in the ConcourseConfig when deploying Concourse instead of using the cluster-config set in the ConfigSet.

Should only be merged after changes to the config are merged.

### DIFF
--- a/concourse/setup.py
+++ b/concourse/setup.py
@@ -181,8 +181,12 @@ def deploy_concourse_landscape(
     config_set = config_factory.cfg_set(cfg_name=config_name)
     concourse_cfg = config_set.concourse()
 
-    # Set the global context to the cluster specified by the given config
-    kubernetes_config = config_set.kubernetes()
+    # Set the global context to the cluster specified in the ConcourseConfig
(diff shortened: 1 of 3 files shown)


concourse/setup.py
```diff
@@ -181,8 +181,12 @@ def deploy_concourse_landscape(
     config_set = config_factory.cfg_set(cfg_name=config_name)
     concourse_cfg = config_set.concourse()
 
-    # Set the global context to the cluster specified by the given config
-    kubernetes_config = config_set.kubernetes()
+    # Set the global context to the cluster specified in the ConcourseConfig
+    kubernetes_config_name = concourse_cfg.kubernetes_cluster_config()
+    kubernetes_config = config_factory._cfg_element(
+        cfg_type_name = 'kubernetes',
+        cfg_name = kubernetes_config_name,
```
ccwienk: I'd inline the k8s-cfg-name retrieval in such trivial cases, but ultimately that's a question of taste

PR #31, opened 2018-06-04
Adjust deploy_and_run_smoketest_pipeline to new repo structure

AndreasBurger:
The move of our smoketest-pipeline to its new location has brought a change to the relative path under which it can be found. Furthermore the name of one of the methods arguments is changed to not refer to the old smoke-test location by its name anymore.

### DIFF
--- a/github/util.py
+++ b/github/util.py
@@ -103,6 +103,18 @@ class GitHubRepositoryHelper(object):
             ref=branch,
         )
 
+    def retrieve_text_file_contents(
+        self,
+        file_path: str,
+        branch: str=None,
+        encoding: str='utf-8',
+    ):
+        if branch is None:
+            branch = self.default_branch
+
+        contents = self.retrieve_file_contents(file_path, branch)
+        return contents.decoded.decode(encoding)
+
     def create_tag(
         self,
         tag_name: str,
--- a/github/util.py
+++ b/github/util.py
@@ -40,40 +40,39 @@ class GitHubRepositoryHelper(object):
 
     def __init__(
         self,
-        github_cfg,
-        repository_owner: str,
-        repository_name: str,
+        github_cfg: GithubConfig,
+        owner: str,
+        name: str,
+        default_branch: str='master',
     ):
         self.github = _create_github_api_object(github_cfg)
         self.repository = self._create_repository(
-            repository_owner=repository_owner,
-            repository_name=repository_name
+            owner=owner,
+            name=name
         )
+        self.default_branch = default_branch
 
-    def _create_repository(
-        self,
-        repository_owner: str,
-        repository_name: str
-    ):
+    def _create_repository(self, owner: str, name: str):
         repository = self.github.repository(
-                owner=repository_owner,
-                repository=repository_name
+                owner=owner,
+                repository=name
         )
-        if not repository:
-            util.fail("Could not retrieve repository {owner}/{name}".format(owner=repository_owner, name=repository_name))
         return repository
 
     def create_or_update_file(
         self,
-        repository_branch: str,
-        repository_version_file_path: str,
+        file_path: str,
         file_contents: str,
-        commit_message: str
+        commit_message: str,
+        branch: str=None,
     )-> str:
+        if branch is None:
+            branch = self.default_branch
+
         try:
             contents = self.repository.file_contents(
-                path=repository_version_file_path,
-                ref=repository_branch
+                path=file_path,
+                ref=branch,
             )
         except NotFoundError:
             contents = None # file did not yet exist
@@ -82,20 +81,21 @@ class GitHubRepositoryHelper(object):
             decoded_contents = contents.decoded.decode('utf-8')
             if decoded_contents == file_contents:
                 # Nothing to do
-                return util.info("Repository file contents are identical to passed file contents.")
+                return util.info('Repository file contents are identical to passed file contents.')
             else:
-                response = contents.update(commit_message, file_contents.encode('utf-8'), branch=repository_branch)
+                response = contents.update(
+                    message=commit_message,
+                    content=file_contents.encode('utf-8'),
+                    branch=branch,
+                )
         else:
             response = self.repository.create_file(
-                path=repository_version_file_path,
+                path=file_path,
                 message=commit_message,
                 content=file_contents.encode('utf-8'),
-                branch=repository_branch
+                branch=branch,
             )
-        if not response:
-            util.fail('failed to update or create file (missing privileges?)')
-
-        return response["commit"].sha
+        return response['commit'].sha
 
     def create_tag(
         self,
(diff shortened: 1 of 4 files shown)


github/util.py
```diff
@@ -95,6 +95,19 @@ def create_or_update_file(
 
         return response["commit"].sha
 
+    def retrieve_text_file_contents(
+        self,
+        repository_branch: str,
+        repository_path: str,
+        encoding: str = 'utf-8',
```
ccwienk: still no spaces around the equals character ('=') according to PEP-8 - this did not change during the past few weeks, afaik.

github/util.py
```diff
@@ -95,6 +95,19 @@ def create_or_update_file(
 
         return response["commit"].sha
 
+    def retrieve_text_file_contents(
```
ccwienk: use this new helper method in 'create_or_update_file', as otherwise, the introduction of this method would duplicate code.

while doing so, please consider:

1. renaming GithubHelper to something more appropriate (like GithubRepositoryHelper)
2. passing the repository branch via the c'tor (and rm it from all offered methods)
3. consider accepting a github_cfg object in the c'tor rather than a github_api object (need to have a look at existing code to judge whether this is a good idea)

integrationtest.py
```diff
@@ -45,13 +48,29 @@ def deploy_and_run_smoketest_pipeline(
...
     calcdir = lambda path: os.path.join(cc_pipelines_repo_dir, path)
 
-    pipeline_definition_file = calcdir('definitions/test/cc-smoketest.yaml')
     template_path = calcdir('templates')
     template_include_dir = cc_pipelines_repo_dir
     pipeline_name = 'cc-smoketest'
     job_name = 'cc-smoketest-master-head-update-job'
 
-    pipeline_definition = parse_yaml_file(pipeline_definition_file, as_snd=False)
+    pipeline_definition = parse_yaml_file(pipeline_definition_file, )
```
ccwienk: according to PEP-8, it is a discouraged practice to write a trailing comma in single-lined argument lists. Also: why did you change this back to a SND? This does not belong to the semantics of this commit either way.

ccwienk: thinking of it: this cannot work, can it? pipeline_deifnition_file is no longer defined, as line 48 is removed. so this looks like a bug to me. did you actually test this?

integrationtest.py
```diff
@@ -45,13 +48,29 @@ def deploy_and_run_smoketest_pipeline(
...
+
+    # obtain pipeline-definition from github at hardcoded location
+    github_cfg = config_set.github('github_wdf_sap_corp')
```
ccwienk: no need to hard-code the github_cfg name - use the default github_cfg

integrationtest.py
```diff
@@ -45,13 +48,29 @@ def deploy_and_run_smoketest_pipeline(
...
-    pipeline_definition_file = calcdir('definitions/test/cc-smoketest.yaml')
     template_path = calcdir('templates')
     template_include_dir = cc_pipelines_repo_dir
     pipeline_name = 'cc-smoketest'
     job_name = 'cc-smoketest-master-head-update-job'
 
-    pipeline_definition = parse_yaml_file(pipeline_definition_file, as_snd=False)
+    pipeline_definition = parse_yaml_file(pipeline_definition_file, )
+
+    # obtain pipeline-definition from github at hardcoded location
```
ccwienk: retrieve or obtain? please choose consistent wording. I'd figure it to be less adequate to use 'obtain', though.

github/util.py
```diff
@@ -41,12 +41,14 @@ def __init__(
         github: GitHub,
         repository_owner: str,
         repository_name: str,
+        repository_branch: str='master',
```
ccwienk: rename to 'default_branch' (optional parameter, initialised with master (as is already the case)
consider renaming repository_owner / repository_name to owner, name respectively

github/util.py
```diff
@@ -41,12 +41,14 @@ def __init__(
         github: GitHub,
```
ccwienk: this hsould better be named github_api (to be more expressive and consitent with other places in our code base; or: replace with github_cfg

github/util.py
```diff
@@ -63,15 +65,14 @@ def _create_repository(
 
     def create_or_update_file(
         self,
-        repository_branch: str,
```
ccwienk: make this an optional parameter (default value =None - if None, use default_branch)

github/util.py
```diff
@@ -38,40 +38,41 @@ class GitHubRepositoryHelper(object):
...
+    def _create_repository(self, owner: str, name: str):
         repository = self.github.repository(
-                owner=repository_owner,
-                repository=repository_name
+                owner=owner,
+                repository=name
         )
         if not repository:
-            util.fail("Could not retrieve repository {owner}/{name}".format(owner=repository_owner, name=repository_name))
+            util.fail("Could not retrieve repository {owner}/{name}".format(owner=owner, name=name))
```
ccwienk: do not use util.fail outside of cli/* . raise a RuntimeError instead (general rule)

in this case: github3.py raises an exception since version 1.1 - so rm the code block entirely

github/util.py
```diff
@@ -38,40 +38,41 @@ class GitHubRepositoryHelper(object):
...
         return repository
 
     def create_or_update_file(
         self,
-        repository_branch: str,
-        repository_version_file_path: str,
+        path: str,
```
ccwienk: either consistently name arguments short (path, contents) or long (file_path, file_contents);

github/util.py
```diff
@@ -96,6 +92,22 @@ def create_or_update_file(
 
         return response["commit"].sha
 
+    def retrieve_text_file_contents(
+        self,
+        path: str,
```
ccwienk: if you opt to rename path to file_path in the previous commit, also rename it here

github/util.py
```diff
@@ -82,20 +81,24 @@ def create_or_update_file(
...
             if decoded_contents == file_contents:
                 # Nothing to do
-                return util.info("Repository file contents are identical to passed file contents.")
+                return util.info('Repository file contents are identical to passed file contents.')
             else:
-                response = contents.update(commit_message, file_contents.encode('utf-8'), branch=repository_branch)
+                response = self.repository.file_contents(
+                    path=file_path,
+                    ref=branch,
+                ).update(
```
ccwienk: revert this hunk, please

github/util.py
```diff
@@ -70,15 +70,11 @@ def create_or_update_file(
             branch = self.default_branch
 
         try:
-            contents = self.repository.file_contents(
-                path=file_path,
-                ref=branch,
-            )
+            decoded_contents = self.retrieve_text_file_contents(file_path=file_path, branch=branch)
```
ccwienk: def retrieve_file(file_path, branch):
  \# retrieve file via self.repository..... and return it

def retrieve_file_contents(.....):
  contents = self._retrieve_file(...)
  return contents.decode(...)

PR #32, opened 2018-06-04
assign team to repo for sync webhooks

RaphaelVogel:
cli command to assign a team to all repositories and give the team admin rights. This is necessary to be able to sync webhooks

### DIFF
--- a/cli/githubutil.py
+++ b/cli/githubutil.py
@@ -18,7 +18,51 @@ from github3.exceptions import NotFoundError
 
 from util import ctx, not_empty, info, warning, verbose, CliHint
 from github import GithubWebHookSyncer, CONCOURSE_ID
-from github.util import GitHubHelper, _create_github_api_object
+from github.util import (GitHubHelper, _create_github_api_object,
+                         _create_team, _add_user_to_team, _add_all_repos_to_team)
+
+
+def assign_github_team_to_repo(
+    github_cfg_name: str,
+    github_org_name: str,
+    auth_token: CliHint(help="Token from an org admin user. Token must have 'admin:org' scope"),
+    team_name: str='ci'
+):
+    '''
+    Assign team 'team_name' to all repositories in organization 'github_org_name' and
+    give the team admin rights on those repositories. The team will be created if it does not exist
+    and the technical github user (from github_cfg_name) will be assigned to the team.
+    The token of the technical github user must have the privilege to create webhooks (scope admin:repo_hook)
+    The 'auth_token' parameter must belong to an org admin. The token must have 'admin:org' privileges.
+    '''
+    cfg_factory = ctx().cfg_factory()
+    github_cfg = cfg_factory.github(github_cfg_name)
+    github_username = github_cfg.credentials().username()
+
+    github = _create_github_api_object(
+        github_cfg=github_cfg,
+        auth_token=auth_token
+    )
+
+    _create_team(
+        github=github,
+        organization_name=github_org_name,
+        team_name=team_name
+    )
+
+    _add_user_to_team(
+        github=github,
+        organization_name=github_org_name,
+        team_name=team_name,
+        user_name=github_username
+    )
+
+    _add_all_repos_to_team(
+        github=github,
+        organization_name=github_org_name,
+        team_name=team_name
+    )
+
 
 def release_and_prepare_next_dev_cycle(
     github_cfg_name: str,
--- a/github/util.py
+++ b/github/util.py
@@ -15,14 +15,23 @@
 import datetime
 import os
 import sys
+from enum import Enum
 
 from github3.github import GitHub, GitHubEnterprise
 from github3.repos.repo import Repository
-from github3.exceptions import NotFoundError
+from github3.exceptions import NotFoundError, ForbiddenError
+from github3.orgs import Team
 
 import util
 import version
 from model import ConfigFactory, GithubConfig
+
+
+class RepoPermission(Enum):
+    PULL = "pull"
+    PUSH = "push"
+    ADMIN = "admin"
+
 
 class GitHubHelper(object):
     GITHUB_TIMESTAMP_UTC_FORMAT = '%Y-%m-%dT%H:%M:%SZ'
@@ -116,10 +125,12 @@ class GitHubHelper(object):
 
 
 def _create_github_api_object(
-    github_cfg: 'GithubConfig'
+    github_cfg: 'GithubConfig',
+    auth_token: str=None
 ):
+    '''parameter 'auth_token' will override the token from 'github_cfg' '''
     github_url = github_cfg.http_url()
-    github_auth_token = github_cfg.credentials().auth_token()
+    github_auth_token = auth_token if auth_token else github_cfg.credentials().auth_token()
 
     github_verify_ssl = github_cfg.tls_validation()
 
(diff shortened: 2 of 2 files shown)


cli/githubutil.py
```diff
@@ -19,6 +19,49 @@
...
+    github_cfg_name: str,
+    github_org_name: str,
+    auth_token: CliHint(help="Token from an org admin user. Token must have 'admin:org' scope"),
+    team_name: str='ci'
+):
+    '''
+    Assign team 'team_name' to all repositories in organization 'github_org_name' and
+    give the team admin rights on those repositories. The team will be created if it does not exist
+    and the technical github user (from github_cfg_name) will be assigned to the team.
+    The token of the technical user must have the privilege to create webhooks (scope admin:repo_hook)
```
ccwienk: it might be worthwile repeating that the given auth_token must grant 'admin:org' privileges (as already noted in the cli-hint)

cli/githubutil.py
```diff
@@ -19,6 +19,49 @@
 from util import ctx, not_empty, info, warning, verbose, CliHint
 from github import GithubWebHookSyncer, CONCOURSE_ID
 from github.util import GitHubHelper, _create_github_api_object
+from github.util import _create_team, _add_user_to_team, _add_all_repos_to_team
```
ccwienk: please do not duplicate the import

github/util.py
```diff
@@ -116,10 +116,11 @@ def retrieve_email_address(self, user_name):
 
 
 def _create_github_api_object(
-    github_cfg: 'GithubConfig'
+    github_cfg: 'GithubConfig',
+    auth_token: str=None
```
ccwienk: it should be documented that the auth_token argument will overwrite the one from github_cfg.

Although it might be the cleaner approach to patch-in the auth_token into the github_cfg's credentials object and not change this function

github/util.py
```diff
@@ -221,3 +222,82 @@ def retrieve_email(username: str):
...
+def _create_team(
+    github: GitHub,
+    organization_name: str,
+    team_name: str
+):
+    # passed GitHub object must have org. admin authorization to create a team
+    organization = github.organization(organization_name)
+    team_list = list(filter(lambda t: t.name == team_name, organization.teams()))
+    if team_list:
+        util.verbose("Team {name} already exist".format(name=team_name))
```
ccwienk: nit: existS

github/util.py
```diff
@@ -221,3 +222,82 @@ def retrieve_email(username: str):
...
     )
 
+def _create_team(
+    github: GitHub,
+    organization_name: str,
+    team_name: str
+):
+    # passed GitHub object must have org. admin authorization to create a team
+    organization = github.organization(organization_name)
+    team_list = list(filter(lambda t: t.name == team_name, organization.teams()))
```
ccwienk: is we are only interested in the first (and only) match, it might be more elegant to iterate over all teams like so:

for team in organization.teams():
 if team.name == team_name:
   return util.verbose(....)

but in the end this is more of a stylistic question

github/util.py
```diff
@@ -221,3 +222,82 @@ def retrieve_email(username: str):
...
+    if team_list:
+        util.verbose("Team {name} already exist".format(name=team_name))
+        return
+
+    try:
+        organization.create_team(name=team_name)
+        util.info("Team {name} created".format(name=team_name))
+    except ForbiddenError as err:
+        util.fail("{msg} Cannot create team {name} in org {org} due to missing privileges".format(
```
ccwienk: why not name 'msg' 'err' - this would make it easier to spot that the exception is printed out =)

github/util.py
```diff
@@ -221,3 +222,82 @@ def retrieve_email(username: str):
...
+
+def _add_user_to_team(
+    github: GitHub,
+    organization_name: str,
+    team_name: str,
+    user_name: str
+):
+    # passed GitHub object must have org. admin authorization to add a user to a team
+    organization = github.organization(organization_name)
+    team_list = list(filter(lambda t: t.name == team_name, organization.teams()))
```
ccwienk: this is redundant coding.

please do the following:

- create a function 'retrieve_team_by_name'
  - if the team exists, the github team object is returned
  - if it does not, it could (to your preference) return None or raise an Exception (I'd prefer the first)
  - as team names should be unique, a single team instance is returned (not a sequence)
- use this function to replace the redundant team-lookups here and in the function above

github/util.py
```diff
@@ -221,3 +222,82 @@ def retrieve_email(username: str):
...
+
+def _add_all_repos_to_team(
+    github: GitHub,
+    organization_name: str,
+    team_name: str,
+    permission: str='admin'
+):
+    # passed GitHub object must have org. admin authorization to assign team to repo with admin rights
+    organization = github.organization(organization_name)
+    team_list = list(filter(lambda t: t.name == team_name, organization.teams()))
```
ccwienk: ditto

github/util.py
```diff
@@ -221,3 +222,82 @@ def retrieve_email(username: str):
...
+    team = team_list[0]
+    for repo in organization.repositories():
+        if team.has_repository(repo.full_name):
+            util.verbose("Team {teamnname} already assigned to repo {reponame}".format(
+                teamnname=team_name,
+                reponame=repo.full_name
+            ))
+            continue
+
+        team.add_repository(repository=repo.full_name, permission=permission)
```
ccwienk: does thy github3 api not define the allowed permission values? :-(

cli/githubutil.py
RaphaelVogel: Done

cli/githubutil.py
RaphaelVogel: Done

github/util.py
RaphaelVogel: Documentation is done. 
Just patching/exchange the token from the created `github_cfg` object is not possible. One would need to create a new login session with the new token

github/util.py
RaphaelVogel: Done

github/util.py
RaphaelVogel: Done

github/util.py
RaphaelVogel: Done

github/util.py
RaphaelVogel: Done

github/util.py
RaphaelVogel: Done

github/util.py
ccwienk: that's true. but this does not actually contradict my proposal. in this function's first line, a github_api object is created from the passed github_cfg. if the token would have been patched before calling this function, this would subsequently lead to a github_api object using the patched token

github/util.py
```diff
@@ -221,3 +224,95 @@ def retrieve_email(username: str):
...
+    try:
+        organization.create_team(name=team_name)
+        util.info("Team {name} created".format(name=team_name))
+    except ForbiddenError as err:
+        util.fail("{err} Cannot create team {name} in org {org} due to missing privileges".format(
+            err=err,
+            name=team_name,
+            org=organization_name
+        ))
+
```
ccwienk: nit: two lines between functions (methods are separated by one line) according to pep-8

github/util.py
```diff
@@ -221,3 +224,95 @@ def retrieve_email(username: str):
...
+            teamname=team_name
+        ))
+    else:
+        util.fail("Could not add {username} to team {teamname}. Please check for missing privileges".format(
+            username=user_name,
+            teamname=team_name
+        ))
+
+
+class RepoPermission(Enum):
```
ccwienk: shouldn't this class be moved to the top of the module?

github/util.py
```diff
@@ -221,3 +224,95 @@ def retrieve_email(username: str):
...
+    push = "push"
+    admin = "admin"
+
+
+def _add_all_repos_to_team(
+    github: GitHub,
+    organization_name: str,
+    team_name: str,
+    permission: RepoPermission=RepoPermission.admin
+):
```
ccwienk: would be nice to document the function's semantics; for the sake of clear(er) structuring, it could be considered to accept an iterable of repositories in this function and retrieve the organisation's repositories outside of this function. admittedly, retrieving an organisation's repositories is a trivial one-liner.

github/util.py
```diff
@@ -221,3 +224,95 @@ def retrieve_email(username: str):
...
+        team.add_repository(repository=repo.full_name, permission=permission.value)
+        util.info("Added team {teamname} to repository {reponame}".format(
+            teamname=team_name,
+            reponame=repo.full_name
+        ))
+
+def _retrieve_team_by_name(
+    organization: 'github3.orgs.Organization',
+    team_name: str
+) -> 'github3.orgs.ShortTeam':
```
ccwienk: should rather be 'github3.orgs.Team'; also, the symbol should be imported and be referenced "directly", I think. using string literals is a recommended practise to workaround cyclic import dependencies within local code afaik (which is not valid here).

github/util.py
```diff
@@ -221,3 +224,95 @@ def retrieve_email(username: str):
...
+            ))
+            continue
+
+        team.add_repository(repository=repo.full_name, permission=permission.value)
+        util.info("Added team {teamname} to repository {reponame}".format(
+            teamname=team_name,
+            reponame=repo.full_name
+        ))
+
+def _retrieve_team_by_name(
```
ccwienk: _could_ be more expressive, if it '_or_none' or 'if_present' or sth. similar were appended to the name. also, it would be nice to state the contract in a docstring (the fact that None is returned if the team does not exist)

PR #33, opened 2018-06-07
add getter for step-arguments

AndreasBurger:
adds a getter for arguments of the step, if they are configured.

### DIFF
--- a/concourse/pipelines/modelbase.py
+++ b/concourse/pipelines/modelbase.py
@@ -152,9 +152,11 @@ class PipelineStep(ModelBase):
             # image must be a valid docker image reference
             allowed_characters = string.ascii_letters + string.digits +'.-_/:'
             if not all(map(lambda c: c in allowed_characters, image_reference)):
-                fail('forbidden character in image reference: ' + str(image_reference))
+                raise ModelValidationError('forbidden character in image reference: ' + str(image_reference))
             if not ':' in image_reference:
-                fail('image reference must contain colon charater:' + str(image_reference))
+                raise ModelValidationError('image reference must contain colon charater:' + str(image_reference))
+        if isinstance(self.raw.get('arguments', None), dict):
+            raise ModelValidationError('step arguments must be a scalar value or a list')
 
     def __str__(self):
         descr = 'PipelineStep {n} - depends: {d}, inputs: {i}, outputs: {o}'.format(
--- a/concourse/pipelines/modelbase.py
+++ b/concourse/pipelines/modelbase.py
@@ -1,4 +1,5 @@
 import string
+import shlex
 
 from abc import abstractmethod
 
@@ -72,6 +73,32 @@ class PipelineStep(ModelBase):
 
     def image(self):
         return self.raw.get('image', None)
+
+    def command_string(self):
+        '''Calculate and return the combined command-string consisting of the executable and all arguments.
+
+        If no arguments are specified, this method returns the shell-escaped executable as given by
+        `self.execute()`. If there is one argument specified, it is assumed to be properly shell-escaped and
+        the space-seperated concatenation of the shell-escaped executable and the argument is returned.
+        Finally, if a list of arguments is configured for the step, a space-seperated concatenation of the
+        shell-escaped executable and each argument (individually shell-escaped) is returned.
+
+        Returns
+        ------
+        str
+            A properly shell-escaped string consisting of the executable followed by all arguments.
+        '''
+        arguments = self.raw.get('arguments', None)
+        shell_escaped_executable = shlex.quote(self.execute())
+
+        if arguments is None:
+            shell_escaped_arguments = []
+        elif not isinstance(arguments, list):
+            shell_escaped_arguments = [str(arguments)]
+        else:
+            shell_escaped_arguments = [shlex.quote(str(argument)) for argument in arguments]
+
+        return ' '.join([shell_escaped_executable] + shell_escaped_arguments)
 
     def registry(self):
         return self.raw.get('registry', None)


concourse/pipelines/modelbase.py
```diff
@@ -69,10 +70,28 @@ def custom_init(self, raw_dict: dict):
         if raw_dict.get('output_dir', None):
             name = raw_dict['output_dir']
             self.add_output(name + '_path', name + '_path')
+        # normalise step-arguments, if given. For convenience we allow both primitive integers and strings
```
ccwienk: what you mean are _scalar_ types

so state it: 'arguments' may either be a list of scalar values or a single scalar.

there is no need to do extensive type-checking: just convert everything into strings

concourse/pipelines/modelbase.py
```diff
@@ -69,10 +70,28 @@ def custom_init(self, raw_dict: dict):
...
+        # and lists containing entries of those two types.
+        args = raw_dict.get('arguments', None)
```
ccwienk: either call the argument 'args' or name the variable 'arguments' to be more consistent with naming

concourse/pipelines/modelbase.py
```diff
@@ -69,10 +70,28 @@ def custom_init(self, raw_dict: dict):
...
+        if args:
+            if isinstance(args, (str, int)):
```
ccwienk: do this the other way around: check if it is a list; that way, you do not have to restrict the allowed types to str and int
also: there is a third case: it could be a dict. raise a ModelValidationError in this case

concourse/pipelines/modelbase.py
```diff
@@ -69,10 +70,28 @@ def custom_init(self, raw_dict: dict):
...
+                args = [args]
+            for argument in args:
+                if not isinstance(argument, (str, int)):
+                    fail('Step arguments must be strings or integers, found: {t}'.format(t=type(argument)))
```
ccwienk: do not "util.fail" outside of a "cli.py"-context! raise a ModelValidationError. This is a general rule.

concourse/pipelines/modelbase.py
```diff
@@ -73,6 +73,9 @@ def custom_init(self, raw_dict: dict):
     def image(self):
         return self.raw.get('image', None)
 
+    def arguments(self):
+        return self.raw.get('arguments', [])
```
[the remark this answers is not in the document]
ccwienk: rather return None in case there are no arguments
also, you should return the single, shlex-quoted string representation of the arguments

concourse/pipelines/modelbase.py
```diff
@@ -69,10 +70,28 @@ def custom_init(self, raw_dict: dict):
...
+            raw_dict['arguments'] = [str(arg) for arg in args]
+
 
     def image(self):
         return self.raw.get('image', None)
 
+    def arguments(self):
+        return self.raw.get('arguments', [])
+
+    def shell_escaped_arguments(self):
```
ccwienk: inline in arguments

concourse/pipelines/modelbase.py
```diff
@@ -69,10 +70,28 @@ def custom_init(self, raw_dict: dict):
...
+        '''Return a single shell-escaped string of all arguments.'''
+        return shlex.quote(' '.join(self.arguments()))
```
ccwienk: in case the arguments were a scalar value, no quoting should be done

concourse/pipelines/modelbase.py
```diff
@@ -128,6 +147,7 @@ def validate(self):
                 fail('forbidden character in image reference: ' + str(image_reference))
             if not ':' in image_reference:
                 fail('image reference must contain colon charater:' + str(image_reference))
+        args = self.arguments()
```
ccwienk: ?

concourse/pipelines/modelbase.py
```diff
@@ -69,10 +70,27 @@ def custom_init(self, raw_dict: dict):
         if raw_dict.get('output_dir', None):
             name = raw_dict['output_dir']
             self.add_output(name + '_path', name + '_path')
+        # normalise step-arguments, if given.
+        arguments = raw_dict.get('arguments', None)
+        if arguments:
+            if isinstance(arguments, dict):
+                raise ModelValidationError('step arguments must be scalar values or lists containing only scalar values')
```
ccwienk: 'arguments' must be either _a_ scalar value or _a_ list (of scalar values);

concourse/pipelines/modelbase.py
```diff
@@ -69,10 +70,27 @@ def custom_init(self, raw_dict: dict):
...
+            raw_dict['arguments'] = [str(argument) for argument in arguments]
 
     def image(self):
         return self.raw.get('image', None)
 
+    def arguments(self):
+        arguments = self.raw.get('arguments', None)
+        if arguments is None:
+            return arguments
+        if len(arguments) == 1:
```
ccwienk: no! this way, you would erroneously handle a single-elemented list the same way as a scalar

ccwienk:
please rebase this PR. It contains a commit that does not belong ("assign team to repo for...")

concourse/pipelines/modelbase.py
```diff
@@ -73,6 +74,31 @@ def custom_init(self, raw_dict: dict):
...
+        str
+            A properly shell-escaped string consisting of the executable and followed by all arguments.
+        '''
+        arguments = self.raw.get('arguments', None)
+        shell_escaped_executable = shlex.quote(self.execute())
+        if arguments is None:
+            return shell_escaped_executable
+
+        if not isinstance(arguments, list):
+            return ' '.join((shell_escaped_executable, arguments))
```
ccwienk: to be safe, you ought to enclose 'arguments' in a call to str

concourse/pipelines/modelbase.py
```diff
@@ -73,6 +74,31 @@ def custom_init(self, raw_dict: dict):
...
+        Returns
+        ------
+        str
+            A properly shell-escaped string consisting of the executable and followed by all arguments.
+        '''
+        arguments = self.raw.get('arguments', None)
+        shell_escaped_executable = shlex.quote(self.execute())
+        if arguments is None:
+            return shell_escaped_executable
+
```
ccwienk:
```
if arguments is None:
 escaped_arguments = ()
elif not isinstance(arguments, list):
  escaped_arguments = [shlex.quote(str(argument)) for argument in arguments]
else:
 escaped_arguments = [arguments]
return ' '.join([shell_escaped_executable] + escaped_arguments)
```

PR #35, opened 2018-06-18
make webhooks for PR configurable

RaphaelVogel:


model/__init__.py
```diff
@@ -573,6 +575,9 @@ def deploy_delaying_proxy(self):
     def kubernetes_cluster_config(self):
         return self.snd.kubernetes_cluster_config
 
+    def disable_webhook_for_pr(self):
```
ccwienk: I think we should name this a bit more expressively; e.g. disable_github_pr_webhooks; or github_pull_request_webhooks;

although most methods are not documented (because most attribute are rather self-explanatory), in this case it might make sense to add a docstring.

model/__init__.py
```diff
@@ -573,6 +575,9 @@ def deploy_delaying_proxy(self):
...
+        return self.snd.disable_webhook_for_pr
```
ccwienk: while you may use SND's behaviour to return None for absent values, it might still be a good idea to use the more explicit `self.raw.get('disable_webhhok_for_pr', False)` - in particular, this would avoid introducing yet another dependency towards SND; if I properly recall, we all agreed on removing SND eventually ;-)

PR #37, opened 2018-06-22
Upgrade 3.14.1

RaphaelVogel:
Three commits with the following changes:
- Changes in original helm chart values.yaml need to be reflected in code.
- Additional CNAME record is added to nginx ingress
- chart version is bumped

### DIFF
--- a/model/__init__.py
+++ b/model/__init__.py
@@ -582,6 +582,12 @@ class ConcourseConfig(NamedModelElement):
         '''
         return self.raw.get('disable_webhook_for_pr', False)
 
+    def cname_record(self):
+        '''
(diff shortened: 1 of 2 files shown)


model/__init__.py
```diff
@@ -582,6 +582,12 @@ def disable_github_pr_webhooks(self):
         '''
         return self.raw.get('disable_webhook_for_pr', False)
 
+    def cname_record(self):
+        '''
+        Returns the CNAME which resolves to the current active Concourse instance.
+        '''
+        return self.raw.get('cname_record', False)
```
ccwienk: why not None?

RaphaelVogel: Done. Returns now "None" if not set.

PR #42, opened 2018-06-28
refactor: replace SimpleNamespaceDict with dictionary

AndreasBurger:


concourse/client.py
```diff
@@ -544,7 +544,7 @@ def process_events(self, callback=None, filter_for_task_id=None, yield_cb=False)
         for event in client.events():
             if event is None or not event.data or len(event.data.strip()) == 0:
                 return True
-            parsed = SimpleNamespaceDict(json.loads(event.data))
+            parsed = json.loads(event.data)
             data = parsed.data
```
ccwienk: this is a regression; use parse.get('data')

util.py
```diff
@@ -167,7 +156,7 @@ def is_yaml_file(path: CliHints.existing_file()):
 def parse_yaml_file(path: CliHints.existing_file(), as_snd=True):
```
ccwienk: 1. add comment indicating the parameter (as_snd) should be removed (OR: remove it)
2. just return yaml.load(f) (no if/else)

PR #48, opened 2018-07-02
always add default options trait

RaphaelVogel:


concourse/pipelines/model/traits/options.py
```diff
@@ -17,7 +17,9 @@
 
 class OptionsTrait(Trait):
     def build_logs_to_retain(self):
-        return self.raw.get('build_logs_to_retain')
+        if 'build_logs_to_retain' in self.raw:
```
ccwienk: nit: it would be shorter to use
self.raw.get('build_logs_to_retain', 1000)

PR #54, opened 2018-07-05
Add job mapping to sync hooks

AndreasBurger:
Enables adding the job mapping name as additional meta-information to webhooks created by our automation. This meta-information is then used to ensure correct behaviour when configuring webhooks for multiple job mappings.

### DIFF
--- a/concourse/client.py
+++ b/concourse/client.py
@@ -137,6 +137,7 @@ class ConcourseApiRoutes(object):
         query_args = urlencode({
             WebhookQueryAttributes.WEBHOOK_TOKEN_ATTRIBUTE_NAME: query_attributes.webhook_token,
             WebhookQueryAttributes.CONCOURSE_ID_ATTRIBUTE_NAME: query_attributes.concourse_id,
+            WebhookQueryAttributes.JOB_MAPPING_ID_ATTRIBUTE_NAME: query_attributes.job_mapping_id,
         })
         return self._api_url(
             'pipelines',
--- a/concourse/client.py
+++ b/concourse/client.py
@@ -136,12 +136,12 @@ class ConcourseApiRoutes(object):
     ):
         query_args = urlencode({'webhook_token': webhook_token, 'concourse_id': concourse_id})
         return self._api_url(
-          'pipelines',
-          pipeline_name,
-          'resources',
-          resource_name,
-          'check',
-          'webhook'
+            'pipelines',
+            pipeline_name,
+            'resources',
+            resource_name,
+            'check',
+            'webhook'
         ) + '?' + query_args
 
     @ensure_annotations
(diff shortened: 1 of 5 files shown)


github/__init__.py
```diff
@@ -22,7 +22,7 @@
 DEFAULT_HOOK_EVENTS = ['create', 'pull_request', 'push']
 DEFAULT_HOOK_NAME = 'web' # see https://developer.github.com/v3/repos/hooks/
 DEFAULT_HOOK_CONTENT_TYPE = 'json'
-CONCOURSE_ID = 'concourse_id'
+HOOK_CONCOURSE_ID = 'concourse_id'
```
ccwienk: how about CONCOURSE_ID_ATTRIB[UTE] (this is the attribute with which we construct an id with which we identify one concourse instance

github/__init__.py
```diff
@@ -23,6 +23,7 @@
 DEFAULT_HOOK_NAME = 'web' # see https://developer.github.com/v3/repos/hooks/
 DEFAULT_HOOK_CONTENT_TYPE = 'json'
 HOOK_CONCOURSE_ID = 'concourse_id'
+HOOK_JOB_MAPPING_ID = 'job_mapping'
```
ccwienk: JOB_MAPPING_ID_ATTRIB[UTE]

ccwienk: how about defining a type:

class ConcourseJobMappingId(object):
 def __init__(self, concourse_id, job_mapping_id):
  .... # self-assign arguments
 CONCOURSE_ID_ATTRIB = 'concourse_id'
 JOB_MAPPING_ID_ATTRIB = 'job_mapping_id'

concourse/client.py
```diff
@@ -134,8 +134,12 @@ def resource_check_webhook(
         resource_name: str,
         webhook_token: str,
         concourse_id: str,
+        job_mapping_name: str=None,
```
ccwienk: do not allow None

concourse/client.py
```diff
@@ -134,8 +134,12 @@ def resource_check_webhook(
...
     ):
-        query_args = urlencode({'webhook_token': webhook_token, HOOK_CONCOURSE_ID: concourse_id})
+        query = {'webhook_token': webhook_token, HOOK_CONCOURSE_ID: concourse_id}
+        if job_mapping_name is not None:
```
ccwienk: no need to introduce this in a backwards-compatible way - do the full required change in one commit, please

concourse/util.py
```diff
@@ -51,6 +51,7 @@ def sync_webhooks(
   concourse_team:str='kubernetes',
   concourse_pipelines:[str]=None,
   concourse_verify_ssl:bool=False,
+  job_mapping_name:str=None,
```
ccwienk: nope, no need

PR #58, opened 2018-07-10
Fix tests

AndreasBurger:


model/__init__.py
```diff
@@ -138,7 +138,7 @@ def _cfg_element(self, cfg_type_name: str, cfg_name: str):
         # (with the exception of ConfigurationSet)
         configs = self._configs(cfg_type.cfg_type_name())
         if not cfg_name in configs:
-            raise ValueError('no such cfg element: {cn}. Known elements: {es}'.format(
+            raise KeyError('no such cfg element: {cn}. Known elements: {es}'.format(
```
ccwienk: in general, this is a valid fix; but as discussed, rather introduce a new exception 'ConfigElementNotFoundError' in model/base.py and expect this exception in the broken unittest, please

PR #59, opened 2018-07-10
use kubeconfig for destroy_concourse

RaphaelVogel:


concourse/setup.py
```diff
@@ -273,6 +274,44 @@ def deploy_concourse_landscape(
     info('Setting teams on Concourse ...')
     set_teams(config=concourse_cfg)
 
+
+def destroy_concourse_landscape(config_name: str, release_name: str, helm_cmd_path: str):
```
ccwienk: why not pass the cfg_set?
also, I would prefer to determine the helm path within this function (util.which('helm')); could be retained as an optional parameter (although I doubt this will add benefit)

concourse/setup.py
```diff
@@ -273,6 +274,44 @@ def deploy_concourse_landscape(
...
+    helm_env['KUBECONFIG'] = KUBECONFIG_FILE_NAME
+
+    with tempfile.TemporaryDirectory() as temp_dir:
+        with open(os.path.join(temp_dir, KUBECONFIG_FILE_NAME), 'w') as f:
+            yaml.dump(kubernetes_config.kubeconfig(), f)
+
+        try:
+            subprocess.run(
+                [helm_cmd_path, "delete", release_name, "--purge"],
+                env=helm_env, check=True, cwd=temp_dir
```
ccwienk: I think kwargs ought to be written one key-value-pair per line

concourse/setup.py
```diff
@@ -273,6 +274,44 @@ def deploy_concourse_landscape(
...
+            )
+        except CalledProcessError:
+            # ignore sporadic connection timeouts from infrastructure
+            info("Connection to K8s cluster lost. Continue with deleting namespace {ns}".format(
```
ccwienk: should this not rather be a warning?

PR #60, opened 2018-07-10
refactor: CliHints can now pass kwargs to created CliHint-instances

AndreasBurger:


util.py
```diff
@@ -65,20 +65,28 @@ class CliHints(object):
     predefined cli hint instances
     '''
     @staticmethod
-    def existing_file(help_string:str='an existing file'):
```
ccwienk: just rename help_string -> help

PR #65, opened 2018-07-13
update certificate cli command

RaphaelVogel:


model/__init__.py
```diff
@@ -774,5 +774,11 @@ def private_key(self):
     def certificate(self):
         return self.raw.get('certificate')
 
+    def private_key(self, private_key: str):
```
ccwienk: this won't work (overwrites the getter)

cli/concourseutil.py
```diff
@@ -100,6 +110,46 @@ def _display_info(dry_run: bool, operation: str, **kwargs):
...
+    tls_config_element.certificate(certificate)
+    configs = cfg_factory._configs('tls_config')
+
+   # patch tls config dict so that yaml.dump outputs literal strings
+    yaml.add_representer(LiteralStr, literal_str_representer)
+    for k1, v1 in configs.items():
+        for k2, v2 in v1.items():
+            configs[k1][k2] = LiteralStr(configs[k1][k2])
+
+    with open(os.path.join(output_path, "tls_config.yaml"), 'w') as f:
```
ccwienk: use the file from tls_cfg_type

PR #66, opened 2018-07-16
Retry concourse connections

AndreasBurger:


http_requests.py
```diff
@@ -75,6 +75,12 @@ def __init__(
         if basic_auth_username and basic_auth_passwd:
             self.auth = HTTPBasicAuth(basic_auth_username, basic_auth_passwd)
 
+        session = requests.Session()
+        session.mount('http://', default_http_adapter)
```
ccwienk: use schema from cfg? or rather register the adapter "for all"?

http_requests.py
```diff
@@ -75,6 +75,12 @@ def __init__(
...
+        session.mount('https://', default_http_adapter)
+        self.session = session
+
+
```
ccwienk: rm second empty line

PR #68, opened 2018-07-16
check for existence of Tag before attempting release

AndreasBurger:


github/util.py
```diff
@@ -286,6 +286,16 @@ def create_release(
         )
         return release
 
+    def tag_exists(
+        self,
+        tag_name: str,
+    ):
+        try:
+            tag_ref = self.repository.ref('tags/{}'.format(tag_name))
```
ccwienk: nit: a simple str-concat might be easier to read in this case ;-)
also, please use not_none to verify the argument is not none

PR #80, opened 2018-07-20
skip tags that are no proper semver versions

petersutter:
tags with invalid SemVer strings are now skipped. (Also a warning message is printed for skipped tags.) Previously an exception was raised causing the build to fail

### DIFF
--- a/github/release_notes.py
+++ b/github/release_notes.py
@@ -29,11 +29,11 @@ ReleaseNote = namedtuple('ReleaseNote', \
 Category = namedtuple("Category", ["identifier", "title"])
 TargetGroup = namedtuple("TargetGroup", ["identifier", "title"])
 categories = \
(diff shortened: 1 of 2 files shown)


github/release_notes.py
```diff
@@ -29,11 +29,11 @@
...
 categories = \
-    Category(identifier='noteworthy', title='## Most notable changes'), \
-    Category(identifier='improvement', title='## Improvements')
+    Category(identifier='noteworthy', title='# Most notable changes'), \
+    Category(identifier='improvement', title='# Improvements')
 target_groups = \
-    TargetGroup(identifier='user', title='### To end users'), \
-    TargetGroup(identifier='operator', title='### To operations team')
+    TargetGroup(identifier='user', title='## To end users'), \
+    TargetGroup(identifier='operator', title='## To operations team')
```
AndreasBurger: nit: Could be moved to its own commit

PR #84, opened 2018-07-24
add `tag-as-latest` to publish-trait

AndreasBurger:


concourse/pipelines/model/traits/publish.py
```diff
@@ -42,6 +42,9 @@ def registry_name(self):
     def image_reference(self):
         return self.raw['image']
 
+    def tag_as_latest(self):
+        return self.raw.get('tag-as-latest', False)
```
ccwienk: actually, I'd prefer snake-case over kebab-case. also, this is what @vlerenc suggested :-)

PR #88, opened 2018-07-30
added support for draft releases

petersutter:


cli/githubutil.py
```diff
@@ -138,6 +142,12 @@ def release_and_prepare_next_dev_cycle(
         prerelease=prerelease_suffix
     )
 
+    draft_name = draft_release_name_for_version(release_version)
+    draft_release = helper.draft_release_with_name(draft_name)
+    if draft_release:
+        verbose('cleaning up draft release {name}'.format(name=draft_release.name))
```
AndreasBurger: verbose takes no arguments. I suppose you ment 'info()'

cli/githubutil.py
```diff
@@ -176,6 +186,48 @@ def release_and_prepare_next_dev_cycle(
...
+    release_version: str,
+    repo_dir: str=None
+):
+    github_cfg = ctx().cfg_factory().github(github_cfg_name)
+
+    helper = GitHubRepositoryHelper(
+        github_cfg=github_cfg,
+        owner=github_repository_owner,
+        name=github_repository_name,
+        default_branch=repository_branch,
```
AndreasBurger: nit: I'd prefer the comma after the final line everywhere but either way is fine

github/release_notes.py
```diff
@@ -104,16 +77,94 @@ def generate_release_notes(
...
+        lines = MarkdownRenderer._nodes_to_markdown_lines(
+            nodes=nodes,
+            level=level,
+            release_note_objs=release_note_objs
+        )
+        return lines
 
-    def get_header_suffix(
+    @staticmethod
+    def _get_header_suffix(
```
AndreasBurger: Why not instance-methods?

cli/githubutil.py
petersutter: are you sure, or am I missing something? https://github.com/gardener/cc-utils/blob/master/util.py#L121

cli/githubutil.py
petersutter: oh, was probably a copy paste thing as I usually do not add a comma after the last argument :-D
I would rather not change it everywhere with this PR and rather do it in a seperate one, if this is something that you generally do in this project

github/release_notes.py
petersutter: I could have, but currently the methods do not make use of instance variables as most of them are called recursive and the parameters (for the current recursion) have to be passed to the method

cli/githubutil.py
AndreasBurger: I stand corrected. In that case there are only nits from my point of view

petersutter: ok, thanks

cli/githubutil.py
AndreasBurger: Unfortunately we don't have a defined style here. Just leave it as it is then, we have an issue concerning code-style in our backlog.

github/release_notes.py
AndreasBurger: Still, I'd say either define them 'public' (i.e. without the _ ) unless there is a good reason against that,  or make them instance-methods.

PR #89, opened 2018-07-31
ensure release and cron are only triggered if explicitly configured

AndreasBurger:


RaphaelVogel:
Please do not forget to document the behavior in cc-pipelines

PR #94, opened 2018-08-03
Remove hard-coded access to template

AndreasBurger:
Removes the hard-coded access to the template-name for rendered pipelines. The rendered pipelines will now use the template as specified by their fully merged definition, or 'default' as default if none is set.

Do not merge, a final exhaustive test-run is still to be performed

### DIFF
--- a/concourse/pipelines/enumerator.py
+++ b/concourse/pipelines/enumerator.py
@@ -296,6 +296,23 @@ class DefinitionDescriptor(object):
             t=self.concourse_target_team
         )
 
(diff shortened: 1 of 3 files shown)


concourse/pipelines/enumerator.py
```diff
@@ -296,6 +296,23 @@ def concourse_target_key(self):
             t=self.concourse_target_team
         )
 
+    def effective_descriptor(self):
```
RaphaelVogel: Why is definition_descriptor returned and not effective_descriptor?

PR #95, opened 2018-08-06
Support release notes in commit message

petersutter:
it is now possible to add release note texts within the commit message, as alternative to specifying the release notes within a pull request (same format). 

However, specifying the release notes within a pull request should be preferred as the release note text can be altered easily before the release.


petersutter:
FYI @rfranzke

PR #96, opened 2018-08-07
Add draft_release trait

AndreasBurger:
Adds the `draft_release` trait. This trait can be used to tag variants that should create or update draft-release-notes.


cli/githubutil.py
```diff
@@ -195,10 +195,17 @@ def create_or_update_draft_release(
     github_repository_name: str,
     repository_branch: str,
     release_version: str,
+    finalize_version: bool=True,
```
petersutter: can't the caller pass the proper release_version to the * create_or_update_draft_release* method?
It should receive **the same** release_version as the release_and_prepare_next_dev_cycle method. This method also does not alter the release_version variable and takes it like it is for the release

PR #107, opened 2018-09-04
do not notify owners when rendering pipelines using cli

AndreasBurger:


concourse/pipelines/replicator.py
```diff
@@ -385,16 +386,19 @@ def process_results(self, results):
...
-            warning(failed_descriptor.definition_descriptor.pipeline_name)
-            try:
-                self._notify_broken_definition_owners(failed_descriptor)
-            except Exception:
-                warning('an error occurred whilst trying to send error notifications')
-                traceback.print_exc()
-                all_notifications_succeeded = False
-
-        # signall error only if error notifications failed
+        if self._notify_owners:
```
ccwienk: mv the notification coding into a method
call said method iff it is configured (as of the notify_owners attribute)

PR #110, opened 2018-09-06
Fixes for release note generation

petersutter:
- skip release note extraction from PR or commit if text is none. Fixes runtime error
- skip empty lines in release note text
- trim bullet points from release note text

### DIFF
--- a/github/release_notes/renderer.py
+++ b/github/release_notes/renderer.py
@@ -210,7 +210,15 @@ class Renderer(object):
         bullet_points = list()
         for rls_note_obj in rls_note_objs:
             for i, rls_note_line in enumerate(rls_note_obj.text.splitlines()):
-                rls_note_line = _.trim(rls_note_line)
+                # trim '*' or '-' bullet points
+                rls_note_line = _\
+                    .chain(rls_note_line)\
+                    .trim()\
+                    .reg_exp_replace(r'^\* ', '')\
+                    .reg_exp_replace(r'^- ', '')\
+                    .trim()\
+                    .value()
+
                 if not rls_note_line:
                     continue
                 if i == 0:
--- a/test/github/release_notes/renderer_test.py
+++ b/test/github/release_notes/renderer_test.py
@@ -280,6 +280,64 @@ class RendererTest(unittest.TestCase):
             '  * second line2'
         self.assertEqual(expected_md_str, actual_md_str)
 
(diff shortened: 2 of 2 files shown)


github/release_notes/util.py
```diff
@@ -376,6 +376,8 @@ def extract_release_notes(
     cn_current_repo -- component name of the current repository
     """
     release_notes = list()
+    if not text:
```
ccwienk: when is this expected to happen? it seems to me as if it were preferrable to check this on the caller's side

github/release_notes/renderer.py
```diff
@@ -210,6 +210,9 @@ def _to_bullet_points(
         bullet_points = list()
         for rls_note_obj in rls_note_objs:
             for i, rls_note_line in enumerate(rls_note_obj.text.splitlines()):
+                rls_note_line = _.trim(rls_note_line)
+                if not rls_note_line:
```
ccwienk: why not rather use map + filter? this would be more consistent with other code from release_notes

github/release_notes/renderer.py
```diff
@@ -210,7 +210,15 @@ def _to_bullet_points(
         bullet_points = list()
         for rls_note_obj in rls_note_objs:
             for i, rls_note_line in enumerate(rls_note_obj.text.splitlines()):
-                rls_note_line = _.trim(rls_note_line)
+                # trim '*' or '-' bullet points
+                rls_note_line = _\
+                    .chain(rls_note_line)\
+                    .trim()\
+                    .reg_exp_replace(r'^\* ', '')\
```
ccwienk: it would be easier to read this by using lstrip('*') / lstrip('- '); you could even merge this into one single lstrip('*- ')

github/release_notes/util.py
petersutter: We saw issues when a pull request does not have a body, then text is None.

there are multiple callers (currently two) of this method, hence I added it to this central place

github/release_notes/renderer.py
petersutter:
```
>>> '  * test'.lstrip('* ')
'test'
```
OK

```
> '  *test'.lstrip('* ')
'test'
```
NOT OK

also the unit tests would fail

ccwienk: i see

PR #113, opened 2018-09-11
change slack trait attributes

petersutter:
preparation for future support of posting to multiple slack channels

dependency: cc-pipelines pull request 129

### DIFF
--- a/concourse/pipelines/model/traits/slack.py
+++ b/concourse/pipelines/model/traits/slack.py
@@ -14,12 +14,30 @@
 # limitations under the License.
 
 from util import not_none
+from model import ModelBase
 
 from concourse.pipelines.model.step import PipelineStep
 from concourse.pipelines.modelbase import (
   Trait,
-  TraitTransformer
+  TraitTransformer,
+  ModelDefaultsMixin,
+  ModelValidationMixin,
+  ModelValidationError,
 )
+
+
+class ChannelConfig(ModelBase, ModelDefaultsMixin, ModelValidationMixin):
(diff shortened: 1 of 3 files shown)


concourse/pipelines/model/traits/slack.py
```diff
@@ -32,19 +53,39 @@ def _defaults_dict(self):
 
     def _optional_attributes(self):
         return {
-            'slack_channel',
-            'slack_cfg_name',
+            'default_channel',
```
ccwienk: should this not be a required attribute?

concourse/pipelines/model/traits/slack.py
```diff
@@ -32,19 +53,39 @@ def _defaults_dict(self):
...
+
+    def default_channel(self):
+        return self.raw.get('default_channel')
 
     def transformer(self):
         return SlackTraitTransformer(name=self.name)
 
+    def validate(self):
+        super().validate()
+        default_channel = self.default_channel()
```
ccwienk: if this is not a required attribute, this would fail in case no default would be set (because there will probably not be a cfg with 'None' as key

concourse/pipelines/model/traits/slack.py
petersutter: right

petersutter:
rebased changes from master so it can be merged

petersutter:
rebased changes again from master so it can be merged

ccwienk:
please fix codestyle error

PR #123, opened 2018-09-24
Add image-rescan functionality to productutils

AndreasBurger:


protecode/model.py
```diff
@@ -81,3 +81,9 @@ def cve_major_severity(self) -> int:
 class ScanResult(ModelBase):
     def name(self):
         return self.raw.get('name')
+
+    def is_stale(self):
```
ccwienk: what does 'is_stale' mean? might be worth a comment

product/scanning.py
```diff
@@ -20,6 +21,11 @@
 from container.registry import retrieve_container_image
 from .model import ContainerImage, Component, UploadResult, UploadStatus
 
+class UploadMode(Enum):
+    UPDATE = 1
```
ccwienk: I would prefer more expressive names (and potentially a _short_ (!) description of the semantics)

- UPDATE -> UPLOAD_IF_CHANGED (or UPLOAD_DIFF?)
- RESCAN -> "document: trigger rescan for all uploaded artifacts / images. (re-)upload absent artifacts"
- ALWAYS -> I'd prefer FORCE or FORCE_UPLOAD

Also, should this maybe be better named "ScanningMode", as the upload is just an implementation detail, whereas the actual semantics we are after is the scanning

product/scanning.py
```diff
@@ -100,10 +102,14 @@ def retrieve_scan_result(
         product = self._api.scan_result(product_id=product_id)
         return product
 
+    def _retrieve_short_scan_result_from_id(self, product_id: int):
+        return
```
ccwienk: that can't be right

product/scanning.py
```diff
@@ -118,22 +124,41 @@ def upload_image(
 
         if scan_result:
             # check if image version changed
-
             metadata = scan_result.custom_data()
             image_reference = metadata.get('image-reference')
-            image_changed = image_reference == container_image.image_reference()
+            image_changed = image_reference != container_image.image_reference()
```
ccwienk: could you split this (valid and good) refactoring change into a separate commit, please?

product/scanning.py
```diff
@@ -118,22 +124,41 @@ def upload_image(
...
 
-            if not image_changed:
-                # image reference did not change - early exit
+            if not image_changed and upload_mode is UploadMode.UPDATE:
+                # image reference did not change and we need not rescan - early exit
```
ccwienk: nit: I think this is noch proper english grammar.
how about: ..no rescanning was requested

product/scanning.py
```diff
@@ -118,22 +124,41 @@ def upload_image(
...
+                        status, result = self.wait_and_retrieve_upload_result(
+                            wait_for_result,
+                            scan_result.product_id(),
+                            enum_type=ScanStatus,
+                        )
+                        return upload_result(
+                            status = status,
+                            result = result,
+                        )
+                else:
```
ccwienk: explicitly check for the other upload modes; add another else that will raise NotImplemented to detect missing case handling

product/scanning.py
```diff
@@ -149,15 +174,23 @@ def upload_image(
...
+        )
+
+    def wait_and_retrieve_upload_result(self, wait_for_result: bool, product_id: int, enum_type):
         if wait_for_result:
-            result = self._api.wait_for_scan_result(product_id=result.product_id())
+            result = self._api.wait_for_scan_result(product_id=product_id)
 
         if result.status() == ProcessingStatus.BUSY:
-            upload_status = UploadStatus.UPLOADED_PENDING
+            upload_status = enum_type(2) # PENDING
```
ccwienk: ??? why?

product/scanning.py
```diff
@@ -149,15 +174,23 @@ def upload_image(
...
         else:
-            upload_status = UploadStatus.UPLOADED_DONE
+            upload_status = enum_type(4) # DONE
 
-        return upload_result(
-            status=upload_status,
-            result=result
-        )
+        return (upload_status, result)
```
ccwienk: why no longer return an instance of UploadResult?

AndreasBurger:
I've finished and committed some more work. Most likely not up to snuff for a merge, but I'm getting there. My thoughts:

I think the approach you've suggested offline is the right one.
That being said, I'm not 100% content with the way I handle waiting. If you think the general approach (do 'the work' first, handle waiting later if at all required) is sound, I'd like to work on this a little bit more. Unfortunately, short of explicit getters/setter I'm out of immediate ideas.
Furthermore, I'm a bit perturbed by the similarity of the control flow in the `UPLOAD_IF_CHANGED` case, albeit with a bit more structure than before.
Finally, I'm not sure if caching results for processing is the right approach. Another approach would be to retrieve them once and pass them around, but the decoupling of processing and waiting would require some more work to get that done properly.

protecode/model.py
```diff
@@ -81,3 +81,12 @@ def cve_major_severity(self) -> int:
...
+        # Results are 'stale' if there has been a protecode-engine update since the last scan, i.e.
+        # a rescan might find more vulnerabilities
+        return self.raw.get('is_stale')
+
+    def has_binary(self):
+        # Indicates whether the binary has already been deleted on the protecode side
```
ccwienk: nit1: this way it _could_ be misinterpreted that True means the file has been removed
nit2 (also for the previous hunk): why not format this as a docstring?

product/model.py
```diff
@@ -281,9 +281,9 @@ def add_generic_dependency(self, generic_dependency):
 ## upload result model
 
 class UploadStatus(Enum):
-    SKIPPED_ALREADY_EXISTED = 1
-    UPLOADED_PENDING = 2
-    UPLOADED_DONE = 4
+    SKIPPED = 1
```
ccwienk: yes, why not. how about pushing this commit to the master branch already, so the PR may focus around the actual "payload"?

product/scanning.py
```diff
@@ -94,6 +94,14 @@ def retrieve_scan_result(
         product = self._api.scan_result(product_id=product_id)
         return product
 
+    @lru_cache(maxsize=8)
```
ccwienk: why 8?

product/scanning.py
```diff
@@ -94,6 +94,14 @@ def retrieve_scan_result(
...
+    def _retrieve_cached_scan_result(
+        self,
+        container_image: ContainerImage,
+        component: Component,
+    ):
+        return self.retrieve_scan_result(container_image=container_image, component=component)
```
ccwienk: caching may make sense. but why introduce it now? also: why not always cache (as in: why a new function)?

product/scanning.py
```diff
@@ -21,6 +22,19 @@
 from .model import ContainerImage, Component, UploadResult, UploadStatus
 
 
+class ProcessingMode(Enum):
```
ccwienk: I think this should rather be IntEnum (or maybe IntFlag)?
see: https://docs.python.org/3/library/enum.html

product/scanning.py
```diff
@@ -21,6 +22,19 @@
...
+    UPLOAD_IF_CHANGED = 1
+    RESCAN = 2
+    FORCE_UPLOAD = 4
+
+
+class AppStatus(Enum):
```
ccwienk: ditto (IntEnum / IntFlag)

also: I am unhappy with the name. I know protecode calls uploaded things "apps". But this term is much too generic for my taste. It should at least be called ProtecodeApp maybe. There must be a better name

product/scanning.py
```diff
@@ -21,6 +22,19 @@
...
+    UPLOADED = 1
```
ccwienk: _brief_ comments / pydoc strings might be helpful (since we are currently in the course of refactoring as far as the commits I have seen go)

cli/productutil.py
```diff
@@ -28,7 +28,7 @@
 )
 from product.util import merge_products, ComponentDescriptorResolver
 from product.scanning import ProcessingMode
-from protecode.util import upload_images
+from protecode.util import process_images
```
ccwienk: yeah, why not. this commit would also be a candidate for being pushed to master branch already, as it does not add to the PR's semantical focus imo

why do you think 'process' is a better name than 'upload'? both fine with me. i just do not see the added semantics resulting from the change of names

cli/productutil.py
```diff
@@ -82,6 +82,30 @@ def parse_dependency_str(token):
     return parse_dependency_str
 
 
+def rescan_product_images(
```
ccwienk: could the scanning mode not be an argument (so that no new function would need to be defined)?

cli/productutil.py
```diff
@@ -82,6 +82,30 @@ def parse_dependency_str(token):
...
+    protecode_cfg_name: str,
+    product_cfg_file: CliHints.existing_file(),
```
ccwienk: btw (just an idea that just came to my mind): how about a CliHints.yaml_file that would already do the yaml-parsing?

product/scanning.py
```diff
@@ -116,40 +116,172 @@ def _retrieve_cached_scan_result(
...
-            component: Component,
-            wait_for_result: bool=False
-        ):
-        metadata = self._metadata(container_image=container_image, component=component)
+    def process_image(
+        self,
+        container_image: ContainerImage,
+        component: Component,
+        processing_mode: ProcessingMode,
+        wait_for_result: bool=False,
```
ccwienk: tbh, I think this argument was never actually used from callers. so we might just as well remove it
buuuuut: it seems to me that it is now silently ignored, is it not?

product/scanning.py
```diff
@@ -116,40 +116,172 @@ def _retrieve_cached_scan_result(
...
+    ):
+        if processing_mode is ProcessingMode.FORCE_UPLOAD:
+            # Shortcut: Upload image no matter what
+            default_custom_data = self._metadata(
+                container_image=container_image,
+                component=component,
+            )
+            processing_result = self.upload_image_to_protecode(
```
ccwienk: why are there two methods: 'upload_image_to_protecode' and 'upload_image'?

also, I would prefer to separate the logic determining which action to take and the logic that actually takes the action.

in particular, I would very much prefer to not mix the introduction of new functionality (rescan) with refactoring/restructuring

product/scanning.py
```diff
@@ -116,40 +116,172 @@ def _retrieve_cached_scan_result(
...
+
+        else:
+            raise NotImplementedError
+
+        return self.wait_for_result(
+            upload_result=processing_result,
+            wait_for_result=wait_for_result,
+        )
+
+    def wait_for_result(
```
ccwienk: the wait_for_result method seems like a good refactoring to me. however: why does it have a wait_for_result argument? what's the point of calling wait_for_result with this argument set to false? in that case, it would not make sense to call this method at all, would it? unless it did something else. which would indicate that:
a) it should be named differently
b) the part that does other things should be split into a separate method

product/scanning.py
```diff
@@ -116,40 +116,172 @@ def _retrieve_cached_scan_result(
...
+                container_image=container_image,
+                component=component,
+                custom_data=default_custom_data,
+            )
+
+        result = self._retrieve_cached_scan_result(
+            container_image=container_image,
+            component=component,
+        )
+        if app_status & AppStatus.STALE.value == 0:
```
ccwienk: i think this can be done nicer - see previous commit's comments

product/scanning.py
```diff
@@ -116,40 +116,172 @@ def _retrieve_cached_scan_result(
...
+            # Scans are still 'fresh', skip scanning
+            return UploadResult(
+                container_image=container_image,
+                component=component,
+                result=result,
+                status=UploadStatus.SKIPPED,
+            )
+
+        if app_status & AppStatus.BINARY_AVAILABLE.value != 0:
```
ccwienk: ditto

product/scanning.py
```diff
@@ -116,40 +116,172 @@ def _retrieve_cached_scan_result(
...
+        app_status = self.determine_app_status(container_image=container_image, component=component)
+
+        default_custom_data = self._metadata(
             container_image=container_image,
             component=component,
         )
 
-        if scan_result:
-            # check if image version changed
+        if app_status & AppStatus.UPLOADED.value == 0:
```
ccwienk: ditto

product/scanning.py
```diff
@@ -116,40 +116,172 @@ def _retrieve_cached_scan_result(
...
-            metadata = scan_result.custom_data()
-            image_reference = metadata.get('IMAGE_REFERENCE')
-            image_changed = image_reference != container_image.image_reference()
+        result = self._retrieve_cached_scan_result(
+            container_image=container_image,
+            component=component,
+            )
+        custom_data = result.custom_data()
+
+        if app_status & AppStatus.IMAGE_REF_CHANGED.value != 0:
```
ccwienk: ditto

product/scanning.py
AndreasBurger: Caching the scan results allows me to retrieve them for free exactly when I need them instead of passing them to every function if they _might_ be needed.

product/scanning.py
AndreasBurger: Seemed like a good idea to prevent a large cache at the time. I realise now that this is currently not really possible and thus removed the limit.

product/scanning.py
```diff
@@ -66,6 +66,7 @@ def _metadata(
         metadata.update(self._component_metadata(component=component, omit_version=omit_version))
         return metadata
 
+    @lru_cache(maxsize=None)
```
ccwienk: I think I already asked before: why the caching?

product/scanning.py
```diff
@@ -40,6 +39,7 @@ def __init__(self, protecode_api: ProtecodeApi, group_id=None):
         protecode_api.login()
         self._api = not_none(protecode_api)
         self._group_id = group_id
+        self._last_analysis_result = None
```
ccwienk: this makes the PCU thread-unsafe (but it is used in a multithreading ctx)
--> rm the field and pass the value via method-args

product/scanning.py
```diff
@@ -108,40 +108,110 @@ def retrieve_scan_result(
...
+
+        if processing_result.status is UploadStatus.SKIPPED:
+            return processing_result
+
+        if wait_for_result:
+            processing_result = self.wait_for_result(processing_result)
+
+        return self.finalize_result(processing_result)
+
+    def finalize_result(self, upload_result: UploadResult):
```
ccwienk: after removal of optional waiting for result, the entire method should be no longer needed

product/scanning.py
```diff
@@ -108,40 +108,110 @@ def retrieve_scan_result(
...
+    def wait_for_result(self, upload_result: UploadResult):
+        '''Block until the protecode-analysis of the component encapsulated in the given
+        UploadResult is complete.
+        '''
+        product_id = upload_result.result.product_id()
+        analysis_result = self._api.wait_for_scan_result(product_id=product_id)
 
-        upload_result = partial(UploadResult, container_image=container_image, component=component)
+        # return updated status
+        return UploadResult(
```
ccwienk: just patch the result object rather than creating new instances

cli/productutil.py
```diff
@@ -44,12 +50,15 @@ def upload_product_images(
         raw_dict=parse_yaml_file(product_cfg_file)
     )
 
+    processing_mode_value = ProcessingMode[processing_mode.upper()]
```
ccwienk: hacky..

product/scanning.py
```diff
@@ -117,26 +119,40 @@ def retrieve_scan_result(
...
+    def supported_modes(self):
+        return [ProcessingMode.UPLOAD_IF_CHANGED, ProcessingMode.RESCAN]
+
     def _determine_upload_action(
             self,
             container_image: ContainerImage,
             scan_result: AnalysisResult,
     ):
-        if self._processing_mode is not ProcessingMode.UPLOAD_IF_CHANGED:
+        if self._processing_mode not in self.supported_modes():
```
ccwienk: that is not a good idea. actually, this is kind of even worse than not checking the modi at all.

product/scanning.py
```diff
@@ -117,26 +119,40 @@ def retrieve_scan_result(
...
+
+        elif self._processing_mode is ProcessingMode.RESCAN:
+            scan_result_short = self._api.scan_result_short(scan_result.product_id())
+
+            if not scan_result_short.has_binary():
+                return UploadAction.UPLOAD
+            elif scan_result_short.is_stale():
+                return UploadAction.RESCAN
+            else:
+                return UploadAction.SKIP
```
ccwienk: add: else: raise NotImplementedError

cli/productutil.py
```diff
@@ -37,7 +37,7 @@ def upload_product_images(
     protecode_cfg_name: str,
     product_cfg_file: CliHints.existing_file(),
     processing_mode: CliHint(
-        choices=['upload_if_changed', 'rescan']
+        choices=['upload_if_changed', 'rescan', 'force_upload']
```
ccwienk: this is much too cumbersome. I replaced this:

https://github.com/gardener/cc-utils/commit/d3c6d5f0245f7ae52d36f0ebabe6e9c05ac3c222

product/scanning.py
```diff
@@ -31,11 +31,13 @@
 
 class ProcessingMode(Enum):
     UPLOAD_IF_CHANGED = auto()
+    RESCAN = auto()
 
 
 class UploadAction(Flag):
```
ccwienk: change this into a regular enum (since we are not using this as flags)

product/scanning.py
```diff
@@ -117,26 +119,40 @@ def retrieve_scan_result(
         product = self._api.scan_result(product_id=product_id)
         return product
 
+    def supported_modes(self):
```
ccwienk: rm

PR #130, opened 2018-10-09
post release notes after successful release

petersutter:


cli/githubutil.py
```diff
@@ -156,18 +157,17 @@ def release_and_prepare_next_dev_cycle(
...
             github_repository_name=github_repository_name,
             github_cfg=github_cfg,
             repo_dir=repo_dir,
             github_helper=helper,
             repository_branch=repository_branch,
-            slack_cfg_name=slack_cfg_name,
-            slack_channel=slack_channel,
-            release_version=release_version
         )
+        release_notes_md = release_notes.to_markdown()
```
ccwienk: would it not suffice to store the release_notes object (or None) and convert it to a str where required?

Also, please initialise variables with default values before conditionally assigning them with values (and keep the amount of those variables to a minimum).

github/release_notes/util.py
```diff
@@ -40,45 +40,49 @@
...
-                channel=slack_channel,
-                title=title,
-                message=release_notes_md_links
-            )
-        except RuntimeError as e:
-            warning(e)
 
-    release_notes_md =  release_notes.to_markdown()
-    return release_notes_md
+def post_to_slack(
```
ccwienk: should this function not rather be part of the slack module (and pbly accept a string or another object containing slack-specific formatting hints?

github/release_notes/util.py
```diff
@@ -40,45 +40,49 @@
...
+    release_notes: ReleaseNote,
+    github_repository_name: str,
+    slack_cfg_name: str,
+    slack_channel: str,
+    release_version: str
+):
+    post_to_slack = release_notes and slack_cfg_name and slack_channel
```
ccwienk: please do not do this. the caller should check whether or not to post. a function named 'post_to_slack' should be expected to actually do what its name suggests.

github/release_notes/util.py
petersutter: as discussed via phone we leave it as is (with the other changes as discussed and already committed)

PR #134, opened 2018-10-12
Add Concourse version handling

RaphaelVogel:
Add concourse version to ConcourseApi object to distinguish API access to different concourse versions.

Corresponds to https://github.wdf.sap.corp/kubernetes/cc-config/commit/c8f5781eef93a499dce2b8092305634910d4c29f

### DIFF
--- a/cli/concourseutil.py
+++ b/cli/concourseutil.py
@@ -308,14 +308,10 @@ def trigger_resource_check(
     cfg_set = cfg_factory.cfg_set(cfg_name)
     concourse_cfg = cfg_set.concourse()
     team_credentials = concourse_cfg.team_credentials(team_name)
-    api = client.ConcourseApi(
-        base_url=concourse_cfg.external_url(),
-        team_name=team_credentials.teamname(),
-    )
-    api.login(
-        team_credentials.teamname(),
-        team_credentials.username(),
-        team_credentials.passwd(),
+
+    api = client.from_cfg(
+        concourse_cfg=concourse_cfg,
+        team_name=team_credentials.teamname()
     )
     api.trigger_resource_check(
         pipeline_name=pipeline_name,
--- a/concourse/client.py
+++ b/concourse/client.py
@@ -25,7 +25,11 @@ import sseclient
 import util
 
 from github.webhook import WebhookQueryAttributes
-from model.concourse import ConcourseTeamCredentials
+from model.concourse import (
+    ConcourseTeamCredentials,
+    ConcourseApiVersion,
+    ConcourseConfig,
+)
 from http_requests import AuthenticatedRequestBuilder
 from util import warning, not_empty
 
@@ -51,18 +55,45 @@ Other types defined in this module are not intended to be instantiated by users.
 '''
 
 
-def from_cfg(concourse_cfg, team_name: str):
-    concourse_api = ConcourseApi(
-        base_url=concourse_cfg.external_url(),
-        team_name=team_name,
-    )
-    # prepare api (perform a login)
+def from_cfg(concourse_cfg: ConcourseConfig, team_name: str, verify_ssl=False):
+    '''
+    Factory method to get Concourse API object
+    '''
+    base_url = concourse_cfg.external_url()
     team_credentials = concourse_cfg.team_credentials(team_name)
-    concourse_api.login(
-        team=team_name,
-        username=team_credentials.username(),
-        passwd=team_credentials.passwd(),
-    )
+    team_name = team_credentials.teamname()
+    username = team_credentials.username()
+    password = team_credentials.passwd()
+    concourse_version = concourse_cfg.concourse_version()
+
+    if concourse_version is ConcourseApiVersion.V3:
+        routes = ConcourseApiRoutesV3(base_url=base_url, team=team_name)
+        concourse_api = ConcourseApiV3(
+            base_url=base_url,
+            team_name=team_name,
+            team_credentials=team_credentials,
+            username=username,
+            password=password,
+            routes=routes,
+            verify_ssl=verify_ssl,
+        )
+    elif concourse_version is ConcourseApiVersion.V4:
+        routes = ConcourseApiRoutesV4(base_url=base_url, team=team_name)
+        concourse_api = ConcourseApiV4(
+            base_url=base_url,
+            team_name=team_name,
+            team_credentials=team_credentials,
+            username=AUTH_TOKEN_REQUEST_USER,
+            password=AUTH_TOKEN_REQUEST_PWD,
+            routes=routes,
+            verify_ssl=verify_ssl,
+        )
+    else:
+        raise NotImplementedError(
+            "Concourse version {v} not supported".format(v=concourse_version.value)
+        )
+
+    concourse_api.login()
     return concourse_api
 
 
@@ -77,9 +108,13 @@ class SetPipelineResult(Enum):
 
 # GLOBAL DEFINES
 CONCOURSE_API_SUFFIX = 'api/v1'
+# Hard coded oauth user and password
+# https://github.com/concourse/fly/blob/f4592bb32fe38f54018c2f9b1f30266713882c54/commands/login.go#L143
+AUTH_TOKEN_REQUEST_USER = 'fly'
+AUTH_TOKEN_REQUEST_PWD = 'Zmx5'
 
 
-class ConcourseApiRoutes(object):
+class ConcourseApiRoutesBase(object):
     '''
     Constructs concourse REST API endpoint URLs for the given concourse base URL and
     team name.
@@ -103,7 +138,7 @@ class ConcourseApiRoutes(object):
         base_url = self.team_url(team_name) if prefix_team else \
                     urljoin(self.base_url, self.api_suffix)
         # preserve all parts of base url
-        base_url +='/'
+        base_url += '/'
 
         return urljoin(base_url, '/'.join(parts))
 
@@ -112,14 +147,6 @@ class ConcourseApiRoutes(object):
         if not team:
             team = self.team
         return self._api_url('teams', team, prefix_team=False)
-
-    def login(self):
-        return util.urljoin(
-            self.base_url,
-            'auth',
-            'basic',
-            'token' + '?' + urlencode({'team_name': self.team})
-        )
 
     def pipelines(self):
         return self._api_url('pipelines')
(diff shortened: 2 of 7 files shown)


concourse/setup.py
```diff
@@ -167,32 +168,76 @@ def create_instance_specific_helm_values(concourse_cfg: ConcourseConfig):
...
-    instance_specific_values = {
-        'concourse': {
-            'externalURL': external_url,
-            'githubAuth': {
-                'team': creds.github_auth_team(),
-                'authUrl': creds.github_auth_auth_url(),
-                'tokenUrl': creds.github_auth_token_url(),
-                'apiUrl': creds.github_auth_api_url(),
+    concourse_version = concourse_cfg.concourse_version
+    if concourse_version == "3":
```
ccwienk: should the differentiation not rather be done based on the helm-chart version? OTOH, this will obviously work for now; at least a comment might make sense as this could be hard to understand "next time"

concourse/setup.py
```diff
@@ -167,32 +168,76 @@ def create_instance_specific_helm_values(concourse_cfg: ConcourseConfig):
...
+            creds.passwd().encode('utf-8'),
+            bcrypt.gensalt()
+        ).decode('utf-8')
+
+        instance_specific_values = {
+            'concourse': {
+                'web': {
+                    'externalUrl': external_url,
+                    'auth': {
+                        'mainTeam': {
```
ccwienk: `mainTeam` attribute is defined / required by the new helm chart, right?

requirements.txt
```diff
@@ -19,3 +19,4 @@ sseclient-py==1.7
 toposort==1.5
 urllib3==1.22
 containerregistry-ccwienk==0.2.1
+bcrypt==3.1.4
```
ccwienk: this change ought to be separately pushed to master (no review required), followed by a release of cc-job-image + incrementing the default cc-job-image version (so the runtime dependency is present in all relevant execution environments for cc-utils).

cli/concourseutil.py
```diff
@@ -308,9 +308,11 @@ def trigger_resource_check(
     cfg_set = cfg_factory.cfg_set(cfg_name)
     concourse_cfg = cfg_set.concourse()
     team_credentials = concourse_cfg.team_credentials(team_name)
+    concourse_version = concourse_cfg.concourse_version()
     api = client.ConcourseApi(
```
ccwienk: I'd prefer to add a factory function for retrieving a concourse API object of the proper version (and introduce version-specific api-classes). Like so:

```# (in concourse.client.py)
def concourse_api(
  base_url: str,
  team_name: str,
  concourse_version: concourse.client.ConcourseApiVersion
):
 if concourse_version == ConcourseVersion.V3:
  return ConcourseApiV3(...)
elif ...
else: raise NotImplementedError

# maybe rename ConcourseApi -> ConcourseApiBase
class ConcourseApiV3(ConcourseApiBase):
  pass # overwrite any functions that need to be different between V3 and V4

class ConcourseApiV4(ConcourseApiBase):
  pass # overwrite any functions that need to be different between V3 and V4

# same for ApiRoutes

class ConcourseApiVersion(enum.Enum): # --> probably this should go do model/concourse.py (and be used there; i.e. the "version-getter" should already return ConcourseApiVersion instances
 V3 = '3'
 V4 = '4'
```

Implementation note: both Api classes must expose the same methods with the same external behaviour. That way, after creating a ConcourseAPI object, no subsequent for any user of said concourse api instances ougth to be required.

concourse/client.py
```diff
@@ -77,6 +78,8 @@ class SetPipelineResult(Enum):
 
 # GLOBAL DEFINES
 CONCOURSE_API_SUFFIX = 'api/v1'
+AUTH_TOKEN_REQUEST_USER = 'fly'
+AUTH_TOKEN_REQUEST_PWD = 'Zmx5'
```
ccwienk: this feels so very, very wrong. could you at least add a comment linking to the concourse sources from where this was copied?

also, this might be better kept as attribute of model/concourse.py (ConcourseConfig); this way, it is at least somewhat prepared to be configurable (just in case it changes again in the future; could happen with each new cc version, could it not?)

concourse/client.py
```diff
@@ -188,12 +202,14 @@ class ConcourseApi(object):
 
     @param base_url: concourse endpoint (e.g. https://ci.concourse.ci)
     @param team_name: the team name used for authentication
+    @param concourse_version: major version of the target concourse
     @param verify_ssl: whether or not certificate validation is to be done
     '''
     @ensure_annotations
-    def __init__(self, base_url: str, team_name: str, verify_ssl=False):
+    def __init__(self, base_url: str, team_name: str, concourse_version: int, verify_ssl=False):
```
ccwienk: not required if handled by factory function

concourse/client.py
```diff
@@ -188,12 +202,14 @@ class ConcourseApi(object):
...
         self.base_url = base_url
         self.team = team_name
+        self.concourse_version = concourse_version
```
ccwienk: ditto

concourse/client.py
```diff
@@ -188,12 +202,14 @@ class ConcourseApi(object):
...
         self.routes = ConcourseApiRoutes(base_url=base_url, team=team_name)
```
ccwienk: the api routes object should be passed-in by the factory function (in the proper version)

concourse/client.py
```diff
@@ -215,20 +231,42 @@ def _delete(self, url: str):
 
     @ensure_annotations
     def login(self, team: str, username: str, passwd: str):
```
ccwienk: so this method would need to be implemented twice (ConcourseApiV3 + ConcourseApiV4)

concourse/client.py
```diff
@@ -344,33 +382,50 @@ def build_events(self, build_id):
 
     def set_team(self, team_credentials: ConcourseTeamCredentials):
```
ccwienk: ditto

concourse/replicator.py
```diff
@@ -245,6 +245,7 @@ def _concourse_api(concourse_cfg, team_name: str):
     api = client.ConcourseApi(
         base_url=concourse_cfg.ingress_url(),
         team_name=team_credentials.teamname(),
+        concourse_version=concourse_cfg.concourse_version(),
```
ccwienk: this needs to be replaced with a call of the factory function

concourse/setup.py
```diff
@@ -196,7 +196,7 @@ def create_instance_specific_helm_values(concourse_cfg: ConcourseConfig):
                 }
             }
         }
-    elif concourse_version == "4":
+    elif concourse_version == 4:
```
ccwienk: so you changed from str to int. in this case: please already do this in the first commit ;-)

also: please replace with enum (as commented above)

concourse/setup.py
```diff
@@ -547,6 +547,7 @@ def set_teams(config: ConcourseConfig):
     concourse_api = client.ConcourseApi(
```
ccwienk: replace with factory function invocation

concourse/util.py
```diff
@@ -32,13 +32,18 @@
...
     concourse_url: str,
+    concourse_version: int,
     concourse_user: str='kubernetes',
     concourse_passwd: str='kubernetes',
     concourse_team: str='kubernetes',
     concourse_pipelines=None,
     github_url: str=None,
 ):
-    concourse_api = concourse.ConcourseApi(base_url=concourse_url, team_name=concourse_team)
+    concourse_api = concourse.ConcourseApi(
```
ccwienk: again: call concourse api factory function.

concourse/util.py
```diff
@@ -32,13 +32,18 @@
...
+        base_url=concourse_url,
+        team_name=concourse_team,
+        concourse_version=concourse_version,
+    )
     concourse_api.login(
```
ccwienk: btw - considering we basically always need to do a login as a first operation: why not do the login in the factory function already? this would remove a lot of redundant code

concourse/util.py
```diff
@@ -32,13 +32,18 @@
 
 def list_github_resources(
```
ccwienk: hmm.. thinking of it, it might be a good idea to change this function's signature in such a way that it accepts an initialised concourse api instance; this way, there would not be the need to pass-in the version after all..

http_requests.py
```diff
@@ -106,7 +106,11 @@ def _request(self,
             headers.update(kwargs['headers'])
             del kwargs['headers']
         if 'data' in kwargs:
-            headers['content-type'] = 'application/x-yaml'
```
ccwienk: oh dear.. this was somewhat dirty (defaulting to x-yaml mime type). at any rate: would it not be better to simply replace the 'content-type' in case it is given via kwargs, rather than introducing this new special logic?

model/concourse.py
```diff
@@ -91,6 +91,12 @@ def ingress_url(self):
     def helm_chart_version(self):
         return self.raw.get('helm_chart_version')
 
+    def concourse_version(self):
+        return self.raw.get('concourse_version')
```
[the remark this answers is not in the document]
RaphaelVogel: done

model/concourse.py
```diff
@@ -91,6 +91,12 @@ def ingress_url(self):
...
+
+    def github_enterprise_host(self):
+        return self.raw.get('github_enterprise_host')
```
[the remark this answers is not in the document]
RaphaelVogel: it's only required for concourse v4 not for concourse v3.

concourse/setup.py
RaphaelVogel: done

requirements.txt
RaphaelVogel: done

cli/concourseutil.py
RaphaelVogel: Done

concourse/setup.py
RaphaelVogel: Added a comment

model/concourse.py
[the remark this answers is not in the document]
ccwienk: then add it to optional_attributes, please ;-)

also, add it into the validate-method (check for version; raise if v4 and absent)

concourse/setup.py
ccwienk: what about the suggested enum?

model/concourse.py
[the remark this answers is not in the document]
ccwienk: nit: could you sqash this into the commit where the attribute is introduced?

concourse/client.py
```diff
@@ -179,23 +188,72 @@ def build_plan(self, build_id):
         return self._api_url('builds', str(build_id), 'plan', prefix_team=False)
 
 
-class ConcourseApi(object):
+class ConcourseApiRoutesV3(ConcourseApiRoutesBase):
+    '''
+    Implemets Concourse version 3 specific api route methods
```
ccwienk: nit: typo (Implemets)

concourse/client.py
```diff
@@ -179,23 +188,72 @@ def build_plan(self, build_id):
...
+
+    Not intended to be used outside of this module. Instantiated by factory method 'from_cfg'
+    '''
+    def __init__(self, base_url: str, team: str, api_suffix=CONCOURSE_API_SUFFIX):
```
ccwienk: why duplicate the c'tor? looks identical between v3 and v4

concourse/client.py
```diff
@@ -342,6 +383,41 @@ def build_events(self, build_id):
...
+        self._post(url, body='{}')
+
+
+class ConcourseApiV3(ConcourseApiBase):
+    '''
+    Implemets Concourse version 3 specific REST api methods
+
+    Not intended to be used outside of this module. Instantiated by factory method 'from_cfg'
+    '''
+    def __init__(self, concourse_cfg: ConcourseConfig, team_name: str, verify_ssl: str):
```
ccwienk: why not let the factory method pass-in the approriate routes object? then, the c'tor would not have to be duplicated

concourse/client.py
```diff
@@ -394,7 +509,7 @@ class ModelBase(object):
     Not intended to be instantiated by users of this module
     '''
 
-    def __init__(self, raw_dict: dict, concourse_api:ConcourseApi):
+    def __init__(self, raw_dict: dict, concourse_api: ConcourseApiBase):
```
ccwienk: could you add an `@ensure_annotations`. Was missing before, but since you are touching the coding anyways..

http_requests.py
```diff
@@ -107,6 +107,9 @@ def _request(self,
             del kwargs['headers']
         if 'data' in kwargs:
             headers['content-type'] = 'application/x-yaml'
+            if 'encoding' in kwargs:
```
ccwienk: I think I asked this before: why not just support passing content-type via headers?

concourse/client.py
RaphaelVogel: done

concourse/client.py
RaphaelVogel: Done

concourse/client.py
RaphaelVogel: Done

concourse/client.py
RaphaelVogel: it's not just the different routes object, but also different a different way of login login that is done in the c'tor

concourse/client.py
RaphaelVogel: done

concourse/util.py
RaphaelVogel: done

http_requests.py
RaphaelVogel: Done

concourse/setup.py
RaphaelVogel: done

PR #136, opened 2018-10-12
Add reST table

AndreasBurger:


doc/lib/sphinxutil.py
```diff
@@ -53,7 +53,7 @@ def create_section(self, title, content, parent_ids):
         section_node = nodes.section(ids=[ids])
 
         par_node, messages = self.create_paragraph(content)
-        title_node = self._get_subtitle_node(title)
+        title_node = self.create_subtitle(title)
```
ccwienk: please squash

doc/lib/sphinxutil.py
```diff
@@ -73,6 +73,11 @@ def create_bullet_list(self, lines: [str]):
 
         return bullet_list, parse_msgs
 
+    def create_table_builder(self, table_classes:[str] = ['colwidths-auto']):
```
ccwienk: nit: according to python's typing module, the typehin should rather read: `typing.List[str]` (or maybe even better something more generic, such as typing.Iterable[str])

PR #139, opened 2018-10-17
Enable cloning directly into folders

AndreasBurger:


gitutil.py
```diff
@@ -130,14 +130,24 @@ def branch_head(self, branch_name):
...
-def clone_repository(
-        to_path: str,
-        github_cfg,
-        github_repo_path: str
-    ):
-        url = url_with_credentials(github_cfg, github_repo_path)
-        git.Git(to_path).clone('--quiet', url)
+    @staticmethod
+    def clone_into_directory(
+            directory_path: str,
```
ccwienk: how about target_dir[ectory]; or repo[sitory]_dir[ectory]?

gitutil.py
```diff
@@ -130,14 +130,24 @@ def branch_head(self, branch_name):
...
+            github_cfg,
+            github_repo_path: str,
+            branch_name: str = 'Master'
```
ccwienk: I think, this should be written with a lower-cased 'm'

also: how about naming this 'initial_ref' (and setting the default value to refs/heads/master)?
also, consider using None as default value.
there is why: this way, we can discriminate between cases where the caller explicitly gave a ref and those where they did not.
in the first case, also add the '--single-branch' CLI to git-clone. This way, we do not fetch objects not reachable from the given initial ref.

gitutil.py
```diff
@@ -137,7 +137,7 @@ def clone_repository(
         github_repo_path: str
     ):
         url = url_with_credentials(github_cfg, github_repo_path)
-        git.Git(to_path).clone('--quiet', url)
+        git.Git().clone('--quiet', url, to_path)
```
ccwienk: you need to adjust all usages (currently one, if I properly recall). Please grep..

gitutil.py
```diff
@@ -130,14 +130,19 @@ def branch_head(self, branch_name):
...
 
-
-def clone_repository(
-        to_path: str,
-        github_cfg,
-        github_repo_path: str
-    ):
-        url = url_with_credentials(github_cfg, github_repo_path)
-        git.Git().clone('--quiet', url, to_path)
+    @staticmethod
```
ccwienk: this commit changes more than the commit message suggests; making a method static would result in exactly two changes:
- rm 'self' argument
- add @staticmethod decorator

I would also agree to squash this commit e.g. into the previous one, or to keep it as it is. in the latter case, please at least do not "lie" in the commit msg.

Again: also adjust users. Note that doing so will break consuming pipelines. So the change ought pbly be introduced in a backward-compatiable manner. Then a new version of cc-job-image needs to be released and rolled out. Then the usages need to be changed to the new API. then the backward-compatibility coding may be reomoved.

gitutil.py
```diff
@@ -134,10 +134,16 @@ def branch_head(self, branch_name):
...
-            github_repo_path: str
+            github_repo_path: str,
+            branch_name: str = None,
         ) -> 'GitHelper':
             url = url_with_credentials(github_cfg, github_repo_path)
-            git.Git().clone('--quiet', url, target_directory)
+            args = ['--quiet']
+            if branch_name is not None:
+                args += ['--branch', branch_name, '--single-branch']
+            args += [url, target_directory]
```
ccwienk: nit: why not add --quiet here as well?

gitutil.py
```diff
@@ -130,6 +130,24 @@ def branch_head(self, branch_name):
         # XXX added for backward compatibility - remove
         return self.fetch_head(ref=branch_name)
 
+    @staticmethod
+    def clone_into_directory(
```
ccwienk: shorten to clone_into?

gitutil.py
```diff
@@ -130,6 +130,24 @@ def branch_head(self, branch_name):
...
+        target_directory: str,
+        github_cfg,
+        github_repo_path: str,
+        branch_name: str = None,
```
ccwienk: initial_branch? checkout_branch?

PR #140, opened 2018-10-18
Enhance parsing of reST in tables

AndreasBurger:
The way restructuredtext was parsed in table-cells is taken almost verbatim from the implementation provided by Sphinx's `autosummary` directive. As this directive aims to put short summaries (one line) into the cells, it's parsing was presumably not intended to be used with larger restructuredtext blocks.

This PR changes the construction of the ViewList required for parsing to be line by line. Thus, it enables usage of directives and other multi-line constructs.

Note: This also means that indentation and initial newlines of python multi-line-strings are taken into account.


AndreasBurger:
Merged manually due to GitHub woes

PR #148, opened 2018-10-26
Respect triages in Protecode

AndreasBurger:


protecode/model.py
```diff
@@ -68,13 +68,15 @@ def cve(self):
     def cve_severity_str(self):
         return str(self.raw.get('vuln').get('cvss'))
 
+    def has_triages(self):
+        return self.raw.get('triage') is not None
```
ccwienk: please check whether "triage" might also mean: "yes, it is really bad"

ccwienk: also: please check if this check is specific enough.

by which I mean: if protecode app A contains vulnerability foo and is triaged "only for app A" (i.e. protecode still shows the same vulnerability for an app B, even after the assessment), then this check must not return True for component B.

protecode/model.py
```diff
@@ -68,13 +68,15 @@ def cve(self):
...
+
     def cve_major_severity(self) -> int:
-        if self.cve_severity_str():
+        if self.cve_severity_str() and not self.has_triages():
```
ccwienk: this filtering should not be hard-coded. either perform it in the caller determining the report or at least introduce a flag (defaulting to not to filter)

cli/productutil.py
```diff
@@ -43,6 +43,7 @@ def upload_product_images(
     protecode_group_id: int=5,
     parallel_jobs: int=4,
     cve_threshold: int=7,
+    ignore_triage: bool=False,
```
ccwienk: how about ignore_if_triaged (also in other occurrences)?

protecode/model.py
```diff
@@ -46,12 +46,15 @@ def name(self):
...
         return (Vulnerability(raw_dict=raw) for raw in self.raw.get('vulns'))
 
-    def highest_major_cve_severity(self) -> int:
+    def highest_major_cve_severity(self, ignore_triage:bool = False) -> int:
         try:
             return max(
                 map(
                     lambda v: v.cve_major_severity(),
-                    filter(lambda v: not v.historical(), self.vulnerabilities())
+                    [
```
ccwienk: it get even harder to read like this. in particular, I do not think it is a good style to mix and nest filter and list comprehension.

consider introducing an inline function and pass it to the filter invocation.

alternatively, this might be a good place to use our FluentIterable

protecode/model.py
```diff
@@ -46,12 +46,15 @@ def name(self):
     def vulnerabilities(self) -> 'Iterable[Vulnerability]':
```
ccwienk: add filter flag (ignore_if_triaged) here

protecode/model.py
```diff
@@ -46,12 +46,15 @@ def name(self):
...
         return (Vulnerability(raw_dict=raw) for raw in self.raw.get('vulns'))
 
-    def highest_major_cve_severity(self) -> int:
+    def highest_major_cve_severity(self, ignore_triage:bool = False) -> int:
```
ccwienk: convert to static method; pass-in vulnerabilities. apply "determine highest cve" logic to passed-in sequence / iterable of vulns (remove new flag)

protecode/util.py
```diff
@@ -42,12 +43,16 @@ def upload_images(
...
+        vulnerable_components = list()
+
+        for component in analysis_result.components():
+            vulnerabilities = filter(lambda v: not v.historical(), component.vulnerabilities())
+            highest_cve = determine_highest_major_cve_severity(vulnerabilities)
+            vulnerable_components.append((component, highest_cve))
 
         if vulnerable_components:
-            highest_cve = max(map(lambda c: c.highest_major_cve_severity(), vulnerable_components))
+            _, highest_cve = max(vulnerable_components, key=lambda pair: pair[1])
```
ccwienk: why bother creating two-tuples, when all we need is the cve?
how about
```
determine_highest_major_cve_severity(
  map(
    lambda r: determine_highest_major_cve_severity(r),
    map(lambda c: c.vulnerabilities())
  ),
)```

In other words:
for each component
- determine maximum for one component's vulnerabilities
from all thus determined CVE values, determine the highest CVE

File: protecode/util.py [hunk 8]
```diff
@@ -42,14 +43,17 @@ def upload_images(
...
-        if vulnerable_components:
-            highest_cve = max(map(lambda c: c.highest_major_cve_severity(), vulnerable_components))
-            if highest_cve >= cve_threshold:
-                info('Highest found CVE Severity: {cve} - Action required'.format(cve=highest_cve))
+        for component in analysis_result.components():
+            vulnerabilities = filter(lambda v: not v.historical(), component.vulnerabilities())
+            highest_cve = highest_major_cve_severity(vulnerabilities)
+            vulnerability_scores.append(highest_cve)
+
+        highest_cve = max(vulnerability_scores)
```
ccwienk: this does not handle the case that vulnerability_scores may be empty

PR #149, opened 2018-11-05
document release_callback

RaphaelVogel:


doc/traits/release.rst
```diff
@@ -29,6 +29,19 @@ The following operations are performed:
   Declaring this trait changes the default triggering behaviour to "manual"
 
 
+`release_callback` Attribute
+----------------------------
+
+You can add additional changes in the release commit using a callback script defined by the
```
ccwienk: this should rather go to concourse.model.traits.release.py (into the doc attribute of the corresponding AttributeSpec)

Also, please consider using passive sentences to match the other documentation's style. For examples, kindly refer to e.g. this file's ctx:

"The following operations are performed"
"Declaring this trait changes .."

--

E.g. ```an optional callback that is called during release commit creation. The callback is passed the absolute path to the main repository's work tree via environment variable `REPO_DIR`. Any changes left inside the worktree are added to the resulting release commit.```

doc/traits/release.rst
```diff
@@ -39,3 +52,4 @@ Example
       preprocess: 'finalize' # recommended
     release:
       nextversion: 'bump_minor'
+      release_callback: 'release_callback'  # script is relative to main repo root directory
```
ccwienk: please do not call it script. how about just stating 'relative to main repository root'

in general, I do not think it is a good pattern to add lengthy comments to our examples

doc/traits/release.rst
RaphaelVogel: Done

doc/traits/release.rst
RaphaelVogel: Done

PR #151, opened 2018-11-08
Setup webhook dispatcher

RaphaelVogel:
relates to https://github.wdf.sap.corp/kubernetes/cc-config/pull/49

### DIFF
--- a/concourse/setup.py
+++ b/concourse/setup.py
@@ -39,6 +39,9 @@ from model.concourse import (
     ConcourseConfig,
     ConcourseApiVersion,
 )
+from model.webhook_dispatcher import (
+    WebhookDispatcher,
+)
 from model.container_registry import (
     GcrCredentials,
 )
@@ -59,6 +62,7 @@ from util import (
     warning,
     fail,
     which,
+    create_url_from_attributes,
 )
 
 from kubernetes.client import (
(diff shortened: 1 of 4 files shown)


model/concourse.py
```diff
@@ -70,6 +70,9 @@ def image_pull_secret(self):
     def tls_secret_name(self):
         return self.raw.get('tls_secret_name')
 
+    def webhook_dispatcher(self):
```
ccwienk: add to required attributes, please

ccwienk: also: why does concourse "own" the dispatcher? should it not be the other way around? the dispatcher will interact (based upon concouse job mappings) with concourse instance(s) and github repositories

concourse/setup.py
```diff
@@ -560,6 +564,38 @@ def deploy_delaying_proxy(
...
+
+    service_helper = ctx.service_helper()
+    deployment_helper = ctx.deployment_helper()
+    namespace_helper = ctx.namespace_helper()
+    ingress_helper = ctx.ingress_helper()
+
+    namespace_helper.create_if_absent(namespace)
+
+    webhook_service = generate_webhook_dispatcher_service()
+    webhook_deployment = generate_webhook_dispatcher_deployment(
```
ccwienk: general remark: would it make sense to separate the creation of deployment manifests and their appliance (to better support local development scenarios)? for a first version, it is certainly ok to not do this.

RaphaelVogel:
We go for Helm charts

PR #152, opened 2018-11-09
move ingress_host to kubernetes config

RaphaelVogel:


model/concourse.py
```diff
@@ -39,6 +39,9 @@ def all_team_credentials(self):
     def external_url(self):
         return self.raw.get('externalUrl')
 
+    def ingress_prefix(self):
```
ccwienk: nit: ingress_host_prefix?

PR #156, opened 2018-11-15
use configured port for whd deployment

AndreasBurger:


model/webhook_dispatcher.py
```diff
@@ -56,3 +56,6 @@ def kubernetes_config_name(self):
 
     def webhook_dispatcher_config_name(self):
         return self.raw.get('webhook_dispatcher_config')
+
+    def webhook_dispatcher_container_port(self):
+        return self.raw['container_port']
```
RaphaelVogel: should be added as required attribute

PR #165, opened 2018-11-20
Remove old webhook syncing code

AndreasBurger:


ccwienk:
on hold for now, as we might re-use some of the removed contents

PR #176, opened 2018-12-10
Add concourse meta information to the elastic search event payload

OlegLoewen:


cli/elastic.py
```diff
@@ -55,3 +63,16 @@ def store_dir(index: str, directory: util.CliHints.existing_dir(), cfg_name: str
             if file.endswith('.json'):
                 json_files.append(os.path.join(dirpath, file))
     store_files(index, json_files, cfg_name)
+
+
+def get_meta():
```
ccwienk: nit: pbly, this should be marked as "private" (e.g. rename to `_get_meta`)

cli/elastic.py
```diff
@@ -41,9 +47,11 @@ def store_files(index: str, files: [str], cfg_name: str):
 
     for file in files:
         with open(file) as file_handle:
+            json_body = json.load(file_handle)
+            json_body.update(get_meta())
```
ccwienk: I'd prefer to explicitly set the attribute; as in:
```
json_body[<attribute>] = <value>
```

Also, I would suggest two changes:

- fail if body contains a "reserved" attribute (so that we do not silently patch away contents, but provide feedback)
- group injected data below a single (reserved) attribute. in case there is a good reason not do do nesting, let's discuss

also: I think the patching ought to be done at the ccc.elasticearch module rather than here

ccwienk: but again: let's follow-up on this tomorrow. maybe I did not put enough thoughts into this!

PR #180, opened 2018-12-13
include updated defaults.mako in built images

AndreasBurger:
Introduces a prepare-step in the docker built which executes the same script the release-hook uses to bump our defaults.mako.


Dockerfile
```diff
@@ -5,9 +5,13 @@ COPY . /cc/utils/
 # place version file into container's filesystem to make it easier to
 # determine the image version during runtime
 COPY VERSION /metadata/VERSION
+# Also copy it lowercase to mimic our version-step contract (with /metadata/ as VERSION_PATH)
+COPY VERSION /metadata/version
```
ccwienk: haha, nice idea^^

PR #184, opened 2018-12-18
Support images from private repositories in tasks

AndreasBurger:


concourse/templates/default.yaml
```diff
@@ -79,8 +80,8 @@ scan_container_images_step = concourse.steps.step_def('scan_container_images')
 <%namespace file="/resources/meta.mako" import="*"/>
 
 inherit:
-${image_registry_defaults(registry_name='gcr', registry_cfg=gcr)}
-${task_image_resource(registry_name='gcr')}
+${image_registry_defaults(registry_name=default_container_registry_name, registry_cfg=default_container_registry)}
```
ccwienk: interesting.. what does the registry_name mean, actually? I completely forgot about this piece of coding

concourse/resources/image.mako
```diff
@@ -10,3 +10,19 @@ credentials = registry_cfg.credentials()
     password: '${credentials.passwd()}'
     repository: ${image_reference}
 </%def>
+<%def name="task_image_resource(registry_cfg, image_repository, image_tag, indent=0)"
+filter="indent_func(indent),trim">
+<%
+from makoutil import indent_func
+# registry_cfg must be of type ContainerRegistryConfig (cc-utils)
```
ccwienk: good idea to comment this. why not even add it as a type hint. or use util.check_type? or both

concourse/templates/default.yaml
```diff
@@ -162,18 +161,21 @@ source_repo = job_variant.main_repository()
 % if job_step.image():
 <%
 image_reference, tag = job_step.image().split(':', 1)
+## We have been given an image. Depending on whether a registry name is also configured or not we
```
ccwienk: nit: why two ##?

concourse/templates/default.yaml
```diff
@@ -162,18 +161,21 @@ source_repo = job_variant.main_repository()
...
+## set the proper registry config and render all required credentials explicitly
+if job_step.registry():
+  container_registry = config_set.container_registry(job_step.registry())
```
ccwienk: why not just use config_set's default mechanism? if no cfg_element name is given, the default element is returned. this should also work if None is passed.

concourse/resources/defaults.mako
```diff
@@ -17,6 +17,22 @@ credentials = registry_cfg.credentials()
         repository: eu.gcr.io/gardener-project/cc/job-image
         tag: "1.44.0"
 </%def>
+<%def name='task_image_defaults(registry_cfg, indent=0)'
```
ccwienk: I do not like that this duplicates the code block immediately above. should there not rather be a function that accepts credentials and image reference ('repository' in the words of concouse's image resource).
then, the default image might be created by another function that presets those arguments, whereas the same function could be used to create custom (non-default) image resources

PR #186, opened 2019-01-03
add optional timeout attribute to steps

AndreasBurger:


concourse/model/step.py
```diff
@@ -133,6 +133,13 @@ def attrs(pipeline_step):
             variable.
             ''',
         ),
+        AttributeSpec.optional(
+            name='timeout',
+            default={},
```
ccwienk: either use 'None' or maybe -1? or omit

concourse/model/step.py
```diff
@@ -267,6 +274,9 @@ def _add_dependency(self, step: 'PipelineStep'):
     def depends(self):
         return set(self.raw['depends'])
 
+    def timeout(self):
+        return self.raw.get('timeout', None)
```
ccwienk: use attrib-spec to declare default; dereference with raw['timeout']

PR #193, opened 2019-01-14
use our default adapter in protecode client

AndreasBurger:


ccwienk:
@AndreasBurger : please fix linter errs

PR #195, opened 2019-01-15
print a warning when retrying http-requests

AndreasBurger:


http_requests.py
```diff
@@ -21,8 +21,18 @@
 
 from util import warning
 
+
+class LoggingRetry(Retry):
+    def increment(self, *args, **kwargs):
+        # super().increment will either raise an exception indicating that no retry is to
+        # be performed or return a new, modified instance of this class
+        retry = super().increment(*args, **kwargs)
+        warning(f'HTTP request unsuccessful. Retrying ...')
```
ccwienk: is there a way to output request urls + retry-count?

PR #202, opened 2019-01-21
add component-name and effective-version to elastic search client

RaphaelVogel:


ccc/elasticsearch.py
```diff
@@ -47,8 +47,8 @@ def _metadata_dict():
     if not util._running_on_ci():
         return {}
 
-    # XXX do not hard-code meta-dir
-    meta_dir = util.existing_dir(os.path.join(util._root_dir(), 'meta'))
+    meta_dir = util.existing_dir(os.path.join(util._root_dir(), os.environ.get('META')))
```
ccwienk: the '# XXX' is still valid, imo (nit)

ccc/elasticsearch.py
```diff
@@ -78,6 +78,14 @@ def read_attr(name):
         meta_dict['build-name'],
     )
 
+    effective_version = os.environ.get('EFFECTIVE_VERSION')
```
ccwienk: likewise: pease add a comment indicating this ought to be refactored

ccc/elasticsearch.py
```diff
@@ -78,6 +78,14 @@ def read_attr(name):
...
+    if effective_version:
```
ccwienk: this line can not reasonably be reached. of the env var is not set, a KeyError would have been raised

ccc/elasticsearch.py
```diff
@@ -78,6 +78,14 @@ def read_attr(name):
...
+        meta_dict['effective-version'] = effective_version
+
+    component_name = os.environ.get('COMPONENT_NAME')
+    if component_name:
```
ccwienk: ditto

PR #205, opened 2019-01-23
Refactor setup scripts

RaphaelVogel:


cli/landscape.py
```diff
@@ -21,7 +22,8 @@
     CliHints,
     CliHint,
 )
-import concourse.setup as setup
+import landscape_setup.concourse.setup_concourse as setup_concourse
```
ccwienk: would it not suffice to name the modules just 'landscape_setup.concourse', etc.?

PR #215, opened 2019-02-13
Preliminary work for transactional logic for our releases

AndreasBurger:
Restructures steps/release.py in such a way that each action with side effect is isolated from other actions with side effects by encapsulating them in classes.
Currently, this has the same behaviour as before but allows implementing revert/rollback logic for each side effect.

### DIFF
--- a/concourse/steps/release.py
+++ b/concourse/steps/release.py
@@ -1,3 +1,4 @@
+import abc
 import os
 import version
 import pathlib
@@ -30,6 +31,49 @@ def rebase(
 ):
     upstream_commit_sha = git_helper.fetch_head(upstream_ref).hexsha
     git_helper.rebase(commit_ish=upstream_commit_sha)
+
+
+class TransactionalStep(object, metaclass=abc.ABCMeta):
+    '''Abstract base class for operations that are to be executed with transactional semantics.
+
+    Instances represent operations which typically cause external and persistent side effects.
+    Typically, a sequence of (different) steps are grouped in a `Transaction`
+
+    Subclasses *may* overwrite the `validate` method, which performs optional checks that indicate
+    whether the operation would probably fail. Those checks are intended to be run for all steps of
+    a `Transaction` before actually executing it. Validation *must not* cause any persistent side
+    effects to external resources.
+
+    Subclasses *must* overwrite the `apply` method, which performs the actual payload of the step,
+    typically resulting in persistent external side effects. The `apply` method *may* also return
+    an object (e.g.: a `dict`) that is then made available to later steps
+    when part of a `Transaction`.
+
+    Subclasses *must* overwrite the `revert` method, which reverts any persistent external side
+    effects previously created by running the step's `apply` method. This should take into account
+    that the execution of the `apply` method may or may not have succeeded, failed,
+    or failed partially.
+    '''
+    def set_context(self, context: TransactionContext):
+        self._context = context
+
+    def context(self):
+        return self._context
+
+    def validate(self):
+        pass
+
+    @abc.abstractmethod
+    def apply(self):
+        return None
+
+    @abc.abstractmethod
+    def revert(self):
+        pass
+
+    @abc.abstractmethod
+    def name(self):
+        pass
 
 
 def release_and_prepare_next_dev_cycle(
--- a/concourse/steps/release.py
+++ b/concourse/steps/release.py
@@ -249,8 +249,7 @@ class ReleaseCommitStep(TransactionalStep):
         )
 
     def revert(self):
-        # TODO: revert commit
-        return
+        raise NotImplementedError('revert-method is not yet implemented')
 
 
 class ReleaseTagStep(TransactionalStep):
@@ -292,8 +291,7 @@ class ReleaseTagStep(TransactionalStep):
         )
 
     def revert(self):
-        # TODO: Delete tag
-        return
+        raise NotImplementedError('revert-method is not yet implemented')
 
 
 class GitHubReleaseStep(TransactionalStep):
@@ -368,8 +366,7 @@ class GitHubReleaseStep(TransactionalStep):
         }
 
     def revert(self):
-        #TODO: remove release from GitHub
-        return
+        raise NotImplementedError('revert-method is not yet implemented')
 
 
 class PrepareDevCycleStep(TransactionalStep):
@@ -436,8 +433,7 @@ class PrepareDevCycleStep(TransactionalStep):
         )
 
     def revert(self):
-        #TODO: remove created commit
-        return
+        raise NotImplementedError('revert-method is not yet implemented')
 
 
 def release_and_prepare_next_dev_cycle(
--- a/concourse/steps/release.py
+++ b/concourse/steps/release.py
@@ -208,7 +208,8 @@ class ReleaseCommitStep(TransactionalStep):
             self.git_helper._stash_changes()
 
         # update version file
-        self.repository_version_file_path.write_text(self.release_version)
+        with open(self.repository_version_file_path, 'w') as f:
+            f.write(self.release_version)
 
         try:
             # call optional release commit callback
@@ -462,7 +463,6 @@ def release_and_prepare_next_dev_cycle(
     slack_channel: str=None,
     rebase_before_release: bool=False,
 ):
-    repo_dir = existing_dir(repo_dir)
 
     helper = GitHubRepositoryHelper.from_githubrepobranch(githubrepobranch)
     git_helper = GitHelper.from_githubrepobranch(
@@ -470,91 +470,64 @@ def release_and_prepare_next_dev_cycle(
         repo_path=repo_dir,
     )
 
-    if helper.tag_exists(tag_name=release_version):
-        fail(
-            "Cannot create tag '{t}' in preparation for release: Tag already exists".format(
-                t=release_version,
-            )
-        )
-
-    if rebase_before_release:
-        rebase(git_helper=git_helper, upstream_ref=f'refs/heads/{repository_branch}')
-
-    # Fetch release notes and generate markdown to catch errors early
-    release_notes = fetch_release_notes(
-        github_repository_owner=github_repository_owner,
-        github_repository_name=github_repository_name,
-        github_cfg=github_cfg,
-        repo_dir=repo_dir,
-        github_helper=helper,
-        repository_branch=repository_branch,
-    )
-    release_notes_md = release_notes.to_markdown()
-
-    # Do all the version handling upfront to catch errors early
-    # Bump release version and add suffix
-    next_version = version.process_version(
-        version_str=release_version,
-        operation=version_operation
-    )
-    next_version_dev = version.process_version(
-        version_str=next_version,
-        operation='set_prerelease',
-        prerelease=prerelease_suffix
-    )
-
-    release_commit_sha = _create_and_push_release_commit(
-        repo_dir=repo_dir,
+    release_commit_step = ReleaseCommitStep(
         git_helper=git_helper,
-        version_file_path=repository_version_file_path,
+        repo_dir=repo_dir,
         release_version=release_version,
-        target_ref=repository_branch,
-        commit_msg=f'Release {release_version}',
+        repository_version_file_path=repository_version_file_path,
+        repository_branch=githubrepobranch.branch(),
         release_commit_callback=release_commit_callback,
+        rebase_before_release=rebase_before_release,
     )
 
-    helper.create_tag(
-        tag_name=release_version,
-        tag_message="Release " + release_version,
-        repository_reference=release_commit_sha,
+    release_tag_step = ReleaseTagStep(
+        github_helper=github_helper,
+        release_version=release_version,
+        author_email=author_email,
         author_name=author_name,
-        author_email=author_email
-    )
-    release = helper.create_release(
-        tag_name=release_version,
-        body=release_notes_md,
-        draft=False,
-        prerelease=False
     )
 
-    if component_descriptor_file_path:
-        with open(component_descriptor_file_path) as f:
-            # todo: validate descriptor
-            component_descriptor_contents = f.read()
-        release.upload_asset(
-            content_type='application/x-yaml',
-            name=product.model.COMPONENT_DESCRIPTOR_ASSET_NAME,
-            asset=component_descriptor_contents,
-            label=product.model.COMPONENT_DESCRIPTOR_ASSET_NAME,
-        )
-
-    # Prepare version file for next dev cycle
-    helper.create_or_update_file(
-        file_path=repository_version_file_path,
-        file_contents=next_version_dev,
-        commit_message="Prepare next dev cycle " + next_version_dev
+    github_release_step = GitHubReleaseStep(
+        github_helper=github_helper,
+        githubrepobranch=githubrepobranch,
+        repo_dir=repo_dir,
+        release_version=release_version,
+        component_descriptor_file_path=component_descriptor_file_path,
     )
 
+    dev_cycle_commit_step = PrepareDevCycleStep(
+        github_helper=github_helper,
+        repo_dir=repo_dir,
+        repository_version_file_path=repository_version_file_path,
+        release_version=release_version,
+        version_operation=version_operation,
+        prerelease_suffix=prerelease_suffix,
+    )
+
+    release_transaction = Transaction(
+        release_commit_step,
+        release_tag_step,
+        github_release_step,
+        dev_cycle_commit_step,
+    )
+
+    release_transaction.validate()
+    release_transaction.execute()
+
+    # clean up old draft release
     draft_name = draft_release_name_for_version(release_version)
-    draft_release = helper.draft_release_with_name(draft_name)
+    draft_release = github_helper.draft_release_with_name(draft_name)
+
     if draft_release:
-        verbose('cleaning up draft release {name}'.format(name=draft_release.name))
+        # TODO: clean up ALL previously made draft-releases (just in case)
+        info(f'cleaning up draft release {draft_release.name}')
         draft_release.delete()
 
+    # publish release notification to slack
     if slack_cfg_name and slack_channel:
         post_to_slack(
-            release_notes=release_notes,
-            github_repository_name=github_repository_name,
+            release_notes=release_context.release_notes(),
+            github_repository_name=githubrepobranch.github_repo_path(),
             slack_cfg_name=slack_cfg_name,
             slack_channel=slack_channel,
             release_version=release_version,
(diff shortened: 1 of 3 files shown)


concourse/steps/release.py
```diff
@@ -23,19 +28,298 @@
...
-    git_helper.rebase(commit_ish=upstream_commit_sha)
+class ReleaseContext(object):
+    def set_release_commit_sha(self, release_commit_sha):
+        self._release_commit_sha = release_commit_sha
+
+    def release_commit_sha(self):
+        return self._release_commit_sha
+
+
+class ReleaseProcessor(object):
```
ccwienk: how about: TransactionalStep? also add a _short_ coment

concourse/steps/release.py
```diff
@@ -23,19 +28,298 @@
...
+    def __init__(self, release_context: ReleaseContext):
+        self.release_context = release_context
+
+    def validate(self):
```
ccwienk: add @abc.abstractmethod

concourse/steps/release.py
```diff
@@ -23,19 +28,303 @@
...
+        return self._release_commit_sha
+
+
+class TransactionalStep(object):
+    ''' Provides a framework to implement some action in an atomic fashion.
+
+    Instances of this class are intended to perform exactly one change in their apply-method
+    and provide the corresponding undo-functionality in the revert-method
+    '''
+    def __init__(self, release_context: ReleaseContext):
```
ccwienk: so. as I elaborated above. the release-ctx should be named (and crafted) more generically. either by changing as described before. or by (this would be the easies approach) changing it to a generic object, such as a dict or a SimpleNamespace.

concourse/steps/release.py
```diff
@@ -23,19 +28,303 @@
...
+    def revert(self):
+        pass
+
+    def name(self):
+        pass
+
+
+class ReleaseCommitStep(TransactionalStep):
+    def __init__(
+        self,
```
ccwienk: you could also add `*args, **kwargs` and pass this to the super-c'tor

concourse/steps/release.py
```diff
@@ -23,19 +28,303 @@
...
+    def apply(self):
+        raise NotImplementedError
+
+    def revert(self):
+        pass
+
+    def name(self):
+        pass
+
+
```
ccwienk: I have already mentioned this in the documentation proposal. I suggest to add a class representing a sequence of steps. Maybe call it `Transaction`. It should receive the step instances and offer a method to run all steps in a transactional fashion (which you inlined a but further down).

That way (with the changes proposed up to here), we could consider to mv this out of this module and re-use it for other steps. This is however a change for another PR (the moving-out and re-use. the Transaction class should be introduced here).

concourse/steps/release.py
```diff
@@ -23,19 +28,303 @@
...
+    Instances of this class are intended to perform exactly one change in their apply-method
+    and provide the corresponding undo-functionality in the revert-method
+    '''
+    def __init__(self, release_context: ReleaseContext):
+        self.release_context = release_context
+
+    def validate(self):
+        pass
+
+    def apply(self):
```
ccwienk: I suppose apply is fine. Would it be more adequate to call it 'run' or 'execute' instead? Not sure, though. This is just a gut feeling.

concourse/steps/release.py
```diff
@@ -23,19 +28,303 @@
...
+        self.repository_branch = repository_branch
+        self.release_version = not_empty(release_version)
+        self.rebase_before_release = rebase_before_release
+        self.release_commit_callback = release_commit_callback
+
+    def name(self):
+        return "Release Commit Creator"
+
+    def validate(self):
+        info(f'validating { self.name()}')
```
ccwienk: mv any printing to the transaction class (for all steps)

concourse/steps/release.py
```diff
@@ -23,19 +28,303 @@
...
+        )
+        self.git_helper = not_none(git_helper)
+        self.repo_dir = existing_dir(repo_dir)
+        self.repository_branch = repository_branch
+        self.release_version = not_empty(release_version)
+        self.rebase_before_release = rebase_before_release
+        self.release_commit_callback = release_commit_callback
+
+    def name(self):
+        return "Release Commit Creator"
```
ccwienk: might it make more sense to choose names that describe the action (e.g. 'create release commit')

Then, we would have error messages like:
"Operation 'create release commit' failed"

(or status messages like "Operation 'create release commit' succeeded")

concourse/steps/release.py
```diff
@@ -23,19 +28,303 @@
...
+        callback_env['EFFECTIVE_VERSION'] = release_version
+
+        subprocess.run(
+            [release_commit_callback],
+            check=True,
+            env=callback_env,
+        )
+
+    def revert(self):
+        # TODO: revert commit
```
ccwienk: how about omitting the unimplemented method and print a warning instead (in the base class)?
at any rate: do not return silently

concourse/steps/release.py
```diff
@@ -23,19 +28,303 @@
...
+        release_context: ReleaseContext,
+    ):
+        super().__init__(release_context)
+        self.github_helper = not_none(github_helper)
+        self.release_version = not_empty(release_version)
+        self.author_name = not_empty(author_name)
+        self.author_email = not_empty(author_email)
+
+    def validate(self):
+        info(f'validating { self.name()}')
```
ccwienk: mv to transation

concourse/steps/release.py
```diff
@@ -23,19 +28,303 @@
...
+        # if a tag with the given release version already exists github will not let us
+        # create another one
+        if self.github_helper.tag_exists(tag_name=self.release_version):
+            raise RuntimeError(
+                f"Cannot create tag '{self.release_version}' in preparation for release: "
+                "Tag already exists"
+            )
+
+    def name(self):
+        return "Release Tag Creator"
```
ccwienk: create release tag

concourse/steps/release.py
```diff
@@ -23,19 +28,303 @@
...
+        self.github_helper.create_tag(
+            tag_name=self.release_version,
+            tag_message=f'Release {self.release_version}',
+            repository_reference=self.release_context.release_commit_sha(),
+            author_name=self.author_name,
+            author_email=self.author_email
+        )
+
+    def revert(self):
+        # TODO: Delete tag
```
ccwienk: ditto - do not silently return

concourse/steps/release.py
```diff
@@ -23,19 +28,303 @@
...
+
+
+class GitHubReleaseStep(TransactionalStep):
+    def __init__(
+        self,
+        github_helper: GitHubRepositoryHelper,
+        git_helper: GitHelper,
+        githubrepobranch: GitHubRepoBranch,
+        release_version: str,
+        release_context: ReleaseContext,
```
ccwienk: the ctx should be either at the beginning or at the end of the argument list, as it is somewhat "inherited". but maybe, we will not need to pass it like this anyway

concourse/steps/release.py
```diff
@@ -23,19 +28,303 @@
...
+    ):
+        super().__init__(release_context)
+        self.github_helper = not_none(github_helper)
+        self.githubrepobranch = githubrepobranch
+        self.git_helper = not_none(git_helper)
+        self.release_version = not_empty(release_version)
+        self.component_descriptor_file_path = existing_file(component_descriptor_file_path)
+
+    def name(self):
+        return "Release Creator"
```
ccwienk: create github release

concourse/steps/release.py
```diff
@@ -23,19 +28,303 @@
...
+        if self.component_descriptor_file_path:
+            # TODO: proper validation
+            existing_file(self.component_descriptor_file_path)
+            with open(self.component_descriptor_file_path) as f:
+                self.component_descriptor_contents = f.read()
+
+    def apply(
+        self,
+    ):
+        info('Create release')
```
ccwienk: mv to tx

concourse/steps/release.py
```diff
@@ -23,19 +28,303 @@
...
+        if self.component_descriptor_file_path:
+            release.upload_asset(
+                content_type='application/x-yaml',
+                name=product.model.COMPONENT_DESCRIPTOR_ASSET_NAME,
+                asset=self.component_descriptor_contents,
+                label=product.model.COMPONENT_DESCRIPTOR_ASSET_NAME,
+            )
+
+    def revert(self):
+        #TODO: remove release from GitHub
```
ccwienk: do not return silently

concourse/steps/release.py
```diff
@@ -23,19 +28,303 @@
...
+        self.github_helper = not_none(github_helper)
+        self.repo_dir=repo_dir
+        self.repository_version_file_path = repository_version_file_path
+
+        self.release_version = not_empty(release_version)
+        self.version_operation = not_empty(version_operation)
+        self.prerelease_suffix = not_empty(prerelease_suffix)
+
+    def name(self):
+        return "Next Cycle Commit Creator"
```
ccwienk: prepare next dev cycle

concourse/steps/release.py
```diff
@@ -23,22 +29,317 @@
...
+        self.release_commit_callback = release_commit_callback
+
+    def name(self):
+        return "Release Commit Step"
+
+    def validate(self):
+        info(f'validating {self.name()}')
+         # ensure the given callback script exists
+        if self.release_commit_callback:
+            self.release_commit_callback = os.path.join(
```
ccwienk: no! this is a very very bad idea. either omit this in the validation, or mv the code to a method that returns the next version. then call it in `validate` w/o storing it. and call it again in `apply` (persisting it this time).

In case I missed other ocurrences: never ever modify the step object (or anything else) during validation. This is super-bad! Any step *must* successfully execute (at least potentially) w/o running validate before

concourse/steps/release.py
```diff
@@ -23,22 +29,317 @@
...
+        return "Next Cycle Commit Step"
+
+    def validate(self):
+        info(f'validating {self.name()}')
+        existing_file(pathlib.Path(self.repo_dir, self.repository_version_file_path))
+        next_cycle_version = version.process_version(
+            version_str=self.release_version,
+            operation=self.version_operation,
+        )
+        self.next_cycle_dev_version = version.process_version(
```
ccwienk: actually, I meant to place the "this is superbad" comment here. but the diff was updated before i could save. so as mentioned earlier: NEVER modify state during validation

concourse/steps/release.py
```diff
@@ -23,22 +29,317 @@
...
+        info('Creating initial commit for next dev cycle')
+        # Prepare version file for next dev cycle
+        self.github_helper.create_or_update_file(
+            file_path=self.repository_version_file_path,
+            file_contents=self.next_cycle_dev_version,
+            commit_message=f'Prepare next dev cycle {self.next_cycle_dev_version}'
+        )
+
+    def revert(self):
+        #TODO: remove created commit
```
ccwienk: do not return silently

concourse/steps/release.py
```diff
@@ -49,166 +350,98 @@ def release_and_prepare_next_dev_cycle(
...
-        file_contents=next_version_dev,
-        commit_message="Prepare next dev cycle " + next_version_dev
-    )
+    # validate all steps
+    info('validating')
+    for p in steps:
+        p.validate()
+
+    successful_steps = list()
+    for step in steps:
```
ccwienk: firstly, mv this into the transaction class (see above).
secondly, you also must handle any exceptions occurring during reversion. we must try to revert as much as possible, even if errors occur in the process

concourse/steps/release.py
```diff
@@ -49,166 +350,98 @@ def release_and_prepare_next_dev_cycle(
...
+        try:
+            step.apply()
+            successful_steps.append(step)
+        except BaseException as e:
+            warning(f'An error occured while creating the release: {e}')
+            # revert the changes attempted
+            step.revert()
+            # undo all previous changes in reverse order
+            successful_steps.reverse()
+            for s in successful_steps:
```
ccwienk: nit: use reversed(...)

AndreasBurger:
I've implemented the current suggestions/requested changes:
- I've created the suggested `Transaction` class and moved the 'old' release context there after generalising it a bit more (as suggested) to a transaction context. An instance of this object will be provided to each transaction.
- I've removed all state-modifying validation. Now, simple sanity checks are performed in the constructor (e.g. no empty string passed) whereas more sophisticated checks (e.g.: Is the given string *actually* a semver) are performed in the validate() method. The release-note verification, however, had to be moved to the apply() method of the relevant step. Might be worth it to move it to the constructor, actually. I'm not sure yet.

concourse/steps/release.py
```diff
@@ -23,22 +30,392 @@
...
+    def __init__(
+        self,
+        *steps: TransactionalStep,
+    ):
+        # create context object for this transaction
+        self.context = TransactionContext()
+        # validate type of args and set context
+        for step in steps:
+            if not isinstance(step, TransactionalStep):
+                raise RuntimeError('Transactions may only contain instances of TransactionalStep')
```
ccwienk: nit: ValueError would be more appropriate

concourse/steps/release.py
```diff
@@ -23,22 +30,392 @@
...
+            info(f"Validating step '{step.name()}'")
+            step.validate()
+
+    def execute(self):
+        executed_steps = list()
+        for step in self._steps:
+            step_name = step.name()
+            # attempt to execute the step
+            try:
+                info(f"Executing step '{step_name}'")
```
ccwienk: execute or apply? ;-)

concourse/steps/release.py
```diff
@@ -23,22 +30,392 @@
...
+    def _revert(self, steps):
+        # attempt to revert each step. Raise an exception if not all reverts succeeded.
+        all_reverted = True
+        for step in steps:
+            step_name = step.name()
+            try:
+                info(f"Reverting step {step_name}")
+                s.revert()
+            except BaseException as e:
+                all_reverted = False
```
ccwienk: this is in general fine. it might help us to also dump the stacktrace in these cases (`traceback.print_exc()`). Admittedly, this is not very sophisticated. but it tends to be helpful for error analysis..

AndreasBurger:
Done. In detail:
- Unimplemented revert-methods now raise an UnimplementedMethodError
- Release notes are now posted in their own three-step-transaction (Update release notes for the GitHub release -> cleanup old draft release -> post slack release notes [optionally]). Steps in this transaction won't actually do anything when reverting, resulting in a 'succesful' release job with a warning if release notes couldn't be posted, but the release was created. 
Also of note here is commit 98e44daf2616fa5fd08b5157901c9233864db340 since it introduces an 'update release notes' method to the GitHubHelper. It should be `cherry-pick`able to the main branch if it is in order.
- I've cleaned up the commit history some more after the previous review. All changes to the Transaction's revert logic are now included in its initial commit. No sense in producing refactoring commits in a PR.

test/concourse/steps/release_test.py
```diff
@@ -0,0 +1,257 @@
...
+            examinee(release_version='made_up_version').validate()
+
+
+class TestSlackReleaseStep(object):
+    @pytest.fixture()
+    def examinee(self):
+        def constructor(
+            slack_cfg_name='test_config',
+            slack_channel='test_channel',
+            githubrepobranch=MagicMock(),
```
ccwienk: please create real objects here. there is no need to use a mock, since this is just a plain python object contain some strings, is it not?

test/concourse/steps/release_test.py
```diff
@@ -0,0 +1,257 @@
...
+                release_version=release_version,
+            )
+        return constructor
+
+    def test_validation(self, examinee):
+        examinee().validate()
+
+    def test_validation_fails_on_invalid_semver(self, examinee):
+        with pytest.raises(ValueError):
+            examinee(release_version='made_up_version').validate()
```
ccwienk: could you use names such as 'invalid_version' to indicate that the value is intentionally invalid? same for file system paths. the name (if possible to be arbitrarily chosen) should reflect the intention.

concourse/steps/release.py
```diff
@@ -23,22 +31,508 @@
...
+            info(f"Validating step '{step.name()}'")
+            step.validate()
+
+    def execute(self):
+        executed_steps = list()
+        for step in self._steps:
+            step_name = step.name()
+            # attempt to execute the step
+            info(f"Applying step '{step_name}'")
+            executed_steps.append(step)
```
ccwienk: add after successful execution?

concourse/steps/release.py
```diff
@@ -23,22 +31,508 @@
...
+        self.rebase_before_release = rebase_before_release
+
+        repo_dir_absolute = pathlib.Path(not_empty(repo_dir)).resolve()
+        self.repo_dir = str(repo_dir_absolute)
+        self.repository_version_file_path = str(
+            pathlib.Path(repo_dir_absolute, repository_version_file_path)
+        )
+        if release_commit_callback:
+            self.release_commit_callback = str(
+                pathlib.Path(repo_dir_absolute, release_commit_callback)
```
ccwienk: why create a pythlib object just to convert it back into a str? just use os.path.join.

concourse/steps/release.py
```diff
@@ -23,22 +31,508 @@
...
+        repository_branch: str,
+        rebase_before_release: bool,
+        release_commit_callback: str=None,
+    ):
+        self.git_helper = not_none(git_helper)
+        self.repository_branch = not_empty(repository_branch)
+        self.release_version = not_empty(release_version)
+        self.rebase_before_release = rebase_before_release
+
+        repo_dir_absolute = pathlib.Path(not_empty(repo_dir)).resolve()
```
ccwienk: as already pointed out. please just use `str`. OR replace all usages at least in this module with pathlib.Paths. it is certainly not a good idea to mix styles. and since we need to interface with mako, str will be the more approprate type..

concourse/steps/release.py
```diff
@@ -23,22 +31,494 @@
...
+            'release commit sha'
+        )
+        self.github_helper.create_tag(
+            tag_name=self.release_version,
+            tag_message=f'Release {self.release_version}',
+            repository_reference=release_commit_sha,
+            author_name=self.author_name,
+            author_email=self.author_email
+        )
+        return {}
```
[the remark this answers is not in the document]
ccwienk: ?

concourse/steps/release.py
```diff
@@ -23,22 +31,508 @@
...
+            with open(self.component_descriptor_file_path) as f:
+                component_descriptor_contents = f.read()
+                release.upload_asset(
+                    content_type='application/x-yaml',
+                    name=product.model.COMPONENT_DESCRIPTOR_ASSET_NAME,
+                    asset=component_descriptor_contents,
+                    label=product.model.COMPONENT_DESCRIPTOR_ASSET_NAME,
+                )
+
+        return {}
```
ccwienk: why return an empty dict?

PR #219, opened 2019-02-20
Add function to cleanup automatically created draft-releases

AndreasBurger:


ccwienk:
according to @petersutter, this would result in _all_ draft release notes to be removed; e.g. also draft release notes for hotfixes

cli/githubutil.py
```diff
@@ -140,3 +141,101 @@ def release_note_blocks_cli(
         repository_branch=repository_branch,
         commit_range=commit_range
     ).release_note_blocks()
+
+
+def cleanup_draft_releases(
+    github_cfg_name: str,
+    github_repository_owner: str,
```
ccwienk: nit: it might be more convenient to merge this into one arg (repo_path) so callers need to type less ;-)

cli/githubutil.py
```diff
@@ -140,3 +141,101 @@ def release_note_blocks_cli(
...
+    github_repository_name: str,
+    latest_release_version: str,
```
ccwienk: what does latest release version mean (for the semantics of the command)?

cli/githubutil.py
```diff
@@ -140,3 +141,101 @@ def release_note_blocks_cli(
...
+    github_cfg = ctx().cfg_factory().github(github_cfg_name)
+
+    githubrepobranch = GitHubRepoBranch(
+        github_config=github_cfg,
+        repo_owner=github_repository_owner,
+        repo_name=github_repository_name,
+        branch='master',
+    )
+
+    helper = GitHubRepositoryHelper.from_githubrepobranch(
```
ccwienk: at least name it repo_helper or gh_helper, maybe?

cli/githubutil.py
```diff
@@ -140,3 +141,101 @@ def release_note_blocks_cli(
...
+        githubrepobranch=githubrepobranch,
+    )
+    # prepare semver for later comparison
+    latest_semver = semver.parse_version_info(latest_release_version)
```
ccwienk: greatest. latest implies chronological order (which is not the implemented semantics)

cli/githubutil.py
```diff
@@ -140,3 +141,101 @@ def release_note_blocks_cli(
...
+
+    # we're only interested in releases whose name can be parsed using semver and whose semver
```
ccwienk: are we?

cli/githubutil.py
```diff
@@ -140,3 +141,101 @@ def release_note_blocks_cli(
...
+    # contains the prerelease-label 'draft'.
+    def _is_semver_parseable(release):
+        try:
+            release_semver = semver.parse_version_info(release.name)
+        except Exception:
+            return False
+        if release_semver.prerelease != 'draft':
```
ccwienk: this is bad! the function name somewhat implies it is only checked whether the given str is a valid semver. This is a different semantics. Do not mix this. Or if you do mix, name the function differently

cli/githubutil.py
```diff
@@ -140,3 +141,101 @@ def release_note_blocks_cli(
...
+            return False
+        return True
+
+    releases = helper.repository.releases()
+    # filter out non-draft releases
+    draft_releases = filter(lambda r: r.draft, releases)
```
ccwienk: if you use multiple fitlers in a row, chain them using our fluentiterable. or (if you muist) pydash..

cli/githubutil.py
```diff
@@ -140,3 +141,101 @@ def release_note_blocks_cli(
...
+    # filter out draft-releases whose names aren't parseable
+    semver_releases = filter(_is_semver_parseable, draft_releases)
+
+    # returns a dict containing the release and the version_info object obtained
+    # from parsing it using semver.
+    def _to_release_and_semver(release):
```
ccwienk: this seems a bit cumbersome

cli/githubutil.py
```diff
@@ -140,3 +141,101 @@ def release_note_blocks_cli(
...
+        return {
+            'release': release,
+            'version info': semver.parse_version_info(release.name),
+            }
+
+    # convert every release to dict containing release and semver (to avoid parsing the version
+    # repeatedly). This will be needed more than once, so persist it
+    draft_releases_and_semver = list(map(_to_release_and_semver, semver_releases))
```
ccwienk: use a list comprehension if you need a list anyway. also, if using functional approach, construct processing chains with fluentiterable. otherwise, this just makes things harder to read by adding more boilerplate

cli/githubutil.py
```diff
@@ -140,3 +141,101 @@ def release_note_blocks_cli(
...
+
+    # first, look at 'mainline' releases, i.e. releases with patch-version 0
+    def _is_main_release(release_and_semver):
+        return release_and_semver['version info'].patch == 0
+
+    main_draft_releases = filter(_is_main_release, draft_releases_and_semver)
```
ccwienk: ditto

cli/githubutil.py
```diff
@@ -140,3 +141,101 @@ def release_note_blocks_cli(
...
+    # need to persist this list since we iterate repeatedly over it
+    hotfix_draft_releases = list(filter(_is_hotfix_release, draft_releases_and_semver))
+
+    # prepare a filter function that checks for a given hotfix version that there is no
+    # hotfix release with the same major and minor version, but a more recent
+    # patch version. In effect, this returns true for all but the most recent patch-release of
+    # a 'patch-branch'
+    def _is_outdated_hotfix_release(release_and_semver):
+        release_semver = release_and_semver['version info']
+        for hotfix_release_and_semver in hotfix_draft_releases:
```
ccwienk: do now you start using for-loops? please choose a more consistent style..

version.py
```diff
@@ -13,10 +13,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from collections import defaultdict
 import os
 import semver
 import sys
 
+from typing import Iterable, Set
```
ccwienk: nit: we usually import multiple symbols using 
```
from xxx import (
  symbol1,
  symbol2,
)
```

Also, in particular for reasonably short module names, I would prefer to just import the module name and dereference the required symbols fully qualified.

version.py
```diff
@@ -117,3 +120,12 @@ def find_latest_version_with_matching_major(reference_version: semver.VersionInf
             if not latest_candidate or latest_candidate < candidate:
                 latest_candidate = candidate
     return latest_candidate
+
+
+def partition_by_major_and_minor(
+    versions: Iterable[semver.VersionInfo],
+) -> Iterable[Set[semver.VersionInfo]]:
+    partitions = defaultdict(set)
```
ccwienk: good idea to use defaultdict (which I did not know so far) :-)

version.py
```diff
@@ -117,3 +120,12 @@ def find_latest_version_with_matching_major(reference_version: semver.VersionInf
...
+    for version_info in versions:
+        partitions[f'{version_info.major}.{version_info.minor}'].add(version_info)
```
ccwienk: I think you should rather use tuples as keys. or maybe even semver objects, maybe?

cli/githubutil.py
```diff
@@ -140,3 +142,99 @@ def release_note_blocks_cli(
...
+    github_helper = GitHubRepositoryHelper.from_githubrepobranch(
+        githubrepobranch=githubrepobranch,
+    )
+
+    greatest_semver = semver.parse_version_info(greatest_release_version)
+
+    def _is_autogenerated_draft_release(release):
+        try:
+            release_semver = semver.parse_version_info(release.name)
+        except Exception:
```
ccwienk: please only catch ValueError
also: log a warning (util.warning);

cli/githubutil.py
```diff
@@ -140,3 +142,99 @@ def release_note_blocks_cli(
...
+            return False
+        if release_semver.prerelease != 'draft':
+            return False
+        return True
+
+    autogenerated_draft_releases = [
+        release for release in github_helper.repository.releases()
+        if release.draft and _is_autogenerated_draft_release(release)
```
ccwienk: I'm confused: in `is_autoge..._draft_release`, you check whether release_semver.prerelease == 'draft'. here, it is checked whether release.draft is true-ish. is this redundant? can there be cases where the two checks bear different results? if so, would this not indicate an error condition? either in the cfg, or in our coding?

cli/githubutil.py
```diff
@@ -140,3 +142,99 @@ def release_note_blocks_cli(
...
+    ]
+
+    version_infos = [
+        semver.parse_version_info(release.name) for release in autogenerated_draft_releases
+    ]
+
+    def _process_partition(partition):
```
ccwienk: nit: maybe if would be more expressive to call them `version_partitions` (or sets, maybe?)
also, 'process' is too generic. the name should express what the function does

cli/githubutil.py
```diff
@@ -140,3 +142,99 @@ def release_note_blocks_cli(
...
+        if len(partition) == 1:
+            version_info = partition.pop()
+            if version_info < greatest_semver and version_info.patch == 0:
```
ccwienk: I thought we would not need to check for patch being equal to zero after processing the version partitions?

cli/githubutil.py
```diff
@@ -140,3 +142,99 @@ def release_note_blocks_cli(
...
+                yield version_info
+        else:
+            sorted_partition = sorted(partition, reverse=True)
+            yield from [
+                version_info
+                for version_info in sorted_partition[1:]
```
ccwienk: why not sort the version partitions already in the partitioning function in version.py? either using lists (or rather tuples, presumably), or sorted sets

cli/githubutil.py
```diff
@@ -140,3 +142,99 @@ def release_note_blocks_cli(
...
+                if version_info < greatest_semver
+            ]
+
+    outdated_version_infos = list()
+    for partition in partition_by_major_and_minor(version_infos):
```
ccwienk: could this not be inlined below, where the outdated_draft_releases are collected?

github/util.py
```diff
@@ -53,12 +53,15 @@ def __init__(
         github_config: GithubConfig,
         repo_owner: str,
         repo_name: str,
-        branch: str,
+        branch: str = None,
```
ccwienk: hmm.. starting to think we require an additional type `GitHubRepo` (which GHRB might extend). I don't think it makes much sense to have a type that explicitly states it be a container type whose instances represent "a specific branch of a given GitHub repository" and then allow certain instances to _not_ specify a branch.

ccwienk: it might also be an option to rename it to just `GitHubRepo` and then make the branch name optional.

version.py
```diff
@@ -117,3 +120,12 @@ def find_latest_version_with_matching_major(reference_version: semver.VersionInf
             if not latest_candidate or latest_candidate < candidate:
                 latest_candidate = candidate
     return latest_candidate
+
+
+def partition_by_major_and_minor(
```
ccwienk: a _brief_ docstr would be appropriate, I suppose
also, consider sorting the versions (maybe use order set..)

version.py
```diff
@@ -117,3 +120,12 @@ def find_latest_version_with_matching_major(reference_version: semver.VersionInf
...
+    versions: Iterable[semver.VersionInfo],
+) -> Iterable[Set[semver.VersionInfo]]:
+    partitions = defaultdict(set)
+    for version_info in versions:
+        partitions[f'{version_info.major}.{version_info.minor}'].add(version_info)
+    return partitions.values()
```
ccwienk: hmm.. have you considered returning the dict? OTOH, this way, we do not have to put too much thought into the chosen key, as it is not exposed to callers.

cli/githubutil.py
```diff
@@ -140,3 +140,22 @@ def release_note_blocks_cli(
...
+    github_repository_owner: str,
+    github_repository_name: str,
+    release_names: CliHint(type=str, nargs='+'),
+):
+    github_cfg = ctx().cfg_factory().github(github_cfg_name)
+    githubrepobranch = GitHubRepoBranch(
+        github_config=github_cfg,
+        repo_owner=github_repository_owner,
+        repo_name=github_repository_name,
+        branch=None,
```
ccwienk: rework so we do not require a ghrb object

cli/githubutil.py
```diff
@@ -140,3 +140,22 @@ def release_note_blocks_cli(
         repository_branch=repository_branch,
         commit_range=commit_range
     ).release_note_blocks()
+
+
+def delete_releases(
+    github_cfg_name: str,
+    github_repository_owner: str,
+    github_repository_name: str,
+    release_names: CliHint(type=str, nargs='+'),
```
ccwienk: consider making this consistent with our other coding; we use type=[str] elsewhere so far. this would introduce a new cli-style

cli/githubutil.py
```diff
@@ -142,6 +142,26 @@ def release_note_blocks_cli(
...
+    github_cfg_name: str,
+    github_repository_owner: str,
+    github_repository_name: str,
+):
+    github_cfg = ctx().cfg_factory().github(github_cfg_name)
+    githubrepobranch = GitHubRepoBranch(
+        github_config=github_cfg,
+        repo_owner=github_repository_owner,
+        repo_name=github_repository_name,
+        branch=None,
```
ccwienk: see previous commit. make this work w/o gbrb

cli/githubutil.py
```diff
@@ -142,6 +142,26 @@ def release_note_blocks_cli(
...
+    )
+    github_helper = GitHubRepositoryHelper.from_githubrepobranch(
+        githubrepobranch=githubrepobranch,
+    )
+    release_list = [release for release in github_helper.repository.releases() if release.draft]
```
ccwienk: nit: why not get the names in the list-comprehension already?

PR #223, opened 2019-02-26
add meta resource per job

RaphaelVogel:
creates an own meta resource per job/variant.


test/concourse/steps/notification_test.py
```diff
@@ -87,6 +100,7 @@ def setUp(self):
         self.tmp_dir = tempfile.TemporaryDirectory()
         self.meta_dir = os.path.join(self.tmp_dir.name, 'meta')
         os.mkdir(self.meta_dir)
+        os.environ['META'] = 'meta'
```
ccwienk: this will permanently alter the environment (also for other test executions). it would be better to at least restore the previous state after test execution (maybe in the tearDown method)

PR #228, opened 2019-02-28
Implement dev_cycle_callback

adracus:
add possibility for dev_cycle_callback
propagate to release templates
add tests

**@ colleagues**: PTAL, hope this matches your current coding style.

### DIFF
--- a/concourse/model/traits/release.py
+++ b/concourse/model/traits/release.py
@@ -48,6 +48,16 @@ ATTRIBUTES = (
         ''',
         type=bool,
     ),
+    AttributeSpec.optional(
+        name='next_version_callback',
+        default=None,
+        doc='''
+        an optional callback that is called during next version commit creation.
+        The callback is passed the absolute path to the main repository's work tree via environment
+        variable `REPO_DIR`.
+        Any changes left inside the worktree are added to the resulting release commit.
+        ''',
+    ),
 )
 
 
@@ -69,6 +79,9 @@ class ReleaseTrait(Trait):
 
     def release_callback_path(self):
         return self.raw['release_callback']
+
+    def next_version_callback_path(self):
+        return self.raw['next_version_callback']
 
     def rebase_before_release(self):
         return self.raw['rebase_before_release']
--- a/concourse/steps/release.mako
+++ b/concourse/steps/release.mako
@@ -30,6 +30,7 @@ if has_component_descriptor_trait:
   )
 
 release_callback_path = release_trait.release_callback_path()
+next_version_callback_path = release_trait.next_version_callback_path()
 %>
 
 ${step_lib('release')}
@@ -60,6 +61,9 @@ release_and_prepare_next_dev_cycle(
   % endif
   % if release_callback_path:
   release_commit_callback='${release_callback_path}',
+  % endif
+  % if next_version_callback_path:
+  next_version_callback='${next_version_callback_path}',
   % endif
   rebase_before_release=${release_trait.rebase_before_release()},
   githubrepobranch=githubrepobranch,
--- a/concourse/steps/release.py
+++ b/concourse/steps/release.py
@@ -166,6 +166,7 @@ class ReleaseCommitsStep(TransactionalStep):
         version_operation: str,
         prerelease_suffix: str,
         release_commit_callback: str=None,
+        next_version_callback: str=None,
     ):
         self.git_helper = not_none(git_helper)
         self.repository_branch = not_empty(repository_branch)
@@ -181,6 +182,7 @@ class ReleaseCommitsStep(TransactionalStep):
             repo_dir_absolute,
             repository_version_file_path,
         )
+
         if release_commit_callback:
             self.release_commit_callback = os.path.join(
                 repo_dir_absolute,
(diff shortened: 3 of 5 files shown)


concourse/steps/release.py
```diff
@@ -390,17 +402,49 @@ def validate(self):
...
+        try:
+            # call optional dev cycle commit callback
+            if self.dev_cycle_commit_callback:
+                self._invoke_dev_cycle_commit_callback(
+                    dev_cycle_commit_callback=self.dev_cycle_commit_callback,
+                    repo_dir=self.repo_dir,
+                    release_version=next_cycle_dev_version,
+                )
+
+            dev_cycle_commit = self.github_helper.index_to_commit(
```
ccwienk: almost.. I admit, our API is a bit confusing there. There are two helpers: `GitHubRepositoryHelper` (from github/util.py) and GitHelper (from gitutil.py). The `index_to_commit` method is exposed from `GitHelper`. So you will also need to instantiate a `GitHelper` instance and use it.

nit: also, maybe it would be adequate to mv the callback-invoking into a separate (private) method

adracus: I implemented a method `_invoke_callback` that has the required characteristics.
I also now make use of the `git_helper` in addition.

adracus:
Thanks for having a look! I've addressed the comments.

concourse/steps/release.mako
```diff
@@ -30,6 +30,7 @@ if has_component_descriptor_trait:
   )
 
 release_callback_path = release_trait.release_callback_path()
+dev_cycle_commit_callback = release_trait.dev_cycle_callback_path()
```
RaphaelVogel: This seems not correct.
shouldn't it be `dev_cycle_callback_path` ?

concourse/steps/release.py
```diff
@@ -390,17 +407,49 @@ def validate(self):
     def apply(
         self,
     ):
+        # clean repository if required
+        worktree_dirty = bool(self.github_helper._changed_file_paths())
```
RaphaelVogel: During test I get the error message:
```
WARNING: An error occured while applying step 'Create Next Cycle Commit': 'GitHubRepositoryHelper' object has no attribute '_changed_file_paths'
  File "<string>", line 128, in execute
  File "<string>", line 415, in apply
AttributeError: 'GitHubRepositoryHelper' object has no attribute '_changed_file_paths'
```

concourse/model/traits/release.py
```diff
@@ -48,6 +48,15 @@
         ''',
         type=bool,
     ),
+    AttributeSpec.optional(
+        name='dev_cycle_callback',
```
ccwienk: how about naming this 'next_version_callback' (or next_version_callback_path)? Would that maybe be more intuitive? At least I would not understand the meaning of 'dev_cycle_callback' (I was unhappy with out internal variable names there, before)

concourse/steps/release.py
ccwienk: _changed_file_paths is also a symbol only available from `git_helper` (not github_helper)

concourse/steps/release.py
```diff
@@ -390,17 +407,49 @@ def validate(self):
...
+        if worktree_dirty:
+            self.github_helper._stash_changes()
```
ccwienk: git_helper

concourse/steps/release.py
```diff
@@ -390,17 +407,49 @@ def validate(self):
...
+                )
+
+            dev_cycle_commit = self.git_helper.index_to_commit(
+                message=f'Prepare next dev cycle {next_cycle_dev_version}',
+            )
+
+            dev_cycle_commit_sha = dev_cycle_commit.hexsha
+
+            # forward head to new commit
+            self.github_helper.repo.head.set_commit(dev_cycle_commit_sha)
```
ccwienk: git_helper

concourse/steps/release.py
```diff
@@ -390,17 +407,49 @@ def validate(self):
...
+
+            # Push release commit to remote
+            self.github_helper.push(
```
ccwienk: git_helper

concourse/steps/release.py
```diff
@@ -390,17 +407,49 @@ def validate(self):
...
+                from_ref=dev_cycle_commit_sha,
+                to_ref=self.github_helper.default_branch,
+                use_ssh=True,
+            )
+        finally:
+            if worktree_dirty and self.github_helper._has_stash():
```
ccwienk: git_helper
yes, both our API is confusing in that regard AND it sucks not to have static typing :-/

concourse/steps/release.py
```diff
@@ -390,17 +407,49 @@ def validate(self):
...
+                self.github_helper._pop_stash()
```
ccwienk: git_helper

concourse/model/traits/release.py
adracus: Thanks, I also found the name to be a bit counter-intuitive. I went with your proposal and changed it to `next_version_callback`.

concourse/steps/release.py
adracus: Yup, was related to the `git_helper` `github_helper` issue :sweat_smile:

adracus:
Thanks again for having a look! I addressed your comments.

RaphaelVogel:
Tests fail, see: https://concourse.ci.infra.dev.k8s.ondemand.com/teams/test/pipelines/rv-demo/jobs/master-manuel-release-job/builds/12
Will investigate....

PR #231, opened 2019-03-07
Implement revert method for all steps

AndreasBurger:
This PR implements the revert stubs for all subclasses of TransactionStep in our release trait coding.

It also includes the necessary changes/additions to previously existing functions.


AndreasBurger:
Some test-code was left over, will remove and reopen

PR #233, opened 2019-03-07
Modify list_draft_releases to list outdated draft releases

AndreasBurger:


cli/githubutil.py
```diff
@@ -146,18 +150,110 @@ def list_draft_releases(
...
     github_repository_name: str,
+    greatest_release_version: CliHint(
+        help=(
+            'Greatest version (according to Semver) up to which (exclusive) '
+            'to consider draft releases, if outdated draft releases are to be determined'
+        ),
+    )=None,
+    list_outdated_draft_releases: bool = False,
 ):
+    '''List all draft releases in a GitHub repository. Optionally, attempt to print only outdated
```
ccwienk: nit: optionally filter heuristically for outdated draft releases. The printing itself should always succeed, so stating that we attempt to print is somewhat misleading imo.

cli/githubutil.py
```diff
@@ -146,18 +150,110 @@ def list_draft_releases(
...
+    draft releases.
+
+    This function prints a list of names of draft releases found in the given GitHub repository.
+    If the `--list-outdated-draft-releases` flag is set, the printed list is reduced to outdated
```
ccwienk: please do not add irrelevant implementation details to the documentation. it is irrelevant that the releases are added into a list.

state that the flag will heuristically filter all "still relevant" drafts releases, thus outputting draft releases that are considered to be obsolete (and thus subject to being cleaned up).

cli/githubutil.py
```diff
@@ -146,18 +150,110 @@ def list_draft_releases(
     github_cfg_name: str,
     github_repository_owner: str,
     github_repository_name: str,
+    greatest_release_version: CliHint(
```
ccwienk: would it not suffice to determine the greatest released version (a functionality which we already have ready-to-use)?

cli/githubutil.py
```diff
@@ -146,18 +150,110 @@ def list_draft_releases(
...
+            3a: it is a hotfix draft release AND
+            3b: there is a hotfix draft release of greater version (according to semver)
+                with the same major and minor version
+
+    Hotfix draft release in this context are draft releases with a semver patch version that is
+    not equal to 0.
+    '''
+    if list_outdated_draft_releases and not greatest_release_version:
+        raise ValueError(
+            "'--greatest-release-version' must be given if '--list-outdated-draft-releases' is set"
```
ccwienk: as commented above. why not just omit the greatest-version parameter and determine the greatest released version instead.

cli/githubutil.py
```diff
@@ -146,18 +150,110 @@ def list_draft_releases(
...
+        )
     github_cfg = ctx().cfg_factory().github(github_cfg_name)
     github_helper = GitHubRepositoryHelper(
         owner=github_repository_owner,
         name=github_repository_name,
         github_cfg=github_cfg,
     )
     release_list = [release for release in github_helper.repository.releases() if release.draft]
```
ccwienk: nit: releases

cli/githubutil.py
```diff
@@ -146,18 +150,110 @@ def list_draft_releases(
...
+    if list_outdated_draft_releases:
+        release_list = _outdated_draft_releases(
+            draft_releases=release_list,
+            greatest_release_version=greatest_release_version,
+        )
     for draft_release in release_list:
         print(draft_release.name)
 
 
+def _outdated_draft_releases(
```
ccwienk: i think this function should not reside below `cli/`. the cli module should only contain the required logic to expose cc-utils's API via CLI.
--> move to github/util.py?

version.py
```diff
@@ -117,3 +123,15 @@ def find_latest_version_with_matching_major(reference_version: semver.VersionInf
             if not latest_candidate or latest_candidate < candidate:
                 latest_candidate = candidate
     return latest_candidate
+
+
+def partition_by_major_and_minor(
+    versions: Iterable[semver.VersionInfo],
+) -> Iterable[Set[semver.VersionInfo]]:
+    '''partition an iterable of semver VersionInfos by their joined major and minor version
```
ccwienk: partition an iterable [..] into sorted sequences of VersionInfo, grouped by equal major and minor components. Sorting is done in ascending order, according to semver semantics. Duplicates are discarded.

version.py
```diff
@@ -117,3 +123,15 @@ def find_latest_version_with_matching_major(reference_version: semver.VersionInf
...
+    '''
+    partitions = defaultdict(set)
+    for version_info in versions:
+        partitions[(version_info.major,version_info.minor)].add(version_info)
+    sorted_partitions = {key:sorted(value, reverse=True) for key,value in partitions.items()}
```
ccwienk: as already discussed some days ago: the partitions themselves should be sorted.

cli.py
```diff
@@ -39,12 +39,12 @@ def print_import_errs():
     # Create a custom width formatter by fixing two arguments for the default formatter class,
     # namely 'width' (defaults to 80 - 2) and 'max_help_position' (defaults to 24)
     FORMATTER_CLASS = functools.partial(
-        argparse.HelpFormatter,
```
ccwienk: put this change into a separate PR

ccwienk: also: why?

PR #234, opened 2019-03-14
improve documentation

RaphaelVogel:


concourse/model/traits/notifications.py
```diff
@@ -32,16 +33,39 @@
 )
 
 
-class NotificationTriggeringPolicy(enum.Enum):
+class NotificationTriggeringPolicy(AttribSpecMixin, enum.Enum):
```
ccwienk: as just discussed via slack: extending this enum seems like a clever idea. my only objection would be that - different to all other attrib-specs we so far have - the names / attributes here do not refer to actual sub-attributes, but to values of one common attribute. How does the rendered documentation page look like?

doc/lib/attributes.py
```diff
@@ -76,6 +77,8 @@ def attr_to_table_row(attr_spec, prefix=None):
                 # recurse to child element
                 if issubclass(type_, model.NamedModelElement):
                     child_element = type_(name='<user-chosen>', raw_dict={})
```
ccwienk: as discussed yesterday: there should be no need to instantiate the elements. just use `type_` in all cases

concourse/model/traits/notifications.py
RaphaelVogel: Changed the heading into `Attribute Values` in case of enum

doc/lib/attributes.py
RaphaelVogel: removed instantiation of elements

PR #235, opened 2019-03-22
Report status of publish step to pull requests

AndreasBurger:


concourse/model/step.py
```diff
@@ -156,12 +156,14 @@ def __init__(
         self,
         name,
         is_synthetic=False,
+        always_notify_pull_requests=False,
```
ccwienk: I don't think it is appropriate to add another formal parameter for this case. how about simply re-using the "is_synthetic" flag? That should do what we want, right?

concourse/model/step.py
```diff
@@ -156,12 +156,14 @@ def __init__(
...
         script_type=ScriptType.BOURNE_SHELL,
         *args,
         **kwargs
     ):
         self.name = name
         self.is_synthetic = is_synthetic
+        self.always_notify_pull_requests = always_notify_pull_requests
```
ccwienk: ditto

concourse/model/step.py
```diff
@@ -321,3 +323,6 @@ def __str__(self):
             o=self.outputs(),
         )
         return descr
+
+    def report_status_to_pull_requests(self):
```
ccwienk: ditto

concourse/model/traits/publish.py
```diff
@@ -152,7 +152,12 @@ def __init__(self, trait: PublishTrait, *args, **kwargs):
 
     def inject_steps(self):
         # 'publish' step
-        publish_step = PipelineStep(name='publish', raw_dict={}, is_synthetic=True)
+        publish_step = PipelineStep(
+            name='publish',
+            raw_dict={},
+            is_synthetic=True,
+            always_notify_pull_requests=True,
```
ccwienk: ditto

concourse/model/step.py
```diff
@@ -166,6 +172,7 @@ def __init__(
         self._outputs_dict = {}
         self._inputs_dict = {}
         self._publish_to_dict = {}
+        self.notification_policy = notification_policy
```
ccwienk: please define a getter and make the field "private"

concourse/model/step.py
```diff
@@ -298,6 +305,9 @@ def set_timeout(self, duration_string: str):
     def retries(self):
         return self.raw['retries']
 
+    def notify_pull_requests(self):
```
ccwienk: nit: name it notification_policy and pass the enum value

concourse/model/step.py
```diff
@@ -151,11 +151,17 @@ def attrs(pipeline_step):
...
+    NO_NOTIFICATION = enum.auto()
+    NOTIFY_PULL_REQUESTS = enum.auto()
+
+
 class PipelineStep(ModelBase):
     def __init__(
         self,
         name,
         is_synthetic=False,
+        notification_policy=StepNotificationPolicy.NOTIFY_PULL_REQUESTS,
```
ccwienk: please rm the default value, as discussed. explicit is better than implicit

concourse/templates/default.yaml
```diff
@@ -374,13 +375,20 @@ if not publish_options:
...
+if (
+    notification_policy is not StepNotificationPolicy.NOTIFY_PULL_REQUESTS
+    and notification_policy is not StepNotificationPolicy.NO_NOTIFICATION
+):
+raise ValueError(f'unsupported StepNotificationPolicy: {job_step.notification_policy()}')
+
 %>
 - do:
-% if not job_step.is_synthetic:
+% if job_step.notification_policy() is StepNotificationPolicy.NOTIFY_PULL_REQUESTS:
```
ccwienk: very good. I like the new variant

PR #236, opened 2019-03-26
Model validation

RaphaelVogel:


model/__init__.py
```diff
@@ -191,6 +191,9 @@ def to_module_info(mi):
             kwargs['name'] = cfg_name
 
         element_instance = element_type(**kwargs)
+        validate_function = getattr(element_instance, "validate", None)
```
ccwienk: just call the method. if an instance should not have it, this is an error (which we shall find with the tests you are adding with this change :-))

test/model_validation_test.py
```diff
@@ -15,7 +15,10 @@
 
 import unittest
 
-import model.concourse as examinee
+from model import (
```
ccwienk: as discussed, please adhere to our naming convention for tests. split the tests into separate modules, each named as the module under test with a suffix '_test'. e.g. `model/concourse.py` --> `test/model/concourse_test.py`

test/model_validation_test.py
```diff
@@ -115,6 +118,62 @@ def test_validation_fails_on_missing_key(self):
             with self.subTest(key=key):
                 test_dict = self.raw_dict.copy()
                 test_dict.pop(key)
-                element = examinee.BasicCredentials(test_dict)
+                element = concourse.BasicCredentials(test_dict)
                 with self.assertRaises(ModelValidationError):
                     element.validate()
+
+
+class AwsModelTest(unittest.TestCase):
```
ccwienk: I think it might be less boiler-platy to just use functions to write those tests (as opposed to the old-style junit-like class-based approach)

just use e.g.

```
minimal_dict = {.......}

def test_validation_fails_on_missing_required_key():
  ....
```

model/__init__.py
RaphaelVogel: done

test/model_validation_test.py
RaphaelVogel: done

PR #237, opened 2019-03-26
Add missing documentation

AndreasBurger:


product/scanning.py
```diff
@@ -22,16 +22,50 @@
...
+                doc='''
+                    trigger a (re)scan of container images. Images will be uploaded if they are not
+                    already present.
+                ''',
+                type=str,
+            ),
+            AttributeSpec.optional(
+                name=cls.FORCE_UPLOAD.value,
+                default=False,
+                doc='upload the container images. This will :strong:`always` upload the images.',
```
ccwienk: nit: how about just: "always upload"

product/scanning.py
```diff
@@ -22,16 +22,50 @@
...
+                    upload the container images. This will :strong:`not` upload the images if they
+                    are already present. Scans will be triggered only for images that are uploaded.
+                ''',
+                type=str,
+            ),
+            AttributeSpec.optional(
+                name=cls.RESCAN.value,
+                default=False,
+                doc='''
+                    trigger a (re)scan of container images. Images will be uploaded if they are not
```
ccwienk: nit: unless they are..

ccwienk: also: what precisely was the difference between UPLOAD_IF_CHANGED and RESCAN again? I don't remember. and I do not understand the difference from the (new) documentation tbh

product/scanning.py
```diff
@@ -22,16 +22,50 @@
...
     UPLOAD_IF_CHANGED = 'upload_if_changed'
     RESCAN = 'rescan'
     FORCE_UPLOAD = 'force_upload'
 
+    @classmethod
+    def _attribute_specs(cls):
+        return (
+            AttributeSpec.optional(
+                name=cls.UPLOAD_IF_CHANGED.value,
+                default=True,
```
ccwienk: i'm still not happy with the mis-use of the 'default' attribute for enums

concourse/model/traits/release.py
```diff
@@ -67,10 +85,11 @@ class ReleaseNotesPolicy(enum.Enum):
     ),
     AttributeSpec.optional(
         name='release_notes_policy',
-        default=ReleaseNotesPolicy.DEFAULT,
+        default=ReleaseNotesPolicy.DEFAULT.value,
```
ccwienk: nit: would it not be better to use the "real" value as default value (the enum instance) and do the conversion in the getter exposing the attr? this holds also true for other occurrences..

concourse/model/traits/image_scan.py
```diff
@@ -81,6 +81,7 @@ def exclude_image_references(self):
         name='processing_mode',
         default='upload_if_changed',
         doc='Protecode processing mode', # todo: document allowed values
```
ccwienk: nit: todo is no longer required ;-)

product/scanning.py
AndreasBurger: UPLOAD_IF_CHANGED uploads and, thus, rescans only those images whose image descriptor changed. RESCAN causes all given container images to be rescanned, uploading those that are not available anymore.

I've tried to clarify it a bit.

product/scanning.py
AndreasBurger: Repalced with a lesser mis-use. Probably worth discussing further tomorrow morning.

AndreasBurger: After another look: I guess it's actually used as intended now.

PR #239, opened 2019-03-29
extend whd to prune and restart concourse workers

RaphaelVogel:
In case of problems with a concourse worker, the alertmanager sends a POST
request to the WHD. There it is checked, if a worker is in state 'stalled', and if so
the worker is pruned and the concourse pod is restarted.

### DIFF
--- a/whd/worker.py
+++ b/whd/worker.py
@@ -13,38 +13,18 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import logging
-from flask import Flask
-from flask_restful import Api
+from flask import Response
+from flask_restful import Resource
 
-from .webhook import GithubWebhook
-from .worker import WorkerResurrector
+import concourse.util
 from model.webhook_dispatcher import WebhookDispatcherConfig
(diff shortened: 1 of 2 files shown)


whd/worker.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+class WorkerResurrector(Resource):
+    def __init__(self, whd_cfg: WebhookDispatcherConfig):
+        self.whd_cfg = whd_cfg
+        self.cfg_factory = util.ctx().cfg_factory()
+
+    # called from Prometheus Alert Manager. Indicates that concourse worker(s) got restarted
+    def post(self):
+        for concourse_client, kubernetes_client in self._concourse_and_kubernetes_clients():
```
ccwienk: I think the payload coding should not be part of whd/, but rather be placed into a util module (concourse/..)

RaphaelVogel: done

RaphaelVogel:
decided not to trigger resurrection over WHD

PR #242, opened 2019-04-05
add resurrector

RaphaelVogel:


concourse/util.py
```diff
@@ -90,3 +92,52 @@ def _enumerate_github_org_configs(job_mapping_set: JobMappingSet,):
...
         for github_org_config in github_org_configs:
             yield (github_org_config.org_name(), github_org_config.github_cfg_name())
+
+
+def resurrect_pods(
+    namespace: str,
+    label_selector: str,
+    concourse_client,
+    kubernetes_client,
+):
```
ccwienk: add a few words as to what this function is all about? as in: concourse pods tend to crash and need to be pruned to help with the self-healing. no need to a lengthy essay, but a few hints might be good methinks

concourse/util.py
```diff
@@ -90,3 +92,52 @@ def _enumerate_github_org_configs(job_mapping_set: JobMappingSet,):
...
+                                    f'Container of pod {pod.metadata.name} died, '
+                                    f'try to resurrect'
+                                )
+                                _resurrect_pod(
+                                    concourse_client,
+                                    kubernetes_client,
+                                    pod.metadata.name,
+                                    namespace,
+                                )
+        except Exception:
```
ccwienk: use more specific exception?

landscape_setup/monitoring.py
```diff
@@ -147,3 +163,52 @@ def create_postgresql_helm_values(
...
+            selector=V1LabelSelector(match_labels=labels),
+            template=V1PodTemplateSpec(
+                metadata=V1ObjectMeta(labels=labels),
+                spec=V1PodSpec(
+                    containers=[
+                        V1Container(
+                            image='eu.gcr.io/gardener-project/cc/job-image:latest',
+                            image_pull_policy='Always',
+                            name=name,
+                            command=['python3'],
```
ccwienk: cannot /cc/utils/cli.py be the executable? this would have the slight benefit of not hard-coding it's a python3 application

RaphaelVogel: done

concourse/util.py
RaphaelVogel: done

PR #243, opened 2019-04-15
Automatically set required labels for pull requests created by org members

AndreasBurger:


whd/dispatcher.py
```diff
@@ -107,6 +110,14 @@ def dispatch_pullrequest_event(self, pr_event):
                 concourse_api=concourse_api,
                 event=pr_event,
             ))
+
+            if len(resources) == 0:
+                continue
+
+            if pr_event.action() is PullRequestAction.OPENED:
+                required_labels = {resource.source['label'] for resource in resources}
```
ccwienk: handle case of absent label attr (hint: use dict.get('label', None) + if not None)

ccwienk: also: please mv to separate method to ensure consistent level of abstraction

whd/dispatcher.py
```diff
@@ -122,6 +133,30 @@ def _trigger_resource_check(self, concourse_api, resources):
...
+            github_cfg=github_cfg,
+        )
+        if github_helper.is_pr_created_by_org_member(pr_number):
+            app.logger.info(
+                f"New pull request by member of '{owner}' in '{repository_path}' found. "
+                f"Setting required labels '{labels}'."
+            )
+            github_helper.add_labels_to_pull_request(pr_number, *labels)
+        else:
+            app.logger.info(
```
ccwienk: debug

PR #245, opened 2019-04-15
Also install MitM-Proxy when installing Concourse Landscape

AndreasBurger:
Also requires changes to our config

### DIFF
--- a/landscape_setup/concourse.py
+++ b/landscape_setup/concourse.py
@@ -26,6 +26,8 @@ from subprocess import CalledProcessError
 
 import yaml
 import concourse.client as client
+import semver
+
 from landscape_setup import kube_ctx
 from landscape_setup.utils import (
     ensure_helm_setup,
@@ -43,7 +45,9 @@ from model.concourse import (
 from model.container_registry import (
     GcrCredentials,
 )
-
+from model.proxy import(
+    ProxyConfig
+)
 from model.secrets_server import (
     SecretsServerConfig,
 )
@@ -110,6 +114,36 @@ def create_image_pull_secret(
             namespace=namespace,
             image_pull_secret_name=image_pull_secret_name
         )
+
+
+# Constants related to the MitM-Proxy installation.
+# The name under which the config map will be stored in K8s
+MITM_CONFIG_CONFIGMAP_NAME = 'mitm-config'
+
+
+@ensure_annotations
+def create_proxy_configmaps(
+    proxy_cfg: ProxyConfig,
+    namespace: str,
+):
+    """Create the config map that contains the configuration of the mitm-proxy"""
+    not_empty(namespace)
+
+    ctx = kube_ctx
+    namespace_helper = ctx.namespace_helper()
+    namespace_helper.create_if_absent(namespace)
+
+    config_map_helper = ctx.config_map_helper()
+
+    mitm_proxy_config = proxy_cfg.mitm_proxy().config()
+
+    config_map_helper.create_or_update_config_map(
+        namespace=namespace,
+        name=MITM_CONFIG_CONFIGMAP_NAME,
+        data={
+            'config.yaml': yaml.dump(mitm_proxy_config),
+        }
+    )
 
 
 def create_instance_specific_helm_values(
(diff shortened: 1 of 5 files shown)


landscape_setup/concourse.py
```diff
@@ -43,7 +43,7 @@
 from model.container_registry import (
     GcrCredentials,
 )
-
+from model.proxy import ProxyConfig
```
ccwienk: nit: code style

landscape_setup/concourse.py
```diff
@@ -111,6 +111,38 @@ def create_image_pull_secret(
             image_pull_secret_name=image_pull_secret_name
         )
 
+MITM_CONFIG_CONFIGMAP_NAME = 'mitm-config'
+MITM_SCRIPTS_CONFIGMAP_NAME = 'mitm-scripts'
+MITM_SCRIPT_DIR = '/.mitmscripts'
```
ccwienk: please add a brief comment explaining where said images reside.
also: why not use '/mitmscripts' (root directory)?
also: why use hidden directories?

landscape_setup/concourse.py
```diff
@@ -111,6 +111,38 @@ def create_image_pull_secret(
...
+MITM_CONFIG_DIR = '/.mitmproxy'
+
+
+@ensure_annotations
+def create_proxy_configmaps(
+    proxy_cfg: ProxyConfig,
+    namespace: str,
+):
+    """Create the config map that contains the configuration of the mitm-proxy"""
+    not_none(proxy_cfg)
```
ccwienk: no need to also check for non-none-ness - this is already done by @ensure_annotations

landscape_setup/concourse.py
```diff
@@ -111,6 +111,38 @@ def create_image_pull_secret(
...
+    config_map_helper = ctx.config_map_helper()
+
+    mitm_proxy_config = proxy_cfg.mitm_proxy().config()
+
+    config_yaml = yaml.dump(mitm_proxy_config)
+    config_map_helper.create_or_update_config_map(
+        namespace=namespace,
+        name=MITM_CONFIG_CONFIGMAP_NAME,
+        data={
+            'config.yaml': config_yaml,
```
ccwienk: inline yaml.dump here?

landscape_setup/concourse.py
```diff
@@ -181,6 +213,67 @@ def create_instance_specific_helm_values(
         raise NotImplementedError(
             "Concourse version {v} not supported".format(v=concourse_version)
         )
+    # Add proxy sidecars to instance specific values.
+    # NOTE: Only works for helm chart version 4.2.2 or greater
```
ccwienk: could we replace this with a runtime-check?

landscape_setup/concourse.py
```diff
@@ -181,6 +213,67 @@ def create_instance_specific_helm_values(
...
+        instance_specific_values = add_proxy_values(
+            concourse_cfg=concourse_cfg,
+            config_factory=config_factory,
+            instance_specific_values=instance_specific_values,
+        )
+
+    return instance_specific_values
+
+
+def add_proxy_values(
```
ccwienk: add ensure_annotations?

landscape_setup/concourse.py
```diff
@@ -181,6 +213,67 @@ def create_instance_specific_helm_values(
...
+        }],
+        'securityContext': {
+            'privileged': True,
+        },
+    },{
+        'name': 'mitm-proxy',
+        'image': f'{mitm_cfg.image_name()}:{mitm_cfg.image_tag()}',
+        'env': [{
+                'name': 'CONFIG_DIR',
+                'value': MITM_CONFIG_DIR,
```
ccwienk: if the MITM_CONFIG_DIR variable is only used inside this function's scope, consider moving its definition into the function, as opposed to making it a module variable

landscape_setup/concourse.py
```diff
@@ -181,6 +213,67 @@ def create_instance_specific_helm_values(
...
+        }],
+        'ports': [{
+            'containerPort': mitm_cfg.config()["listen_port"],
+            'hostPort': mitm_cfg.config()["listen_port"],
+            'protocol': 'TCP',
+        }],
+        'volumeMounts': [{
+            'name': 'mitm-config',
```
ccwienk: is not there a (module-"global" variable) for that?

landscape_setup/concourse.py
```diff
@@ -181,6 +213,67 @@ def create_instance_specific_helm_values(
...
+            'mountPath': MITM_CONFIG_DIR,
+        }],
+    }]
+    additional_volumes = [{
+        'name':'mitm-config',
```
ccwienk: ditto

landscape_setup/concourse.py
```diff
@@ -181,6 +213,67 @@ def create_instance_specific_helm_values(
...
+        'configMap': {'name': MITM_CONFIG_CONFIGMAP_NAME},
+    }]
+    # add new values to dict without replacing existing ones
+    vals = instance_specific_values.get('worker', {})
+    vals.update(
```
ccwienk: why not just assign the attributes (vals['sidecarContainers'] = sidecar_containers..)

landscape_setup/concourse.py
```diff
@@ -181,6 +213,67 @@ def create_instance_specific_helm_values(
...
+        {
+            'sidecarContainers': sidecar_containers,
+            'additionalVolumes': additional_volumes,
+        }
+    )
+    instance_specific_values['worker']= vals
```
ccwienk: nit: missing whitespace around eq

model/proxy.py
```diff
@@ -0,0 +1,54 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from model.base import (
+    ModelBase,
+)
+
+
+class DockerImageConfig(ModelBase):
```
ccwienk: nit: containerImage?

model/proxy.py
```diff
@@ -0,0 +1,54 @@
...
+            return super._required_attributes() + [
+                'image_name',
+                'image_tag',
+            ]
+
+    def image_name(self):
+        return self.raw['image_name']
+
+    def image_tag(self):
+        return self.raw['image_tag']
```
ccwienk: nit: since it's required on the consumer-side anyways, why not add a getter for 'image_reference' (returning <name>:<tag>)?

model/proxy.py
```diff
@@ -0,0 +1,54 @@
...
+    def _required_attributes(self):
+            return super._required_attributes() + [
+                'mitm_proxy',
+                'setup_iptables',
+            ]
+
+    def mitm_proxy(self):
+        return MitmProxyConfig(raw_dict=self.raw['mitm_proxy'])
+
+    def setup_iptables(self):
```
ccwienk: could you find a better name?

model/proxy.py
```diff
@@ -0,0 +1,54 @@
...
+            ]
+
+    def image_name(self):
+        return self.raw['image_name']
+
+    def image_tag(self):
+        return self.raw['image_tag']
+
+
+class ProxyConfig(ModelBase):
```
ccwienk: it's a bit unfortunate that we have a "proxyConfig", which in turn bears a "mitmProxyConfig". at least add a brief comment on what the proxy-cfg is for? or maybe think of a better name. maybe networkpolicy-cfg. oder just network-cfg?

model/proxy.py
```diff
@@ -0,0 +1,54 @@
...
+    def mitm_proxy(self):
+        return MitmProxyConfig(raw_dict=self.raw['mitm_proxy'])
+
+    def setup_iptables(self):
+        return DockerImageConfig(raw_dict=self.raw['setup_iptables'])
+
+
+class MitmProxyConfig(DockerImageConfig):
+    def _required_attributes(self):
+        return super._required_attributes() + ['config']
```
ccwienk: nit: maybe it would be the better pattern to `yield from` super()._required_attributes() and than just add another yield for the additional attrs.

PR #258, opened 2019-05-16
Make posting of release notes to Slack its own transaction

AndreasBurger:
Prevents the removal of all release notes if only the act of posting to Slack fails.

### DIFF
--- a/concourse/steps/release.py
+++ b/concourse/steps/release.py
@@ -25,11 +25,11 @@ from github.util import (
 )
 import product.model
 from github.release_notes.util import (
-    fetch_release_notes,
-    post_to_slack,
     delete_file_from_slack,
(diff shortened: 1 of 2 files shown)


concourse/steps/release.py
```diff
@@ -679,16 +684,23 @@ def release_and_prepare_next_dev_cycle(
         cleanup_draft_releases_step,
     ]
 
+    release_notes_transaction = Transaction(*release_notes_steps)
+    release_notes_transaction.validate()
+    if not release_notes_transaction.execute():
+        raise RuntimeError('An error occurred while publishing the release notes.')
+
     if slack_cfg_name and slack_channel:
```
ccwienk: nit: could we do an early exit here?

PR #274, opened 2019-05-27
Refactor Image Scan Trait: Separate vulnerability and AV scanning

AndreasBurger:


concourse/steps/scan_container_images.py
```diff
@@ -84,12 +84,11 @@ def has_results(self):
...
         parts = []
         parts.append(self._mail_disclaimer())
-        parts.append(protecode_results_table(
-            protecode_cfg=self._protecode_cfg,
-            upload_results=self._protecode_results,
-            )
-        )
-        parts.append(self._clamav_report())
+
+        if self._protecode_results:
```
ccwienk: that's something I did think about (but not yet implement): we need to take into account whether or not a scan has been done.

id est: if both protecode and clamav scans have been done. protecode yielded some results, but no malware was found. then there should be a part in the mail that states about the successful virus scan. OTOH, if only protecode scans have been done, it should _not_ be (falsely) stated that a clamav scan was done. Compare to my remark about applying a factory-ish pattern

concourse/steps/scan_container_images.py
```diff
@@ -211,3 +213,23 @@ def result_to_tuple(upload_result: UploadResult):
       tablefmt='html',
     )
     return table
+
+
+def create_license_report(license_report):
```
ccwienk: separate to own commit - feel free to merge to master already (the entire move, of course)

concourse/model/traits/image_scan.py
```diff
@@ -143,6 +144,12 @@ class Notify(enum.Enum):
         type=ProtecodeScanCfg,
         doc='if present, perform protecode scanning',
     ),
+    AttributeSpec.optional(
+        name='clam_av',
```
ccwienk: I'd prefer for `clam_av` to also feature configuration options (similar to protecode)

ccwienk: as discussed (a bit earlier today), use `scanning_mode` or sth. similar, which should be an enum, with one value 'DEFAULT/default', which should be the default (for now)

concourse/model/traits/image_scan.py
```diff
@@ -161,11 +170,22 @@ def email_recipients(self):
...
+
+    def clam_av(self):
+        return self.raw['clam_av']
 
     def transformer(self):
         return ImageScanTraitTransformer(trait=self)
 
+    def validate(self):
+        super().validate()
+        if not (self.protecode() or self.clam_av()):
```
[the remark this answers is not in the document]
AndreasBurger: For _exactly_ 2 values, I prefer this syntax

concourse/steps/scan_container_images.mako
```diff
@@ -121,29 +110,33 @@ if images_with_potential_viruses:
...
+if not protecode_results and not images_with_potential_viruses:
   sys.exit(0)
+
 email_recipients = ${image_scan_trait.email_recipients()}
 
 email_recipients = tuple(
   mail_recipients(
     notification_policy='${image_scan_trait.notify().value}',
     root_component_name='${component_trait.component_name()}',
+% if protecode_scan:
```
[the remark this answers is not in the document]
AndreasBurger: as discussed, this will be done in a later pr

concourse/model/traits/image_scan.py
AndreasBurger: As `clam_av` now has a configuration (albeit one whose values are not used) I've opted not to create a default scanning for now.

concourse/model/traits/image_scan.py
```diff
@@ -72,7 +73,7 @@
     AttributeSpec.optional(
         name='protecode_cfg_name',
         default=None,
-        doc='protecode cfg name to use (see cc-utils)',
+        doc='protecode cfg name to use (see cc-config)',
```
ccwienk: good catch

model/clamav.py
```diff
@@ -0,0 +1,23 @@
...
+# limitations under the License.
+
+from model.base import NamedModelElement
+
+
+class ClamAVConfig(NamedModelElement):
+    '''A stub for the ClamAV config.
+
+    Not intended to be instantiated by users of this module
+    '''
```
ccwienk: I think you need to add a `pass` here.

PR #279, opened 2019-05-31
make set-teams work with concourse 5.2

RaphaelVogel:


concourse/client/__init__.py
```diff
@@ -50,7 +52,7 @@
 
 [0] https://github.com/concourse/concourse/issues/1122
 [1] https://concourse.ci
-[2] https://github.com/concourse/atc/blob/master/routes.go
+[2] https://github.com/concourse/concourse/blob/master/atc/routes.go
```
ccwienk: good catch

concourse/client/api.py
```diff
@@ -83,9 +82,21 @@ def _post(self, url: str, body: str="", headers={}):
     def _delete(self, url: str):
         return self.request_builder.delete(url)
 
-    @abstractmethod
-    def login(self, team: str, username: str, passwd: str):
-        raise NotImplementedError
+    def login(self, username: str, passwd: str):
```
ccwienk: so this is the same for v4 and v5?
for the sake of focussing this PR's scope, could you mv this refactoring to a separate commit (may be pushed to master w/o review)

PR #291, opened 2019-06-18
Add filter config to mitm config

AndreasBurger:


model/proxy.py
```diff
@@ -86,8 +86,31 @@ def _required_attributes(self):
...
+
+
+class MitmFilterConfig(ModelBase):
+    def _required_attributes(self):
+        yield from super()._required_attributes()
+        yield from [
+            'allow',
+        ]
+
+    def allowed_hosts(self):
```
ccwienk: how about "whitelisted_hosts" + add getter for "blacklisted_hosts", please

model/proxy.py
```diff
@@ -86,8 +86,31 @@ def _required_attributes(self):
...
+        if not config:
+            return None
+        return MitmFilterConfig(config)
+
+
+class MitmFilterConfig(ModelBase):
+    def _required_attributes(self):
+        yield from super()._required_attributes()
+        yield from [
+            'allow',
```
ccwienk: whitelisted_hosts? (+blacklisted_hosts')

ccwienk: also: I think this attribute should be optional (empty set / tuple if not configured)

model/proxy.py
```diff
@@ -86,8 +86,31 @@ def _required_attributes(self):
...
         ]
 
+    def _optional_attributes(self):
+        yield from super()._optional_attributes()
+        yield from [
+            'filter_config',
+        ]
+
     def config(self):
         return self.raw['config']
```
ccwienk: what's `config`?

model/proxy.py
```diff
@@ -86,8 +86,31 @@ def _required_attributes(self):
             'logging',
         ]
 
+    def _optional_attributes(self):
+        yield from super()._optional_attributes()
+        yield from [
+            'filter_config',
```
ccwienk: might it be enough to inline the one (or two) attrs of filter_config?

AndreasBurger: For consistency and (admittedly a _very_ small concern) extensibility I'd rather not.

PR #295, opened 2019-06-25
refactor: split cli-parts of config module to cli/

ccwienk:
Cleanup some old legacy (hacky symlink).

- split `config` module into two parts:
  - toplevel module containing secretsServerClient
  - cli/ module containing exposed command line tools

I think SecretsServerClient should not be used anywhere outside our codebase, so in a subsequent cleanup step, the toplevel module `config` could be resolved entirely (either moved into `ccc`, or into a more appropriate location).


ccwienk:
note: e2e-test before merging (_after_ rolling out new release of cc-job-image)

PR #298, opened 2019-06-27
Move els-logging to daemon-thread

AndreasBurger:


http_requests.py
```diff
@@ -75,41 +77,71 @@ def mount_default_adapter(
     return session
 
 
+def _create_logging_queue():
+    '''Set up a logging thread and return the queue used for communicating with it.
+    '''
+    config_set_name = 'internal_active'
```
ccwienk: do not hard-code cfg set name

http_requests.py
```diff
@@ -75,41 +77,71 @@ def mount_default_adapter(
...
+    els_index_name = 'github_access_stacktrace'
+    logging_queue = queue.SimpleQueue()
+    els_logging_client = _create_logging_client(config_set_name)
+    thread = threading.Thread(
```
ccwienk: queue and thread ought to be module attrs (either safeguard the queue with a lock, or initialise it immediately)

http_requests.py
```diff
@@ -75,41 +77,71 @@ def mount_default_adapter(
...
+        name='els_logging_daemon',
+        daemon=True,
+        target=_log_entry_consumer,
+        args=(logging_queue, els_logging_client, els_index_name),
```
ccwienk: use kwargs

http_requests.py
```diff
@@ -75,40 +77,70 @@ def mount_default_adapter(
...
+    config_set_name = util.check_env('CONCOURSE_CURRENT_CFG')
+    config_set = ctx().cfg_factory().cfg_set(config_set_name)
+    try:
+        elastic_cfg = config_set.elasticsearch()
+    except KeyError:
+        return None
+    return ccc.elasticsearch.from_cfg(elasticsearch_cfg=elastic_cfg)
+
+
+def _log_entry_consumer(
```
ccwienk: naming nit: this is a function, not a class. I think functions ought rather be named with verbs as opposed to substantices (which are conventionally used for class names or instances)

http_requests.py
```diff
@@ -75,40 +77,70 @@ def mount_default_adapter(
...
+                elastic_client.store_document(
+                    index=elastic_index_name,
+                    body=log_item,
+                )
+            except Exception as e:
+                info(f'Could not log stack trace information: {e}')
+
+
+if util._running_on_ci():
+    ELS_LOGGING_INDEX_NAME = 'github_access_stacktrace' # TODO: Make configurable
```
ccwienk: still, this is not what I meant with my previous comment. also, why UPPERCASE?

please first create a separate commit / PR that factors out the creation / retrieval of elasticsearch client into a separate function, then update this PR so that it only introduces the asynchronous processing

http_requests.py
```diff
@@ -75,40 +77,70 @@ def mount_default_adapter(
     return session
 
 
+def _create_logging_client():
+    '''Create an ELS Client for the default els in the active config set and return it
```
ccwienk: nit: the first sentence is not correct. and it should not be correct. and thus the name is not well-chosen.

what this function _should_ do is return (and if necessary create beforehand) the singleton instance of the logging client or `None` if no such client is available (e.g. because execution environment is not central CI/CD).

on the other hand, this function - albeit private - should not be used by anyone but the consumer thread anyways, so I do not consider it a good idea to implement it in the first place

starting to look in more detail in this PR. why is this code added to `http_request.py`? this is not the right place at all! the asynchronous handling should of course be done (in a transparent manner) in the elasticsearch client (`ccc.elasticsearch`, I think?)

PR #300, opened 2019-07-04
Prepare routes for future ClamAV client

AndreasBurger:


clamav/client/routes.py
```diff
@@ -0,0 +1,48 @@
...
+    '''
+    @ensure_annotations
+    def __init__(self, base_url: str):
+        '''
+        @param base_url: the ClamAV cluster-service URL
+        '''
+        self.base_url = base_url
+
+    def _api_url(self, *parts, **kwargs):
+        return urljoin(self.base_url, '/'.join(parts))
```
ccwienk: you do not need to join '/'. just pass the parts. and use util.urljoin, pls.

PR #304, opened 2019-07-05
Do not use uppercase for boolean values

RaphaelVogel:


concourse/resources/github.mako
```diff
@@ -26,9 +26,9 @@ disable_tls_validation = not github_cfg.tls_validation()
   source:
     uri: ${github_cfg.ssh_url()}/${repo_cfg.repo_path()}
     branch: ${repo_cfg.branch()}
-    disable_ci_skip: ${repo_cfg.disable_ci_skip()}
-    skip_ssl_verification: ${disable_tls_validation}
-    no_ssl_verify: ${disable_tls_validation}
+    disable_ci_skip: ${'true' if repo_cfg.disable_ci_skip() else 'false'}
```
ccwienk: why? yaml permits the upper-cased variant: https://yaml.org/type/bool.html
at least, use `json.dumps`. but then again: why?

RaphaelVogel:
Searching for root cause of not working [skip ci]. This https://github.com/concourse/git-resource/blob/master/assets/check#L77
make me wonder if this could be the reason for problems (check for lowercase "true")

PR #305, opened 2019-07-09
no resource check if 'skip ci' in commit msg

RaphaelVogel:
in case `[skip-ci]` or `[ci skip]` is in commit message, we do not want to check the resource.
We only check the resource, if `disable_ci_skip` is set to True

### DIFF
--- a/whd/dispatcher.py
+++ b/whd/dispatcher.py
@@ -183,6 +183,13 @@ class GithubWebhookDispatcher(object):
             if isinstance(event, PushEvent):
                 if not event.ref().endswith(ghs.branch_name()):
                     continue
+            # TODO: remove for Concourse 6.0 see https://github.com/concourse/concourse/issues/3463
+            if any(skip in event.commit_message() for skip in ('[skip ci]', '[ci skip]')):
+                if not ghs.disable_ci_skip():
+                    app.logger.info(
+                        f"Do not trigger resource {resource.name}. Found [skip ci] or [ci skip]"
+                    )
+                    continue
 
             yield resource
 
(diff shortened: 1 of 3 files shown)


whd/dispatcher.py
```diff
@@ -175,6 +175,11 @@ def _matching_resources(self, concourse_api, event):
             if not resource.type == resource_type:
                 continue
             ghs = resource.github_source()
+            if any(skip in event.commit_message() for skip in ('[skip ci]', '[ci skip]')):
```
ccwienk: consider using `reutil` (from cc-utils)

whd/dispatcher.py
```diff
@@ -175,6 +175,11 @@ def _matching_resources(self, concourse_api, event):
...
+                if not ghs.disable_ci_skip():
+                    continue
+                if ghs.disable_ci_skip() in ('False', 'false'):
```
ccwienk: use `yaml.safe_load`. preferrable do the parsing in the getter method and return a boolean value from it (this is what the model classes should do..)

whd/model.py
```diff
@@ -67,6 +67,11 @@ def modified_paths(self):
             return ()
         yield from head_commit.get('modified', ())
 
+    def commit_message(self):
+        head_commit = self.raw.get('head_commit', None)
```
ccwienk: i think this is too defensive. fail if no head_commit is available

ccwienk:
almost forgot: add `TODO / FIXME` in code stating that this special handling should be removed with the advent of concourse 6, please

PR #306, opened 2019-07-09
Add clamav client to cc-clients module

AndreasBurger:


ccc/clamav.py
```diff
@@ -0,0 +1,80 @@
...
+
+    @staticmethod
+    def from_config_name(
+        clamav_config_name: str,
+    ):
+        cfg_factory = util.ctx().cfg_factory()
+        clamav_config = cfg_factory.clamav(clamav_config_name)
+        return ClamAVClient(clamav_config=clamav_config)
+
+    def scan_container_image(self, image_reference: str) -> typing.Tuple[ClamAVScanResult, str]:
```
ccwienk: i think this method should either be mv'ed to clamav.client.api (which might be better-named client). ccc should just contain factory functions/methods

although admittedly, we are not 100% consistent in this regard. e.g. elasticsearch and secrets-server both "live" here. but otoh they are fairly small and do not have additional modules.

clamav/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import util
+
+from model.clamav import ClamAVConfig
+from .client import ClamAVClient
+from .routes import ClamAVRoutes
+
+
+def client_from_config(clamav_config: ClamAVConfig):
```
ccwienk: mv to ccc.clamav?

clamav/client.py
```diff
@@ -61,3 +66,37 @@ def health(self):
         url = self.routes.health()
         response = self._request(self._session.get, url)
         return ClamAVHealth(response.json())
+
+    def scan_container_image(self, image_reference: str) -> typing.Tuple[ClamAVScanResult, str]:
```
ccwienk: nit/tip:
```
FilePath = str
...
typing.Tuple[ClamAVScanResult, FilePath]
```
would make the signature more expressive.

as an alternative, why not augment the ClamAvScanResult obj with the path? We do not need to restrict our model classes to be 100% identical to ClamAV's raw data

clamav/client.py
```diff
@@ -61,3 +66,37 @@ def health(self):
...
+        '''Fetch and scan the container image with the given image reference using ClamAV
+
+        Returns a tuple containing ClamAVs results and a string specifying the file which contained
+        the finding.
+        '''
+        logger.debug(f'scanning container image {image_reference}')
+        with tarfile.open(
```
ccwienk: this is copy-paste from clamav.util. please do not copy-paste. refactor the function instead. either pass-in the scanning function, or yield the objects to scan from it

clamav/client.py
```diff
@@ -61,3 +66,37 @@ def health(self):
...
+                            continue
+                        else:
+                            # early exit on first match
+                            return (
+                                scan_result,
+                                f'{ti.name}:{inner_ti.name}'
+                            )
+                logger.debug(f'{image_reference}:{ti.name} looks clean')
+            logger.debug(f'image looked clean: {image_reference}')
+            return None, None # no match
```
ccwienk: omit

model/clamav.py
```diff
@@ -36,8 +36,12 @@ def service_name(self):
     def service_port(self):
         return self.raw.get('service_port')
 
-    def cluster_url(self):
-        return f'{self.service_name}.{self.namespace()}.svc.cluster.local:{self.service_port()}'
+    def service_url(self):
+        return 'http://{service_name}.{namespace}.svc.cluster.local:{port}'.format(
```
ccwienk: use urlunparse

clamav/model.py
```diff
@@ -87,7 +88,17 @@ def kernel_info(self) -> str:
...
 
 
-class ClamAVScanResult(ModelBase):
+class ScanResult(object, metaclass=abc.ABCMeta):
+    @abc.abstractmethod
+    def malware_detected(self) -> bool:
+        pass
+
+    @abc.abstractmethod
+    def finding(self):
```
ccwienk: let's call this e.g. "virus_signature"

clamav/util.py
```diff
@@ -57,22 +63,43 @@ def clamd_client():
...
-    return status, signature_or_none
+
+    return LegacyScanResult(
+        status=status,
+        signature=signature_or_none,
+    )
 
 
-def scan_container_image(image_reference: str):
+def scan_container_image(
```
ccwienk: please rewrite in such a way that this function remains as it is, but split out the part of retrieving container image + layers; yield layers + path; this function is to be re-used. the actual scanning part to be "duplicated"
no need to introduce the result objects for the (soon-to-be) legacy code

clamav/model.py
```diff
@@ -97,10 +97,10 @@ def encrypted_content_detected(self) -> bool:
     def scan_size_octet(self) -> int:
         return self.raw['scanSize']
 
-    def finding(self) -> str:
-        '''Return a string describing ClamAV's findings (e.g.: "Eicar-Test-Signature")
+    def virus_signature(self):
+        '''Return a string describing ClamAV's findings, if any (e.g.: "Eicar-Test-Signature")
```
ccwienk: nit: document return value in case no virus was found

clamav/util.py
```diff
@@ -70,32 +71,47 @@ def scan_stream(fileobj):
     return status, signature_or_none
 
 
-def scan_container_image(image_reference: str):
-    logger.debug(f'scanning container image {image_reference}')
+def yield_files_from_container_image(
```
ccwienk: please do not name functions after implementation details. stick with more conventional and abstract names.

For example, `iter` or `enumerate` would be more conventional prefixes. maybe `iter_files`.

clamav/util.py
```diff
@@ -70,32 +71,47 @@ def scan_stream(fileobj):
...
+    container_image_reference: str,
+) -> typing.Tuple[typing.IO, str]:
+    '''Yields tuples of file and path from the container image
```
ccwienk: please do not just re-iterate the keywords used in the function. this is not helping in comparison to just reading the source code.

instead, please document the abstract semantics.

which are:
- as input, a container image reference is accepted
- the container image is retrieved
- the contained layers are extracted in the order they are contained
- for each layer, two-tuples of contained file paths + file-like objects are returned (as a generator / iterable)

side note: it would also be acceptable for me to not add documentation. but if we do, i consider it very important for the documentation to be correct, precise, and actually add value compared to if the documentation were omitted.

clamav/util.py
```diff
@@ -70,32 +71,47 @@ def scan_stream(fileobj):
...
+    The paths yielded are relative paths from inside the container image.
+    '''
     with tarfile.open(
         mode='r|',
-        fileobj=container.registry.retrieve_container_image(image_reference)
-    ) as tf:
-        for ti in tf:
+        fileobj=container.registry.retrieve_container_image(container_image_reference)
+    ) as tar_file:
+        for file_object in tar_file:
```
ccwienk: NO! please read the documentation on `tarfile` first. the object returned is a `TarInfo` object. _NOT_ a FileObject. this is misleading!

clamav/client.py
```diff
@@ -61,3 +62,14 @@ def health(self):
         url = self.routes.health()
         response = self._request(self._session.get, url)
         return ClamAVHealth(response.json())
+
+    def scan_container_image(self, image_reference: str):
+        '''Fetch and scan the container image with the given image reference using ClamAV
+        '''
+        logger.debug(f'scanning container image {image_reference}')
+        for content, relative_image_path in iter_container_image(image_reference):
```
ccwienk: path

clamav/client.py
```diff
@@ -61,3 +62,14 @@ def health(self):
...
+            scan_result = self.scan(image_reference)
+            if not scan_result.malware_detected():
+                continue
+            else:
+                return scan_result, relative_image_path
```
ccwienk: consider yielding

clamav/util.py
```diff
@@ -70,32 +71,43 @@ def scan_stream(fileobj):
     return status, signature_or_none
 
 
-def scan_container_image(image_reference: str):
-    logger.debug(f'scanning container image {image_reference}')
-
+def iter_container_image(
```
ccwienk: image_files?

clamav/util.py
```diff
@@ -70,32 +71,43 @@ def scan_stream(fileobj):
     return status, signature_or_none
 
 
-def scan_container_image(image_reference: str):
-    logger.debug(f'scanning container image {image_reference}')
-
+def iter_image_files(
+    container_image_reference: str,
+) -> typing.Tuple[typing.IO, str]:
```
ccwienk: it is not _one_ tuple, but an iterable yielding such-typed tuples. if adding type hints, please be precise. otherwise, they are misleading and harm more than help.

model/clamav.py
```diff
@@ -36,8 +37,15 @@ def service_name(self):
     def service_port(self):
         return self.raw.get('service_port')
 
-    def cluster_url(self):
-        return f'{self.service_name}.{self.namespace()}.svc.cluster.local:{self.service_port()}'
+    def service_url(self):
```
ccwienk: please separate this commit from this PR, and merge it to master

clamav/client/api.py
```diff
@@ -33,7 +33,7 @@ class ClamAVApi(object):
     @ensure_annotations
     def __init__(
         self,
-        routes:ClamAVRoutes,
+        routes: ClamAVRoutes,
```
ccwienk: please separate this commit from this PR; merge it to master. cosmetic changes like that do not need to be reviewed. we have a linter for that

clamav/client.py
```diff
@@ -27,9 +27,7 @@
 )
 
 
-class ClamAVApi(object):
-    '''Implements ClamAV REST API functionality.
-    '''
+class ClamAVClient(object):
```
ccwienk: please push this commit to master. or sqash into appropriate commit from this PR

PR #313, opened 2019-07-15
Consider GitHub instance when determining target team

AndreasBurger:
Fixes the issue that the webhook dispatcher may deploy Pipelines to the wrong Team, if multiple JobMappings are configured

### DIFF
--- a/concourse/enumerator.py
+++ b/concourse/enumerator.py
@@ -267,15 +267,37 @@ class GithubRepositoryDefinitionEnumerator(GithubDefinitionEnumeratorBase):
         self.cfg_set = not_none(cfg_set)
         concourse_cfg = cfg_set.concourse()
         job_mapping_set = cfg_set.job_mapping(concourse_cfg.job_mapping_cfg_name())
-
-        # hack: use first mapping that matches
         org = self._repository_url.path.lstrip('/').split('/')[0]
+        self.job_mapping = self._find_matching_job_mapping(
+            job_mapping_set,
+            org,
+            self._repository_url.hostname,
+        )
+
+
+    def _find_matching_job_mapping(self,
+        job_mapping_set,
+        github_org_name: str,
+        repository_hostname: str,
+    ):
+        matching_mapping = None
         for job_mapping in job_mapping_set.job_mappings().values():
-            if org in [o.org_name() for o in job_mapping.github_organisations()]:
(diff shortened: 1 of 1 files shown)


concourse/enumerator.py
```diff
@@ -265,6 +265,11 @@ class GithubRepositoryDefinitionEnumerator(GithubDefinitionEnumeratorBase):
     def __init__(self, repository_url:str, cfg_set):
         self._repository_url = urlparse(not_none(repository_url))
         self.cfg_set = not_none(cfg_set)
+        self.github_cfg = github_cfg_for_hostname(
```
ccwienk: please separate refactoring commit from this PR

concourse/enumerator.py
```diff
@@ -267,15 +267,23 @@ def __init__(self, repository_url:str, cfg_set):
         self.cfg_set = not_none(cfg_set)
         concourse_cfg = cfg_set.concourse()
         job_mapping_set = cfg_set.job_mapping(concourse_cfg.job_mapping_cfg_name())
-
-        # hack: use first mapping that matches
+        github_cfg = github_cfg_for_hostname(
```
ccwienk: this will not work. do not lookup githubcfg from hostname. use the github-cfgs from the job-mappings

concourse/enumerator.py
```diff
@@ -267,15 +267,42 @@ def __init__(self, repository_url:str, cfg_set):
...
         job_mapping_set = cfg_set.job_mapping(concourse_cfg.job_mapping_cfg_name())
-
-        # hack: use first mapping that matches
         org = self._repository_url.path.lstrip('/').split('/')[0]
-        for job_mapping in job_mapping_set.job_mappings().values():
-            if org in [o.org_name() for o in job_mapping.github_organisations()]:
-                self.job_mapping = job_mapping
-                break
-        else:
+        matching_mappings = list(self._find_matching_job_mappings(
```
ccwienk: as discussed yesterday, it makes not too much sense to use a generator, just to put it into a list at the only user

consider the following pattern:
```
def find_element(condition):
  candidate = None
  for element in elements:
    if condition(element):
      if candiate:
        # signal error - not unambigous
      candidate = element
  return candiate # optionally check if found and raise, depending on your coding style preference
```

PR #314, opened 2019-07-17
Lookup target team

ccwienk:
see commit msg

### DIFF
--- a/concourse/enumerator.py
+++ b/concourse/enumerator.py
@@ -264,16 +264,20 @@ class GithubDefinitionEnumeratorBase(DefinitionEnumerator):
 class GithubRepositoryDefinitionEnumerator(GithubDefinitionEnumeratorBase):
     def __init__(self, repository_url:str, cfg_set):
         self._repository_url = urlparse(not_none(repository_url))
+        self._repo_host = self._repository_url.hostname
         self.cfg_set = not_none(cfg_set)
         concourse_cfg = cfg_set.concourse()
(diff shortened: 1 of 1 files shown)


concourse/enumerator.py
```diff
@@ -264,16 +264,18 @@ def _scan_repository_for_definitions(
...
         concourse_cfg = cfg_set.concourse()
         job_mapping_set = cfg_set.job_mapping(concourse_cfg.job_mapping_cfg_name())
 
-        # hack: use first mapping that matches
         org = self._repository_url.path.lstrip('/').split('/')[0]
         for job_mapping in job_mapping_set.job_mappings().values():
             if org in [o.org_name() for o in job_mapping.github_organisations()]:
-                self.job_mapping = job_mapping
-                break
+                github_cfg = cfg_set.github(org.org_name())
```
AndreasBurger: This does not work, as org is a string here.

Only way I could come up with is something akin to:

```
        for job_mapping in job_mapping_set.job_mappings().values():
            for organisation in job_mapping.github_organisations():
                if org != organisation.org_name():
                    continue
                github_cfg = cfg_set.github(organisation.github_cfg_name())
                if github_cfg.matches_hostname(self._repo_host):
                    self.job_mapping = job_mapping
                    break
```
which, of course requires its own function due to the nested loops.

ccwienk:
sry, I accidentally closed the PR

concourse/enumerator.py
ccwienk: not so - see my amended commit

PR #315, opened 2019-07-18
Implement SSE workaround for ClamAV scanning

AndreasBurger:
Due to short connection-timeouts we need to keep sending SSE to keep the connection alive when scanning large files.

This PR adds support for the SSE sent by our ClamAV service.

### DIFF
--- a/clamav/client.py
+++ b/clamav/client.py
@@ -21,10 +21,11 @@ from .routes import ClamAVRoutes
 from clamav.util import iter_image_files
 
 from .model import (
+    ClamAVHealth,
     ClamAVInfo,
     ClamAVMonitoringInfo,
+    ClamAVScanEventClient,
     ClamAVScanResult,
-    ClamAVHealth,
 )
 
 logger = logging.getLogger(__name__)
@@ -57,6 +58,13 @@ class ClamAVClient(object):
         url = self.routes.scan()
         response = self._request(self._session.post, url=url, data=data)
         return ClamAVScanResult(response.json())
+
+    def sse_scan(self, data):
+        url = self.routes.sse_scan()
+        client = ClamAVScanEventClient(
+            self._request(self._session.post, url=url, data=data, stream=True)
+        )
+        return client.process_events()
 
     def health(self):
         url = self.routes.health()
--- a/clamav/model.py
+++ b/clamav/model.py
@@ -14,6 +14,8 @@
 # limitations under the License.
 
 import enum
+import json
+import sseclient
 import typing
 
 from model import ModelBase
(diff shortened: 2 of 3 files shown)


clamav/client.py
```diff
@@ -25,6 +25,7 @@
     ClamAVMonitoringInfo,
     ClamAVScanResult,
     ClamAVHealth,
+    ClamAVScanEventClient,
```
ccwienk: nit: order alphabetically

clamav/model.py
```diff
@@ -123,3 +125,45 @@ def age(self) -> int:
...
+        Our ClamAV service will send exactly one event which is of one of two types:
+            1. An event of type 'error' containing an error-code (a HTTP/1.1 Status Code) and a
+                message, in case an error was encountered when scanning.
+            2. An event of type 'result' containing the scan result.
+
+        This method blocks until an event is received and then returns an instance of the
+        corresponding model class.
+        '''
+        event = next(self.client.events())
+        if event.event == ClamAVScanEventTypes.ERROR.value:
```
ccwienk: instantiate enum using it's c'tor. no need to redundantly read the enum's value. also, this implicitly enforces the value be known

clamav/model.py
```diff
@@ -123,3 +125,45 @@ def age(self) -> int:
...
+            return ClamAVError(json.loads(event.data))
+        elif event.event == ClamAVScanEventTypes.RESULT.value:
+            return ClamAVScanResult(json.loads(event.data))
+        else:
+            raise RuntimeError(
```
ccwienk: not required if using enum c'tor

clamav/routes.py
```diff
@@ -35,6 +35,9 @@ def _api_url(self, *parts, **kwargs):
     def scan(self):
         return self._api_url('scan')
 
+    def sse_scan(self):
+        return self._api_url('sse', 'scan')
```
ccwienk: so the route actually reads '/sse/scan'? well ok

clamav/model.py
ccwienk: I _think_ it is recommended practise to check for identity. decreases readability for me, though, so your choice.

clamav/model.py
```diff
@@ -123,3 +125,45 @@ def age(self) -> int:
...
+    def status_code(self) -> int:
+        '''Return the HTTP/1.1 status code that was given by ClamAV
+        '''
+        return self.raw['code']
+
+    def message(self) -> str:
+        return self.raw['message']
+
+
+class ClamAVScanEventClient(object):
```
ccwienk: a brief hint as to what this client is for might help. naming nit: maybe better call it ClamAVScanEventSSEClient? not sure, though. gets lengthy..

clamav/model.py
```diff
@@ -123,3 +125,45 @@ def age(self) -> int:
...
+
+        Our ClamAV service will send exactly one event which is of one of two types:
+            1. An event of type 'error' containing an error-code (a HTTP/1.1 Status Code) and a
+                message, in case an error was encountered when scanning.
+            2. An event of type 'result' containing the scan result.
+
+        This method blocks until an event is received and then returns an instance of the
+        corresponding model class.
+        '''
+        event = next(self.client.events())
```
ccwienk: is it not necessary to do this in a loop, until the event stream is terminated? to me this looks as if you would merely read the first event and then return

clamav/model.py
```diff
@@ -123,3 +125,43 @@ def age(self) -> int:
...
+        '''Return the HTTP/1.1 status code that was given by ClamAV
+        '''
+        return self.raw['code']
+
+    def message(self) -> str:
+        return self.raw['message']
+
+
+class ClamAVScanEventClient(object):
+    '''Simple client to handle SSE events sent by our k8s ClamAV installation
```
ccwienk: omit "simple".

PR #316, opened 2019-07-23
Use in-cluster ClamAV service

AndreasBurger:
Switches our pipelines to using the ClamAV service deployed in our cluster.

### DIFF
--- a/clamav/model.py
+++ b/clamav/model.py
@@ -132,14 +132,19 @@ class ClamAVScanEventTypes(enum.Enum):
     RESULT = 'result'
 
 
-class ClamAVError(ModelBase):
-    def status_code(self) -> int:
-        '''Return the HTTP/1.1 status code that was given by ClamAV
-        '''
-        return self.raw['code']
+class ClamAVError(Exception):
+    def __init__(self, error_code: int, error_message: str):
+        super().__init__(
+            f'Received error response from ClamAV: {error_code} - {error_message}'
+        )
+        self._error_code = error_code
+        self._error_message = error_message
 
-    def message(self) -> str:
-        return self.raw['message']
+    def error_code(self) -> int:
+        return self._error_code
+
+    def error_message(self) -> str:
+        return self._error_message
 
 
 class ClamAVScanEventClient(object):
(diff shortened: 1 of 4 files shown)


concourse/steps/scan_container_images.py
```diff
@@ -183,12 +184,24 @@ def comp_filter(component):
...
+    clamav_client = ccc.clamav.client_from_config_name(clamav_config_name)
     for image_reference in image_references:
-        status, signature = clamav.util.scan_container_image(image_reference=image_reference)
-        if clamav.util.result_ok(status=status, signature=signature):
+        scan_result = clamav_client.scan_container_image(image_reference=image_reference)
+        if clamav_client.result_ok(scan_result):
             continue
-        yield (image_reference, f'{status}: {signature}')
+
+        if isinstance(scan_result, ClamAVScanResult):
```
ccwienk: hmm.. I think, it is not ideal style to use runtime type inspection for this kind of flow control.

To me, it looks it would be more appropriate to raise `ClamAVError` (make it an exception). You could still handle the special case (HTTP-422) within the exception handling block, otherwise just re-raise

clamav/model.py
```diff
@@ -152,20 +157,21 @@ class ClamAVScanEventClient(object):
...
         '''Process the events sent by our ClamAV service
 
         Our ClamAV service will send exactly one SSE-event which is of one of two types:
             1. An event of type 'error', with its data containing an error-code (a HTTP/1.1 Status
                 Code) and a message, in case an error was encountered when scanning.
             2. An event of type 'result' containing the scan result in its data.
 
-        This method blocks until an event is received and then returns an instance of the
-        corresponding model class.
+        This method blocks until an event is received and then either returns the scan result or
```
ccwienk: nit: indicate that the method blocks until the scanning is done (either with a finding/no finding, or an error)

concourse/steps/scan_container_images.py
```diff
@@ -183,12 +184,19 @@ def comp_filter(component):
...
+        try:
+            scan_result, _ = clamav_client.scan_container_image(image_reference=image_reference)
+            if not scan_result.malware_detected():
+                continue
+            yield (image_reference, f'Malware detected: {scan_result.virus_signature()}')
+        except ClamAVError as e:
+            if e.error_code() == 422:
+                yield (image_reference, f'Scan aborted: {e.error_message()}')
+            else:
+                raise e
```
ccwienk: style nit: you could omit the argument (not important)

PR #321, opened 2019-08-13
Transport triages not only when uploading images

AndreasBurger:


product/scanning.py
```diff
@@ -70,15 +70,16 @@ def _attribute_specs(cls):
...
         self.rescan = rescan
         self.wait = wait
+        self.transport_triages = transport_triages
 
-    SKIP = (False, False, False)
-    UPLOAD = (True, False, True)
-    RESCAN = (False, True, True)
-    WAIT_FOR_RESULT = (False, False, True)
+    SKIP = (False, False, False, True)
+    UPLOAD = (True, False, True, False)
```
ccwienk: please set last flag to `True`; add special handling for execution order

PR #326, opened 2019-08-16
Make access to PR information easier

AndreasBurger:
Adds getters for special git-config entries created by the pull request resource on pull request jobs.
Also adds convenience-method for fetching all labels on pull requests via GitHubRepoHelper.

### DIFF
--- a/gitutil.py
+++ b/gitutil.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 
 import contextlib
+import enum
 import functools
 import os
 import subprocess
@@ -36,6 +37,13 @@ def _ssh_auth_env(github_cfg):
     id_only = '-o "IdentitiesOnly yes"'
     os.environ['GIT_SSH_COMMAND'] = f'ssh -i {tmp_id} {suppress_hostcheck} {id_only}'
     return tmp_id
+
+
+class ConfigLevel(enum.Enum):
+    LOCAL = 'repository'
+    GLOBAL = 'global'
+    SYSTEM = 'system'
+    DEFAULT = None
 
 
 class GitHelper(object):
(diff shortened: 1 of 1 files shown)


github/util.py
```diff
@@ -498,6 +498,10 @@ def is_org_member(self, organization_name, user_login):
         organization = self.github.organization(organization_name)
         return organization.is_member(user_login)
 
+    def fetch_labels_on_pr(self, pull_request_number):
```
ccwienk: nit: type hint?

gitutil.py
```diff
@@ -38,6 +39,12 @@ def _ssh_auth_env(github_cfg):
     return tmp_id
 
 
+class ConfigLevel(enum.Enum):
+    LOCAL = 'repository'
+    GLOBAL = 'global'
+    SYSTEM = 'system'
```
ccwienk: add `NONE = None` (or maybe `DEFAULT = None`?)

gitutil.py
```diff
@@ -151,6 +158,11 @@ def _has_stash(self):
     def _pop_stash(self):
         self.repo.git.stash('pop', '--quiet')
 
+    def _config_value(self, section, option, level:ConfigLevel=None):
```
ccwienk: use `NONE / DEFAULT` to get rid of the special-case below

gitutil.py
```diff
@@ -151,6 +158,11 @@ def _has_stash(self):
...
+        if level:
+            level = level.value
```
ccwienk: not required - always use level.value

btw: why is this required? I know what the cfg levels mean on git-side. just wondering why we need to discriminate now.

AndreasBurger: I dont think it is needed for the introduced changes, as the api mentions
> If None, all applicable levels will be used

I'd rather do it properly, however.

github/util.py
AndreasBurger: On further thought, I agree. I've removed it, as the benefit is miniscule

PR #327, opened 2019-08-22
Drop duplicate release notes

### LINKED ISSUE #323: Duplicates in release notes
I saw almost every release note duplicated in one of our internal landscapes.
@petersutter can you check this please?

petersutter:
Fixes #323

### DIFF
--- a/github/release_notes/model.py
+++ b/github/release_notes/model.py
@@ -13,34 +13,26 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from collections import namedtuple
+from dataclasses import dataclass
 from pydash import _
 
 from product.model import ComponentName
 from util import check_type
 
-ReleaseNote = namedtuple('ReleaseNote', [
-    "category_id",
-    "target_group_id",
-    "text",
-    "reference",
-    "user_login",
-    "is_current_repo",
-    "from_same_github_instance",
-    "cn_source_repo"
-])
 
-Commit = namedtuple('Commit', [
-    "hash",
-    "subject",
-    "message"
-])
+@dataclass
+class Commit:
+    hash: str
+    subject: str
+    message: str
 
-ReferenceType = namedtuple('ReferenceType', [
-    'identifier', # reference type identifier in release notes block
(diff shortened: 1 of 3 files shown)


github/release_notes/model.py
```diff
@@ -28,19 +29,60 @@
     "is_current_repo",
     "from_same_github_instance",
     "cn_source_repo"
-])
+])):
+    def __eq__(self, other):
```
ccwienk: why is this required? namedtuple instances should already have `__eq__` implementations with this sematics (equal iff all attrs are equal).

also, consider switching to data classes (new feature in python3.7):

https://docs.python.org/3/library/dataclasses.html

github/release_notes/model.py
```diff
@@ -28,19 +29,60 @@
...
+
 
-Commit = namedtuple('Commit', [
+class Commit(namedtuple('Commit', [
     "hash",
     "subject",
     "message"
-])
+])):
+    def __eq__(self, other):
```
ccwienk: also: should not be required.

github/release_notes/model.py
```diff
@@ -55,10 +97,19 @@
 )
 REFERENCE_TYPES = [REF_TYPE_PULL_REQUEST, REF_TYPE_COMMIT]
 
-Reference = namedtuple('Reference', [
+
+class Reference(namedtuple('Reference', [
```
ccwienk: ditto

github/release_notes/renderer.py
```diff
@@ -77,7 +77,7 @@ def get_or_call(obj, path):
 
 class Renderer(object):
     def __init__(self, release_note_objs: [ReleaseNote]):
-        self.rls_note_objs = release_note_objs
+        self.rls_note_objs = _.uniq(release_note_objs)
```
ccwienk: i assume this single change should be sufficient for the intended fix

petersutter:
yes correct, thanks. I have addressed your comments and also went with `dataclasses`. PTAL

rfranzke:
ping

PR #335, opened 2019-09-13
Changes to GithubConfig protocol handling

AndreasBurger:
GithubConfig now specifies all available protocols as list instead of only explicitly giving the preferred one.

Also adds additional validation and adjusts tests.

### DIFF
--- a/model/github.py
+++ b/model/github.py
@@ -22,10 +22,11 @@ from urllib.parse import urlparse
 from model.base import (
     BasicCredentials,
     NamedModelElement,
+    ModelValidationError,
 )
 
 
-class PreferredProtocol(enum.Enum):
+class Protocol(enum.Enum):
     SSH = 'ssh'
     HTTPS = 'https'
 
@@ -39,9 +40,13 @@ class GithubConfig(NamedModelElement):
         return set(self.raw.get('purpose_labels', ()))
 
     def ssh_url(self):
+        if Protocol.SSH not in self.available_protocols():
+            raise RuntimeError(f"SSH protocol is not available for Github config '{self.name()}'")
         return self.raw.get('sshUrl')
 
     def http_url(self):
+        if Protocol.HTTPS not in self.available_protocols():
+            raise RuntimeError(f"HTTPS protocol is not available for Github config '{self.name()}'")
         return self.raw.get('httpUrl')
 
     def api_url(self):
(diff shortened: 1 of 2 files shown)


model/github.py
```diff
@@ -22,10 +22,11 @@
 from model.base import (
     BasicCredentials,
     NamedModelElement,
+    ModelValidationError,
 )
 
 
-class PreferredProtocol(enum.Enum):
+class GitProtocol(enum.Enum):
```
ccwienk: just protocol, please

model/github.py
```diff
@@ -72,23 +82,34 @@ def matches_hostname(self, host_name):
...
             'disable_tls_validation',
-            'httpUrl',
-            'sshUrl',
             'webhook_token',
         ]
 
     def validate(self):
         super().validate()
+
+        available_protocols = self.available_protocols()
```
ccwienk: check len(available_protocols) > 0

model/github.py
```diff
@@ -72,23 +82,34 @@ def matches_hostname(self, host_name):
...
+        if GitProtocol.SSH in available_protocols and not self.ssh_url():
+            raise ModelValidationError(
+                f'Ssh url is missing for Github config {self.name}'
```
ccwienk: SSH (nit)

model/github.py
```diff
@@ -72,23 +82,34 @@ def matches_hostname(self, host_name):
...
+            )
+        if GitProtocol.HTTPS in available_protocols and not self.http_url():
+            raise ModelValidationError(
+                f'Http url is missing for Github config {self.name}'
```
ccwienk: HTTP (nit)

model/github.py
```diff
@@ -54,7 +59,12 @@ def webhook_secret(self):
         return self.raw.get('webhook_token')
 
     def preferred_protocol(self):
-        return PreferredProtocol(self.raw.get('preferred_protocol', PreferredProtocol.SSH))
+        return self.available_protocols()[0]
+
+    def available_protocols(self):
+        '''Return available git protocols, in order of preference
```
ccwienk: nit: document preference order.. (most preferred first..)..

model/github.py
```diff
@@ -72,23 +82,39 @@ def matches_hostname(self, host_name):
...
-            'httpUrl',
-            'sshUrl',
             'webhook_token',
         ]
 
     def validate(self):
         super().validate()
+
+        available_protocols = self.available_protocols()
+        if len(available_protocols) == 0:
```
ccwienk: nit: better to check for < 1 (as a general rule. len might be defined to always return `>=0`, but there may be other cases. better to be as restrictive as possible when doing validation checks

model/github.py
```diff
@@ -22,10 +22,11 @@
...
     BasicCredentials,
     NamedModelElement,
+    ModelValidationError,
 )
 
 
-class PreferredProtocol(enum.Enum):
+class Protocol(enum.Enum):
     SSH = 'ssh'
     HTTPS = 'https'
```
ccwienk: yet another nit: the procotol is `HTTP`, even if it is used w/ transport-level encryption (TLS)

PR #339, opened 2019-09-18
Continue ClamAV scan after first finding

AndreasBurger:


clamav/client.py
```diff
@@ -76,9 +77,16 @@ def scan_container_image(self, image_reference: str):
...
-                return scan_result, path
-        return ClamAVScanResult({'malwareDetected':False}), ''
+            try:
+                scan_result = self.sse_scan(content)
+                if not scan_result.malware_detected():
+                    continue
+                else:
+                    yield (scan_result, path)
+            except ClamAVError as e:
+                if e.error_code() == 422:
```
ccwienk: could you at least define a module variable with a more expressive name for this integer constant (and use it here in lieu of the literal)?

PR #343, opened 2019-09-19
DO NOT MERGE call trigger of resources in threads

RaphaelVogel:


whd/dispatcher.py
```diff
@@ -137,15 +138,20 @@ def _trigger_resource_check(self, concourse_api, resources):
                 continue
             resources.insert(0, resource)
 
+        executor = ThreadPoolExecutor()
         for resource in resources:
-            app.logger.info('triggering resource check for: ' + resource.name)
```
ccwienk: through parallelisation, you break the guaranteed execution order (see comment a few lines above)

whd/webhook.py
```diff
@@ -39,6 +44,12 @@ def __init__(
         self.dispatcher = GithubWebhookDispatcher(cfg_set=cfg_set, whd_cfg=whd_cfg)
 
     def post(self):
+        executor.submit(self._post)
+        # resource checking can take long, so we execute in an own thread and return
+        time.sleep(3)
```
ccwienk: no.. just wait for the submitted task's result with a timeout; ignore the timeout-exception (or rather, print a warning..). that#s it

whd/webhook.py
```diff
@@ -39,6 +44,12 @@ def __init__(
...
+        return "Ok"
```
ccwienk: I think this can be omitted

whd/webhook.py
```diff
@@ -39,6 +45,14 @@ def __init__(
         self.dispatcher = GithubWebhookDispatcher(cfg_set=cfg_set, whd_cfg=whd_cfg)
 
     def post(self):
+        # resource checking can take long, so we execute in an own thread
```
ccwienk: please state _why_ it's a problem the updating tennds to take long (btw, it is also true to state that any subsequent operations take potentially too long - it does not matter which at this point), namely we want to mark github-webhooks as "delivered successfully" even though we might not succeed in performing all tasks we want / need to upon the webhook request, so that we can identify actual delivery issues from github-reporting

RaphaelVogel: all done

PR #344, opened 2019-09-20
catch resource check exceptions

RaphaelVogel:


concourse/replicator.py
```diff
@@ -453,10 +453,14 @@ def _initialise_new_pipeline_resources(self, concourse_api, results):
...
             )
 
-            FluentIterable(concourse_api.pipeline_resources(pipeline_name)) \
-            .filter(lambda resource: resource.has_webhook_token()) \
-            .map(lambda resource: trigger_pipeline_resource_check(resource_name=resource.name)) \
-            .as_list()
+            try:
+                FluentIterable(concourse_api.pipeline_resources(pipeline_name)) \
+                .filter(lambda resource: resource.has_webhook_token()) \
+                .map(lambda resource: trigger_pipeline_resource_check(resource_name=resource.name)) \
```
ccwienk: the exception handling ought to be done inside the `trigger_pipeline_resource_check` method / function. otherwise, this will still break upon the first exception

PR #345, opened 2019-09-20
add pause pipeline methods to concourse API

RaphaelVogel:


concourse/client/api.py
```diff
@@ -160,6 +160,14 @@ def unpause_pipeline(self, pipeline_name: str):
                 body=""
         )
 
+    @ensure_annotations
+    def pause_pipeline(self, pipeline_name: str):
+        pause_url = self.routes.pause_pipeline(pipeline_name)
+        self.request_builder.put(
+                pause_url,
+                body=""
```
ccwienk: nit: can this not be omitted (empty body)?

ccwienk: also: sure it's `PUT` rather than `POST`?

RaphaelVogel: yes PUT I checked go routes

PR #347, opened 2019-09-23
Add function to retrieve own build from within CI

AndreasBurger:


concourse/util.py
```diff
@@ -116,3 +139,60 @@ def resurrect_pods(
...
+    # api accesses, use only the 10 most recent.
+    job_builds = client.job_builds(pipeline_metadata.pipeline_name, pipeline_metadata.job_name)[:10]
+
+    for build in job_builds:
+        if not build.status() is concourse.client.model.BuildStatus.RUNNING:
+            continue
+        build_events = build.events()
+        build_plan = build.plan()
+        meta_task_id = build_plan.task_id(concourse.model.traits.meta.META_STEP_NAME)
+        for line in itertools.islice(build_events.iter_buildlog(meta_task_id), 0, 40):
```
RaphaelVogel: It works, as long as we guarantee, that our metadata JSON is printed without line-breaks.

PR #348, opened 2019-09-24
Refactor email.mako

AndreasBurger:
Move out most of the logic. Determine parameters to set in the template instead.


RaphaelVogel:
lgtm

PR #355, opened 2019-10-07
Do not add build metadata automatically in els client

AndreasBurger:
Removes build metadata from information that is automatically added when running on CI. The build
metadata is replaced by the build's UUID.

### DIFF
--- a/ccc/elasticsearch.py
+++ b/ccc/elasticsearch.py
@@ -49,32 +49,17 @@ def _metadata_dict():
     if not util._running_on_ci():
         return {}
 
-    build = concourse.util.find_own_running_build()
     pipeline_metadata = concourse.util.get_pipeline_metadata()
     config_set = util.ctx().cfg_factory().cfg_set(pipeline_metadata.current_config_set_name)
     concourse_cfg = config_set.concourse()
 
     meta_dict = {
-      'build-id': build.id(),
-      'build-name': str(build.build_number()),
-      'build-job-name': pipeline_metadata.job_name,
-      'build-team-name': pipeline_metadata.team_name,
-      'build-pipeline-name': pipeline_metadata.pipeline_name,
-      'atc-external-url': concourse_cfg.external_url(),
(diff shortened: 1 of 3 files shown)


ccc/elasticsearch.py
```diff
@@ -49,33 +49,18 @@ def _metadata_dict():
     if not util._running_on_ci():
         return {}
 
-    build = concourse.util.find_own_running_build()
     pipeline_metadata = concourse.util.get_pipeline_metadata()
     config_set = util.ctx().cfg_factory().cfg_set(pipeline_metadata.current_config_set_name)
     concourse_cfg = config_set.concourse()
 
     meta_dict = {
-      'build-id': build.id(),
```
ccwienk: change indentation in separate commit

ccc/elasticsearch.py
```diff
@@ -49,33 +49,18 @@ def _metadata_dict():
...
-      'atc-external-url': concourse_cfg.external_url(),
+        'build-uuid': pipeline_metadata.build_uuid,
+        'build-job-name': pipeline_metadata.job_name,
+        'build-team-name': pipeline_metadata.team_name,
+        'build-pipeline-name': pipeline_metadata.pipeline_name,
+        'atc-external-url': concourse_cfg.external_url(),
     }
 
-    # XXX deduplicate; mv to concourse package
-    meta_dict['concourse_url'] = util.urljoin(
```
ccwienk: i understand it is no longer possible to calculate concourse(-job)-url w/o build-id. However, this is a potentially severe incompatible contract change.

we cannot merge this change w/o previously clarifying this with our users, as also already discussed earlier.

pbly, we should first provide an equivalent functionality, e.g. by registering each build job in our "delivery-db" and exposing all relevant metadata through delivery-svc.

at any, rate, at this point in time, this PR cannot be accepted.

AndreasBurger:
Closed (at least until work on delivery-svc has progressed)

PR #357, opened 2019-10-09
Add util function to obtain active gardener-cicd-libs-version

AndreasBurger:
__note__: Also contains https://github.com/gardener/cc-utils/pull/356 and must not be merged before

### DIFF
--- a/ci/util.py
+++ b/ci/util.py
@@ -21,6 +21,7 @@ import yaml
 import yamllint
 import yamllint.config
 
+import pkg_resources
 import termcolor
 
 from urllib.parse import urlunparse
@@ -330,6 +331,17 @@ def merge_dicts(base: dict, other: dict, list_semantics='merge'):
     from copy import deepcopy
     # copy dicts, so they remain unmodified
     return merger.merge(deepcopy(base), deepcopy(other))
+
+
+DISTRIBUTION_NAME = 'gardener-cicd-libs'
+
+
+def gardener_cicd_libs_version():
+    '''Return the version of the active gardener-cicd-libs distribution
+    '''
+    for distribution in pkg_resources.working_set:
+        if distribution.key == DISTRIBUTION_NAME:
+            return distribution.version
 
 
 class FluentIterable(object):


ci/util.py
```diff
@@ -332,6 +333,17 @@ def merge_dicts(base: dict, other: dict, list_semantics='merge'):
...
 
 
+DISTRIBUTION_NAME = 'gardener-cicd-libs'
+
+
+def gardener_cicd_libs_version():
+    '''Return the version of the active gardener-cicd-libs distribution
+    '''
+    for distribution in pkg_resources.working_set:
+        if distribution.key == DISTRIBUTION_NAME:
```
ccwienk: inline?

also: raise if not found (document + catch on caller-side (lss))

ci/util.py
```diff
@@ -332,6 +333,17 @@ def merge_dicts(base: dict, other: dict, list_semantics='merge'):
     return merger.merge(deepcopy(base), deepcopy(other))
 
 
+DISTRIBUTION_NAME = 'gardener-cicd-libs'
+
+
+def gardener_cicd_libs_version():
+    '''Return the version of the active gardener-cicd-libs distribution
+    '''
+    for distribution in pkg_resources.working_set:
```
ccwienk: why doing this cumbersome lookup? there is already a sane way to access entries by key:
```
import pkg_resources as pr
ws = pr.working_set
ws.by_key['gardener-cicd-libs']

gardener-cicd-libs 1.417.0 (/usr/lib/python3.7/site-packages)
```
How I found out about this? Well, I looked briefly into the documentation, because it was hard for me to believe that this module does not expose any more convenient way for such a straight-forward query. Turns out it does.

ci/util.py
```diff
@@ -332,6 +333,17 @@ def merge_dicts(base: dict, other: dict, list_semantics='merge'):
     return merger.merge(deepcopy(base), deepcopy(other))
 
 
+DISTRIBUTION_NAME = 'gardener-cicd-libs'
```
ccwienk: no.

AndreasBurger:
closed; will be done (as discussed) by reading version from our versionfile rather than using an undocumented attrib.

PR #358, opened 2019-10-11
switch mapping-order in steps

AndreasBurger:
At some point in the past, we started to use `<env_var_name>:<value>` mappings in our step in-/outputs backwards, i.e. we'd render the env_var name as value and the value as name.

Fortunately, these two values are the same, most of the time. Nevertheless, this PR 'fixes' the mapping to map env_var names to values and changes all usage of our step in-/outputs to be more expressive.

### DIFF
--- a/concourse/model/step.py
+++ b/concourse/model/step.py
@@ -217,14 +217,14 @@ class PipelineStep(ModelBase):
         raw_dict['depends'] = set(raw_dict['depends'])
         if raw_dict.get('output_dir', None):
             name = raw_dict['output_dir']
-            self.add_output(name + '_path', name + '_path')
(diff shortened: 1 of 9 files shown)


concourse/model/step.py
```diff
@@ -292,9 +292,9 @@ def outputs(self):
         return self._outputs_dict
 
     def add_output(self, name, variable_name):
-        if name in self._outputs_dict:
-            raise ValueError('output already exists: ' + str(name))
-        self._outputs_dict[name] = variable_name
+        if variable_name in self._outputs_dict:
+            raise ValueError('output already exists: ' + str(variable_name))
```
ccwienk: nit: while at it, rewrite to use format-str

PR #362, opened 2019-10-15
Replace hardcoded env var names

AndreasBurger:


ci/util.py
```diff
@@ -332,6 +332,11 @@ def merge_dicts(base: dict, other: dict, list_semantics='merge'):
     return merger.merge(deepcopy(base), deepcopy(other))
 
 
+def sane_env_var_name(name: str):
```
ccwienk: leave a copy of the moved function where it was + add a @deprecated annotation to it (remove after relasing a.s.o.)

concourse/model/traits/meta.py
```diff
@@ -25,8 +25,8 @@
   ScriptType,
 )
 
-META_INFO_ENV_VAR_NAME = 'META'
-META_INFO_DIR_NAME = 'meta'
+ENV_VAR_NAME = 'meta'
```
ccwienk: renaming / shortening is fine with me. however does not match the PR's title

concourse/steps/component_descriptor.mako
```diff
@@ -3,12 +3,18 @@
...
+import concourse.model.traits.version
+import concourse.model.traits.component_descriptor
 from makoutil import indent_func
 from concourse.steps import step_lib
+
 descriptor_trait = job_variant.trait('component_descriptor')
 main_repo = job_variant.main_repository()
 main_repo_path_env_var = main_repo.logical_name().replace('-', '_').upper() + '_PATH'
 
+version_file_path = job_step.input(concourse.model.traits.version.ENV_VAR_NAME) + '/version'
```
ccwienk: ?????
use os.path.join

ccwienk: also - we already had this discussion - several time. do. not. introduce this unnecessary indirection using env vars.

AndreasBurger: First of, good catch - mistake when adjusting, probably taken over from https://github.com/gardener/cc-utils/blob/41c03c4a9454118586e597e241774265a9277b90/concourse/steps/draft_release.mako#L8

Second of, I'm going to remove the commits concerning the env-var stuff alltogether. I have stated my intent to 'remove hard-coded in/output names' several times, but considering the second remark that seems moot. I see that the pr name is misleading, however. My apologies, I'll take a closer look at GHs suggestions in the future.

AndreasBurger:
Closed; After working on it not much that semantically belongs to _this_ pr remains. Will create a new PR

PR #369, opened 2019-10-31
make yaml linter configurable

RaphaelVogel:


ci/util.py
```diff
@@ -207,10 +208,13 @@ def parse_yaml_file(path, lint=False):
         return yaml.load(f, Loader=yaml.SafeLoader)
 
 
-def lint_yaml_file(path):
+def lint_yaml_file(path, linter_config: Dict=None):
```
ccwienk: nit: why not use `dict` as type hint?
typing.Dict is only there to be used to express key/value types, e.g. `Dict[str, int]`, I think.

also, why not specify the default-cfg as default value (linter_config={'extends': 'relaxed'}), thus adhering better to python's common practice of defining default values?

RaphaelVogel: done

PR #370, opened 2019-11-05
Add (optional) validation to dependency parsing

AndreasBurger:


product/model.py
```diff
@@ -357,6 +357,18 @@ def _optional_attributes(self):
     def dependencies(self):
         return ComponentDependencies(raw_dict=self.raw['dependencies'])
 
+    def add_container_image_dependency(self, dependency):
```
ccwienk: add comment indicating those are convenience methods

cli/gardener_ci/productutil.py
```diff
@@ -195,19 +278,23 @@ def add_dependencies(
...
-        component_deps.add_component_dependency(component_ref)
-    for image_dep in container_image_dependencies:
-        component_deps.add_container_image_dependency(image_dep)
-    for web_dep in web_dependencies:
-        component_deps.add_web_dependency(web_dep)
-    for generic_dep in generic_dependencies:
-        component_deps.add_generic_dependency(generic_dep)
+        ))
+
+    # maintain old behaviour
```
ccwienk: maybe give a bit more context; e.g. "default value for backwards compatibility")?

cli/gardener_ci/productutil.py
```diff
@@ -195,19 +278,23 @@ def add_dependencies(
...
+    if not validation_policies:
+        validation_policies=[ValidationPolicy.FORBID_EXTRA_ATTRIBUTES]
+
+    _add_dependencies(
```
ccwienk: as discussed:

split this into two functions:

- parse_dependencies (which returns a sequence (e.g. a generator, but this is up to you, of course)) of dependency-model-objects
- add_dependencies (which accepts said sequence + validation options + target component descriptor)

add-dependencies should handle the proper dispatching of the dependency-objects, depending on type. as discussed, it might make sense to add such a method to the dependencies model class

PR #372, opened 2019-11-06
add hyperlink in protocode email

RaphaelVogel:


concourse/steps/scan_container_images.py
```diff
@@ -236,10 +236,11 @@ def result_to_tuple(upload_result: UploadResult):
         custom_data = analysis_result.custom_data()
         if custom_data is not None:
           image_reference = custom_data.get('IMAGE_REFERENCE')
+          image_reference_url = f'<a href="https://{image_reference}">{image_reference}</a>'
         else:
-          image_reference = None
+          image_reference_url = None
 
-        return [link_to_analysis_url, greatest_cve, image_reference]
+        return [link_to_analysis_url, greatest_cve, image_reference_url]
```
AndreasBurger: Note: The table header should probably also be adjusted. Also, maybe add the url as extra column.

Otherwise lgtm

AndreasBurger: After discussion: Not an issue; is consistent with how we handle the component name.

PR #374, opened 2019-11-08
fix: expand components

ccwienk:


protecode/util.py
```diff
@@ -78,7 +78,7 @@ def _upload_tasks():
 
         for component_name, components in component_groups.items():
             for image_group in product.util._grouped_effective_images(
-                components,
+                *components,
                 component_descriptor=component_descriptor,
             ):
                 # XXX HACK: arbitrarily use first component for filtering
```
RaphaelVogel: at this point components is still a set of `ComponentReference`: Trying to access a set with `component = components[0]` dumps

ccwienk: done

PR #375, opened 2019-11-11
Do not bump version or create commit if `nextversion` is NOOP

RaphaelVogel:


concourse/steps/release.py
```diff
@@ -274,6 +274,10 @@ def apply(self):
         next_cycle_commit = self._add_all_and_create_commit(
             message=self._next_dev_cycle_commit_message(next_cycle_dev_version)
         )
+
+        if not next_cycle_commit:
```
ccwienk: to me, this indicates that this step does too many things.

instead of doing an early-exit here, rather split the transaction-step into multiple parts:

- one step doing the rebase (if configured)
  - note: if no rebase to be done, do not add the step that would to the rebase. do not add an early exit to the rebase-step
- one step creating the release-commit (if required according to configuration)
- one step creating the "next version bump" (if required according to configuration)

concourse/steps/release.py
```diff
@@ -349,6 +353,8 @@ def _add_all_and_create_commit(self, message: str):
         commit = self.git_helper.index_to_commit(
             message=message,
         )
+        if not commit:
```
ccwienk: by splitting this step into multiple steps, this early-exit will no longer be required

concourse/steps/release.py
```diff
@@ -361,12 +367,16 @@ def _calculate_next_cycle_dev_version(
         version_operation: str,
         prerelease_suffix: str,
     ):
-        # calculate the next version and append the prerelease suffix
+        # calculate the next version
+        version_str = version.process_version(
+            version_str=release_version,
+            operation=version_operation,
+        )
+        if version_operation == version.NOOP:
```
ccwienk: not sure I follow. but if the version-operation is a noop, thy not return it immediately (w/o first processing it)?

gitutil.py
```diff
@@ -142,8 +142,12 @@ def index_to_commit(self, message, parent_commits=None):
         The worktree remains unchanged after the method returns.
 
         @param parent_commits: optional iterable of parent commits; head is used if absent
-        @return the git.Commit object representing the newly created commit
+        @return the git.Commit object representing the newly created commit or
+        None if no changes are done
         '''
+        if not self.repo.is_dirty():
```
ccwienk: not sure I agree.

firstly, the caller should rather check whether or not it makes sense to create a commit (I think?)
secondly, if the index is empty, this should imo create an empty commit (i mean, why not?)

PR #377, opened 2019-11-12
split release commit and next dev cycle commit

RaphaelVogel:


RaphaelVogel:
@ccwienk , please take a look.

concourse/steps/release.py
```diff
@@ -209,93 +200,169 @@ def __init__(
...
 
-        # second, prepare next dev cycle commit
-        next_cycle_dev_version = self._calculate_next_cycle_dev_version(
+        # Push commit to remote
+        self.git_helper.push(
+            from_ref=release_commit.hexsha,
+            to_ref=self.repository_branch,
+        )
+        return {
+            'release commit sha': release_commit.hexsha,
```
ccwienk: I'm still not very convinced of the idea to use dict attrs containing whitespace. or of returning data like this at all. but this is not subject to this PR

concourse/steps/release.py
```diff
@@ -209,93 +200,169 @@ def __init__(
...
+        }
+
+    def revert(self):
+        if not self.context().has_output(self.name()):
+            # push unsuccessful, nothing to do
```
ccwienk: nit: might log this

concourse/steps/release.py
```diff
@@ -209,93 +200,169 @@ def __init__(
...
 
-        # first, prepare release commit
+        # prepare release commit
         with open(self.repository_version_file_path, 'w') as f:
             f.write(self.release_version)
 
         # call optional release commit callback
         if self.release_commit_callback:
-            self._invoke_callback(
+            _invoke_callback(
```
ccwienk: why did you convert this into a module function?

RaphaelVogel: Yes it's because of code duplication. `_invoke_callback` is a generic function, which is called once for a Release Commit Callback and once for Next Dev Cycle Commit Callback.

ccwienk: ok, understood

PR #379, opened 2019-11-25
do not create dev cycle commit if operation is NOOP

RaphaelVogel:


concourse/steps/release.py
```diff
@@ -715,17 +715,18 @@ def release_and_prepare_next_dev_cycle(
...
-        repo_dir=repo_dir,
-        release_version=release_version,
-        repository_version_file_path=repository_version_file_path,
-        repository_branch=githubrepobranch.branch(),
-        version_operation=version_operation,
-        prerelease_suffix=prerelease_suffix,
-        next_version_callback=next_version_callback,
-    )
-    step_list.append(next_cycle_commit_step)
+    if version_operation != version.NOOP:
```
ccwienk: nit: rather use identity comparison (`if version_operation is not version.NOOP`)

ccwienk: my bad. 'tis not an enum

PR #387, opened 2019-12-09
Do not upload images that are being scanned

AndreasBurger:


product/scanning.py
```diff
@@ -288,6 +289,21 @@ def enumerate_reference_triages():
...
+        products_with_busy_scans = [
+            p for p in scan_results if p.product_id() in busy_scans
+        ]
+
+        for product in products_with_busy_scans:
+            for container_image in container_image_group.images():
+                if product.custom_data().get('IMAGE_VERSION') == container_image.version():
+                    images_currently_being_scanned.append(container_image)
+
+        images_to_upload = [i for i in images_to_upload if i not in images_currently_being_scanned]
```
ccwienk: see below

product/scanning.py
```diff
@@ -288,6 +289,21 @@ def enumerate_reference_triages():
...
+
         # upload new images
         for container_image in images_to_upload:
             scan_result = self._upload_image(
```
ccwienk: as discussed: try-except for http-409 (and only this, please).
in except block, retrieve product using self.retrieve_scan_result; then wait for it

PR #394, opened 2020-01-14
Fix release notes

AndreasBurger:
With the latest change to the release trait release notes are not correctly processed. This happens because the tagged commits are not directly reachable from master anymore. Thus all release notes since the most recent release commit on the release-branch are always fetched.

This PRaims to fix this by considering the parents of the release-commits (who are still on the release branch).


ccwienk:
as discussed yesterday, let us as a short-term mitigation revert relevant release-jobs to the "old" tagging semantics to make the current release-notes handling work again, and then (mid-term) overhaul release-notes handling

PR #395, opened 2020-01-22
Add debug output to protecode triage transporting

AndreasBurger:


protecode/model.py
```diff
@@ -69,6 +72,15 @@ def license(self) -> 'License':
             return None
         return License(raw_dict=license_raw)
 
+    def __repr__(self):
+        return (
+            f'{self.__class__.__name__}: {self.name()} '
+            f'{self.version() or "Version not detected"}'
```
ccwienk: can this happen? is it not enough to rely on str(None)?

product/scanning.py
```diff
@@ -417,6 +417,7 @@ def _upload_image(
 
     def _transport_triages(self, triages, product_id):
         for triage in triages:
+            info(f"adding triage '{triage}' to product with id {product_id}')")
```
ccwienk: this will cause too much noise. we might log to a file, maybe. or print the triage-IDs, if required
+ consider switching to python's default `logging` module

protecode/model.py
```diff
@@ -50,7 +50,10 @@ def custom_data(self):
         return self.raw.get('custom_data')
 
     def __repr__(self):
-        return f'{self.__class__.__name__}: {self.name()}({self.product_id()})'
+        return f'{self.__class__.__name__}: {self.display_name()}({self.product_id()})'
```
ccwienk: +1

protecode/model.py
```diff
@@ -50,7 +50,10 @@ def custom_data(self):
...
+
+    def __str__(self):
```
ccwienk: according to https://docs.python.org/3/reference/datamodel.html#object.__repr__, this should not be necessary. instead, consider changing the super-class's `__str__` to `__repr__`

protecode/model.py
```diff
@@ -69,6 +72,15 @@ def license(self) -> 'License':
             return None
         return License(raw_dict=license_raw)
 
+    def __repr__(self):
+        return (
+            f'{self.__class__.__name__}: {self.name()} '
+            f'{self.version() or "Version not detected"}'
+        )
+
+    def __str__(self):
```
ccwienk: ditto

protecode/model.py
```diff
@@ -125,6 +137,12 @@ def cve_major_severity(self, cvss_version) -> int:
         else:
             return -1
 
+    def __repr__(self):
+        return f'{self.__class__.__name__}: {self.cve()}'
```
ccwienk: prefix w/ CVE / CVSS(vx)

protecode/model.py
```diff
@@ -135,14 +153,17 @@ class TriageScope(Enum):
...
+        return self.raw['id']
+
     def vulnerability_id(self):
         return self.raw['vuln_id']
 
     def component_name(self):
         return self.raw['component']
 
     def component_version(self):
-        return self.raw['version']
```
ccwienk: check usages (None-handling..)

protecode/model.py
```diff
@@ -153,6 +174,16 @@ def reason(self):
...
 
+    def __repr__(self):
+        return (
+            f'{self.__class__.__name__}: {self.id()} '
+            f'({self.component_name()} {self.component_version() or "Version not detected"}, '
+            f'{self.vulnerability_id()})'
+        )
+
+    def __str__(self):
+        return repr(self)
```
ccwienk: ditto

PR #398, opened 2020-01-28
add function to retrieve current config-set

RaphaelVogel:


ci/util.py
```diff
@@ -322,6 +322,12 @@ def _running_on_ci():
     return 'CC_ROOT_DIR' in os.environ
 
 
+def current_config_set():
```
ccwienk: + `_name` - or return the cfg set (from cfg_factory)?

RaphaelVogel: done

PR #399, opened 2020-02-12
Improvements to triage import

AndreasBurger:
Contains two sets of changes:

1.: When importing triages for vulnerabilities, sometimes the component version is not known to protecode. This prevents any triages to be set on the component. The first change sets a dummy version for these components so that the vulnerabilities can be automatically triaged.

2.: A method that imports the version of components from gcr iff their version is unknown to protecode and the name matches.

### DIFF
--- a/protecode/client.py
+++ b/protecode/client.py
@@ -29,6 +29,7 @@ from .model import (
     ScanResult,
     Triage,
     TriageScope,
+    VersionOverrideScope,
 )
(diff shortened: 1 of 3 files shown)


AndreasBurger:
As discussed with @ccwienk: removed second commit (version import from GCR), now only contains automatic override of version in order to be able to import triages

PR #401, opened 2020-02-13
Consider an images digest instead of its version when handling protecode scans

AndreasBurger:
We only considered the image's version for the final determination whether a given Protecode product corresponds to a container image and vice versa.
Use the image digest instead to ensure that we always detect changes to the image which would make a rescan necessary.

### DIFF
--- a/product/model.py
+++ b/product/model.py
@@ -14,13 +14,16 @@
 # limitations under the License.
 
 import abc
-import urllib.parse
+import functools
 import typing
+import urllib.parse
 from enum import Enum
 
 from model.base import ModelBase, ModelValidationError
 from protecode.model import AnalysisResult
 from ci.util import not_none, urljoin, check_type
+
+import container.registry
 import version as ver
 
 #############################################################################
@@ -395,6 +398,13 @@ class ContainerImage(ContainerImageReference):
 
     def image_name(self):
         return self.raw.get('name')
+
+    @functools.lru_cache
+    def image_reference_with_digest(self):
+        return container.registry.to_hash_reference(self.image_reference())
+
+    def image_digest(self):
+        return f'@{self.image_reference_with_digest().split("@")[1]}'
 
     def validate(self):
         img_ref = check_type(self.image_reference(), str)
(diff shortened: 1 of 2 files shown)


container/registry.py
```diff
@@ -169,6 +169,7 @@ def _image_exists(image_reference: str) -> bool:
   return False
 
 
+@functools.lru_cache
```
ccwienk: hmmmmm....

tbh, I think this is a bad idea. it may be a constructed corner case. but what if the digest _does_ change while we still hold the old, cached value? in that case we would miss this.

protecode/scanning_util.py
```diff
@@ -157,6 +158,9 @@ def _image_ref_metadata(self, container_image, omit_version):
         if not omit_version:
             metadata_dict['IMAGE_REFERENCE'] = container_image.image_reference()
             metadata_dict['IMAGE_VERSION'] = container_image.version()
+            metadata_dict['IMAGE_DIGEST'] = container.registry.to_hash_reference(
```
ccwienk: could you just store the digest (including the algorithm, i.d. everything behind the color (`:`))?
the full image reference (including the digest) could - as discussed yesterday - also be stored (better name required^^)
maybe DIGEST_IMAGE_REFERENCE?

also: how about adding a method to `ContainerImage`, that exposes the hashy-reference-value? this method could/should also apply caching
additional value-add: we could/should check whether the image-reference already contains a digest-tag, in which case no lookup would be required.

container/registry.py
AndreasBurger: Fair enough. I don't think that we have many instances of this coding running long enough for this to be a likely occurrence, but I figure that it'd be a real head-scratcher if it _does_ happen.
I'll remove the cache.

ccwienk: you never know what functionality we shall mv to `whd` in the future. which is a long-running instance

product/model.py
```diff
@@ -396,6 +399,13 @@ def image_reference(self):
     def image_name(self):
         return self.raw.get('name')
 
+    @functools.lru_cache
+    def content_addressable_id(self):
```
ccwienk: not sure I like the name. how about `digest_image_reference`?
or `image_reference_with_digest`
or maybe adding a parameter to `image_reference` (e.g tag='symbolic'|'digest')
in the latter case, only the `image_digest` function would require caching

PR #402, opened 2020-02-26
Allow attempting to send mail without credentials

AndreasBurger:


model/email.py
```diff
@@ -39,13 +39,17 @@ def sender_name(self):
     def credentials(self):
         return EmailCredentials(self.raw.get('credentials'))
 
+    def has_credentials(self):
+        return self.raw.get('credentials') is not None
```
ccwienk: how about checking for truthyness (and still require the attr to be present)? in that case, we would explicitly configure `credentials: None` (or `nil`, or `null`, or `false`, or `~`), which seems more expressive than simply omitting the attr

PR #408, opened 2020-04-02
ConcourseClient: Improve logging

AndreasBurger:
Replace cryptic error msg for resources of custom Concourse resource-types with a more helpful one.

### DIFF
--- a/concourse/client/api.py
+++ b/concourse/client/api.py
@@ -219,11 +219,28 @@ class ConcourseApiBase(object):
         return BuildEvents(response, self)
 
     @ensure_annotations
-    def trigger_resource_check(self, pipeline_name: str, resource_name: str):
+    def trigger_resource_check(self, pipeline_name: str, resource_name: str, retries:int=5):
         url = self.routes.resource_check(pipeline_name=pipeline_name, resource_name=resource_name)
+
         # Resource checks are triggered by a POST with an empty JSON-document as body against
         # the resource's check-url
-        self._post(url, body='{}')
+        response = self.request_builder.post(url, check_http_code=False, body='{}', headers={})
+
+        if response.ok:
+            return
+
+        if response.status_code == 500 and response.reason.startswith('parent type has no version'):
+            if retries > 0:
+                self.trigger_resource_check(pipeline_name, resource_name, retries-1)
+            else:
+                response.reason = (
+                    f"Unable to check resource '{resource_name}' in pipeline "
+                    f"'{pipeline_name}'. If this pipeline was recently deployed, please "
+                    "try again in a minute."
+                )
+                response.raise_for_status()
+        else:
+            response.raise_for_status()
 
     @ensure_annotations
     def resource_versions(self, pipeline_name: str, resource_name: str):


concourse/client/api.py
```diff
@@ -221,9 +221,24 @@ def build_events(self, build_id):
...
         url = self.routes.resource_check(pipeline_name=pipeline_name, resource_name=resource_name)
-        # Resource checks are triggered by a POST with an empty JSON-document as body against
-        # the resource's check-url
-        self._post(url, body='{}')
+        max_retries = 5
+
+        def _try_resource_check(current_attempt):
+            # Resource checks are triggered by a POST with an empty JSON-document as body against
+            # the resource's check-url
+            response = self.request_builder.post(url, check_http_code=False, body='{}', headers={})
```
ccwienk, line 229: how about passing-through the `check_http_code` argument in `self._post`?

concourse/client/api.py
```diff
@@ -221,9 +221,24 @@ def build_events(self, build_id):
     @ensure_annotations
     def trigger_resource_check(self, pipeline_name: str, resource_name: str):
```
ccwienk, line 222: I would have preferred (also, but not also) to decrease nesting, to specify the retries here. this way, it would:

a) be explicit that retries are done
b) configurable (a caller could decide not to do retries)

Also, did I mention the reduced nesting?

-> init the `retries` as default value, like so:
```
trigger_resource_check(...... retries=5) # or retries_left?
```

concourse/client/api.py
```diff
@@ -221,9 +221,24 @@ def build_events(self, build_id):
...
-        # Resource checks are triggered by a POST with an empty JSON-document as body against
-        # the resource's check-url
-        self._post(url, body='{}')
+        max_retries = 5
+
+        def _try_resource_check(current_attempt):
+            # Resource checks are triggered by a POST with an empty JSON-document as body against
+            # the resource's check-url
+            response = self.request_builder.post(url, check_http_code=False, body='{}', headers={})
+            if response.status_code == 500:
```
ccwienk, line 230: do an early exit:
```
if response.ok:
  return # early exit, nothing to do
```

concourse/client/api.py
ccwienk, line 229: another nit: we should use `json={}` (i know this was not there in the original code)

concourse/client/api.py
```diff
@@ -221,9 +221,24 @@ def build_events(self, build_id):
...
+                if response.reason.startswith('parent type has no version'):
+                    if current_attempt <= max_retries:
+                        return _try_resource_check(current_attempt+1)
+                    else:
+                        raise RuntimeError(
+                            f"Unable to check resource '{resource_name}' in pipeline "
```
ccwienk, line 236: hm.. i do not like the duplication (plus: two different ways to raise)

concourse/client/api.py
```diff
@@ -221,9 +221,24 @@ def build_events(self, build_id):
...
+                            f"'{pipeline_name}'. If this pipeline was recently deployed, please "
+                            "try again in a minute."
+                        )
+            else:
```
ccwienk, line 240: i think this else could be removed

how about determining the error-type (500 + reason-str), and store it (e.g. `is_pending_initialisation = True`, or similar).

then either early-exit for a retry, or raise (either with special, or generic error-msg)? should be doable w/o nesting deeper than two

ccwienk:
lgtm

PR #409, opened 2020-04-06
Lint pipeline defs and branch cfgs while replicating

AndreasBurger:
To avoid breaking pipeline replication, linting errors in the branch cfg currently only print a warning.

### DIFF
--- a/ci/util.py
+++ b/ci/util.py
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import collections
 import os
 import pathlib
 import shutil
@@ -28,6 +29,47 @@ from urllib.parse import urlunparse
 
 class Failure(RuntimeError, ValueError):
     pass
+
+
+class LintingError(Failure):
+    pass
+
+
+class LintingResult:
+    def __init__(self, problems):
+        problems_dict = collections.defaultdict(list)
+        for p in problems:
+            problems_dict[self._normalise_problem_level(p)].append(p)
+        self.problems_dict = problems_dict
+
+    def _normalise_problem_level(self, p):
+        # Yamllint uses a dict with mappings for both int->level and level->int. Consistently use
+        # the int representation
+        if isinstance(p.level, str):
+            return yamllint.linter.PROBLEM_LEVELS[p.level]
+        else:
+            return p.level
+
+    def problems(self):
+        return self.problems_dict.items()
+
+    def has_problems(self):
+        return bool(self.problems_dict)
+
+    def max_level(self):
+        if self.problems:
+            max(self.problems_dict.keys())
+        return 0
+
+    def __str__(self):
+        if not self.has_problems():
+            return "No linting problems found"
+
+        return '\n'.join(s for s in [
+            f'[{yamllint.linter.PROBLEM_LEVELS[l]}] {p}'
+            for l in self.problems_dict.keys()
+            for p in self.problems_dict[l]
+        ])
 
 
 def _set_cli(is_cli: bool):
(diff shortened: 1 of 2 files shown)


concourse/enumerator.py
```diff
@@ -235,7 +256,11 @@ def _scan_repository_for_definitions(
 
             verbose('from repo: ' + repository.name + ':' + branch_name)
             try:
-                definitions = load_yaml(definitions.decoded.decode('utf-8'))
+                decoded_definitions = definitions.decoded.decode('utf-8')
+                info(f'Linting pipeline_definitions for {repository}')
+                lint_yaml(decoded_definitions, config=REPLICATION_LINTING_CFG)
+                definitions = load_yaml(decoded_definitions)
```
ccwienk, line 262: how about adding a `lint` (+lint-cfg) flag to `load_yaml`?

concourse/enumerator.py
```diff
@@ -192,9 +208,14 @@ def _branch_cfg_or_none(
                 path='branch.cfg',
                 ref='refs/meta/ci',
             ).decoded.decode('utf-8')
-            return BranchCfg(raw_dict=load_yaml(branch_cfg))
+            info(f'Linting branch cfg for {repository}')
+            lint_yaml(branch_cfg, config=REPLICATION_LINTING_CFG)
         except NotFoundError:
             return None # no branch cfg present
+        except LintingError as e:
```
ccwienk, line 215: I agree with introducing a new exception. However: why not fail?

concourse/enumerator.py
```diff
@@ -24,20 +24,36 @@
...
 from model.base import ModelBase, NamedModelElement
 from concourse.factory import RawPipelineDefinitionDescriptor
 import ccc.github
 
 
+REPLICATION_LINTING_CFG = {
+    'extends': 'relaxed',
+    'rules': {
+        'key-duplicates': {
+            'level': 'error',
```
ccwienk, line 48: is this not the default?
```d051236 ~/src/cc-utils$ yamllint -drelaxed x.yaml                        
x.yaml
  2:1       error    duplication of key "a" in mapping  (key-duplicates)

d051236 ~/src/cc-utils$ cat x.a                                          
cat: x.a: No such file or directory
d051236 ~/src/cc-utils$ cat x.yaml                                       
a: 42
a: 10
```

concourse/enumerator.py
```diff
@@ -24,20 +24,36 @@
...
+        },
+        'new-line-at-end-of-file': {
+            'level': 'warning',
```
ccwienk, line 51: if you remove this relaxation, we could use the default profile, which I would prefer tbh

ci/util.py
```diff
@@ -262,24 +266,43 @@ def _count_elements(value, count=0, max_elements_count=100000):
...
-            elif level > yamllint.linter.PROBLEM_LEVELS['warning']:
-                error(str(problem))
-
-            worst_level = max(worst_level, level)
-    if worst_level >= yamllint.linter.PROBLEM_LEVELS['error']:
-        fail('found errors whilst linting (see above)')
+        lint_yaml(f.read(), linter_config)
+
+
+def _parse_yamllint_problem_level(problem):
```
ccwienk, line 275: nit: this is not _parsing_, but more a normalisation / casting

ci/util.py
```diff
@@ -262,24 +266,43 @@ def _count_elements(value, count=0, max_elements_count=100000):
...
+    cfg = yamllint.config.YamlLintConfig(yaml.dump(config))
+
+    problems = [p for p in yamllint.linter.run(input=input, conf=cfg)]
+    warnings = [
+        str(p) for p in problems
+        if _parse_yamllint_problem_level(p) == yamllint.linter.PROBLEM_LEVELS['warning']
+    ]
+    errors = [
+        str(p) for p in problems
+        if _parse_yamllint_problem_level(p) > yamllint.linter.PROBLEM_LEVELS['warning']
```
ccwienk, line 294: this is wrong

ci/util.py
```diff
@@ -262,24 +266,43 @@ def _count_elements(value, count=0, max_elements_count=100000):
...
+def _parse_yamllint_problem_level(problem):
+    # Yamllint uses a dict with mappings for both int->level and level-> int. Consistently use
+    # the int representation
+    if isinstance(problem.level, str):
+        return yamllint.linter.PROBLEM_LEVELS[problem.level]
+    else:
+        return problem.level
+
+
+def lint_yaml(input, config={'extends': 'relaxed'}):
```
ccwienk, line 284: i think the processing / conversion of yamllint's results should be separated into a dedicated function, if you are already refactoring it. this function could return a dataclass, or maybe a dict with findings grouped by severity

lint_yaml, and lint_yaml_file would then instrument this, and - depending on the result - raise.

concourse/enumerator.py
AndreasBurger: It is. Not a good day to read the docs, I guess. Removed

ci/util.py
AndreasBurger: I disagree.

ci/util.py
```diff
@@ -30,6 +31,45 @@ class Failure(RuntimeError, ValueError):
...
+
+    def problems(self, level:int):
+        return self.problems_dict.get(level, ())
+
+    def has_problems(self):
+        return bool(self.problems)
+
+    def max_level(self):
+        if self.problems:
+            return sorted(self.problems.keys())[-1]
```
ccwienk, line 61: how about `max`?

ci/util.py
```diff
@@ -30,6 +31,45 @@ class Failure(RuntimeError, ValueError):
...
+        return 0
```
ccwienk, line 62: -1?

ci/util.py
```diff
@@ -30,6 +31,45 @@ class Failure(RuntimeError, ValueError):
...
+
+    def problems_iter(self):
```
ccwienk, line 64: tbh, I think this is a bit heavyweight, since it basically exposes the (almost same) functionality of a plain `dict.items()`

ci/util.py
```diff
@@ -262,24 +302,44 @@ def _count_elements(value, count=0, max_elements_count=100000):
...
+
+def lint_yaml(input, config={'extends': 'relaxed'}):
+
+    linting_result = _lint_yaml(input=input, config=config)
+
+    if not linting_result.has_problems():
+        return
+
+    if linting_result.max_level() <= yamllint.linter.PROBLEM_LEVELS['warning']:
+        warning('Found some problems while linting:\n\t' + '\n\t'.join(
```
ccwienk, line 336: nit: rather use f-str? also, rather deduplicate the msg, please
also: why not add `__str__` to linting_result. would make a lot more sense to me than the `problems_iter` method.

concourse/enumerator.py
ccwienk: mh?

concourse/enumerator.py
```diff
@@ -235,7 +243,11 @@ def _scan_repository_for_definitions(
 
             verbose('from repo: ' + repository.name + ':' + branch_name)
             try:
-                definitions = load_yaml(definitions.decoded.decode('utf-8'))
+                decoded_definitions = definitions.decoded.decode('utf-8')
+                info(f'Linting pipeline_definitions for {repository}')
```
ccwienk, line 247: how about also stating the branch-name here?

concourse/enumerator.py
AndreasBurger: originally, the Idea was to avoid failing our central replication. I've moved the exception handling so that this is done in a more elegant fashion

concourse/enumerator.py
AndreasBurger: done.
Not _strictly_ part of this PR, however.

ci/util.py
```diff
@@ -30,6 +31,47 @@ class Failure(RuntimeError, ValueError):
...
+    def _normalise_problem_level(self, p):
+        # Yamllint uses a dict with mappings for both int->level and level->int. Consistently use
+        # the int representation
+        if isinstance(p.level, str):
+            return yamllint.linter.PROBLEM_LEVELS[p.level]
+        else:
+            return p.level
+
+    def problems(self):
+        return self.problems_dict.items()
```
ccwienk, line 54: O.o
why not simple access the problems-dict?

PR #413, opened 2020-04-17
fix whd erros

RaphaelVogel:
* access the labels in the PR webhook correctly
* do not trigger resource if PullRequestAction.OPENED or SYNCHRONIZE
  only needs to be done if action = PullRequestAction.LABELED

Fixes #

**Release note**:

```improvement operator

```


whd/dispatcher.py
```diff
@@ -122,6 +122,7 @@ def dispatch_pullrequest_event(self, pr_event):
 
             if pr_event.action() in [PullRequestAction.OPENED, PullRequestAction.SYNCHRONIZE]:
                 self._set_pr_labels(pr_event, resources)
+                continue
```
ccwienk, line 125: coulod you add a brief comment explaining why it is ok to skip here?

PR #416, opened 2020-04-23
fix setup.base.py

RaphaelVogel:

Fixes #

**Release note**:

```improvement operator

```

### DIFF
--- a/setup.base.py
+++ b/setup.base.py
@@ -5,21 +5,14 @@ own_dir = os.path.abspath(os.path.dirname(__file__))
 
 
 def requirements():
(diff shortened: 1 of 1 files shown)


setup.base.py
```diff
@@ -5,21 +5,17 @@
...
-            if not 'yaml' in line:
-                continue
-            if not 'termcolor' in line:
-                continue
-            if not 'urllib3' in line:
-                continue
-            if not 'elasticsearch' in line:
-                continue
-            yield line
+            if any(s in line for s in in_list):
```
AndreasBurger, line 15: how about:
```
if any(s in line for s in in_list):
  yield line
```

PR #418, opened 2020-04-30
Add limited support for reloading the cfg_set in whd

AndreasBurger:
When trying to render a pipeline for a new JobMapping that is not yet part of the cfg_set the WHD is given on initialisation, the WHD will encounter an AttributeError. Use this error as signal to attempt
a single reload of the cfg_set.

### DIFF
--- a/whd/dispatcher.py
+++ b/whd/dispatcher.py
@@ -19,7 +19,17 @@ import logging
 import time
 import traceback
 
+import ccc.elasticsearch
+import ccc.secrets_server
+import ci.util
+import concourse.client
+
+from .pipelines import update_repository_pipelines
+from concourse.enumerator import JobMappingNotFoundError
+from github.util import GitHubRepositoryHelper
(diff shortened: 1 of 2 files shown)


whd/dispatcher.py
```diff
@@ -84,11 +87,27 @@ def dispatch_push_event(self, push_event):
...
-                cfg_set=self.cfg_set,
-                whd_cfg=self.whd_cfg,
-            )
+            try:
+                update_repository_pipelines(
+                    repo_url=push_event.repository().repository_url(),
+                    cfg_set=self.cfg_set,
+                    whd_cfg=self.whd_cfg,
+                )
+            except AttributeError as ae:
```
ccwienk, line 96: as just discussed:
- check for the missing attribute (`hasattr`)
- define a new exception (e.g. `OutdatedCfgError`)
- raise this specific exception iff the specific attr is amiss
- do the retry only upon this specific exception

PR #419, opened 2020-05-12
Fix determine previous build

AndreasBurger:
From version `6.0.0` on, Concourse supports re-triggering builds. These builds are assigned non-integer build names, e.g. `5.4`. Previously, our notification logic assumed that these build-names are integers.
With this PR, the 20 most recent builds are examined to determine the predecessor of a given build.

### DIFF
--- a/concourse/client/api.py
+++ b/concourse/client/api.py
@@ -189,6 +189,22 @@ class ConcourseApiBase(object):
         return builds
 
     @ensure_annotations
+    def previous_build(self, pipeline_name: str, job_name: str, build_id: int):
+        '''Iterates over the 20 most recent builds and returns the predecessor of the one with
+        the given build id, or `None` if the predecessor could not be determined.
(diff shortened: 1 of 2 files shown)


concourse/client/api.py
```diff
@@ -188,6 +188,22 @@ def job_builds(self, pipeline_name: str, job_name: str):
...
+    @ensure_annotations
+    def previous_build(self, pipeline_name: str, job_name: str, build_id: int):
+        '''Iterates over the 20 most recent builds and returns the predecessor of the one with
+        the given build id, or `None` if the predecessor could not be determined.
+        '''
+        builds = self.job_builds(pipeline_name, job_name)
+
+        subsequent_build_encountered = False
+        for build in builds[:20]:
+            if subsequent_build_encountered:
```
ccwienk, line 200: ??? why not immediately return?

PR #423, opened 2020-06-03
apply email blacklist before send

RaphaelVogel:

Fixes #

**Release note**:

```improvement operator

```

### DIFF
--- a/mailutil.py
+++ b/mailutil.py
@@ -127,6 +127,8 @@ def _send_mail(
     )
 
     recipients.update(cc_recipients)
+    recipients = email_cfg.filter_recipients(recipients)
+
     smtp_server.send_message(msg=mail, to_addrs=recipients)  # from_addr is taken from header
 
 
--- a/model/email.py
+++ b/model/email.py
@@ -12,6 +12,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import typing
 
 from model.base import (
     BasicCredentials,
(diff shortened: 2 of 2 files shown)


mailutil.py
```diff
@@ -127,6 +127,11 @@ def _send_mail(
     )
 
     recipients.update(cc_recipients)
+
+    if email_cfg.has_blacklist():
+        blacklist = email_cfg.blacklist()
+        recipients = list(filter(lambda r: r not in blacklist, recipients))
```
ccwienk, line 133: please use a list comprehension

mailutil.py
```diff
@@ -127,6 +127,11 @@ def _send_mail(
     )
 
     recipients.update(cc_recipients)
+
+    if email_cfg.has_blacklist():
```
ccwienk, line 131: rather add an attr `blacklist_entries`. booleans are almost never a good idea

OR (pbly even better): add a method `filter_recipients` to the cfg model class

model/email.py
```diff
@@ -44,6 +44,14 @@ def has_credentials(self):
             return True
         return False
 
+    def blacklist(self):
+        return self.raw.get('blacklist')
+
+    def has_blacklist(self):
```
ccwienk, line 50: not required

model/email.py
```diff
@@ -44,6 +44,14 @@ def has_credentials(self):
             return True
         return False
 
+    def blacklist(self):
+        return self.raw.get('blacklist')
```
ccwienk, line 48: default to empty collection of absent - problem solved, and one codepath less

also, as discussed: add a method doing the filtering, instead

model/email.py
```diff
@@ -44,6 +45,12 @@ def has_credentials(self):
             return True
         return False
 
+    def filter_recipients(self, recipients:typing.Iterable[str]):
+        blacklist = self.raw.get('blacklist')
```
ccwienk, line 49: nit: this `if` is not needed, if using the empty tuple as fallback value, like so:

```
blacklist = self.get.get('blacklist', ())
```

PR #429, opened 2020-06-17
Start multible processes and threads

RaphaelVogel:

Fixes #

**Release note**:

```improvement operator

```

### DIFF
--- a/cli/gardener_ci/whdutil.py
+++ b/cli/gardener_ci/whdutil.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 
 import logging
+import multiprocessing
 
 import ci.util
 ci.util.ctx().configure_default_logging(stdout_level=logging.INFO)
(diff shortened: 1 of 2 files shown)


whd/webhook.py
```diff
@@ -33,7 +34,7 @@ def __init__(
         self.whd_cfg = whd_cfg
         self.dispatcher = GithubWebhookDispatcher(cfg_set=cfg_set, whd_cfg=whd_cfg)
 
-    def on_post(self, req, resp):
+    def on_post_handling(self, req, resp):
```
ccwienk, line 37: firstly, please do not name the method like that - as per falcon's convention, this method name suggests that this be hooked to a POST with "suffix" `_handling`. marking the method as private would be the better choice (e.g. _on_post)

also, this is not the suggested semantics, as this will not return early and thus avoid timeouts.

please do the following:

- identify the long-running part of the coding
- define an inline function handling this part
- start a thread, passing said inline function as target
- after starting the thread, return http-200

cli/gardener_ci/whdutil.py
```diff
@@ -41,11 +42,12 @@ def start_whd(
 
     if production:
         import bjoern
+        multiprocessing.set_start_method('fork')
 
         def serve():
             bjoern.run(app, any_interface, port, reuse_port=True)
         for _ in range(workers - 1):
-            serve()
+            multiprocessing.Process(target=serve).start()
```
ccwienk, line 50: for the sake of readability, please assign the process to a variable, first

PR #430, opened 2020-06-23
Pinning

RaphaelVogel:

Fixes #

**Release note**:

```improvement operator

```

### DIFF
--- a/concourse/client/model.py
+++ b/concourse/client/model.py
@@ -46,6 +46,9 @@ class ResourceVersion(ModelBase):
     Wraps a single result returned from concourse's `<resource>/versions` route.
     Both `metadata` and `version` adhere to a schema specific to the resource type.
     '''
+    def id(self):
+        return self.raw['id']
+
     def type(self):
         return self.raw['type']
 
@@ -76,9 +79,37 @@ class PipelineConfig(object):
             warning('Pipeline did not contain resource definitions: {p}'.format(p=name))
             raise ValueError()
         self.resources = map(lambda r: Resource(r, self), resources)
+        jobs = self.raw.get('jobs', None)
+        if not jobs:
+            warning(f'Pipeline {name} did not contain job definitions')
+            raise ValueError()
+        self.jobs = (Job(job, self) for job in jobs)
 
     def resources_of_types(self, types):
         return filter(lambda r: r.type in types, self.resources)
+
+
+class Job:
+    '''
+    Wrapper around the dictionary representing a job as part of a
+    concourse.PipelineConfig
+
+    Not intended to be instantiated by users of this module
+    '''
+    @ensure_annotations
+    def __init__(self, raw: dict, pipeline: PipelineConfig):
+        self.pipeline = pipeline
+        self.concourse_api = pipeline.concourse_api
+        self.raw = raw
+        self.name = raw['name']
+        self.plan = raw['plan']
+
+    def is_triggered_by_resource(self, resource_name: str):
+        get_steps = [item for item in self.plan if 'get' in item]
+        for get_step in get_steps:
+            if get_step['get'] == resource_name and get_step['trigger']:
+                return True
+        return False
 
 
 class Resource(object):
@@ -221,6 +252,24 @@ class BuildPlan(ModelBase):
                             return task_id
         return find_tid(plan)
 
+    def contains_version_ref(self, resource_version_ref: str):
+        def has_version_ref(p):
+            if 'ref' in p:
+                ref = p.get('ref')
+                if ref == resource_version_ref:
+                    return True  # end recursion
+            for k, v in p.items():
+                if isinstance(v, dict):
+                    if has_version_ref(v):
+                        return True
+                if isinstance(v, list):
+                    for element in v:
+                        if has_version_ref(element):
+                            return True
+            return False
+
+        return has_version_ref(self.raw.get('plan'))
+
 
 class BuildEvents(object):
     '''
@@ -324,6 +373,24 @@ class Worker(ModelBase):
 
     def name(self):
         return self.raw['name']
+
+
+class PipelineResource:
+    '''
+    Wrapper around the dictionary representing a pipeline resource
+
+    Not intended to be instantiated by users of this module
+    '''
+    @ensure_annotations
+    def __init__(self, raw:dict, concourse_api):
+        self.concourse_api = concourse_api
+        self.raw = raw
+        self.name = raw['name']
+
+    def is_pinned(self):
+        if self.raw.get('pinned_version'):
+            return True
+        return False
 
 
 class BuildStatus(Enum):
--- a/whd/dispatcher.py
+++ b/whd/dispatcher.py
@@ -25,6 +25,7 @@ import ccc.secrets_server
 import ci.util
 import concourse.client
 
+from concourse.client.model import BuildStatus
 from .pipelines import update_repository_pipelines
 from concourse.enumerator import JobMappingNotFoundError
 from github.util import GitHubRepositoryHelper
@@ -153,6 +154,11 @@ class GithubWebhookDispatcher(object):
 
                 logger.info(f'triggering resource check for PR number: {pr_event.number()}')
                 self._trigger_resource_check(concourse_api=concourse_api, resources=resources)
+                self._ensure_pull_requests_jobs_are_run(
+                    concourse_api=concourse_api,
+                    matching_resources=resources,
+                    pr_event=pr_event
+                )
 
         thread = threading.Thread(target=_set_labels)
         thread.start()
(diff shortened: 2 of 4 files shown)


whd/dispatcher.py
```diff
@@ -211,6 +212,57 @@ def _set_pr_labels(self, pr_event, resources):
...
+                for resource_version in self._resource_version_ids_not_build(
+                    resource_versions, builds
+                ):
+                    logger.info(
+                        f"Pin resource version {resource_version.version()} "
+                        f"and trigger build for job {job.name}"
+                    )
+                    self._pin_trigger_unpin(job, resource.name, resource_version.id())
+
+    def _resource_version_ids_not_build(self, resource_versions, builds):
```
ccwienk, line 242: at the very least, this should be named `...builT` (past tense)

whd/dispatcher.py
```diff
@@ -211,6 +212,57 @@ def _set_pr_labels(self, pr_event, resources):
...
+        for resource_version in resource_versions[:10]:
+            resource_version_in_build = False
+            for build in builds[-10:]:
```
ccwienk, line 245: rather process from most recent to less recent; also: add a comment stating _why_ we only consider the last couple of build executions (also: rather more than 10; maybe as much as 50, even? I do not think this will cause a performance issue - as opposed to iterating over _all_ build executions, of course)

whd/dispatcher.py
```diff
@@ -211,6 +212,57 @@ def _set_pr_labels(self, pr_event, resources):
...
+                build_plan = build.plan()
+                if build_plan.contains_version_ref(resource_version.version()['ref']):
+                    resource_version_in_build = True
+                    break
+            if not resource_version_in_build:
+                yield resource_version
+
+    def _pin_trigger_unpin(job, resource_name, resource_version_id):
```
ccwienk, line 253: could you please add some conjunctions to the method's name, to make its semantics easier to grasp?
-> pin_and_trigger_and_unpin

or maybe (less lowlevel): trigger_execution_with_pinned_resource_version

also, since those methods (I am referring to most parts of this PR) implement rather complex, and unintuitive semantics, I would very much vote for adding _brief_ doc-strs explaining said semantics

whd/dispatcher.py
```diff
@@ -211,6 +212,57 @@ def _set_pr_labels(self, pr_event, resources):
...
+        concourse_api = job.concourse_api
+        concourse_api.pin_resource_version(
```
ccwienk, line 255: this needs to be guarded against a case where the resource version has already been pinned

whd/dispatcher.py
```diff
@@ -211,6 +212,57 @@ def _set_pr_labels(self, pr_event, resources):
...
+            pipeline_name=job.pipeline.name,
+            resource_name=resource_name,
+            resource_version_id=resource_version_id,
+        )
+        concourse_api.trigger_build(job.pipeline, job.name)
+        concourse_api.unpin_resource(
```
ccwienk, line 261: consider using a ctx mgr (which could also handle the guarding against another pinning)

concourse/client/model.py
```diff
@@ -76,11 +79,39 @@ def __init__(self, raw:dict, concourse_api, name:str):
             warning('Pipeline did not contain resource definitions: {p}'.format(p=name))
             raise ValueError()
         self.resources = map(lambda r: Resource(r, self), resources)
+        jobs = self.raw.get('jobs', None)
+        if not jobs:
+            warning(f'Pipeline {name} did not contain job definitions')
+            raise ValueError()
+        self.jobs = map(lambda j: Job(j, self), jobs)
```
ccwienk, line 86: please use list comprehension

concourse/client/model.py
```diff
@@ -76,11 +79,39 @@ def __init__(self, raw:dict, concourse_api, name:str):
...
 
     def resources_of_types(self, types):
         return filter(lambda r: r.type in types, self.resources)
 
 
+class Job(object):
```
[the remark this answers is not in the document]
ccwienk: ..

concourse/client/model.py
```diff
@@ -76,11 +79,39 @@ def __init__(self, raw:dict, concourse_api, name:str):
...
+    concourse.PipelineConfig
+
+    Not intended to be instantiated by users of this module
+    '''
+    @ensure_annotations
+    def __init__(self, raw: dict, pipeline: PipelineConfig):
+        self.pipeline = pipeline
+        self.concourse_api = pipeline.concourse_api
+        self.raw = raw
+        self.name = raw['name']
```
ccwienk, line 104: if not using a dataclass, rather stick to using getters to be consistent with the other code in this module

concourse/client/model.py
```diff
@@ -76,11 +79,39 @@ def __init__(self, raw:dict, concourse_api, name:str):
...
+        self.plan = raw['plan']
+
+    def has_triggering_resource_in_get_step(self, resource_name: str):
```
ccwienk, line 107: how about naming it `is_triggered_by_resource`?

concourse/client/model.py
```diff
@@ -76,11 +79,39 @@ def __init__(self, raw:dict, concourse_api, name:str):
...
+        get_steps = filter(lambda p: 'get' in p, self.plan)
```
ccwienk, line 108: use a generator comprehension, inline it

concourse/client/model.py
```diff
@@ -221,6 +252,24 @@ def find_tid(p):
                             return task_id
         return find_tid(plan)
 
+    def contains_version_ref(self, resource_version_ref: str):
+        def has_version_ref(p):
```
ccwienk, line 256: is it really necessary to recursively process this?

concourse/client/model.py
```diff
@@ -326,6 +375,24 @@ def name(self):
...
+
+    Not intended to be instantiated by users of this module
+    '''
+    @ensure_annotations
+    def __init__(self, raw:dict, concourse_api):
+        self.concourse_api = concourse_api
+        self.raw = raw
+        self.name = raw['name']
+
+    def has_pinned_version(self):
```
ccwienk, line 390: nit: rather name this `is_pinned`? as this is the semantics that this means

also, it might be useful to expose the `pinned_version`, actually (e.g. to check if the pinned version is maybe the one we would like to pin in the first place)

whd/dispatcher.py
```diff
@@ -211,6 +213,106 @@ def _set_pr_labels(self, pr_event, resources):
                     f" in '{repository_path}' found. Ignoring, since they are not an org member'."
                 )
 
+    def _start_not_triggered_jobs(self, concourse_api, resources):
```
ccwienk, line 216: please rename this method, as already remarked before. this is a very confusing name (for me).

how about: `ensure_pull_requests_jobs_are_run` or sth similar?

also, this method is nested way too deeply, and does too many things. the filtering for PR-jobs e.g. could be done before.

ideally, there should be:
- one method that accepts a PR, and determines all related PR-resource versions + jobs triggered by it
- one method that accepts a PR-resource version and a job, that will ensure said job is executed exactly once for the given PR resource version


as a general note, please refrain from using `filter` / `map` built-ins, and rather use dict, set, list, generator comprehensions (depending on what is needed), unless there is a good reason not to

whd/dispatcher.py
```diff
@@ -211,6 +213,106 @@ def _set_pr_labels(self, pr_event, resources):
...
+        check all PR resource versions for which no build exists. If a resource version has
+        no build, pin this resource version, trigger the corresponding job and unpin the resource
+        '''
+        for resource in resources:
+            if resource.type != "pull-request":
+                continue
+
+            resource_versions = concourse_api.resource_versions(
+                pipeline_name=resource.pipeline.name,
+                resource_name=resource.name
```
ccwienk, line 227: as noted above: we know the specific PR for which we want to do this. so there is no need to process _all_ resources, is there?

concourse/client/model.py
```diff
@@ -76,11 +79,39 @@ def __init__(self, raw:dict, concourse_api, name:str):
             warning('Pipeline did not contain resource definitions: {p}'.format(p=name))
             raise ValueError()
         self.resources = map(lambda r: Resource(r, self), resources)
+        jobs = self.raw.get('jobs', None)
+        if not jobs:
+            warning(f'Pipeline {name} did not contain job definitions')
+            raise ValueError()
+        self.jobs = (Job(job, self) for job in jobs)
```
ccwienk, line 86: this is a generator comprehension. which means that the result may only be iterated over _once_. I do not think this is a good idea (hence my suggestion to use list comprehension earler..)

whd/dispatcher.py
```diff
@@ -211,6 +217,128 @@ def _set_pr_labels(self, pr_event, resources):
...
+                            build_number,
+                            concourse_api,
+                            retries,
+                            sleep_sec=sleep_sec*1.2
+                        )
+
+        concourse_api = job.concourse_api
+        pipeline_name = job.pipeline.name
+        version_pin_lock = threading.Lock()
+        with version_pin_lock:
```
ccwienk, line 329: why the lock? there should be no multithreading involved here, should there?

concourse/client/model.py
```diff
@@ -76,11 +79,39 @@ def __init__(self, raw:dict, concourse_api, name:str):
...
+
+    Not intended to be instantiated by users of this module
+    '''
+    @ensure_annotations
+    def __init__(self, raw: dict, pipeline: PipelineConfig):
+        self.pipeline = pipeline
+        self.concourse_api = pipeline.concourse_api
+        self.raw = raw
+        self.name = raw['name']
+        self.plan = raw['plan']
```
ccwienk, line 105: add a model class also for `Plan`, please

concourse/client/model.py
```diff
@@ -76,11 +79,39 @@ def __init__(self, raw:dict, concourse_api, name:str):
...
+
+    def is_triggered_by_resource(self, resource_name: str):
+        get_steps = [item for item in self.plan if 'get' in item]
+        for get_step in get_steps:
+            if get_step['get'] == resource_name and get_step['trigger']:
```
ccwienk, line 110: avoid processing raw `dicts`, please (see comment above)

concourse/client/model.py
```diff
@@ -326,6 +375,24 @@ def name(self):
...
+
+    Not intended to be instantiated by users of this module
+    '''
+    @ensure_annotations
+    def __init__(self, raw:dict, concourse_api):
+        self.concourse_api = concourse_api
+        self.raw = raw
+        self.name = raw['name']
+
+    def is_pinned(self):
```
ccwienk, line 390: as already mentioned: why not expose the pinned version?

whd/dispatcher.py
```diff
@@ -213,6 +219,130 @@ def _set_pr_labels(self, pr_event, resources):
                     f" in '{repository_path}' found. Ignoring, since they are not an org member'."
                 )
 
+    def _resource_versions_and_jobs(self, concourse_api, matching_resources, pr_event):
+        resource = [r for r in matching_resources if r.type == "pull-request"]
+        if len(resource) != 1:
```
ccwienk, line 224: firstly: please handle _both_ unexpected error cases. otherwise, this will make debugging unnecessarily hard!
secondly: why would it be an error or unexpected to find more than one resource?
thirdly: why would it be unexpected to find none?

ALSO: after reading through `dispatcher.py#_matching_resources()`, I found that:

1. resources are already filtered by type. so why do the filtering again? This is confusing, and unnecessary!
2. resources are already being filtered by repository. so again: why filter again. why validate again? are there reasons to believe that said method is erroneous? In that case, please fix that one, instead of (seemingly) working around a bug there (I would not be aware that there is a bug, tbh)

concourse/client/model.py
ccwienk: let me rephrase: please handle the different expected cases explicitly. I understood that there are a few explicit possible nesting variants. please handle those explicitly. please do _not_ process raw `dicts`, create model classes instead.

whd/dispatcher.py
```diff
@@ -213,6 +219,130 @@ def _set_pr_labels(self, pr_event, resources):
...
+            logger.warning(f"none or more than on resource found for PR {pr_event.number()}")
+            return
+
+        resource = resource[0]
```
ccwienk, line 228: no!
there can be more than one resource version

whd/dispatcher.py
```diff
@@ -213,6 +219,130 @@ def _set_pr_labels(self, pr_event, resources):
...
+                f"Pin resource version {resource_version.version()} "
+                f"and trigger build for job {job.name}"
+            )
+            self._trigger_build_with_pinned_resource_version(
+                job=job,
+                resource_name=resource.name,
+                resource_version_id=resource_version.id()
+            )
+
+    def _resource_version_ids_not_built(self, resource_versions, builds):
```
ccwienk, line 282: could you please rather use a more generic term such as `triggered`, or `run`, or `executed`. not all jobs are builds. also, what the jobs do is completely irrelevant at this point. we care about them to be triggered.

whd/dispatcher.py
```diff
@@ -213,6 +219,130 @@ def _set_pr_labels(self, pr_event, resources):
...
+        '''
+        yield resource versions which have not been built by checking if the resource version
```
ccwienk, line 284: nit: the resource versions are not built. we want to find those jobs that should have been triggered for a given resource version, that were not yet triggered. It is very hard to understand this redundant filtering tbh.

whd/dispatcher.py
```diff
@@ -213,6 +219,130 @@ def _set_pr_labels(self, pr_event, resources):
...
+                    break
+            if not resource_version_in_build:
+                yield resource_version
+
+    def _trigger_build_with_pinned_resource_version(self, job, resource_name, resource_version_id):
+        '''
+        pin the passed resource version, trigger the corresponding job and unpin resourcce
+        version again
+        '''
+        def _job_running(
```
ccwienk, line 302: this is a very confusing name. what is tried to do is to trigger a job, right? so why not call it try_trigger_job, or sth. similar, instead of calling it _job_running, what I would expect to be a function that checks IF the given job is running (which it also does)?

also: is it really necessary to use recursion here?

at any rate, I do not think the sleep should be done at the beginning
i also do not think the retries should be decremented before the first check. the usual pattern I know is to first check for whether we should abort, abort if so, otherwise execute (try), then sleep (in case of the target state not yet reached, then decrement, then run the next try).

whd/dispatcher.py
```diff
@@ -213,6 +219,130 @@ def _set_pr_labels(self, pr_event, resources):
...
+            pipeline_name, job_name, build_number, concourse_api, retries=10, sleep_sec=3
+        ):
+            time.sleep(sleep_sec)
+            retries -= 1
+            if retries < 0:
+                logger.warning(
+                    f"Job {job_name} with build number {build_number} cannot be started"
```
ccwienk, line 309: nit: _could_ not be started
using present tense indicates to me that it was apparent from the start that triggering is not possible (for whatever reason). what actually happened is that we failed to trigger the job after several attempts (which might be worth mentioning)

whd/dispatcher.py
```diff
@@ -213,6 +219,130 @@ def _set_pr_labels(self, pr_event, resources):
...
+                            concourse_api,
+                            retries,
+                            sleep_sec=sleep_sec*1.2
+                        )
+
+        concourse_api = job.concourse_api
+        pipeline_name = job.pipeline.name
+        version_pin_lock = threading.Lock()
+        with version_pin_lock:
+            concourse_api.pin_resource_version(
```
ccwienk, line 334: as already discussed earlier:

- check whether the version is already pinned
  - if so: is it already pinned to _our_ version
    - if so: no pinning required - just trigger
  - if not: wait for the version to be unpinned again
   - do this using try-sleep-retry.... for some 1..2m maybe
   - however, check whether the version was maybe pinned by someone else than us (a human..)
    - if so, there is no point in waiting - just bail out
    - if so, we should have also written a comment (in particular: add a timestamp)
      - check the timestamp: if it is too old (say 5m+), unpin it (probably this was a leftover / stems from a bug)

Also, as discussed earlier, there must be a transactional semantics for the resource-pinning:

Use (preferrably) a context mgr. Or - if you do not like using context mgrs, for whichever reason - use a try-finally (which is semantically equivalent).

If the resource is not yet pinned (see above for cases where it is):

- pin it
- add a machine-readable note to it, stating our username, maybe, and a comment (e.g. auto-pinning for PR xyz, ..) and a timestamp ("now")
- try to run the job(s) - there could be more than one
- do this with a timeout
- if either all jobs were successfully triggered, or the timeout expires:
- unpin the version again

rfranzke:
Hello, can we somehow speed up getting this PR merged? The CI/CD support for PRs in g/g is not working since ~2w now, affecting productivity and increasing manual effort.

RaphaelVogel:
we can close it

PR #434, opened 2020-06-26
add method to return jobs not been triggered

RaphaelVogel:

Fixes #

**Release note**:

```improvement operator

```

### DIFF
--- a/whd/pinning.py
+++ b/whd/pinning.py
@@ -1,0 +1,103 @@
+# Copyright (c) 2019-2020 SAP SE or an SAP affiliate company. All rights reserved. This file is
+# licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import logging
+import typing
+from dataclasses import dataclass
+
+from .model import(
+    PullRequestEvent,
+)
+from concourse.client.model import (
+    ResourceVersion,
+    Job,
+    Resource,
+)
+
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass
+class JobResourceVersion:
+    job: Job
+    resource: Resource
+    resource_version: ResourceVersion
+
+
+def jobs_not_triggered(
+    self,
+    concourse_api,
+    resources: typing.Sequence[Resource],
+    pr_event: PullRequestEvent,
+) -> JobResourceVersion:
+    '''
+    @return all jobs and the corresponding resource versions, which have not been triggered
+    '''
+    for resource in resources:
+        pipeline_name = resource.pipeline.name
+        resource_versions = concourse_api.resource_versions(
+            pipeline_name=pipeline_name,
+            resource_name=resource.name,
+            pr_number=pr_event.number(),
+        )
+        pr_resource_versions = [
+            rv for rv in resource_versions if rv.version()['pr'] == str(pr_event.number())
+        ]
+        if not pr_resource_versions:
+            logger.warning(
+                f"no resource versions found for pipeline {pipeline_name} "
+                f"resource {resource.name} and PR {pr_event.number()}"
+            )
+            return
+
+        # get all jobs which should be triggered by the resource
+        jobs_which_should_be_triggered = [
+            job for job in resource.pipeline.jobs if job.is_triggered_by_resource(resource.name)
+        ]
+
+        # check if there is a pr resource version which has not triggered a job.
+        # If so return the job and the corresponding resource version
+        for job in jobs_which_should_be_triggered:
+            builds = concourse_api.job_builds(pipeline_name, job.name)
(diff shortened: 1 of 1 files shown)


whd/dispatcher.py
```diff
@@ -318,6 +318,56 @@ def is_up_to_date(resource, resource_versions):
...
+        for resource_version in resource_versions:
+            resource_version_in_build = False
+            for build in reversed(builds[-30:]):
+                build_plan = build.plan()
+                if build_plan.contains_version_ref(resource_version.version()['ref']):
+                    resource_version_in_build = True
+                    break
+            if not resource_version_in_build:
+                yield resource_version
+
```
AndreasBurger, line 370: Haven't tested it, but think about something like
```python
def _jobs_not_triggered(self, resource_versions, builds):
    builds_to_consider = reversed(builds[-30:])

    yield from [
        rv for rv in resource_versions
        if not any([
            b.plan().contains_version_ref(rv.version()['ref']) for b in builds_to_consider
        ])
    ]
```
This would be a bit quicker to understand ,in my opinion.

Not a deal-breaker, though. I'll leave it up to you.

whd/dispatcher.py
```diff
@@ -318,6 +318,56 @@ def is_up_to_date(resource, resource_versions):
...
+            resource_versions = concourse_api.resource_versions(
+                pipeline_name=pipeline_name,
+                resource_name=resource.name
+            )
+            pr_resource_versions = [
+                rv for rv in resource_versions if rv.version()['pr'] == str(pr_event.number())
+            ]
+            if not pr_resource_versions:
+                logger.warning(
+                    f"no resource versions found for pipeline {resource.pipeline.name} "
```
AndreasBurger, line 336: Please either consistently use `pipeline_name` or inline l.328 and rm l.326

whd/dispatcher.py
```diff
@@ -318,6 +318,56 @@ def is_up_to_date(resource, resource_versions):
             sleep_seconds=sleep_seconds*1.2,
         )
 
+    def _jobs_not_triggered(self, concourse_api, matching_resources, pr_event):
+        '''
+        return all jobs and the corresponding resource versions, which have not been triggered
```
ccwienk, line 323: nit: @return?

whd/dispatcher.py
ccwienk, line 336: nit: use `{resource.pipeline.name=}`

whd/dispatcher.py
```diff
@@ -318,6 +318,56 @@ def is_up_to_date(resource, resource_versions):
...
+        '''
+        for resource in matching_resources:
+            pipeline_name = resource.pipeline.name
+            resource_versions = concourse_api.resource_versions(
+                pipeline_name=pipeline_name,
+                resource_name=resource.name
+            )
+            pr_resource_versions = [
+                rv for rv in resource_versions if rv.version()['pr'] == str(pr_event.number())
```
ccwienk, line 332: as discussed previously: is it really necessary to filter _again_?

whd/dispatcher.py
```diff
@@ -318,6 +318,56 @@ def is_up_to_date(resource, resource_versions):
...
+            ]
+            if not pr_resource_versions:
+                logger.warning(
+                    f"no resource versions found for pipeline {resource.pipeline.name} "
+                    f"resource {resource.name} and PR {pr_event.number()}"
+                )
+                return
+
+            # get all jobs which get triggered by the resource
+            jobs_triggered_by_resource = [
```
ccwienk, line 342: it is the jobs that _should_ be triggered by the resource. so please reflect this in the name (and comment!)

whd/dispatcher.py
```diff
@@ -318,6 +318,56 @@ def is_up_to_date(resource, resource_versions):
...
+                job for job in resource.pipeline.jobs if job.is_triggered_by_resource(resource.name)
+            ]
+
+            # check if there is a pr resource version which has not triggered a job.
+            # If so return the job and the resource version
+            for job in jobs_triggered_by_resource:
+                builds = concourse_api.job_builds(pipeline_name, job.name)
```
ccwienk, line 349: again, I would figure this filtering should rather be moved into the api class, or moved to a util function. please do not make the dispatcher too large. the implementation details of how to tell whether or not a job was already triggered should not be this class's concern

whd/dispatcher.py
```diff
@@ -318,6 +318,56 @@ def is_up_to_date(resource, resource_versions):
...
+                for resource_version in self._resource_versions_without_build(
+                    resource_versions=pr_resource_versions,
+                    builds=builds,
+                ):
+                    yield (job, resource, resource_version)
```
ccwienk, line 354: instead of returning a three-tuple, rather return a dataclass or a namedtuple

whd/dispatcher.py
```diff
@@ -318,6 +318,56 @@ def is_up_to_date(resource, resource_versions):
...
+
+    def _resource_versions_without_build(self, resource_versions, builds):
```
ccwienk, line 356: add typehints?

whd/dispatcher.py
RaphaelVogel: we have all matching resources not the matching resource versions. Therefore we need to filter for the resource version with. Will add this filter to the API though.

whd/dispatcher.py
RaphaelVogel: done

whd/dispatcher.py
RaphaelVogel: done

whd/dispatcher.py
RaphaelVogel: will move all "pinning" code to own module

whd/dispatcher.py
RaphaelVogel: done

PR #441, opened 2020-07-07
Add support for optional initial-dev-cycle-commit prefix

AndreasBurger:
Adds a new attribute to the release trait, `next_cycle_commit_message_prefix`, that allows one to add a prefix to the commit message of the auto-generated initial-dev-cycle-commit. The prefix will usually be `[ci skip]`, I suspect.

### DIFF
--- a/concourse/model/traits/release.py
+++ b/concourse/model/traits/release.py
@@ -171,6 +171,15 @@ ATTRIBUTES = (
         an optional prefix for release commit messages
         ''',
         type=str,
+    ),
+        AttributeSpec.optional(
+        name='next_version_commit_message_prefix',
+        default=None,
+        doc='''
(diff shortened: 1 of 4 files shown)


concourse/model/traits/release.py
```diff
@@ -171,6 +171,14 @@ def _attribute_specs(cls):
         an optional prefix for release commit messages
         ''',
         type=str,
+    ),
+        AttributeSpec.optional(
+        name='next_cycle_commit_message_prefix',
```
ccwienk, line 176: how about just calling this `bump_commit_message_prefix`?
I consider `next_cycle` to be less commonly understood

ccwienk: as just discussed, for consistency reasons, let us opt for `nextversion_commitmsg_prefix`, or similar

PR #442, opened 2020-07-08
retry and wait for job to be triggered

RaphaelVogel:

Fixes #

**Release note**:

```improvement operator

```

### DIFF
--- a/concourse/client/util.py
+++ b/concourse/client/util.py
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import time
 import typing
 
 from whd.model import (
@@ -67,6 +68,33 @@ def determine_jobs_to_be_triggered(
     )
 
 
+def wait_for_job_to_be_triggered(
+    job: Job,
+    resource_version: ResourceVersion,
+    concourse_api,
+    retries: int=8,
+    sleep_time_seconds: int=5,
+) -> bool:
+    '''
+    There is some delay between the update of a resource and the start of the corresponding job.
+    Therefore we have to wait some time to decide if the job has been triggered or not.
+
+    @return True if the job has been triggered
+    '''
+    while retries >= 0:
+        if has_job_been_triggered(
+            job=job,
+            resource_version=resource_version,
+            concourse_api=concourse_api,
+        ):
+            return True
+
+        time.sleep(sleep_time_seconds)
+        retries -= 1
+
+    return False
+
+
 def has_job_been_triggered(
     job: Job,
     resource_version: ResourceVersion,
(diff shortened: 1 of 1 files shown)


concourse/client/util.py
```diff
@@ -71,14 +72,22 @@ def has_job_been_triggered(
     job: Job,
     resource_version: ResourceVersion,
     concourse_api,
+    retry: int=6,
```
ccwienk, line 75: nit: please use plural (retries)

concourse/client/util.py
```diff
@@ -71,14 +72,22 @@ def has_job_been_triggered(
...
+    sleep_sec: int=5,
```
ccwienk, line 76: please do not abbreviate the unit (seconds); this is inconsistent with our codebase, and also I do not think the benefit of saving a few keystrokes is worth the potentially worse readability

in other places in our codebase, such a symbol would be named `sleep_time_seconds`

concourse/client/util.py
```diff
@@ -71,14 +72,22 @@ def has_job_been_triggered(
     job: Job,
```
ccwienk, line 72: please do not mix two semantics in one method.

firstly: why do we need to wait after all? this is very much worth stating
secondly: retain this method as it was, with the original semantics, which was: if the job was already triggered, return True, otherwise return False; then add a second method, that uses this method to use the first method repeatedly with some waiting involved

concourse/client/util.py
```diff
@@ -71,14 +72,22 @@ def has_job_been_triggered(
...
-            return True
-    return False
+    job_has_been_triggered = False
+    while retry >= 0:
+        builds = concourse_api.job_builds(job.pipeline.name, job.name)
+        for build in builds:
+            build_plan = build.plan()
+            if build_plan.contains_version_ref(resource_version.version()['ref']):
+                job_has_been_triggered = True
+                break
```
ccwienk, line 86: use return; why make it so unnecessarily cumbersome?

concourse/client/util.py
```diff
@@ -67,6 +68,33 @@ def determine_jobs_to_be_triggered(
     )
 
 
+def job_has_been_triggered(
```
ccwienk, line 71: this name is confusing, as it is merely a permutation of `has_job_been_triggered`

please try to choose expressive names that correlate with the implemented semantics

as I understand this function, it will _wait_, if required, for some time, until a given job of interest is triggered. so a better name should include this.

e.g. `wait_for_job_execution`, or `wait_for_job_to_be_triggered`, or sth. of the like

RaphaelVogel: done

PR #445, opened 2020-07-08
process not triggered jobs

RaphaelVogel:

Fixes #

**Release note**:

```improvement operator

```

### DIFF
--- a/concourse/client/util.py
+++ b/concourse/client/util.py
@@ -13,18 +13,55 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import dataclasses
+import json
 import time
 import typing
 
+import dateutil.parser
+
+from dacite import from_dict
+from dacite.exceptions import DaciteError
+from contextlib import contextmanager
+from dataclasses import dataclass
+from datetime import (
+    datetime,
+    timedelta,
+)
+
+from ci.util import (
+    info,
+    warning,
+)
 from whd.model import (
     PullRequestEvent,
 )
 from .model import (
     Job,
     PipelineConfigResource,
+    PipelineResource,
     ResourceType,
     ResourceVersion,
+    PinnedPRVersion,
+    PinnedGitVersion,
+    PinnedTimeVersion,
 )
+
+
+class PinningFailedError(Exception):
+    pass
+
+
+class PinningUnnecessaryError(Exception):
+    pass
+
+
+@dataclass
+class PinComment:
+    pin_timestamp: str
+    version: typing.Union[PinnedPRVersion, PinnedGitVersion, PinnedTimeVersion]
+    next_retry: str
+    comment: str
 
 
 def determine_pr_resource_versions(
@@ -72,7 +109,7 @@ def wait_for_job_to_be_triggered(
     job: Job,
     resource_version: ResourceVersion,
     concourse_api,
-    retries: int=8,
+    retries: int=10,
     sleep_time_seconds: int=5,
 ) -> bool:
     '''
@@ -122,3 +159,168 @@ def jobs_not_triggered(
             for pr_resource_version in pr_resource_versions:
                 if not wait_for_job_to_be_triggered(job, pr_resource_version, concourse_api):
                     yield (job, pr_resource, pr_resource_version)
+
+
+def pin_resource_and_trigger_build(
+    job: Job,
+    resource: PipelineConfigResource,
+    resource_version: ResourceVersion,
+    concourse_api,
+    retries: int=3,
+):
+    # TODO: check whether resource is still pinned for our version
+    with ensure_resource_unpinned(
+        resource=resource,
+        resource_version=resource_version,
+        concourse_api=concourse_api,
+    ):
+        retries_for_job_trigger = 10
+        sleep_seconds_between_attempts = 6
+        wait_until_next_retry_seconds = retries_for_job_trigger * sleep_seconds_between_attempts
+
+        # pin resource and add comment
+        _pin_and_comment_resource(
+            wait_until_next_retry_seconds=wait_until_next_retry_seconds,
+            resource=resource,
+            resource_version=resource_version,
+            concourse_api=concourse_api,
+        )
+
+        for count in range(retries):
+            concourse_api.trigger_build(
+                pipeline_name=resource.pipeline_name(),
+                job_name=job.pipeline_name,
+            )
+            if wait_for_job_to_be_triggered(
+                job=job,
+                resource_version=resource_version,
+                concourse_api=concourse_api,
+                retries=retries_for_job_trigger,
+                sleep_time_seconds=sleep_seconds_between_attempts,
+            ):
+                info(f'{job.name=} for {resource_version.version()=} has been triggered')
+                return
+
+            else:
+                # job did not start, update pin comment with new retry timestamp
+                _pin_and_comment_resource(
+                    wait_until_next_retry_seconds=wait_until_next_retry_seconds,
+                    resource=resource,
+                    resource_version=resource_version,
+                    concourse_api=concourse_api,
+                )
+                warning(
+                    f'job {job.name} for resource version {resource_version.version()} '
+                    f'could not be triggered. Retry {count + 1}/{retries}'
+                )
+
+        warning(
+            f'job {job.name} for resource version {resource_version.version()} '
+            f'could not be triggered. Giving up after {retries} retries'
+        )
+
+
+def _pin_and_comment_resource(
+    wait_until_next_retry_seconds: int,
+    resource: PipelineConfigResource,
+    resource_version: ResourceVersion,
+    concourse_api,
+):
+    now = datetime.now()
+    next_retry = now + timedelta(seconds=wait_until_next_retry_seconds)
+    pin_comment = PinComment(
+        pin_timestamp=now.isoformat(),
+        version=resource_version.version(),
+        next_retry=next_retry.isoformat(),
+        comment=(
+            "pinned by technical user 'concourse', please do not unpin before "
+            f"{next_retry.isoformat()}"
+        ),
+    )
+    concourse_api.pin_and_comment_resource_version(
+        pipeline_name=resource.pipeline_name(),
+        resource_name=resource.name,
+        resource_version_id=resource_version.id(),
+        comment=json.dumps(dataclasses.asdict(pin_comment)),
+    )
+
+
+@contextmanager
+def ensure_resource_unpinned(
+    resource: PipelineResource,
+    resource_version: ResourceVersion,
+    concourse_api,
+):
+    try:
+        _ensure_resource_unpinned(
+            resource=resource,
+            resource_version=resource_version,
+            concourse_api=concourse_api,
+        )
+        yield
+    finally:
+        concourse_api.unpin_resource(
+            pipeline_name=resource.pipeline_name(),
+            resource_name=resource.name,
+        )
+        info(f'successfully unpinned {resource.name=}')
+
+
+def _ensure_resource_unpinned(
+    resource: PipelineResource,
+    resource_version: ResourceVersion,
+    concourse_api,
+    maximal_checktime_seconds: int = 240,
+):
+    # maximal time to wait for pinned resource, after that we give up
+    latest_unpin_checktime = datetime.now() + timedelta(seconds=maximal_checktime_seconds)
+
+    while latest_unpin_checktime >= datetime.now():
+        cc_resource = concourse_api.resource(
+            pipeline_name=resource.pipeline_name(),
+            resource_name=resource.name,
+        )
+
+        if not cc_resource.is_pinned():
+            return
+
+        sleep_time = 60
+        try:
+            pin_comment = from_dict(
+                data_class=PinComment,
+                data=json.loads(cc_resource.pin_comment()),
+            )
+
+            next_retry_time = dateutil.parser.isoparse(pin_comment.next_retry)
+            sleep_time = min(
+                (next_retry_time - datetime.now()).seconds + 5,
+                maximal_checktime_seconds,
+            )
+            if next_retry_time <= datetime.now():
+                warning(
+                    'found resource which was not unpinned although timestamp expired: '
+                    f'{resource.pipeline_name()=} {resource.name=}. Unpinning resource now'
+                )
+                concourse_api.unpin_resource(
+                    pipeline_name=resource.pipeline_name(),
+                    resource_name=resource.name,
+                )
+                return
+
+        except json.JSONDecodeError:
+            warning(
+                f'Resource comment for {resource.name=} of {resource.pipeline_name()=} '
+                'could not be parsed - most likely pinned by human user'
+            )
+        except DaciteError:
+            warning(
+                f'Resource comment for {resource.name=} of {resource.pipeline_name()=} '
+                'could not be instantiated'
+            )
+
+        time.sleep(sleep_time)
+
+    raise PinningFailedError(
+        f'Tried at least {latest_unpin_checktime} seconds to unpin '
+        f'{resource.name=} of pipeline {resource.pipeline_name()=}'
+    )


concourse/client/util.py
```diff
@@ -122,3 +128,73 @@ def jobs_not_triggered(
             for pr_resource_version in pr_resource_versions:
                 if not wait_for_job_to_be_triggered(job, pr_resource_version, concourse_api):
                     yield (job, pr_resource, pr_resource_version)
+
+
+def process_not_triggered_job(
```
AndreasBurger, line 133: Consider renaming:
```suggestion
def process_untriggered_job(
```

concourse/client/util.py
```diff
@@ -122,3 +128,73 @@ def jobs_not_triggered(
...
+    job: Job,
+    resource: PipelineConfigResource,
+    resource_version: ResourceVersion,
+    concourse_api
```
AndreasBurger, line 137:
```suggestion
    concourse_api,
```

concourse/client/util.py
```diff
@@ -122,3 +128,73 @@ def jobs_not_triggered(
...
+            resource=cc_resource,
+            resource_version=resource_version,
+            concourse_api=concourse_api,
+        )
+    else:
+        # to be implemeted
+        pass
+
+
+def pin_resource_trigger_job_unpin(
```
AndreasBurger, line 156: I think this name is confusing. I'd suggest a more general name (e.g. `handle_resource_pinning` or maybe better `ensure_job_run` or similar).

concourse/client/util.py
```diff
@@ -122,3 +128,73 @@ def jobs_not_triggered(
...
+    job: Job,
+    resource: PipelineResource,
+    resource_version: ResourceVersion,
+    concourse_api
```
AndreasBurger, line 160:
```suggestion
    concourse_api,
```

concourse/client/util.py
RaphaelVogel: done

concourse/client/util.py
RaphaelVogel: done

concourse/client/util.py
RaphaelVogel: done

concourse/client/util.py
RaphaelVogel: done

concourse/client/util.py
```diff
@@ -122,3 +129,73 @@ def jobs_not_triggered(
...
+
+
+def handle_resource_pinning(
+    job: Job,
+    resource: PipelineResource,
+    resource_version: ResourceVersion,
+    concourse_api,
+):
+    pin_comment = {
+        'datetime': datetime.now(),
```
ccwienk, line 164: it would be easier to read (for humans..) if you formatted this to an isocalendar date

concourse/client/util.py
```diff
@@ -122,3 +129,73 @@ def jobs_not_triggered(
...
+        'resource_type': resource.type().value,
+        'version_data': resource_version.version()
```
ccwienk, line 166: add a human-friendly comment as well. such as "automatically pinned by {username}", so it is more obvious what this is all about. maybe even add a stmt such as "please leave alone for ~5m" (just as an idea)

concourse/client/util.py
```diff
@@ -122,3 +129,73 @@ def jobs_not_triggered(
...
+
+        concourse_api.trigger_build(
+            pipeline_name=resource.pipeline_name(),
+            job_name=job.pipeline_name,
+        )
+
+        if wait_for_job_to_be_triggered(
+            job=job,
+            resource_version=resource_version,
+            concourse_api=concourse_api,
```
ccwienk, line 185: should this method not be parameterised with max-wait-time / max-retries?

concourse/client/util.py
```diff
@@ -122,3 +129,73 @@ def jobs_not_triggered(
...
+            resource=cc_resource,
+            resource_version=resource_version,
+            concourse_api=concourse_api,
+        )
+    else:
+        # to be implemeted
+        pass
+
+
+def handle_resource_pinning(
```
ccwienk, line 157: as already mentioned, I think it would make sense to turn this into a ContextManager

concourse/client/util.py
```diff
@@ -122,3 +129,73 @@ def jobs_not_triggered(
...
+    resource: PipelineConfigResource,
+    resource_version: ResourceVersion,
+    concourse_api,
+):
+    cc_resource = concourse_api.resource(
+        pipeline_name=resource.pipeline_name(),
+        resource_name=resource.name,
+    )
+
+    if not cc_resource.is_pinned():
```
ccwienk, line 145: I think, this should also be surrounded by a "while <condition -no-met> - sleep (if retries left) - try"

concourse/client/util.py
```diff
@@ -122,3 +129,73 @@ def jobs_not_triggered(
...
+        handle_resource_pinning(
+            job=job,
+            resource=cc_resource,
+            resource_version=resource_version,
+            concourse_api=concourse_api,
+        )
+    else:
+        # to be implemeted
```
ccwienk, line 153: as the comment already says..

concourse/client/util.py
ccwienk: +1 for a clearer name. as already commented, please consider turning this to a context-mgr. this is i think very much what context-mgrs are intended for. also, the naming would be easier.

consider this code snippet:
```
while True:
  try:
    with pin_resource(...args):
     trigger_job...
     break # finally, we succeeded
  except PinningFailed:
   # too bad, let's retry..
  sleep(....)
  if no_retries_left:
    bail out / fail / error
```

concourse/client/util.py
```diff
@@ -13,20 +13,35 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import json
 import time
 import typing
+from contextlib import contextmanager
```
AndreasBurger, line 19:
```suggestion
import typing

from contextlib import contextmanager
```

concourse/client/util.py
```diff
@@ -13,20 +13,35 @@
...
+from datetime import (
+    datetime,
+    timedelta,
+)
 
+from ci.util import (
```
AndreasBurger, line 25:
```suggestion
)
from ci.util import (
```

concourse/client/util.py
```diff
@@ -122,3 +137,118 @@ def jobs_not_triggered(
             for pr_resource_version in pr_resource_versions:
                 if not wait_for_job_to_be_triggered(job, pr_resource_version, concourse_api):
                     yield (job, pr_resource, pr_resource_version)
+
+
+def process_untriggered_job(
```
ccwienk, line 142: please call the function to more closely match its semantics

concourse/client/util.py
```diff
@@ -122,3 +137,118 @@ def jobs_not_triggered(
...
+        )
+        yield
+    finally:
+        concourse_api.unpin_resource(
+            pipeline_name=resource.pipeline_name(),
+            resource_name=resource.name,
+        )
+
+
+def _ensure_resource_unpinned(
```
ccwienk, line 217: ensure is a bit misleading, as the function will not actively unpin it. this function rather waits for the resource to become unpinned

concourse/client/util.py
```diff
@@ -122,3 +137,118 @@ def jobs_not_triggered(
...
+            pipeline_name=resource.pipeline_name(),
+            resource_name=resource.name,
+        )
+
+        if not cc_resource.is_pinned():
+            return
+
+        pin_comment = cc_resource.pin_comment()
+        try:
+            # if comment is json format, we assume it was pinned by us
```
ccwienk, line 234: please do not only take into account the fact whether or not it is json, but rather also check for the attributes. it could be that case that someone else write some different json

concourse/client/util.py
```diff
@@ -122,3 +137,118 @@ def jobs_not_triggered(
...
+            parsed_comment = json.loads(pin_comment)
+            pin_time = datetime.fromisoformat(parsed_comment['datetime'])
+            if pin_time + timedelta(minutes=3) < datetime.now():
```
ccwienk, line 237: since we are writing and reading the comment, please add a dataclass

concourse/client/util.py
```diff
@@ -122,3 +137,118 @@ def jobs_not_triggered(
...
+                # resource pinned by us but unpin did not happen for more than 3 minutes -> cleanup
+                concourse_api.unpin_resource(
+                    pipeline_name=resource.pipeline_name(),
+                    resource_name=resource.name,
+                )
+                return
+        except:
+            # resource pinned by someone else, wait and try again
+            pass
```
AndreasBurger, line 246: This catches _all_ raised errors. 
```suggestion
        except JSONDecodeError:
            # resource pinned by someone else, wait and try again
            pass
```
As for handling other errors: Not sure. I think it is already retrying on connection errors, so that should be fine.

concourse/client/util.py
```diff
@@ -122,3 +137,118 @@ def jobs_not_triggered(
...
+            pin_time = datetime.fromisoformat(parsed_comment['datetime'])
+            if pin_time + timedelta(minutes=3) < datetime.now():
+                # resource pinned by us but unpin did not happen for more than 3 minutes -> cleanup
+                concourse_api.unpin_resource(
+                    pipeline_name=resource.pipeline_name(),
+                    resource_name=resource.name,
+                )
+                return
+        except:
+            # resource pinned by someone else, wait and try again
```
ccwienk, line 245: please do not do this. catch-all is rarely a good idea
instead, explicitly do a try-except for the json-parsing + deserialisation part, and explicitly handle the case that parsing failed, or structure was unexpected. log a warning in that case.

also: is it possible to find out about the user that did the pinning? if so, please log this, as well. consider writing an informative email to said user

concourse/client/util.py
```diff
@@ -13,20 +13,35 @@
...
 from .model import (
     Job,
     PipelineConfigResource,
+    PipelineResource,
     ResourceType,
     ResourceVersion,
 )
 
 
+class PinningFailed(Exception):
```
AndreasBurger, line 41: I think these are called `xyError` by convention (except the base Exception).
```suggestion
class PinningFailedError(Exception):
```

concourse/client/util.py
```diff
@@ -122,3 +137,118 @@ def jobs_not_triggered(
...
+                return
+        except:
+            # resource pinned by someone else, wait and try again
+            pass
+
+        retries -= 1
+        time.sleep(sleep_time_seconds)
+
+    raise PinningFailed(
+        f'Tried {retries * sleep_time_seconds}s to unpin resource {resource.name} '
```
ccwienk, line 252: consider also logging a warning - this could prove very helpful

concourse/client/util.py
```diff
@@ -122,3 +137,118 @@ def jobs_not_triggered(
...
+            return
+
+        pin_comment = cc_resource.pin_comment()
+        try:
+            # if comment is json format, we assume it was pinned by us
+            parsed_comment = json.loads(pin_comment)
+            pin_time = datetime.fromisoformat(parsed_comment['datetime'])
+            if pin_time + timedelta(minutes=3) < datetime.now():
+                # resource pinned by us but unpin did not happen for more than 3 minutes -> cleanup
+                concourse_api.unpin_resource(
```
ccwienk, line 239: this should definitely be logged as a warning, as it indicates a bug (previous job did not free its "lock"). also, there should be a coupling of maximum retries and maximum waiting time. currently, this is just hardcoded at two places, which I find to be far from ideal.

-> consider writing remaining retries into pinning comment. is it possible to update a pinning comment w/o releasing the pin? if so, please update each time the triggering was retried, and update the "last-tried" timestamp (+ store the sleep-time). in this case, we do not even have to hardcode anything here.

instead, we can check:

- last time of update
- sleep-time

if last-time-of-update + sleep-time > now, then something went wrong.

concourse/client/util.py
```diff
@@ -122,3 +137,118 @@ def jobs_not_triggered(
...
+                else:
+                    warning(
+                        f'job {job.name} for resource version {resource_version.version()} '
+                        'could not be triggered'
+                    )
+                return
+        except PinningFailed:
+            retries -= 1
+
+    warning(f'Could not process untriggered job {job.name}')
```
ccwienk, line 181: this is misleading. why not explicitly state that we failed to trigger the job? processing is a very generic term. why not be precise and explicit?

ccwienk: also (stilistic nit), why not make use of python3.8's new f-str feature: `f'failed to trigger {job.name=}'`. this is exactly as informative, and requires less typing

concourse/client/util.py
AndreasBurger: It does unpin if the current pin was performed by a technical user and the resource was not unpinned again after a set interval.

concourse/client/util.py
RaphaelVogel: User who pinned the resource is not available via CC API.

concourse/client/util.py
```diff
@@ -122,3 +152,159 @@ def jobs_not_triggered(
...
+                resource=resource,
+                resource_version=resource_version,
+                concourse_api=concourse_api,
+            ):
+                concourse_api.trigger_build(
+                    pipeline_name=resource.pipeline_name(),
+                    job_name=job.pipeline_name,
+                )
+
+                job_retry = 10
```
ccwienk, line 177: nit: the plural form is written `job_retries`

concourse/client/util.py
```diff
@@ -122,3 +152,159 @@ def jobs_not_triggered(
...
+                    retries=job_retry,
+                    sleep_time_seconds=job_sleep_seconds,
+                ):
+                    info(
+                        f'job {job.name} for resource version {resource_version.version()} '
+                        'has been triggered'
+                    )
+                else:
+                    pin_comment = PinComment(
+                        pin_timestamp=datetime.now().isoformat(),
```
ccwienk, line 192: please assign "now" to a variable, so the two usages match exactly. i know this is not strictly required, however, I consider this to be cleaner.

concourse/client/util.py
```diff
@@ -122,3 +152,159 @@ def jobs_not_triggered(
...
+                        version_data=resource_version.version(),
```
ccwienk, line 193: why `version_data`. rather than just `version`?

concourse/client/util.py
```diff
@@ -122,3 +152,159 @@ def jobs_not_triggered(
...
+                        next_retry=(datetime.now() + timedelta(
+                            seconds=(job_retry + 1) * job_sleep_seconds,
+                        )).isoformat(),
+                        comment="pinned by technical user, please do not unpin manually"
```
ccwienk, line 197: can we please add our username?

ccwienk: also, the comment should be changed to: do not unpin before (<earliest date after which it is safe to unpin>)

concourse/client/util.py
```diff
@@ -122,3 +152,159 @@ def jobs_not_triggered(
...
+            pipeline_name=resource.pipeline_name(),
+            resource_name=resource.name,
+        )
+
+        if not cc_resource.is_pinned():
+            return
+
+        pin_comment_data = cc_resource.pin_comment()
+        try:
+            pin_comment = PinComment(**json.loads(pin_comment_data))
```
ccwienk, line 276: use dacite.from_dict

concourse/client/util.py
```diff
@@ -122,3 +152,159 @@ def jobs_not_triggered(
...
+            if pin_comment.version_data == resource_version.version():
+                raise PinningUnnecessaryError("Resource is already pinned for the same version")
+
+            next_retry_time = datetime.fromisoformat(pin_comment.next_retry)
```
ccwienk, line 280: rather use dateutil.parser.isoparse - datetime.fromisoformat is buggy (not fully compliant to iso8601)

concourse/client/util.py
```diff
@@ -122,3 +152,159 @@ def jobs_not_triggered(
...
+        except json.JSONDecodeError:
+            # resource comment not parsable. Resource is pinned by someone else
+            info(
+                f'Resource comment for {resource.name=} of {resource.pipeline_name()=} '
+                'could not be parsed'
+            )
+        except TypeError:
+            # Not able to instantiate. Resource pinned by someone else
+            info(
+                f'Resource comment for {resource.name=} of {resource.pipeline_name()=} '
```
ccwienk, line 300: also dump the pinning comment?

concourse/client/util.py
```diff
@@ -122,3 +152,159 @@ def jobs_not_triggered(
...
+                )
+                concourse_api.unpin_resource(
+                    pipeline_name=resource.pipeline_name(),
+                    resource_name=resource.name,
+                )
+                return
+        except json.JSONDecodeError:
+            # resource comment not parsable. Resource is pinned by someone else
+            info(
+                f'Resource comment for {resource.name=} of {resource.pipeline_name()=} '
```
ccwienk, line 294: dump comment

concourse/client/util.py
```diff
@@ -122,3 +152,159 @@ def jobs_not_triggered(
...
+            return
+
+        pin_comment_data = cc_resource.pin_comment()
+        try:
+            pin_comment = PinComment(**json.loads(pin_comment_data))
+            if pin_comment.version_data == resource_version.version():
+                raise PinningUnnecessaryError("Resource is already pinned for the same version")
+
+            next_retry_time = datetime.fromisoformat(pin_comment.next_retry)
+            if next_retry_time < datetime.now():
```
ccwienk, line 281: please add ~5s..10s of buffer

concourse/client/util.py
```diff
@@ -122,3 +152,159 @@ def jobs_not_triggered(
...
+        cc_resource = concourse_api.resource(
+            pipeline_name=resource.pipeline_name(),
+            resource_name=resource.name,
+        )
+
+        if not cc_resource.is_pinned():
+            return
+
+        pin_comment_data = cc_resource.pin_comment()
+        try:
```
ccwienk, line 275: factor out deserialisation to separate (private) function, please

concourse/client/util.py
```diff
@@ -122,3 +152,159 @@ def jobs_not_triggered(
...
+def pin_resource(
+    resource: PipelineResource,
+    resource_version: ResourceVersion,
+    concourse_api,
+):
+    _ensure_resource_unpinned(
+        resource=resource,
+        resource_version=resource_version,
+        concourse_api=concourse_api,
+        retries=10,
```
ccwienk, line 233: I think the sleep-time + retries-amount should no longer be required to be passed; instead, use the existing metadata, as elaborated in a different comment yesterday

concourse/client/util.py
```diff
@@ -122,3 +152,159 @@ def jobs_not_triggered(
...
+        )
+
+        if not cc_resource.is_pinned():
+            return
+
+        pin_comment_data = cc_resource.pin_comment()
+        try:
+            pin_comment = PinComment(**json.loads(pin_comment_data))
+            if pin_comment.version_data == resource_version.version():
+                raise PinningUnnecessaryError("Resource is already pinned for the same version")
```
ccwienk, line 278: not sure I agree with this to be signalled as an error;

also: we should at least check whether the lock has become stale (i.e. timestamp too old)

concourse/client/util.py
RaphaelVogel: inline not possible we use it 2 times

concourse/client/util.py
RaphaelVogel: we already have set the buffer when writing

concourse/client/util.py
```diff
@@ -13,20 +13,54 @@
...
+
+
+class PinningUnnecessaryError(Exception):
+    pass
+
+
+@dataclass
+class PinComment:
+    pin_timestamp: str
+    version: dict
```
ccwienk, line 59: which structure does `version` have?
-> + add a dataclass for it, as well?

concourse/client/util.py
```diff
@@ -122,3 +156,163 @@ def jobs_not_triggered(
...
+                )
+
+        warning(
+            f'job {job.name} for resource version {resource_version.version()} '
+            f'could not be triggered. Giving up after {retries} retries'
+        )
+
+
+def _pin_and_comment_resource(
+    next_retry_in_seconds: int,
```
ccwienk, line 220: naming nit: rather name it `wait_until_next_retry_seconds`?

concourse/client/util.py
```diff
@@ -122,3 +156,163 @@ def jobs_not_triggered(
...
+        info(f'successfully unpinned {resource.name=}')
+
+
+def _ensure_resource_unpinned(
+    resource: PipelineResource,
+    resource_version: ResourceVersion,
+    concourse_api,
+):
+    previous_retry_time = None
+    maximal_unpin_checktime = datetime.now() + timedelta(minutes=5)
```
ccwienk, line 271: rather call it _latest_

also, please add a brief comment, as to the implementation idea (allow detecting stale lock from a branch that died / failed to cleanup)

ccwienk, line 271: make this time a parameter (it is ok to add a default value, of course)?

also: maybe 5m are a bit too long..

concourse/client/util.py
```diff
@@ -122,3 +156,163 @@ def jobs_not_triggered(
...
+
+        try:
+            pin_comment = from_dict(
+                data_class=PinComment,
+                data=json.loads(cc_resource.pin_comment()),
+            )
+            if pin_comment.version == resource_version.version():
+                raise PinningUnnecessaryError("Resource is already pinned for the same version")
+
+            next_retry_time = dateutil.parser.isoparse(pin_comment.next_retry)
```
ccwienk, line 290: I think we should add some margin

concourse/client/util.py
```diff
@@ -122,3 +156,163 @@ def jobs_not_triggered(
...
+            info(
+                f'Resource comment for {resource.name=} of {resource.pipeline_name()=} '
+                'could not be instantiated'
+            )
+
+        previous_retry_time = next_retry_time
+        time.sleep((next_retry_time - datetime.now()).seconds + 5)
+
+    raise PinningFailedError(
+        f'Tried at least 5 minutes to unpin '
```
ccwienk, line 316: use the actual value (time in minutes), rather than hard-coding.

also, rather use seconds?

concourse/client/util.py
```diff
@@ -122,3 +156,163 @@ def jobs_not_triggered(
...
+                'could not be parsed - most likely pinned by human user'
+            )
+        except DaciteError:
+            info(
+                f'Resource comment for {resource.name=} of {resource.pipeline_name()=} '
+                'could not be instantiated'
+            )
+
+        previous_retry_time = next_retry_time
+        time.sleep((next_retry_time - datetime.now()).seconds + 5)
```
ccwienk, line 313: add comment as to why we add a margin

ccwienk: also, there should be a maximum sleep-time

RaphaelVogel: sleep time will be maxed by parameter `wait_until_next_retry_seconds` in function `pin_resource_and_trigger_build`

ccwienk: yeah, but you do not check for it here. if the timestamp would e.g. be a year in the future, this code would wait for a year

ccwienk: also, just checking for whether or not the timestamp changed is not even required (and adds an unnecesary code path). admittedly, this could be used as an _optimisation_. However, let us please not add optimisations right in the first shot. please let us keep this PR focussed, so we (and in particular me) have a chance of reviewing / understanding it properly. the diff is already too big for my taste, and adding optional optimisations is not helping. add optimisations later (after getting the first version merged), if required.

how about simply going for:

```
def acquire_resource_pin(....):
  while "timeout not exceeded and retries left":
    if "resource is unpinned":
      "pin_it_and_return"
    "read_resource_pinning_comment"
    if "resource_pinning_comment is not "from us" (e.g. from a human)":
      "write warning to log + return"
    else: # "resource_pinning_comment _is_ "from us""
       "check announced release timestamp" (note: this time could / should be published whilst locking)
       if "announced release timestamp is in the past":
         log a warning + unpin + pin + return
       else: # "announced release timestamp is in the future":
        wait for max(max_wait_time, announced_release_timestamp)
        # max_wait_time is merely a "safety guard" to prevent erroneous timestamps from causing DOS
```

PR #449, opened 2020-07-14
Add support for Concourse 6.3.0 API changes

AndreasBurger:
Adds support for the changes in the login flow in Concourse version `6.3.0`. Without this, our automations won't be able to authenticate against Concourse.

### DIFF
--- a/concourse/client/api.py
+++ b/concourse/client/api.py
@@ -22,6 +22,8 @@ from urllib3.exceptions import InsecureRequestWarning
 
 from .routes import (
     ConcourseApiRoutesBase,
+    ConcourseApiRoutesV5,
+    ConcourseApiRoutesV6_3_0,
 )
 from .model import (
     Build,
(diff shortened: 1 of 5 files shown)


model/concourse.py
```diff
@@ -81,8 +84,15 @@ def ingress_url(self):
     def helm_chart_version(self):
         return self.raw.get('helm_chart_version')
 
-    def concourse_version(self) -> ConcourseApiVersion:
-        return ConcourseApiVersion(self.raw.get('concourse_version'))
+    def concourse_api_version(self) -> ConcourseApiVersion:
```
ccwienk, line 87: naming nit: `(calculate_)compatible_api_version`?

concourse/client/api.py
```diff
@@ -331,3 +379,8 @@ def set_team(self, concourse_team: ConcourseTeam):
 
         team_url = self.routes.team_url(concourse_team.teamname())
         self._put(team_url, json.dumps(body))
+
+
+class ConcourseApiV6_3_0(ConcourseApiV5):
```
ccwienk, line 384: as just discussed: factor out common methods for v5 and v6 to a third class, and have both versions inherit it, rather than having v6_3 inherit from v5

as an _alternative_, I would also agree with copy-pasting the common methods (since will remove V5 after switching to v6_3 anway)

PR #465, opened 2020-08-18
Add license scanning to protecode trait

AndreasBurger:
Allows configuration of licenses to check for in protecode scans and add information about them to generated report-mails

### DIFF
--- a/concourse/model/traits/image_scan.py
+++ b/concourse/model/traits/image_scan.py
@@ -84,6 +84,26 @@ PROTECODE_ATTRS = (
         doc='CVSS version used to evaluate the severity of vulnerabilities',
         type=CVSSVersion,
     ),
+    AttributeSpec.optional(
+        name='allowed_licenses',
+        default=[],
+        doc=(
+            'A list of regular expressions. If configured, licenses detected by protecode '
+            'that do not match at least one of regular expressions will result in a report mail '
+            'being sent. If not configured or empty, **all** licenses will be accepted.'
+        ),
+        type=list,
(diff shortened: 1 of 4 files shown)


concourse/model/traits/image_scan.py
```diff
@@ -84,6 +84,26 @@
...
+        doc=(
+            'A list of regular expressions. If configured, licenses detected by protecode '
+            'that do not match at least one of regular expressions will result in a report mail '
+            'being sent. If not configured (i.e. `None`), **all** licenses will be accepted.'
+        ),
+        type=list,
+    ),
+    AttributeSpec.optional(
+        name='prohibited_licenses',
+        default=[],
```
ccwienk, line 99: for consistency reasons, use either `None`, or an empty sequence in both cases. for avoiding unexpected side-effects, use immutable sequence (aka tuple)

ccwienk: also: what happens if both allowed and prohibited licenses are defined (should be documented)?

I would _expect_ the following behavour:

1. allowed licences are defined, but prohibited licences are not defined
-> any license that is not explicitly allowed, is assumed to be prohibited

2. prohibited licenses are defined, but allowed licences are not
-> any license that is not explicitly prohibited is allowed

3. both allowed and prohibited licences are defined
-> seems difficult to define a reasonable semantics
we might in this case differentiate:
- allowed licenses are allowed (no-brainer)
- prohibited licenses are prohibilted (also no-brainer)
- all other licences are "unclassified"

PR #471, opened 2020-08-28
Make code compatible to new google libs versions

AndreasBurger:
Make code compatible to new versions and also restrict the google libs to current major version in requirements.txt

### DIFF
--- a/ccc/grafeas.py
+++ b/ccc/grafeas.py
@@ -39,26 +39,14 @@ def grafeas_client(container_registry_cfg: model.container_registry.ContainerReg
     default_oauth_scope = (
         'https://www.googleapis.com/auth/cloud-platform',
     )
-
-    try:
-        service_address = ContainerAnalysisClient.SERVICE_ADDRESS
-        transport = GrafeasGrpcTransport(
-            address=service_address,
-            scopes=default_oauth_scope, # XXX hard-code for now
-            credentials=credentials.service_account_credentials(),
-        )
-
-    except AttributeError:
-        service_address = ContainerAnalysisClient.DEFAULT_ENDPOINT
-        service_port = 443
-
-        transport = GrafeasGrpcTransport(
-            host=f'{service_address}:{service_port}',
-            scopes=default_oauth_scope, # XXX hard-code for now
-            credentials=credentials.service_account_credentials(),
-        )
-
-    return GrafeasClient(transport=transport)
+    service_address = ContainerAnalysisClient.DEFAULT_ENDPOINT
+    transport = GrafeasGrpcTransport(
+        host=service_address,
+        credentials=credentials.service_account_credentials().with_scopes(default_oauth_scope),
+    )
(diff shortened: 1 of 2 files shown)


ccc/grafeas.py
```diff
@@ -1,24 +1,13 @@
 import logging
 import urllib.parse
 
-try:
-    from grafeas.grafeas_v1.gapic.transports.grafeas_grpc_transport import GrafeasGrpcTransport
```
ccwienk, line 5: do we not need to do this (again) in a two-step change (first, upgrade dependencies + release, then remove backward-compatibility coding, and release again)?

ccc/grafeas.py
```diff
@@ -39,26 +28,14 @@ def grafeas_client(container_registry_cfg: model.container_registry.ContainerReg
...
-
-        transport = GrafeasGrpcTransport(
-            host=f'{service_address}:{service_port}',
-            scopes=default_oauth_scope, # XXX hard-code for now
-            credentials=credentials.service_account_credentials(),
-        )
-
-    return GrafeasClient(transport=transport)
+    service_address = ContainerAnalysisClient.DEFAULT_ENDPOINT
+    transport = GrafeasGrpcTransport(
```
ccwienk, line 32: much nicer like this, with my ugly hack gone :-)

ccc/grafeas.py
```diff
@@ -103,7 +80,10 @@ def scan_available(
 
     filter_str = f'resourceUrl = "https://{hash_reference}" AND kind="DISCOVERY"'
     try:
-        results = list(client.list_occurrences(f'projects/{project_name}', filter_=filter_str))
+        results = list(client.list_occurrences(
+            parent=f'projects/{project_name}',
+            filter=filter_str,
```
ccwienk, line 85: so they actually renamed this ugly parameter name (filter_ -> filter)?

requirements.txt
```diff
@@ -18,17 +18,18 @@ ensure<1.0.0
...
-grpc-google-iam-v1==0.12.3
+google-api-core>=1.0.0, <2.0.0
+google-api-python-client>=1.0.0, <2.0.0
+google-auth>=1.0.0, <2.0.0
+google-auth-httplib2<1.0.0
+google-cloud-containeranalysis>=2.0.0, <3.0.0
+google-cloud-core>=1.0.0, <2.0.0
+google-cloud-storage>=1.0.0, <2.0.0
+google-crc32c>=1.0.0, <2.0.0
+google-resumable-media>=1.0.0, <2.0.0
```
ccwienk, line 29: just as a remark (not necessarily a request for changes): you may omit the patch-level and minor components of version restrictions afaik

ccwienk (commented):
please check my first comment, otherwise lgtm

ccwienk:
addendum: the failing tests seem to prove my point - please retain the backwards-compatibility coding (and add another such change for the renamed parameter name - thx)

PR #472, opened 2020-08-31
Simplify Documentation for Enums

AndreasBurger:
Introduces a new Enum-Type that is easier to document.

Only converted one enum so far, thus this is a draft PR.

### DIFF
--- a/concourse/model/base.py
+++ b/concourse/model/base.py
@@ -14,12 +14,14 @@
 # limitations under the License.
 
 
+import typing
+
 from abc import abstractmethod
 from enum import Enum
-import typing
 from collections import namedtuple
 
 import ci.util
+
 from model.base import (
     ModelValidationError,
     ModelValidationMixin,
--- a/concourse/model/base.py
+++ b/concourse/model/base.py
@@ -17,12 +17,27 @@
 from abc import abstractmethod
 from enum import Enum
 import typing
+from collections import namedtuple
 
 import ci.util
 from model.base import (
     ModelValidationError,
     ModelValidationMixin,
 )
+
+
+class EnumWithDocumentation(Enum):
+    def __new__(cls, value, docstring=None):
+        obj = object.__new__(cls)
+        obj._value_ = value
+        obj.__doc__ = docstring
+        return obj
+
+    def __str__(self):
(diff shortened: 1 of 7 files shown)


concourse/model/base.py
```diff
@@ -25,6 +25,17 @@
 )
 
 
+class EnumWithDocumentation(Enum):
+    def __new__(cls, value, docstring=None):
```
ccwienk, line 29: would it not suffice to overwrite the `__init__` method?

concourse/model/traits/component_descriptor.py
```diff
@@ -27,39 +27,24 @@
...
     TraitTransformer,
 )
 
 
-class ValidationPolicy(AttribSpecMixin, enum.Enum):
-    NOT_EMPTY = "not_empty"
-    FORBID_EXTRA_ATTRIBUTES = "forbid_extra_attributes"
+class ValidationPolicy(EnumWithDocumentation):
+    NOT_EMPTY = (
+        'not_empty',
```
ccwienk, line 40: I did not try: is it possible to pass those as named parameters?

ccwienk (commented):
looks good so far. consider also handling plain enums, which may in many cases be a natural choice with even less boilerplate

concourse/model/base.py
AndreasBurger: No, because I am changing the way value assignment works for this enum, `__new__` is the way to go as per doc.

concourse/model/traits/component_descriptor.py
AndreasBurger: Unfortunately not. As a workaround, NamedTuples should provide the added explicitness. I'll experiment with them.

concourse/model/base.py
ccwienk: you are right; I did not read through the documentation thorough enough yesterday, sry for that

concourse/model/traits/component_descriptor.py
ccwienk: ok - as discussed earlier: it is OK with me to use two-tuples;

ccwienk:
overall, lgtm

PR #474, opened 2020-09-11
Switch to generated containeranalysis client

AndreasBurger:
Replace the current grafeas-client with an autogenerated one to reduce grpcio-introduced problems we face.

Final testing still to be done and I still need to think about how much exception handling is truly necessary, but code itself shouldn't change much.

Also, tests are currently broken due to the removal of the old client.


ccc/grafeas_model.py
```diff
@@ -0,0 +1,133 @@
...
+
+def parse_list_occurrences_response(response: dict) -> ListOccurrencesResponse:
+    '''Parses a dict representing a response to a list_occurrences request to dataclasses
+    '''
+    return dacite.from_dict(
+        data_class=ListOccurrencesResponse,
+        data=response,
+        config=dacite.Config(
+            type_hooks={
+                Severity: lambda x: Severity[x],
```
ccwienk, line 128: ahh, so that's how to use type-hooks - great, so I do not have to tal at the documentation for that one :-)

protecode/scanning_util.py
```diff
@@ -474,17 +474,17 @@ def _import_triages_from_gcr(self, scan_result: AnalysisResult):
         worst_effective_vuln = Severity.SEVERITY_UNSPECIFIED
         try:
             vulnerabilities_from_grafeas = list(
-                ccc.grafeas.filter_vulnerabilities(
+                client.filter_vulnerabilities(
```
ccwienk, line 477: rather name it grafeas_client, maybe?

protecode/scanning_util.py
```diff
@@ -506,49 +506,46 @@ def find_worst_vuln(
...
-                    wurst_cve = max(wurst_cve, v.cvss_score)
-                    wurst_eff = max(
-                        wurst_eff,
-                        Severity(getattr(v, 'effective_severity', wurst_eff))
-                    )
+                    worst_cve = max(worst_cve, v.cvssScore)
+                    worst_effective_severity = max(worst_effective_severity, v.effectiveSeverity)
 
-            return found_it, wurst_cve, wurst_eff
+            return found_it, worst_cve, worst_effective_severity
```
ccwienk, line 532: oh dear - I never fixed that ^^

ccwienk (changes requested):

ccc/gcp.py
```diff
@@ -40,3 +58,156 @@ def cloud_storage_client(gcp_cfg: str, *args, **kwargs):
...
+            parsed_response = parse_list_occurrences_response(response)
+            yield from parsed_response.occurrences
+
+            request = occurrences.list_next(request, response)
+
+    def _determine_gcr_parameters(self, image_reference):
+        project_name = urllib.parse.urlparse(image_reference).path.split('/')[1]
+        try:
+            hash_reference = container.registry.to_hash_reference(image_reference)
+        except Exception as e:
```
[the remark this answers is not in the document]
AndreasBurger: Looked at it again and restored the old error handling (catch everything and wrap it into a VulnerabilitiesRetrievalFailed error that will then be handled by the caller)

PR #476, opened 2020-09-17
Add whitesource-api-extension deployment

zkdev:
This PR adds the deployment of `whitesource-api-extension` to the `landscape_setup`.
Addtionally the mail reporting is implemented.

Fixes #

**Release note**:
NONE

```improvement operator

```


kube/ctx.py
```diff
@@ -58,8 +60,10 @@ def get_kubecfg(self):
             fail('KUBECONFIG env var must be set')
         return config.load_kube_config(existing_file(kubeconfig))
 
-    def set_kubecfg(self, kubeconfig_dict: dict):
+    def set_kubecfg(self, kubeconfig_dict: typing.Union[dict, model.kubernetes.KubernetesConfig]):
```
ccwienk, line 63: should this really be part of your diff? Oo

mail/template_mailer.py
```diff
@@ -31,7 +32,8 @@ def create_mail(
         recipients: [str],
         text: str,
         cc_recipients: [str]=[],
-        mimetype: str='text'
+        mimetype: str='text',
+        pdfs=[],
```
ccwienk, line 36: so PDFs are dicts w/ {pdf_bytes, filename}?

I would prefer to make this one tiny bit more generic: add mimetype to the dict (or rather: define a dataclass..). That way, other attachments could also be passed

whitesource/util.py
```diff
@@ -1,10 +1,150 @@
...
+    above = {}
+    below = {}
+
+    for lib, dic in report.items():
+        if float(dic['CVSS-V3']) > threshold:
+            above[lib] = dic
+        else:
+            below[lib] = dic
+
+    def _sort_struct(struct):
```
ccwienk, line 34: I think `struct` is a rather unconventional name in python; also, it is a special kind of struct (-> a vulnerability-something - yes, good names are hard^^)

additional nit: if you would sort already in line 28, you could save one line of code (and might actually inline the sorting function, thus saving some more boilerplate, and not have to think about a good name) :-P

whitesource/util.py
```diff
@@ -1,10 +1,150 @@
...
+        return sorted(struct.items(),
+                      key=lambda k_v: k_v[1]['CVSS-V3'],
+                      reverse=True)
+
+    # sort tables descending by CVSS-V3
+    below = _sort_struct(below)
+    above = _sort_struct(above)
+
+    ttable_header = ('Component', 'Greatest CVSS-V3', 'Corresponding CVE')
+    ttables = []
```
ccwienk, line 44: why ttables?

whitesource/util.py
```diff
@@ -1,10 +1,150 @@
...
+            tablefmt=tablefmt,
+            colalign=('left', 'center', 'center')
+        )
+
+        ttables.append(ttable)
+
+    return ttables
+
+
+def assemble_mail_body(tables: list,
```
ccwienk, line 70: nit: typing.List
also: it would be nice to specify the element types....

whitesource/util.py
```diff
@@ -1,10 +1,150 @@
...
+    for project in projects["projects"]:
+        pname = project["projectName"]
+        ptoken = project["projectToken"]
+
+        # get vulnerability report per project
+        ci.util.info(f'retrieving project vulnerability report for {pname}')
+        pvr = client.get_project_vulnerability_report(project_token=ptoken)
+
+        # find greatest cve per project
+        for vul in json.loads(pvr.content)["vulnerabilities"]:
```
ccwienk, line 121: same nit: it would be nicer, of the client would return a dict, rather than a json-str

whitesource/util.py
```diff
@@ -1,10 +1,150 @@
...
+            try:
+                if report.get(pname) is None or vul["cvss3_score"] > report[pname]["CVSS-V3"]:
+                    report[pname] = {
+                        "CVSS-V3": vul["cvss3_score"],
+                        "CVE": vul["name"],
+                    }
+            except KeyError:
+                # WS are ignored
```
ccwienk, line 129: what's WS?
in which case will those key-errs occur, and why is it okay to ignore them silently?

I know, it is a pythonic pattern to "ask for forgiveness"; however, in this case, there is quite some risk that this will hide actual implementation errors, as there are a lot of dict-element-accesses (btw: did I already mention dataclasses and dacite do a great job in deserialising dicts?

ccwienk (changes requested):
some nits, some questions

whitesource/util.py
zkdev: tabulate-tables

whitesource/util.py
zkdev: 'Whereas the NVD publishes security vulnerabilities using a "CVE-" prefixed identifier, WhiteSource classifies non-NVD security vulnerabilities using a "WS-" prefix' - https://whitesource.atlassian.net/wiki/spaces/WD/pages/470188110/Notice+Regarding+CVE+and+WS+Vulnerability+Identifiers

Unfortunately the `getProjectVulnerabilityReport` from the WhiteSource API includes some inconsistency. While all `CVE*` entries have a key called `cvss3_score` some `WS*` entries call the corresponding key either `cvss3_score` or just `score`. I adjusted the logic to look for both possibilities.

concourse/steps/scan_sources.mako
```diff
@@ -29,12 +29,16 @@ scan_sources_and_notify(
...
 
 % if whitesource_cfg:
+component_name = "${component_trait.component_name()}"
 scan_component_with_whitesource(
     whitesource_cfg_name='${whitesource_cfg.cfg_name()}',
     product_token='${whitesource_cfg.product_token()}',
     component_descriptor_path=component_descriptor_path(),
     extra_whitesource_config={},
     requester_mail='${email_recipients[0]}',
+    cve_threshold=,'${whitesource_cfg.cve_threshold()}',
```
ccwienk, line 39: nit: should not this rather be a float/int?
also, I believe there is a syntax error (comma after equals sign)

concourse/steps/scan_sources.mako
```diff
@@ -29,12 +29,16 @@ scan_sources_and_notify(
 % endif
 
 % if whitesource_cfg:
+component_name = "${component_trait.component_name()}"
```
ccwienk, line 32: nit: please consistently use single quotes

concourse/steps/scan_sources.py
```diff
@@ -197,54 +199,99 @@ def scan_component_with_whitesource(whitesource_cfg_name: str,
                                     product_token: str,
                                     component_descriptor_path: str,
                                     extra_whitesource_config: dict,
-                                    requester_mail: str):
+                                    requester_mail: str,
+                                    cve_threshold: float,
+                                    component_name: str,
+                                    notification_recipients=[],
+                                    send_notificaton=True,):
```
ccwienk, line 206: nit: please place closing paranthesis in separate line, so that adding / reordering new lines is actually more convenient

concourse/steps/scan_sources.py
```diff
@@ -197,54 +199,99 @@ def scan_component_with_whitesource(whitesource_cfg_name: str,
...
         ci.util.parse_yaml_file(component_descriptor_path)
     )
-    ci.util.info('component descriptor parsed\n')
 
     # for component in []: "components matching component_descriptor"
     for component in component_descriptor.components():
 
         # create whitesource component
-        ci.util.info('preparing POST project for {}...'.format(component.name()))
+        ci.util.info('preparing POST project for {}'.format(component.name()))
```
ccwienk, line 222: this is certainly copy-pasta. at any rate, could you change this to format-str?

concourse/steps/scan_sources.py
```diff
@@ -197,54 +199,99 @@ def scan_component_with_whitesource(whitesource_cfg_name: str,
...
             post_project_object.whitesource_client.post_product(
                 product_token=post_project_object.product_token,
                 component_name=post_project_object.component_name.name(),
+                component_version=post_project_object.component_version,
                 requester_email=requester_mail,
                 extra_whitesource_config=extra_whitesource_config,
                 file=tmp_file)
-            ci.util.info('project posted\n')
+
+        # get all projects
```
ccwienk, line 253: nit: this comment is rather redundant, seeing that it is placed immediately before calling a method named `get_all_projects` :-)

concourse/steps/scan_sources.py
```diff
@@ -197,54 +199,99 @@ def scan_component_with_whitesource(whitesource_cfg_name: str,
...
+
+        if send_notificaton:
+            # generate reporting table for notification
+            tables = whitesource.util.generate_reporting_tables(report=report,
+                                                                threshold=cve_threshold,
+                                                                tablefmt="html")
+
+            # get product risk report
+            ci.util.info('retrieving product risk report')
+            prr = client.get_product_risk_report(product_token=product_token).content
```
ccwienk, line 276: simiar as to the other comment last time (mv to-json-conversion into client method), I would also in this case rather return the str, as opposed to exposing the http-request response, which:

a) makes consumption unncessarily cumbersome
b) exposes implementation details that a user of the client should not (have) to know or care about

concourse/steps/scan_sources.py
```diff
@@ -197,54 +199,99 @@ def scan_component_with_whitesource(whitesource_cfg_name: str,
...
+
+            # assemble html body
```
ccwienk, line 278: again, I think this comment kind of repeats the code..

concourse/steps/scan_sources.py
```diff
@@ -197,54 +199,99 @@ def scan_component_with_whitesource(whitesource_cfg_name: str,
...
+            body = whitesource.util.assemble_mail_body(tables=tables,
+                                                       threshold=cve_threshold)
+
+            # send mail
+            ci.util.info('sending notification')
+            dto = datetime.now()
```
ccwienk, line 284: nit: why `dto` - as opposed to e.g. `now`?

whitesource/util.py
```diff
@@ -1,10 +1,167 @@
...
+
+    # monkeypatch: disable html escaping
+    tabulate.htmlescape = lambda x: x
+
+    # split respecting CVSS-V3 threshold
+    above = {}
+    below = {}
+
+    for lib, dic in report.items():
+        if float(dic['CVSS-V3']) > threshold:
```
ccwienk, line 30: I still would have preferred to use a dataclass - or at least deduplicate the accessor-function returning the cvss-float :-/

whitesource/util.py
```diff
@@ -1,10 +1,167 @@
...
+            above[lib] = dic
+        else:
+            below[lib] = dic
```
ccwienk, line 33: btw, this approach has the disadvantage that conflicts will not be detected (equally-named libs..)
also, I would prefer to use more expressive names, e.g. `lib_name`, i.p. seeing that the report is simply a dict (is that the case?)

ccwienk: as a general "rule-of-thumb", I think that if using less strict typing, at least more expressive names / naming conventions ought to be used

whitesource/util.py
```diff
@@ -1,10 +1,167 @@
...
+
+    def _sort_cve_list(struct):
```
ccwienk, line 35: I really do not like a generic parameter name such as `struct` :disappointed:

whitesource/util.py
```diff
@@ -1,10 +1,167 @@
...
+
+        # find greatest cve per project
+        for vul in pvr["vulnerabilities"]:
+            try:
+                report = _add_cve_entry(report=report,
+                                        cvss_key_name="cvss3_score",
+                                        pname=pname,
+                                        vul=vul)
+            except KeyError:
+                # https://github.com/gardener/cc-utils/pull/476#discussion_r490231239
```
ccwienk, line 129: it might be nice to add also the quoted link and part of the response to this piece of code, as it - as discussed - is not obvious (I think)

ccwienk (changes requested):

whitesource/util.py
```diff
@@ -1,10 +1,167 @@
...
+                </a>
+                page. Appropriate filters have to applied manually, "Gardener" is a matching keyword.
+            </p>
+        </div>
+    '''
+
+
+def find_greatest_cve(projects,
+                      client):
+    report = {}
```
zkdev, line 110: introduce dataclass

ccwienk (approved):

PR #480, opened 2020-09-20
Declare license and copyright information using REUSE

msohn:
We now use https://reuse.software/ to declare license and copyright
information in a machine readable format.

The copyright statement for SAP open source projects was updated
to give credit to project contributors not employed by SAP.

Fixes #

**Release note**:

```noteworthy developer
License and copyright information is now specified in REUSE format.
```


ccwienk:
@msohn : sounds like a good approach.

Is it possible to inject the copyright hints to the _end_ of the source code files? They are usually of no interest at all, but they repeatedly appear most prominent when opening a file w/o stored editing position, which I experience to be quite annoying..

msohn:
https://reuse.software/tutorial/#step-2 says the copyright and license should be a in header.
What you propose would be a footer.

With the new format it's 3 lines, or if we remove the empty blank line, just 2.
IDEs can fold comments, not sure which editor or IDE you use.

msohn:
Also see the discussion in https://github.com/gardener/aws-lb-readvertiser/pull/14.
The main open issue is full automation of generating license/copyright information for source dependencies (e.g. in golang vendor/ folder). I have a semi-automated approach right now, and filed a feature request for scancode-toolkit to add support for exporting copyright/license information in debian copyright format used in .reuse/dep5.

ccwienk:
> https://reuse.software/tutorial/#step-2 says the copyright and license should be a in header.
> What you propose would be a footer.
> 
> With the new format it's 3 lines, or if we remove the empty blank line, just 2.
> IDEs can fold comments, not sure which editor or IDE you use.

I use `vim`. However, I think it is a weak argument that one could workaround the disadvantages of cluttering source files by using tools. As I see it, _nobody_ working on the code has any interest in this license information, but it is the first thing everyone working on the code sees. I do not see why it would make any difference to put the license information in the footer.

Also, using the header has yet another disadvantage: it will break ´shebang` lines (this happened already the last time when bulk-insertions of this license header were done to this repository).

As for the mako (`.mako` in most cases) files, also, we will need to escape the comments using two hash characters (`##`), lest they will be included in the rendering results. Is it possible to configure this?

There is yet another commenting style required for the sphinx (`.rst` in most cases) files.

Why is it even necessary to add this information redundantly into each individual source file in the first place? The files are grouped together by residing in the same git repository. Would it not suffice to declare once that the license conditions apply for each individual file in the repository, and maybe add a stmt that if using only parts of the repository, the license information must then (and only then) be incorporated into the individual files?

To me, this seems like cargo-cult-compliance.

msohn:
This is a short machine-readable form of implementing what the Apache-2.0 license says itself in
https://www.apache.org/licenses/LICENSE-2.0#apply

All the open source foundations and prominent projects I am aware of are putting license headers in all source code files.
This helps downstream consumers who copy some code to keep track what is the license and who are the original copyright holders. Sometimes only a few files are copied to another project, then this practice ensures this information isn't lost.
There are several Gardener repositories which have copied some source code from other projects (mostly from kubernetes). 

See e.g. 
http://www.gnu.org/licenses/gpl-howto.html
https://www.apache.org/legal/src-headers.html
https://www.eclipse.org/projects/handbook/#ip-copyright-headers
https://wiki.openstack.org/wiki/LegalIssuesFAQ#Copyright_Headers

Also CNCF is strongly recommending this, see
https://github.com/open-telemetry/community/issues/113
https://github.com/cncf/foundation/blob/master/copyright-notices.md

Also Linux kernel adopted SPDX license headers, see
https://www.kernel.org/doc/html/v4.18/process/license-rules.html, 
and here is an example
https://github.com/torvalds/linux/blob/master/kernel/dma.c

For file formats where this is impossible or hard it's sufficient to list their license and copyright in .reuse/dep5.
I'd say use the reuse addheader command to add license/copyright statements in the standard format and
for all file formats not supported by this tool list this information in .reuse/dep5.
Here is the corresponding reuse command:

```bash
reuse addheader --copyright="SAP SE or an SAP affiliate company and Gardener contributors" --license=Apache-2.0 <file path>
```

AFAIK this command respects shebang lines.

ccwienk:
@msohn : I am aware that many projects to this (hence my term "cargo-cult compliance". However, I still would like to challenge whether this is really necessary to be done like that, for the aforementioned reasons.

PR #487, opened 2020-10-16
Add transport tool for component descriptor v2

lcavajani:

This commit adds the version 2 of the tool originally created by Christian Cwienk.

It adds the possibility to use the tool offline in airgap environment.

The tool only supports ociImage resource type for the moment.

I have also updated the unit tests and added new ones for the
downloaders and uploaders.

**Release note**:

```improvement operator
Add transport tool for component descriptor v2
```


gardener-robot-ci-3:
Thank you @lcavajani for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

transport/processing/processing.py
```diff
@@ -0,0 +1,321 @@
...
+                itertools.repeat(component, filters_count),
+                itertools.repeat(container_image, filters_count),
+            )
+        )
+
+    def process(self, component, container_image):
+        if not self.matches(component, container_image):
+            return None
+
+        logging.info(
```
ccwienk, line 58: nit: I think it is a better practice to use named loggers, like so:

```

# ...

logger.info('something happend')
```

transport/transport.py
```diff
@@ -0,0 +1,147 @@
...
+
+
+def parse_args():
+    parent_parser = argparse.ArgumentParser(
+        description='Transport tool for component descriptor v2',
+    )
+
+    subparsers = parent_parser.add_subparsers(dest="action", help="Actions", required=True)
+    parser_download = subparsers.add_parser(
+            name="download",
```
ccwienk, line 22: might make sense to extract the actions to an enum (since they are being checked for later)

transport/processing/processing_component.py
```diff
@@ -0,0 +1,153 @@
...
+                component_version=version,
+                ctx_repo_base_url=ctx_base_url,
+        )
+    except container.registry.OciImageNotFoundException as err_not_found:
+        ci.util.error(err_not_found)
+        sys.exit(1)
+
+
+def _resolve_cd_from_references(component_descriptor: cm.ComponentDescriptor):
+    ctx_base_url = component_descriptor.component.repositoryContexts[0].baseUrl
```
ccwienk, line 34: I think this should be `-1` (the _last_ ctx-repo-URL)

transport/processing/processing_component.py
```diff
@@ -0,0 +1,153 @@
...
+
+    def __enumerate_cd_from_references(
```
ccwienk, line 36: nit: why name it w/ dunder-prefix? it is only visible locally, anyways

transport/processing/processing_component.py
```diff
@@ -0,0 +1,153 @@
...
+            descriptor_path = ComponentTool.gen_yaml_file_path(oci_ref)
+            if os.path.isfile(descriptor_path):
+                yield parse_component_descriptor(descriptor_path)
+            else:
+                yield _download_descriptor(comp_ref.componentName, comp_ref.version, ctx_base_url)
+
+    for cd_from_ref in __enumerate_cd_from_references(component_descriptor, ctx_base_url):
+        yield (cd_from_ref.component.name, cd_from_ref)
+        # Recurse if a component descriptor has itself other references
+        if cd_from_ref.component.componentReferences != []:
```
ccwienk, line 55: nit: I think relying on python's boolean-conversions would be more pythonic here (`if  cd_from_ref.component.componentReferences: ...`)

transport/processing/processing_model.py
```diff
@@ -0,0 +1,7 @@
+import collections
+
+ProcessingJob = collections.namedtuple(
+    'ProcessingJob',
+    ['component', 'container_image', 'download_request',
```
ccwienk, line 5: codestyl-nit: in our codebas, we prefer to have a single item per line, or all in the same line (not mix styles)
also, consider using @dataclasses.dataclass - those are (I think) easier to read, while introducing minimal boilerplate overhead (and allow for more flexibilty)

transport/processing/processors.py
```diff
@@ -0,0 +1,48 @@
...
+import functools
+import os
+
+import container.model
+import container.util
+
+OWN_DIR = os.path.abspath(os.path.dirname(__file__))
+
+
+class ProcessorBase():
```
ccwienk, line 11: nit: parantheses are optional in python3 (-> `class ProcessorBase:`)

transport/processing/uploaders.py
```diff
@@ -0,0 +1,76 @@
...
+
+class PrefixUploader():
+    def __init__(
+        self,
+        prefix,
+        context_url,
+        mangle=True,
+        **kwargs
+    ):
+        super().__init__(**kwargs)
```
ccwienk, line 22: I think both the invocation of super-class's initialised, and the ```**kwargs`` parameter should be removed
also: consider using a dataclass (not 100% sure whether this is a good match in this case, though)?

ccwienk (commented):
overall lgtm - some stylistic nits (see inline comments)

transport/processing/processing_component.py
lcavajani: right, good catch

transport/processing/processing_component.py
lcavajani: no reason ;) fixed

transport/processing/processors.py
lcavajani: great, I thought only `object` was not required anymore

transport/processing/uploaders.py
lcavajani: I've switched to a dataclass, I think the code looks better

transport/transport.py
lcavajani: definitely makes sense ;)

lcavajani:
@ccwienk Thanks for the review ! I've addressed your comments and rebased ;)

transport/processing/processing_component.py
```diff
@@ -0,0 +1,160 @@
...
+        sys.exit(1)
+
+
+def _resolve_cd_from_references(component_descriptor: cm.ComponentDescriptor):
+    ctx_base_url = component_descriptor.component.repositoryContexts[-1].baseUrl
+
+    def enumerate_cd_from_references(
+            component_descriptor: cm.ComponentDescriptor,
+            ctx_base_url: str
+        ):
```
AndreasBurger, line 43: nit: wrong indentation

```suggestion
    def enumerate_cd_from_references(
        component_descriptor: cm.ComponentDescriptor,
        ctx_base_url: str,
    ):
```

transport/processing/processing_component.py
```diff
@@ -0,0 +1,160 @@
...
+        if cd_from_ref.component.componentReferences:
+            yield from _resolve_cd_from_references(cd_from_ref)
+
+
+@dataclass
+class ComponentTool:
+    name: str
+    version: str
+    ctx_base_url: str
+    descriptor: cm.ComponentDescriptor = None
```
AndreasBurger, line 68: consider `typing.Optional[cm.ComponentDescriptor]`

transport/processing/processing_component.py
AndreasBurger: On second thought: That might be old code of ours. If so disregard the comment.

transport/processing/processing.py
```diff
@@ -0,0 +1,332 @@
...
+    NewResources = collections.namedtuple(
+        'NewResources',
+        ['resources', 'expected']
+    )
+
+    # only modify the OCI_REGISTRY resources
+    def _new_resources(resources: cm.AccessType) -> NewResources:
+        return NewResources(
+            resources=list(r for r in resources if r.access.type != cm.AccessType.OCI_REGISTRY),
+            expected=len(resources)
```
AndreasBurger, line 258: Maybe naming it something like `expected_resource_count` or something like that would be more explicit - not sure.
@ccwienk is usually better at naming stuff than me.

Also, could you point me to where that is incremented?

transport/processing.cfg
```diff
@@ -0,0 +1,108 @@
+# vim: set syntax=yaml:
```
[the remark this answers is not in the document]
lcavajani: it does not do any harm anyway but I will change the extension of the config file to `.yaml` so the format is more obvious.

transport/processing/uploaders.py
ccwienk: glad to hear you seem to like the change :-)

ci/util.py
```diff
@@ -230,6 +231,65 @@ def none(value):
...
 
+class Checksum():
+    '''
+    Manage checksum from files
+
+    Args:
+        algo (str): Algorithm to use, default to SHA256
+    '''
+    def __init__(self, algo="sha256"):
+        self.algo = getattr(hashlib, algo)()
```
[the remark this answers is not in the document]
ccwienk: @AndreasBurger : +1 for human-friendly error-messages

however, i think the even cleaner way would be to use `hashlib.new`, which also does this validation and raises ValueError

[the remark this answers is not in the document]
ccwienk:
```>>> hashlib.new(name='foo')
  File "/usr/lib/python3.8/hashlib.py", line 157, in __hash_new
    return _hashlib.new(name, data)
ValueError: unsupported hash type

During handling of the above exception, another exception occurred:

  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.8/hashlib.py", line 163, in __hash_new
    return __get_builtin_constructor(name)(data)
  File "/usr/lib/python3.8/hashlib.py", line 120, in __get_builtin_constructor
    raise ValueError('unsupported hash type ' + name)
ValueError: unsupported hash type foo
>>> hashlib.new(name='sha256')
<sha256 HASH object @ 0x7fa3c2d67810>
```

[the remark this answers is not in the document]
ccwienk: I took the liberty to patch this myself :-)

transport/processing.cfg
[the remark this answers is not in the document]
ccwienk: @AndreasBurger : actually, I would welcome it to add this information to our entire codebase.. :-)

transport/processing/processing.py
ccwienk: additional nit: rather use a `list-comprehension` (square brackets []), rather than a generator-comprehension and immediate conversion into a list.

lcavajani:
I will address the comment in a new PR, thanks again for the review !

transport/processing.cfg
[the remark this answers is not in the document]
lcavajani: yes, `# vim: set syntax=yaml sw=4 ts=4 expandtab:` would do the trick if `vim` does not detect automatically the indent

PR #488, opened 2020-10-20
Address PR comments for transport tool v2

lcavajani:

Refactor process resources with a class to improve readability.

Fix typing issues.

Set vim parameters in config file.

Signed-off-by: lcavajani <lcavajani@suse.com>

PR: https://github.com/gardener/cc-utils/pull/487

**Release note**:

```improvement operator
NONE
```


gardener-robot-ci-1:
Thank you @lcavajani for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

transport/processing/processing.py
```diff
@@ -133,9 +132,9 @@ def _uploader(uploader_cfg: dict):
 
 
 def processing_pipeline(
-    processing_cfg: dict,
-    shared_processors: dict={},
-    shared_uploaders: dict={},
+        processing_cfg: dict,
+        shared_processors: dict = {},
```
ccwienk, line 136: general remark: it can lead to unexpected side-effects if using mutable values as default values.

consider e.g.:

```
def foo(x,a={}):
 for k,v in x.items():
  a[k] = v
 print(a)
>>> foo({1:1})
{1: 1}
>>> foo({2:1})
{1: 1, 2: 1}
>>> foo({3:1})
{1: 1, 2: 1, 3: 1}
```

ccwienk: that being said: in this particular case, it will pbly not be an issue (+ we do have several similar constructs in cc-utils, currently - I only learnt about this not foo far back, when reading through @dataclasses..)

transport/processing/processing_component.py
```diff
@@ -1,9 +1,10 @@
+from dataclasses import dataclass, field
```
ccwienk, line 1: yet another stylistic remark: I personally prefer to stick with `import a.b.c` import-style, and use import-aliases if import-names would become too long (pbly this is not a generally-approved practice in python-community - so from-import is in general also fine w/ me)

however, we do typically sort our imports alphabetically (not sure, however, how this fits with mixing import and from-import..) :-/

transport/processing/processors.py
```diff
@@ -21,13 +21,13 @@ def process(self, processing_job):
 
 class FileFilter(ProcessorBase):
     def __init__(
-        self,
-        filter_files,
+            self,
```
ccwienk, line 24: nit: are you sure the additional indentation is "correct"? I would have thunk it was better before..

ccwienk (commented):
some nits, otherwise lgtm - ty

transport/processing/processors.py
lcavajani: I was a bit surprised as well but this is in PEP 8, basically we need an extra level of indentation to distinguish arguments from the rest.

transport/processing/processing.py
lcavajani: hmm that's a really good point in general I was not aware of.

It is not an issue in this case because we set a default value before, when we look in the processing configuration. Since they are not useful, I will remove them.

transport/processing/processors.py
ccwienk: hmm, okay, if pep8 says so, I shall not question this :-)

PR #494, opened 2020-11-10
Rename Concourse-Pipeline objects

AndreasBurger:
Pipeline-object-names with an underscore in them are marked as deprecated and will raise an error in a future, as of yet unspecified Concourse version.

This PR contains two commits in preparation for this:
- Pipeline-Objects whose names are generated now use `.` as separator instead of `_`
- The step-name of our `component_descriptor` step was hard-coded in many places. This was changed to help with future work.

Still left to do, for a future PR:
- Switch all our hard-coded step names (e.g.: `scan_images`). This would require fixing all pipelines that define custom steps dependent on our steps.

### DIFF
--- a/concourse/model/traits/component_descriptor.py
+++ b/concourse/model/traits/component_descriptor.py
@@ -47,10 +47,13 @@ class ValidationPolicy(EnumWithDocumentation):
     )
 
 
+DEFAULT_COMPONENT_DESCRIPTOR_STEP_NAME = 'component_descriptor'
+
+
 ATTRIBUTES = (
     AttributeSpec.optional(
         name='step',
-        default={'name': 'component_descriptor'},
+        default={'name': DEFAULT_COMPONENT_DESCRIPTOR_STEP_NAME},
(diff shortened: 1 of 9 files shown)


concourse/model/traits/component_descriptor.py
```diff
@@ -47,10 +47,13 @@ class ValidationPolicy(EnumWithDocumentation):
     )
 
 
+DEFAULT_COMPONENT_DESCRIPTOR_STEP_NAME = 'component_descriptor'
```
ccwienk, line 50: nit: it is actually _the_ component-descriptor-step-name (we cannot change this anymore easily)

ccwienk (approved):
+1 for deduplicating the component-descriptor-step-name

otherwise lgtm

concourse/model/traits/component_descriptor.py
AndreasBurger: We cannot (or rather should not), but it is a default that can be overridden by the User (see l.56 below)

PR #498, opened 2020-11-27
Add initial support for labels to protecode scanning

AndreasBurger:


protecode/util.py
```diff
@@ -114,13 +115,22 @@ def group_resources(components):
             for resource_name, resources in resource_groups.items():
                 yield resources
 
+        def _has_required_scan_label(resource):
+            for label in product.v2.scan_labels(resource.access):
```
ccwienk, line 119: I think it might make sense to add a util-function that will already do the filtering (similar to `scan_labels`). in fact, there is already such a method :-)))

--> https://github.com/gardener/component-spec/blob/master/bindings-python/gci/componentmodel.py#L105

protecode/util.py
```diff
@@ -114,13 +115,22 @@ def group_resources(components):
...
+            return False
+
         for component_name, components in component_groups.items():
             for grouped_resources in group_resources(components):
                 # all components in a component group share a name
                 component = next(iter(components))
                 resources = [
                     r for r in grouped_resources
                     if image_reference_filter(component, r)
+                    or _has_required_scan_label(r)
```
ccwienk, line 133: not sure - I think we also need to look into the label's value, do we not? :-)

At the very least (I did not forget we discussed to start with a minimal increment, and not go for the full-blown solution in this first step), I think the semantics should be like so:

```
foreach component:
  filter relevant resources (ignore sources for now):
   if resource-type !== oci-image: skip (+ add TODO - we should in general handle other types "in the future")
   if resource-not-has-"binary-scan"-label: use default behaviour (i.e. scan it)
   if resource-has-"binary-scan"-label: act according to label (either do-scan, or not-do-scan - future extension: path-filters)
```

I think/hope this should not require a lot of changes to this PR. On additional note, we also need to consider a better way for expressing the "resource-groups" (that we could do in a later PR - let us keep the first one focussed and small)

ccwienk (changes requested):

protecode/util.py
```diff
@@ -114,13 +116,25 @@ def group_resources(components):
...
                 yield resources
 
+        def _filter_resources_to_scan(component, resource):
+            # check whether the trait was configured to filter out the resource
+            configured_image_reference_filter_response = image_reference_filter(component, resource)
+            if not configured_image_reference_filter_response:
+                return False
+
+            # check for scanning labels on resource in cd
+            if label := resource.find_label(name=sdo.labels.ScanLabelName.BINARY_SCAN):
```
jkrayl, line 126: does this work? You are passing the enum value instead of a string. Do you handle the ValueError exception?

PR #499, opened 2020-11-27
Ws v2 labeling

jkrayl:


ccc/github.py
```diff
@@ -226,56 +224,11 @@ def log_stack_trace_information_hook(resp, *args, **kwargs):
...
-    else:
-        raise ValueError
-
-
-def github_api_from_gh_access(access: cm.GithubAccess):
+def github_api_from_gh_access(
+    access: cm.GithubAccess,
+) -> typing.Union[github3.github.GitHub, github3.github.GitHubEnterprise]:
     if access.type is not cm.AccessType.GITHUB:
         raise ValueError
```
ccwienk, line 231: nit: I think it would be a good idea to pass `access.type` as argument for `ValueError` (so the wrong/unexpected value is printed as part of the stack-trace)

concourse/steps/scan_sources.py
```diff
@@ -63,98 +66,156 @@ def scan_component_with_whitesource(
...
 
     # parse component_descriptor
     ci.util.info('parsing component descriptor')
     component_descriptor = cm.ComponentDescriptor.from_dict(
         ci.util.parse_yaml_file(component_descriptor_path)
     )
 
-    # for component in []: 'components matching component_descriptor'
-    for component in product.v2.components(component_descriptor_v2=component_descriptor):
+    product_name = component_descriptor.component.name
```
ccwienk, line 79: in a subsequent change, we might want to mv the naming-logic (component.name -> product_name) out of the template (into a util- or model-function in whitesource-pkg). At least in case we rely on a certain naming scheme, and this becomes part of a contract we use.

concourse/steps/scan_sources.py
```diff
@@ -63,98 +66,156 @@ def scan_component_with_whitesource(
...
+    notify_users(
+        notification_recipients=notification_recipients,
+        cve_threshold=cve_threshold,
+        ws_client=ws_client,
+        product_name=product_name,
+        product_token=product_token,
+    )
+
+
+def scan_artifact_with_ws(
```
ccwienk, line 107: hmm... this is a lot of code in our template. we should consider moving it into the `whitesource`-pkg, maybe (keep the interface between template and our libs small..)

concourse/steps/scan_sources.py
```diff
@@ -63,98 +66,156 @@ def scan_component_with_whitesource(
...
+        # download whitesource component
+        clogger.info('downloading component for scan')
+        component_filename = whitesource.component.download_component(
+            clogger=clogger,
+            github_repo=github_repo,
+            path_filter_func=path_filter_func,
+            ref=commit_hash,
+            target=tmp_file,
+        )
+        # needed
```
ccwienk, line 148: :-)

don't change the following line, lest things no longer work

concourse/steps/scan_sources.py
```diff
@@ -63,98 +66,156 @@ def scan_component_with_whitesource(
...
             threshold=cve_threshold,
-            tablefmt='simple',
+            tablefmt='html',
+        )
+
+        # get product risk report
+        ci.util.info('retrieving product risk report')
+        prr = ws_client.get_product_risk_report(product_token=product_token)
+
+        # assemble html body
```
ccwienk, line 198: nit: I think this comment might be omitted, as it merely repeats the called function's name (which is an indication that said function is well-named)

concourse/steps/scan_sources.py
```diff
@@ -63,98 +66,156 @@ def scan_component_with_whitesource(
...
-            whitesource.util.send_mail(
-                body=body,
-                recipients=notification_recipients,
-                component_name=component.name,
-                attachments=[attachment],
-            )
+        # send mail
+        ci.util.info('sending notification')
+        now = datetime.now()
+        fname = f'{product_name}-{now.year}.{now.month}-{now.day}-product-risk-report.pdf'
```
ccwienk, line 207: firstly: consider inlining (only used once)
secondly: how about `datetime.datetime.now().strftime('%Y-%m-%d')`? Admittedly, this is not actually a lot shorter, and not necessarily easier to read.. so the approach chosen here might be the better choice

whitesource/component.py
```diff
@@ -1,77 +1,84 @@
...
-    return PostProjectObject(
-        whitesource_client=whitesource_client,
-        github_api=github_api,
-        product_token=product_token,
-        component=component,
-        github_repo=github_repo,
-        component_version=component.version,
-    )
+def _get_ws_label_from_source(labels: typing.List[cm.Label]) -> sdo.labels.SourceIdHint:
+    for label in labels:
```
ccwienk, line 15: consider https://github.com/gardener/component-spec/blob/master/bindings-python/gci/componentmodel.py#L105 :-)

ccwienk: maybe we could add a `type`-attr to said find-function that one can pass a conversion-callable to also handle the deserialisation, wdyt?

ccwienk (changes requested):

concourse/steps/scan_sources.py
ccwienk: I just realised (by accident..) that this approach does have one disadvantage: it will not include a leading `9` for months and days < 10. -> either ensure a leading zero, or use strftime, please..

ccc/github.py
jkrayl: I added the complete access as argument

concourse/steps/scan_sources.py
jkrayl: moved the code to the util package

concourse/steps/scan_sources.py
jkrayl: you are right. I removed a couple of obvious comments

whitesource/component.py
jkrayl: the function now uses the find_label method :)

jkrayl: To handle the deserialisation inside the find_label function would be good

AndreasBurger (approved):

ccwienk (approved):

PR #503, opened 2020-12-03
Attach pdf scan-reports to component-owner mails

AndreasBurger:


concourse/steps/scan_container_images.py
```diff
@@ -152,6 +153,32 @@ def mail_body(self):
 
         return ''.join(parts)
 
+    def pdf_report_attachments(self):
+        attachments = []
+
+        def report_filename(upload_result):
+            return upload_result.component.name.replace('.','_').replace('/', '_') + '-report.pdf'
```
ccwienk, line 160: not opposed to doing that - but why replace `.`? the `/`-part, is of course obviously a good idea

protecode/scanning_util.py
```diff
@@ -402,10 +402,15 @@ def enumerate_reference_triages():
         # yield results
         for protecode_app in protecode_apps_to_consider:
             scan_result = self._api.scan_result(protecode_app.product_id())
+
+            def pdf_report_callback():
```
ccwienk, line 406: naming-remark: this is not actually a callback, is it?

a brief remark as to _why_ we pass the "promise"-like callable might be a good idea (avoid loading everything into memory longer than necessary)

ccwienk (commented):
one remark w.r.t. naming - otherwise lgtm

PR #504, opened 2020-12-04
Prepare codepath for CTF use

AndreasBurger:


concourse/steps/component_descriptor.mako
```diff
@@ -147,6 +150,7 @@ subproc_env['MAIN_REPO_DIR'] = main_repo_path
 subproc_env['BASE_DEFINITION_PATH'] = base_descriptor_file_v2
 subproc_env['COMPONENT_DESCRIPTOR_PATH'] = v2_outfile
 subproc_env['COMPONENT_DESCRIPTOR_PATH_V1'] = descriptor_path
+subproc_env['COMPONENT_DESCRIPTOR_PATH_CTF'] = ctf_out_dir
```
ccwienk, line 153: I think it should rather only have `CTF` in the name (e.g. `CTF_PATH`)

concourse/steps/component_descriptor.mako
```diff
@@ -113,6 +113,9 @@ v2_outfile = os.path.join(
   component_descriptor_fname(schema_version=gci.componentmodel.SchemaVersion.V2),
 )
 
+ctf_out_dir = os.path.join(descriptor_out_dir, 'ctf')
```
ccwienk, line 116: how about using a more expressive name for the directory? nobody will have to type it (except for you in this particular codeline^^)..

how about sth like `cnudie_transport_format-out.d`?

concourse/steps/component_descriptor.mako
```diff
@@ -174,22 +178,28 @@ subprocess.run(
...
-  fail(f'no descriptor file was found at: {v2_outfile=}')
+if os.path.isdir(ctf_out_dir) and os.path.listdir(ctf_out_dir):
+  if os.path.isfile(v2_outfile):
+    fail('Both CTF dir and component-descriptor-v2-file found.')
+  else:
+    pass
+else:
+  # ensure the script actually created an output
+  if not os.path.isfile(v2_outfile):
+    fail(f'no descriptor file was found at: {v2_outfile=}')
```
ccwienk, line 189: I think this check is redundant, as `parse_yaml_file` will also fail if the file be absent

AndreasBurger: Strictly speaking that was not changed/introduced by this PR, but I'll refactor it

AndreasBurger: On second thought - I prefer the more expressive error message.

ccwienk: `parse_yaml_file` does call `existing_file`, does it not?

ccwienk: so you refer to the `no descriptor file` part? ok, fine w/ me

concourse/steps/component_descriptor.mako
```diff
@@ -113,6 +113,9 @@ v2_outfile = os.path.join(
   component_descriptor_fname(schema_version=gci.componentmodel.SchemaVersion.V2),
 )
 
+cnudie_transport_format_out_dir = os.path.join(descriptor_out_dir, 'ctf')
+os.mkdir(cnudie_transport_format_out_dir)
```
ccwienk, line 117: I think we should not create this directory

PR #505, opened 2020-12-07
Add support for defining labels in the publish-trait

AndreasBurger:


concourse/model/traits/publish.py
```diff
@@ -78,6 +82,11 @@
         default=None,
         doc='only for multistage builds: the target up to which to build',
     ),
+    AttributeSpec.optional(
+        name='component_descriptor_labels',
+        default=[],
+        doc="a list of labels to add to the images' generated resource in the Component Descriptor",
```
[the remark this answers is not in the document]
ccwienk: also: the resource declaration is _injected_ into the _base_ CD

concourse/model/traits/publish.py
```diff
@@ -137,6 +149,21 @@ def resource_name(self):
...
+                try:
+                    dacite.from_dict(
+                        data_class=Label,
+                        data=label,
+                        config=dacite.Config(strict=True),
+                    )
+                except dacite.DaciteError as e:
+                    raise ModelValidationError(
+                        f'Invalid label "{label}".'
+                    ) from e
```
ccwienk, line 165: mhhh, I have to admit that I did not know `raise from` - nice!

concourse/model/traits/publish.py
```diff
@@ -137,6 +149,21 @@ def resource_name(self):
     def _children(self):
         return ()
 
+    def validate(self):
+        super().validate()
+        if labels := self.component_descriptor_labels():
```
ccwienk, line 154: stylistic nit: you might save one level of indentation like so:
```
for label in labels or ():
 ...
```
+ if keeping the approach of checking for a truthy value, it would be better (stricter) to explicitly check for the value being `None` (because, a scalar, false value (such as `false`, or `0`) would be invalid in this ctx).

+ if keeping the if, I would prefer an early-exit (thus also reducing the nesting by one)

concourse/steps/component_descriptor.mako
```diff
@@ -79,6 +79,7 @@ component_v2.resources.append(
       type=cm.AccessType.OCI_REGISTRY,
       imageReference='${image_descriptor.image_reference()}' + ':' + effective_version,
     ),
+    labels=${image_descriptor.component_descriptor_labels()}
```
ccwienk, line 82: does this not require a fallback value (e.g. `None`, or `()`)?

concourse/model/traits/publish.py
```diff
@@ -78,6 +82,11 @@
         default=None,
         doc='only for multistage builds: the target up to which to build',
     ),
+    AttributeSpec.optional(
+        name='component_descriptor_labels',
+        default=[],
```
ccwienk, line 87: is there an issue if we use the empty tuple instead of the empty list as default value? in general, default values should preferrably be immutable..

ccwienk (commented):
some stylistic nits + one potential bug (did not test..)

concourse/steps/component_descriptor.mako
AndreasBurger: That default value is provided by our AttributeSpec implementation.

ccwienk: good point. however: the tuple literal (`()`) is not valid yaml, is it? :-/
I know, I suggested to use tuple instead of list (however, I _asked_ whether there would be a problem if switching to it^^)

ccwienk (changes requested):

PR #506, opened 2020-12-07
use custom oci-client for verbatim image-replication

ccwienk:


oci/client.py
```diff
@@ -276,6 +277,10 @@ def blob(
             timeout=None,
         )
 
+        if absent_ok and res.status_code == 404:
```
AndreasBurger, line 280: consider using
```suggestion
        if absent_ok and res.status_code == request.codes.NOT_FOUND:
```

AndreasBurger (approved):

oci/client.py
ccwienk: sure, good idea

PR #507, opened 2020-12-08
Add support for defining labels in the component-descriptor trait

AndreasBurger:


concourse/model/traits/component_descriptor.py
```diff
@@ -83,6 +85,12 @@
             If not configured, the CICD-landscape's default ctx will be used.
         '''
     ),
+    AttributeSpec.optional(
+        name='component_descriptor_labels',
```
ccwienk, line 89: I would prefer to call it `component_labels` - it is already below the `component_descriptor_trait` (DRY)

concourse/model/traits/component_descriptor.py
```diff
@@ -83,6 +85,12 @@
...
+        default=(),
+        type=list,
```
ccwienk, line 91: `typing.List[cm.Label]`?

concourse/steps/component_descriptor.mako
```diff
@@ -45,6 +45,7 @@ with open(version_file_path) as f:
   effective_version = f.read().strip()
 
 component_name = '${descriptor_trait.component_name()}'
+component_labels = ${descriptor_trait.component_descriptor_labels()}
```
ccwienk, line 48: I think I already asked that: will this work if the attr is not specified (as in: will the result be valid yaml)? caveat (sorry for "tricking" you into this: I think, the tuple literal is not compatible w/ yaml, is it?

ccwienk (changes requested):

jkrayl (approved):

concourse/steps/component_descriptor.mako
AndreasBurger: Yes, pyyaml will dump tuples as lists, but when using `safe_dump` the type-information is lost. On deserialisation the labels will be in a list.

ccwienk: I just tried:
```>>> yaml.dump({'a': ()})
'a: !!python/tuple []\n'
```

so I figure pyyaml will retain the tuple-metadata. As the component-descriptor-spec requires a _list_, I think we should also actually dump an (empty) list

concourse/steps/component_descriptor.py
```diff
@@ -81,7 +82,7 @@ def base_component_descriptor_v2(
         ],
         componentReferences=[], # added later
         resources=[], # added later
-        labels=[], # added later
+        labels=component_labels,
```
ccwienk: if this attr is to be modified later, defaulting it to a tuple is rather unhandy (I already changed this to a list, though)..

ccwienk (approved):

PR #508, opened 2020-12-08
Add support for defining labels on the main repository

AndreasBurger:


concourse/model/resources.py
```diff
@@ -367,3 +385,26 @@ def __str__(self):
             rp=self.repo_path(),
             b=self.branch()
         )
+
+    def validate(self):
+        super().validate()
+        labels = self.component_descriptor_labels()
+        if not isinstance(labels, Iterable):
```
ccwienk, line 392: just iterate over them / convert into a list, and rely on the typechecking that is already done there

you may still emit a specialised error, if you see this as a value-add (I do not tbh)

also, I do not think it is exactly appropriate to raise a MVE. ValueError seems to be more appropriate to me

concourse/model/resources.py
```diff
@@ -367,3 +385,26 @@ def __str__(self):
...
+            raise ModelValidationError(
+                f'Invalid component descriptor labels on repsitory config {self}'
+            )
+        if labels and not self.is_main_repo():
```
ccwienk, line 396: why? of course, all repositories should be possible to be labelled

concourse/model/resources.py
```diff
@@ -367,3 +385,26 @@ def __str__(self):
...
+            raise ModelValidationError(
+                "Definition of labels is only supported on the main repository, i.e. 'repo'"
```
ccwienk, line 398: what is `id est 'repo'` supposed to tell the reader? I do not understand this

concourse/steps/component_descriptor.py
```diff
@@ -38,6 +38,7 @@ def dump_component_descriptor_v2(component_descriptor_v2: cm.ComponentDescriptor
 def base_component_descriptor_v2(
     component_name_v2: str,
     effective_version: str,
+    source_labels: tuple,
```
ccwienk, line 41: please use proper type-hint. we do not require a tuple here.

-> typing.List[cm.Label] (please stick w/ the (new) convention of aliased importing (import gci.componentmodel as cm)

concourse/steps/component_descriptor.py
```diff
@@ -77,6 +78,7 @@ def base_component_descriptor_v2(
               commit=commit,
             ),
             version=effective_version,
+            labels=source_labels,
```
ccwienk, line 81: convert into a list. this on the one-hand-side will immediately validate that whathever is passed in is a least iterable, and also ensure that we always have a copy (thus avoiding unexpected side-effects from modifications done later-on).

ccwienk (changes requested):

PR #509, opened 2020-12-09
inputs from steps for component-descriptor-trait

ccwienk:
With the advent of CTF (CNUDIE Transport Format), it may be relevant to
pass outputs from build-steps to the component-descriptor step (so it
may e.g. upload local resources).


jkrayl (approved):
lgtm

PR #510, opened 2020-12-14
add base component archive to simplify the usage of the component-cli

schrodit:

Added an base archive with the base component descriptor to simplify the usage of the component-cli

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/concourse/steps/component_descriptor.mako
+++ b/concourse/steps/component_descriptor.mako
@@ -159,17 +159,23 @@ else:
   if not is_executable:
     fail('descriptor script file exists but is not executable: ' + descriptor_script)
 
+# dump base_ctf_v2 as valid component archive and pass it to descriptor script
+base_ctf_dir = os.path.join(descriptor_out_dir, 'base_component_ctf')
+os.makedirs(base_ctf_dir, exist_ok=False)
 
 # dump base_descriptor_v2 and pass it to descriptor script
-base_descriptor_file_v2 = os.path.join(descriptor_out_dir, 'base_component_descriptor_v2')
+base_descriptor_file_v2 = os.path.join(
+  descriptor_out_dir,
+  component_descriptor_fname(schema_version=gci.componentmodel.SchemaVersion.V2)
+)
 with open(base_descriptor_file_v2, 'w') as f:
   f.write(dump_component_descriptor_v2(base_descriptor_v2))
-
 
 subproc_env = os.environ.copy()
 subproc_env['${main_repo_path_env_var}'] = main_repo_path
 subproc_env['MAIN_REPO_DIR'] = main_repo_path
 subproc_env['BASE_DEFINITION_PATH'] = base_descriptor_file_v2
+subproc_env['BASE_CTF_PATH'] = base_ctf_dir
 subproc_env['COMPONENT_DESCRIPTOR_PATH'] = v2_outfile
 subproc_env['CTF_PATH'] = ctf_out_path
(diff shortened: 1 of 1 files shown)


concourse/steps/component_descriptor.mako
```diff
@@ -170,11 +170,21 @@ base_descriptor_file_v2 = os.path.join(descriptor_out_dir, 'base_component_descr
 with open(base_descriptor_file_v2, 'w') as f:
   f.write(dump_component_descriptor_v2(base_descriptor_v2))
 
+# dump base_archive_v2 as valid component archive and pass it to descriptor script
+base_archive_dir = os.path.join(descriptor_out_dir, 'base_component_archive')
+os.makedirs(base_archive_dir, exist_ok=True)
```
ccwienk, line 175: nit: as this code runs in a well-defined (containerised..) environment, I would consider the existence of the tgt-dir an error

concourse/steps/component_descriptor.mako
```diff
@@ -170,11 +170,21 @@ base_descriptor_file_v2 = os.path.join(descriptor_out_dir, 'base_component_descr
 with open(base_descriptor_file_v2, 'w') as f:
   f.write(dump_component_descriptor_v2(base_descriptor_v2))
 
+# dump base_archive_v2 as valid component archive and pass it to descriptor script
```
ccwienk, line 173: nit: did we not call this ctf-archive, or sth. like that?

concourse/steps/component_descriptor.mako
```diff
@@ -170,11 +170,21 @@ base_descriptor_file_v2 = os.path.join(descriptor_out_dir, 'base_component_descr
...
+base_archive_dir = os.path.join(descriptor_out_dir, 'base_component_archive')
+os.makedirs(base_archive_dir, exist_ok=True)
+base_archive_descriptor_file = os.path.join(
+  descriptor_out_dir,
+  component_descriptor_fname(schema_version=gci.componentmodel.SchemaVersion.V2)
+)
+with open(base_archive_descriptor_file, 'w') as f:
+  f.write(dump_component_descriptor_v2(base_descriptor_v2))
```
ccwienk, line 181: could you not simply cp (or rather symlink/hardlink) the file from `base_descriptor_file_v2`)?

ccwienk (changes requested):
apart from the nits, I have some concerns w/ this change, because it will now pass the same contents (base component-descriptor) through two different file system locations.

how about simply changing the environment variable name that is currently used to indicate the base-descriptor's name? each cd-script that adheres to the contract (i.e. use the path indicated by the env-var) will work as before; that way, we do no have to pass the file twice (which could also be confusing if e.g. someone were to enter the step-container to e.g. do debugging).

concourse/steps/component_descriptor.mako
schrodit: yes there is a `CTF_PATH` but thats for the output. I wanted to add something similar to the base component descriptors so that if its used it shoiuld be copied to `CTF_PATH`.
I can also name it accordingly `BASE_CTF_PATH`

concourse/steps/component_descriptor.mako
schrodit: added now a `os.link`

schrodit:
I applied your suggestions and now the base component descriptor is only written once to `$DESCRIPTOR_PATH/base_descriptor_file_v2/component_descriptor`.
The `BASE_DEFINITION_PATH` does now also point to that file.

PR #511, opened 2020-12-15
Add support for aborting running jobs on push-events to whd

AndreasBurger:
Allows one to configure whether a build should be aborted on follow-up push. Old behaviour (no aborting) is preserved if no config is given. Aborting can be configured on a per-job basis. Config example:

```yaml
…
  jobs:
    foo:
      abort_outdated_jobs: 'all' # all | forced_only
      steps:
        …
      traits:
        …
…
```

### DIFF
--- a/ci/util.py
+++ b/ci/util.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 
 import collections
+import functools
 import hashlib
 import os
 import pathlib
@@ -496,7 +497,7 @@ def which(cmd_name: str) -> str:
     return cmd_path
 
 
-def merge_dicts(base: dict, other: dict, list_semantics='merge'):
+def merge_dicts(base: dict, *other: dict, list_semantics='merge'):
     '''
     merges copies of the given dict instances and returns the merge result.
     The arguments remain unmodified. However, it must be possible to copy them
@@ -508,10 +509,10 @@ def merge_dicts(base: dict, other: dict, list_semantics='merge'):
     By default, different from the original implementation, a merge will be applied to
     lists. This results in deduplication retaining element order. The elements from `other` are
     appended to those from `base`.
-
     '''
+
     not_none(base)
-    not_none(other)
+    not_empty(other)
 
     from deepmerge import Merger
 
@@ -530,8 +531,12 @@ def merge_dicts(base: dict, other: dict, list_semantics='merge'):
         raise NotImplementedError
 
     from copy import deepcopy
-    # copy dicts, so they remain unmodified
-    return merger.merge(deepcopy(base), deepcopy(other))
+
+    return functools.reduce(
+        lambda b, o: merger.merge(b, deepcopy(o)),
+        [base, *other],
+        {},
+    )
 
 
 class FluentIterable(object):
--- a/concourse/client/model.py
+++ b/concourse/client/model.py
@@ -117,7 +117,7 @@ class Job:
     def is_triggered_by_resource(self, resource_name: str):
         get_steps = self.plan().get_steps()
         for get_step in get_steps:
-            if get_step['get'] == resource_name and get_step['trigger']:
+            if get_step['get'] == resource_name and get_step.get('trigger', False):
                 return True
         return False
 
--- a/concourse/enumerator.py
+++ b/concourse/enumerator.py
@@ -420,6 +420,9 @@ class DefinitionDescriptor:
     def template_name(self):
         return self.pipeline_definition.get('template', 'default')
 
+    def effective_pipeline_name(self):
+        return '-'.join((self.pipeline_name, self.main_repo['branch']))
+
     def concourse_target(self):
         return (self.concourse_target_cfg, self.concourse_target_team)
 
--- a/concourse/model/job.py
+++ b/concourse/model/job.py
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import enum
 import toposort
 
 from concourse.model.base import (
@@ -21,6 +22,12 @@ from concourse.model.base import (
 )
 from ci.util import not_none
 from concourse.model.resources import RepositoryConfig, ResourceIdentifier
+
+
+class AbortObsoleteJobs(enum.Enum):
+    ALWAYS = 'always'
+    ON_FORCE_PUSH_ONLY = 'on_force_push_only'
+    NEVER = 'never'
 
 
 class JobVariant(ModelBase):
@@ -36,10 +43,11 @@ class JobVariant(ModelBase):
 
     def _known_attributes(self):
         return {
-            'steps',
-            'traits',
+            'abort_obsolete_jobs',
             'repo',
             'repos',
+            'steps',
+            'traits',
         }
 
     def _children(self):
--- a/concourse/replicator.py
+++ b/concourse/replicator.py
@@ -156,20 +156,12 @@ class Renderer(object):
         if bg := effective_definition.get('background_image'):
             pipeline_metadata['background_image'] = bg
 
-        # determine pipeline name (if there is main-repo, append the configured branch name)
         for variant in pipeline_metadata.get('definition').variants():
-            # hack: take the first "main_repository" we find
             if not variant.has_main_repository():
-                continue
-            main_repo = variant.main_repository()
-            pipeline_metadata['pipeline_name'] = '-'.join(
-                [pipeline_definition.name, main_repo.branch()]
-            )
-            break
-        else:
-            # fallback in case no main_repository was found
-            pipeline_metadata['pipeline_name'] = pipeline_definition.name
-            main_repo = None
+                raise RuntimeError(
+                    f"No main repository for pipeline definition {pipeline_definition.name}."
+                )
+            pipeline_metadata['pipeline_name'] = definition_descriptor.effective_pipeline_name()
 
         t = mako.template.Template(template_contents, lookup=self.lookup)
 
--- a/whd/dispatcher.py
+++ b/whd/dispatcher.py
@@ -19,6 +19,7 @@ import logging
 import random
 import threading
 import time
+import typing
 import traceback
 
 import ccc.elasticsearch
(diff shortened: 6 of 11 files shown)


concourse/client/__init__.py
```diff
@@ -76,3 +76,11 @@ def from_cfg(concourse_cfg: ConcourseConfig, team_name: str, verify_ssl=False):
         passwd=password,
     )
     return concourse_api
+
+
+@functools.lru_cache()
+@ensure_annotations
+def from_cfg_name(concourse_cfg_name: str, team_name: str, verify_ssl: bool=False):
```
ccwienk, line 83: I think this should go into `ccc.concourse`

ccwienk: also, I would prefer not to expose `verify_ssl` as a parameter. it is already configurable via cc-config (and even that could be considered a bad idea). Let us not make it too easy to disable tls-validation

concourse/client/model.py
```diff
@@ -117,7 +117,7 @@ def plan(self):
     def is_triggered_by_resource(self, resource_name: str):
         get_steps = self.plan().get_steps()
         for get_step in get_steps:
-            if get_step['get'] == resource_name and get_step['trigger']:
+            if get_step['get'] == resource_name and 'trigger' in get_step and get_step['trigger']:
```
ccwienk, line 120: nit: rather use `get_step.get('trigger', False')`

concourse/enumerator.py
```diff
@@ -420,6 +420,9 @@ def __init__(
     def template_name(self):
         return self.pipeline_definition.get('template', 'default')
 
+    def _pipeline_name(self):
+        return '-'.join((self.pipeline_name, self.main_repo['branch']))
```
ccwienk, line 424: - why make this method "private"?
- did you replace all usages of the `name` attr with this method (if not, please do)?

concourse/model/job.py
```diff
@@ -23,6 +24,11 @@
 from concourse.model.resources import RepositoryConfig, ResourceIdentifier
 
 
+class AbortObsoleteJobs(enum.Enum):
```
ccwienk, line 27: in fact, I also would prefer to use `obsolete` (in the PR-description's example, you named the attr `outdated` - let us please make this consistent)

concourse/model/job.py
```diff
@@ -23,6 +24,11 @@
...
+    ALL = 'all'
```
ccwienk, line 28: how about `always` and `on_force_push_only`?

ccwienk: also: should there not rather also be a value `never`?

concourse/replicator.py
```diff
@@ -156,20 +156,12 @@ def _render(self, definition_descriptor):
         if bg := effective_definition.get('background_image'):
             pipeline_metadata['background_image'] = bg
 
-        # determine pipeline name (if there is main-repo, append the configured branch name)
         for variant in pipeline_metadata.get('definition').variants():
-            # hack: take the first "main_repository" we find
             if not variant.has_main_repository():
-                continue
-            main_repo = variant.main_repository()
-            pipeline_metadata['pipeline_name'] = '-'.join(
```
ccwienk, line 165: I know - this is not part of the semantical focus of this PR. but: how about implementing a `dataclass` for `pipeline_metadata`? should (hopefully..) not be too much effort, but would increase maintainability a bit

whd/dispatcher.py
```diff
@@ -139,6 +143,77 @@ def _pipeline_definition_changed(self, push_event):
             return True
         return False
 
+    def determine_affected_pipelines(self, push_event):
+        '''yield the triple (pipeline name, target team, effective definition) for each concourse
```
ccwienk, line 147: I would prefer a dataclass tbh. It is very little boilerplate, but helps maintainability a lot

whd/dispatcher.py
```diff
@@ -139,6 +143,77 @@ def _pipeline_definition_changed(self, push_event):
...
+                    yield from determine_jobs_to_be_triggered(resource)
+
+            for job in multiresource_determine_jobs_to_be_triggered(resources):
+                if (
+                    not pipeline_definition.get(job.name)
+                    or not 'abort_outdated_jobs' in pipeline_definition[job.name]
+                ):
+                    continue
+
+                abort_cfg = AbortObsoleteJobs(pipeline_definition[job.name]['abort_outdated_jobs'])
```
ccwienk, line 198: also here, I would prefer to introduce a dataclass

concourse/enumerator.py
AndreasBurger: Because its a sensible default and to prevent clashing with the `pipeline_name` attr. Thinking about this some more, the two names should be separated more clearly. One is given by the user (e.g. `cc-utils`) and referred to more often than the other which is rendered by us (`cc-utils-master`). The latter seems to be needed mostly in two places during replication.
I'll need to check further, but for all cases I could find we use the name as retrieved from Concourse.

ccwienk: what do you mean by "clashing"? in python, attribute names must be unique. i.e. you cannot "overload" them (have a method named the same as a "plain" attr); I assume you knew that, and this is what you meant by "avoidance"

so if I remember properly, the plain attr `pipeline_name` is the value from the `pipeline_definition` file, whereas the (new) method `_pipeline_name()` calculates the derived pipeline-names actually to be used for the rendered pipelines.

in that case, I would rather name the method to express this semantics. such as `unique_pipeline_name`, or `effective_pipeline_name`, and add a - brief - comment as to why we add e.g. the branch name.

for consistency reasons, we might also consider to do the name-calculation before the object is created; not sure whether the benefit is worth the efforts, tnough

concourse/model/job.py
```diff
@@ -36,10 +43,11 @@ def _attribute_specs(cls):
 
     def _known_attributes(self):
         return {
-            'steps',
-            'traits',
+            'abort_obsolete_jobs',
             'repo',
             'repos',
+            'steps',
+            'traits',
```
ccwienk, line 50: very good :-)

concourse/replicator.py
```diff
@@ -156,20 +156,12 @@ def _render(self, definition_descriptor):
...
-                [pipeline_definition.name, main_repo.branch()]
-            )
-            break
-        else:
-            # fallback in case no main_repository was found
-            pipeline_metadata['pipeline_name'] = pipeline_definition.name
-            main_repo = None
+                raise RuntimeError(
+                    f"No main repository for pipeline definition {pipeline_definition.name}."
```
ccwienk, line 162: +1

whd/dispatcher.py
```diff
@@ -139,6 +145,89 @@ def _pipeline_definition_changed(self, push_event):
             return True
         return False
 
+    def determine_affected_pipelines(self, push_event) -> Pipeline:
```
ccwienk, line 148: nit: typehint should by typing.Generator[Pipeline], should it not?
nit2: I would not necessarily include the yielding into the docstr (this is already documented in the return-typehint..) - not that important, though

whd/dispatcher.py
```diff
@@ -139,6 +145,89 @@ def _pipeline_definition_changed(self, push_event):
...
+        repo_enumerator = concourse.enumerator.GithubRepositoryDefinitionEnumerator(
+            repository_url=push_event.repository().repository_url(),
+            cfg_set=self.cfg_set,
+        )
+
+        for descriptor in repo_enumerator.enumerate_definition_descriptors():
+            # need to merge and consider the effective definition
+            effective_definition = descriptor.pipeline_definition
+            for override in descriptor.override_definitions:
+                effective_definition = ci.util.merge_dicts(effective_definition, override)
```
ccwienk, line 160: mhhh..... I also needed this sometimes by now. how about changing merge_dicts in such a way that it accepts an arbitrary amount of definitions:

```
def merge_dicts(first_definition, other_definitions*, / kwargs): ...
```

note that I was too lazy to lookup the exactly correct syntax :-)

whd/dispatcher.py
```diff
@@ -139,6 +145,89 @@ def _pipeline_definition_changed(self, push_event):
...
+
+            yield Pipeline(
+                pipeline_name=descriptor._pipeline_name(),
```
ccwienk, line 163: I did not check the entire PR, yet. do you also use this new method in the replication-codepath? If not, please do

whd/dispatcher.py
```diff
@@ -139,6 +145,89 @@ def _pipeline_definition_changed(self, push_event):
...
+                )
+
+            pipeline_config = client.pipeline_cfg(pipeline.pipeline_name)
+
+            resources = [
+                r for r in pipeline_config.resources
+                if ResourceType(r.type) in (ResourceType.GIT, ResourceType.PULL_REQUEST)
+            ]
+
+            def multiresource_determine_jobs_to_be_triggered(resources):
```
ccwienk, line 190: what does "multiresource" mean?

whd/model.py
```diff
@@ -73,6 +77,12 @@ def commit_message(self):
             return head_commit.get('message')
         return None
 
+    def is_forced_push(self):
+        return self.raw['forced']
+
+    def before_ref(self):
+        return self.raw['before']
```
ccwienk, line 84: what value does `before` have / what does it mean?

whd/model.py
```diff
@@ -109,3 +119,23 @@ def sender(self):
         the user who performed the event
         '''
         return self.raw['sender']
+
+
+@dataclasses.dataclass
+class Pipeline:
+    pipeline_name: str
+    target_team: str
+    effective_definition: str
```
ccwienk, line 128: is this really a str? Oo

ccwienk (changes requested):

whd/model.py
ccwienk: previous_ref, please

ci/util.py
```diff
@@ -496,7 +497,7 @@ def which(cmd_name: str) -> str:
     return cmd_path
 
 
-def merge_dicts(base: dict, other: dict, list_semantics='merge'):
+def merge_dicts(base: dict, *other: dict, list_semantics='merge'):
```
ccwienk, line 500: I suggest to make `list_semtics` a "named-only" parameter

ccwienk: https://www.python.org/dev/peps/pep-3102/

PR #512, opened 2020-12-21
Support Multiple Merge-Policies and Add Filter Semantic

AndreasBurger:
Adds support for multiple merge-policies to the update-pull-request trait. For each merge policy, a list regex must be given that specifies the components to which this policy should apply.

```yaml
…
  jobs:
    upgrade-component-deps:
      traits:
        component_descriptor: ~
        update_component_deps:
          set_dependency_version_script: '.ci/foo'
          merge_policy:
            manual:
              - github.com/.*
            auto:
              - github.tools.sap/.*
```
If a policy applies to multiple components, an exception will be raised.

### DIFF
--- a/concourse/model/traits/update_component_deps.py
+++ b/concourse/model/traits/update_component_deps.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 
 import enum
+import typing
 
 from concourse.model.step import (
     PipelineStep,
@@ -21,13 +22,15 @@ from concourse.model.step import (
 )
 from concourse.model.base import (
     AttributeSpec,
+    ModelBase,
+    ScriptType,
     Trait,
     TraitTransformer,
-    ScriptType,
 )
 from concourse.model.job import (
     JobVariant,
 )
+from model.base import ModelValidationError
 
 import concourse.model.traits.component_descriptor
 
@@ -40,6 +43,43 @@ class MergePolicy(enum.Enum):
 class UpstreamUpdatePolicy(enum.Enum):
     STRICTLY_FOLLOW = 'strictly_follow'
     ACCEPT_HOTFIXES = 'accept_hotfixes'
+
+
+MERGE_POLICY_CONFIG_ATTRIBUTES = (
+    AttributeSpec.optional(
+        name='component_names',
+        default=[],
+        type=typing.List[str],
+        doc=(
+            'a sequence of regular expressions. This merge policy will be applied to matching '
+            'component names. Matches all component names by default'
+        )
+    ),
+    AttributeSpec.optional(
+        name='merge_mode',
+        default='manual',
+        type=MergePolicy,
+        doc='whether or not created PRs should be automatically merged',
+    ),
+)
+
+
+class MergePolicyConfig(ModelBase):
+    @classmethod
+    def _attribute_specs(cls):
+        return MERGE_POLICY_CONFIG_ATTRIBUTES
+
+    def component_names(self):
+        # handle default here
+        # TODO: refactor default arg handling. User-given values should *replace* defaults, not
+        #       add to them
+        if not self.raw['component_names']:
+            return ['.*']
+
+        return self.raw['component_names']
+
+    def merge_mode(self):
+        return MergePolicy(self.raw['merge_mode'])
 
 
 ATTRIBUTES = (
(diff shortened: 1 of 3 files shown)


ccwienk:
How about adding a list of policies, like so:

```
merge_policies:
 - component_names: ['component-name-.*']
   merge_mode: 'automatic' # | manual
```

concourse/model/traits/update_component_deps.py
```diff
@@ -60,8 +61,11 @@ class UpstreamUpdatePolicy(enum.Enum):
     ),
     AttributeSpec.optional(
         name='merge_policy',
-        default=MergePolicy.MANUAL,
+        default={
+            'manual': ['*',],
```
ccwienk, line 65: - stylistic nit: there should not be a trailing comma in oneline-lists
- if this is a regex (which I hope..), this should rather read `.*`, should it not?
- I would prefer to keep the old attr (`merge_policy`), deprecate it, and introduce a new one (see comment on the PR - name it either `merge_policies`, or just `policies`; it both are present (configured), this is an error
- as a "trick", we might make the component-filter optional (absent filter meaning "match-all")

ccwienk: also: if possible, always use immutable container types (tuple instead of list).

concourse/model/traits/update_component_deps.py
```diff
@@ -91,8 +95,18 @@ def upstream_component_name(self):
...
-    def merge_policy(self) -> MergePolicy:
-        return MergePolicy(self.raw['merge_policy'])
+    def _has_legacy_merge_policy(self):
+        return isinstance(self.raw['merge_policy'], str)
+
+    def merge_policy(self):
+        if self._has_legacy_merge_policy():
+            return {
+                self.raw['merge_policy']: {
+                    'include_component_names': '*',
```
ccwienk, line 105: are you really sure you do not want to use regex?

concourse/steps/update_component_deps.mako
```diff
@@ -102,6 +111,16 @@ for from_ref, to_version in determine_upgrade_prs(
     upgrade_pull_requests=upgrade_pull_requests,
     ctx_repo_base_url='${ctx_repo_base_url}',
 ):
+    configured_merge_policy = [
+        policy for policy, filter_func in merge_policies.items() if filter_func(from_ref)
+    ]
+    if len(configured_merge_policy) > 1:
+        raise RuntimeError(f'More than one merge policy applies to {from_ref.name}')
```
ccwienk, line 118: I think we should be more forgiving here. how about only considering it an error if there is a conflict (two different policies for the same component..)?

concourse/steps/update_component_deps.mako
```diff
@@ -102,6 +111,16 @@ for from_ref, to_version in determine_upgrade_prs(
...
+    elif len(configured_merge_policy) == 0:
+        policy = MergePolicy('manual')
```
ccwienk, line 120: since MergePolicy is an enum (I think - did not look it up), rather use MergePolicy.MANUAL

doc/lib/attributes.py
```diff
@@ -71,9 +71,9 @@ def _type_name(self, type_):
         if isinstance(type_, typing._GenericAlias):
             if type_.__origin__ is dict:
                 key_type, val_type = type_.__args__
-                return type_._name + f'[{self._type_name(key_type)}, {self._type_name(val_type)}]'
+                return 'Dict' + f'[{self._type_name(key_type)}, {self._type_name(val_type)}]'
             elif type_.__origin__ in (list, set):
-                return type_._name + f'[{self._type_name(type_.__args__[0])}]'
+                return 'List' + f'[{self._type_name(type_.__args__[0])}]'
```
ccwienk, line 76: nit: why not include `Dict` and `List` into the formatstr-literal? also: how about `List of`? But I am also ok with your less verbose variant

ccwienk (changes requested):

concourse/model/traits/update_component_deps.py
```diff
@@ -91,8 +135,22 @@ def upstream_component_name(self):
     def upstream_update_policy(self):
         return UpstreamUpdatePolicy(self.raw.get('upstream_update_policy'))
 
-    def merge_policy(self) -> MergePolicy:
-        return MergePolicy(self.raw['merge_policy'])
+    def merge_policy(self):
+        if merge_policy := self.raw.get('merge_policy', None):
+            # translate to new model class to keep code simple
+            return [
```
ccwienk, line 141: mv to merge_policies

PR #513, opened 2020-12-29
Add Component Spec to dependencies

AndreasBurger:
adds gardener component spec to our component-dependencies.

### DIFF
--- a/.ci/component_descriptor
+++ b/.ci/component_descriptor
@@ -1,0 +1,84 @@
+#!/usr/bin/env python3
+
+import dataclasses
+import os
+import yaml
+
+import pkg_resources
+
+import ci.util
+import gci.componentmodel
+
+component_descriptor_base_path = os.path.abspath(ci.util.check_env('BASE_DEFINITION_PATH'))
+component_descriptor_path = os.path.abspath(ci.util.check_env('COMPONENT_DESCRIPTOR_PATH'))
+
+
+COMPONENT_SPEC_PYPI_PROJECT_NAME = 'gardener-component-model'
+COMPONENT_SPEC_COMPONENT_NAME = 'github.com/gardener/component-spec'
+
+
+def parse_component_descriptor():
+    component_descriptor = gci.componentmodel.ComponentDescriptor.from_dict(
+        ci.util.parse_yaml_file(component_descriptor_base_path)
+    )
+    return component_descriptor
+
+
+def component_spec_version():
+    with open(
+        os.path.join(
+            os.path.abspath(ci.util.check_env('MAIN_REPO_DIR')),
+            'requirements.txt',
+        ),
+        'r',
+    ) as requirements_file:
+        parsed_requirements = pkg_resources.parse_requirements(requirements_file)
+        if not (component_spec_requirement := next(
+            (r for r in parsed_requirements if r.project_name == COMPONENT_SPEC_PYPI_PROJECT_NAME),
+            None,
+        )):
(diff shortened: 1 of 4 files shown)


.ci/component_descriptor
```diff
@@ -0,0 +1,89 @@
...
+        )):
+            raise RuntimeError('Unable to find component-spec in requirements.txt')
+        if len(component_spec_requirement.specs) > 1:
+            raise RuntimeError('No exact version specified for component-spec dependency')
+        component_spec_specs = component_spec_requirement.specs[0]
+
+        return component_spec_specs[1]
+
+
+def add_dependence_on_component(
```
ccwienk, line 58: dependencY

.ci/component_descriptor
```diff
@@ -0,0 +1,89 @@
...
+            version=component_version,
+            labels=[],
+        )
+    )
+
+
+component_descriptor = parse_component_descriptor()
+component = component_descriptor.component
+
+add_dependence_on_component(
```
ccwienk, line 78: I think this helper function would not rm a lot of redundancy, even if it were used more than once. Like this, is merely adds boilerplate imo

.ci/component_descriptor
```diff
@@ -0,0 +1,89 @@
...
+    component=component,
+    component_name='github.com/gardener/component-spec',
+    component_version=f'v{component_spec_version()}',
```
ccwienk, line 81: I think we should simply add it into the requirements-file with a v-prefix as well; according to setuptools, it is acceptable / semantically equivalent to do so (although not considered a good practice)

.ci/set_dependency_version
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env python3
+
+import fileinput
+import os
+
+import util
+
+COMPONENT_SPEC_PROJECT_NAME = 'gardener-component-model'
```
ccwienk, line 8: deduplidate?

.ci/set_dependency_version
```diff
@@ -0,0 +1,31 @@
...
+        else:
+            print(line)
+
+
+util.check_env('REPO_DIR')
+
+component_name = util.check_env('DEPENDENCY_NAME')
+component_version = util.check_env('DEPENDENCY_VERSION')
+
+if component_name == 'github.com/gardener/component-spec':
```
ccwienk, line 28: deduplicate

.ci/set_dependency_version
```diff
@@ -0,0 +1,31 @@
...
+    update_project_version(COMPONENT_SPEC_PROJECT_NAME, component_version)
+else:
+    raise NotImplementedError(f"Don't know how to upgrade '{component_name}'")
```
ccwienk, line 31: nit: please always append `=` in format-str-expressions, unless there is a reason against doing so

.ci/set_dependency_version
```diff
@@ -0,0 +1,31 @@
...
+            'requirements.txt',
+        )
+    for line in fileinput.input(path_to_requirements_txt, inplace=True):
+        if line.startswith(project_name):
+            print(f'{project_name}=={project_version}')
+        else:
+            print(line)
+
+
+util.check_env('REPO_DIR')
```
ccwienk, line 23: why use both MAIN_REPO_DIR and REPO_DIR?

ccwienk (changes requested):

.ci/component_descriptor
AndreasBurger: that was intentional. Will change it to `add_component_dependency`

PR #516, opened 2021-01-07
Fix issues in default component descriptor

AndreasBurger:
Changes default component descriptor so that a github-config is always available for additional repositories. The config taken will be the default config for the location of the pipeline if no config is explicitly configured, mimicking the behaviour of the replication.

Also, for additional sources (created by having additional repositories defined) it will now be attempted to determine the _actual_ commit hash before falling back to the main-repositories' commit-hash.

### DIFF
--- a/concourse/steps/component_descriptor.mako
+++ b/concourse/steps/component_descriptor.mako
@@ -9,8 +9,9 @@ descriptor_trait = job_variant.trait('component_descriptor')
 main_repo = job_variant.main_repository()
 main_repo_labels = main_repo.source_labels()
 main_repo_path_env_var = main_repo.logical_name().replace('-', '_').upper() + '_PATH'
-other_repos = [r for r in job_variant.repositories()
- if not r.is_main_repo()
+other_repos = [
+  r for r in job_variant.repositories()
+  if not r.is_main_repo()
 ]
 ctx_repository_base_url = descriptor_trait.ctx_repository_base_url()
 
@@ -34,7 +35,7 @@ cm = gci.componentmodel
 
 from product.model import ComponentDescriptor, Component, ContainerImage, Relation
 from product.util import ComponentDescriptorResolver
-from ci.util import info, fail, parse_yaml_file, ctx
+from ci.util import info, fail, parse_yaml_file, ctx, warning
 import product.v2
 
 ${step_lib('component_descriptor')}
@@ -54,14 +55,7 @@ component_name_v2 = component_name.lower() # OCI demands lowercase
 ctx_repository_base_url = '${descriptor_trait.ctx_repository_base_url()}'
 
 main_repo_path = os.path.abspath('${main_repo.resource_name()}')
-main_git_repo = git.Repo(main_repo_path)
-if not main_git_repo.head.is_valid():
-  commit_hash = None
-else:
-  try:
-    commit_hash = main_git_repo.head.commit.hexsha
-  except:
-    commit_hash = None
+commit_hash = head_commit_hexsha(main_repo_path)
 
 # create base descriptor filled with default values
 base_descriptor_v2 = base_component_descriptor_v2(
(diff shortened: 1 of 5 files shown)


concourse/steps/component_descriptor.mako
```diff
@@ -1,8 +1,9 @@
 <%def
-  name="component_descriptor_step(job_step, job_variant, output_image_descriptors, indent)",
+  name="component_descriptor_step(job_step, job_variant, output_image_descriptors, default_github_config, indent)",
```
ccwienk, line 2: are you sure it is necessary to pass-in the default_gh_cfg? rather calculate if from hostname (ccc.github), if even required in the first place.

concourse/steps/component_descriptor.mako
```diff
@@ -1,8 +1,9 @@
...
   filter="indent_func(indent),trim"
 >
 <%
+import re
```
ccwienk, line 6: nit: leave one empty line after imports from stdlib vs "local" ones

concourse/steps/component_descriptor.mako
```diff
@@ -76,15 +77,28 @@ component_v2 = base_descriptor_v2.component
 
 ## XXX unify w/ injection-method used for main-repository
 % for repository in other_repos:
+<%
+  repo_variable_name = f"repo_{other_repos.index(repository)}"
```
ccwienk, line 81: this is a very hard-to-read codestyle. please do not partially construct variable names like that

concourse/steps/component_descriptor.mako
```diff
@@ -76,15 +77,28 @@ component_v2 = base_descriptor_v2.component
...
+    ${repo_variable_name}_commit_hash = commit_hash
+
 component_v2.sources.append(
   cm.ComponentSource(
     name='${repository.name()}',
     type=cm.SourceType.GIT,
     access=cm.GithubAccess(
       type=cm.AccessType.GITHUB,
-      repoUrl='${repository.repo_hostname()}/${repository.repo_path()}',
+      repoUrl='${repository.repo_hostname() or default_github_config.hostname()}/${repository.repo_path()}',
```
ccwienk, line 99: this cannot be right

concourse/steps/component_descriptor.mako
```diff
@@ -76,15 +77,28 @@ component_v2 = base_descriptor_v2.component
...
       ref='${repository.branch()}',
-      commit=commit_hash,
+      commit=${repo_variable_name}_commit_hash,
```
ccwienk, line 101: as indicated before: please do not do this. this is unnecessarily hard to understand/read

concourse/templates/default.mako
```diff
@@ -405,7 +405,7 @@ else:
   else:
     image_descriptors_for_variant = {}
 %>
-        ${component_descriptor_step(job_step=job_step, job_variant=job_variant, output_image_descriptors=image_descriptors_for_variant, indent=8)}
+        ${component_descriptor_step(job_step=job_step, job_variant=job_variant, output_image_descriptors=image_descriptors_for_variant, default_github_config=github, indent=8)}
```
ccwienk, line 408: please rm this parameter again

test/concourse/steps/component_descriptor_test.py
```diff
@@ -30,6 +30,10 @@ def setUp(self):
         self.component_descriptor_transformer = self.component_descriptor_trait.transformer()
         self.render_step = step_def('component_descriptor')
 
+        github_cfg_mock = unittest.mock.MagicMock()
+        github_cfg_mock.hostname.return_value = 'foo.com'
+        self.default_github_config = github_cfg_mock
```
ccwienk, line 35: should also not be required

test/concourse/steps/component_descriptor_test.py
```diff
@@ -60,6 +64,7 @@ def test_render_and_compile(self):
             job_step=self.job_variant.step('component_descriptor'),
             job_variant=self.job_variant,
             output_image_descriptors={},
+            default_github_config=self.default_github_config,
```
ccwienk, line 67: ditto

ccwienk (changes requested):
I really do have some stylistic concerns with this change.

also, to make reviewing easier, it would help (me) a lot to not merely state that this pr will "fix issues", but give a _highlevel_ and _short_ summary of what it fixes (in this case, I understand it fixes cases where `repoUrl` and `commit` are not properly injected into base-descriptor?)

ccwienk:
+1

PR #520, opened 2021-01-11
Adapt cd v2 comp diff

jkrayl:
adapts v2 component descriptor in component diff alg


product/util.py
```diff
@@ -154,15 +157,52 @@ def merge_products(left_product, right_product):
...
+    cpairs_version_changed: list = dataclasses.field(default_factory=list)
+    # only set when new component is added/removed
+    names_only_left: set = dataclasses.field(default_factory=set)
+    names_only_right: set = dataclasses.field(default_factory=set)
+    # only set on update
+    names_version_changed: set = dataclasses.field(default_factory=set)
+
+
+# for v2 =  pass cd v2 and resolve components
+# make use of caching
```
ccwienk, line 173: are those comments intended to remain here?

product/util.py
```diff
@@ -154,15 +157,52 @@ def merge_products(left_product, right_product):
...
+def diff_products(
+    left_product: cm.ComponentDescriptor,
+    right_product: cm.ComponentDescriptor,
+    ignore_component_names=(),
+    cache_dir=None,
+) -> ComponentDiff:
+    if type(left_product) is not cm.ComponentDescriptor:
+        raise NotImplementedError(
+            f'left product unsupported type {type(left_product)=}.'
```
ccwienk, line 182: ok with me to be this verbose
in general, as this is merely a programming error (not helpful to end-users), I would tend to save some typing.

also, I think a `TypeError` would be more adequate here.

NotImplementedError fits better to indicate situations where the passed value is in general valid, but not (yet) handled. However, I think we do not plan to support different types anytime soon, do we? :-)

product/util.py
```diff
@@ -154,15 +157,52 @@ def merge_products(left_product, right_product):
...
+    names_only_left: set = dataclasses.field(default_factory=set)
+    names_only_right: set = dataclasses.field(default_factory=set)
+    # only set on update
+    names_version_changed: set = dataclasses.field(default_factory=set)
+
+
+# for v2 =  pass cd v2 and resolve components
+# make use of caching
+def diff_products(
+    left_product: cm.ComponentDescriptor,
```
ccwienk, line 175: I know, the parameter names are originally mine (back then, the component-descriptors were called (by me) "product" for some reason)

how about (since this is an incompatible change anyhow) renaming this to `left_component` / `right_component`?

also, we _might_ be a bit nice here, and e.g. accept either a `Component` or a `ComponentDescriptor`, wdyt?

product/util.py
```diff
@@ -154,15 +157,52 @@ def merge_products(left_product, right_product):
...
+    right_product: cm.ComponentDescriptor,
+    ignore_component_names=(),
+    cache_dir=None,
+) -> ComponentDiff:
+    if type(left_product) is not cm.ComponentDescriptor:
```
ccwienk, line 180: I think it would better to use `isinstance` - this way, we would also accept (hypothetical) subtypes

product/util.py
```diff
@@ -154,15 +157,52 @@ def merge_products(left_product, right_product):
...
+    left_components: typing.Generator[cm.Component] = product.v2.components(
+        component_descriptor_v2=left_product,
+        cache_dir=cache_dir,
+    )
+    right_components: typing.Generator[cm.Component] = product.v2.components(
+        component_descriptor_v2=right_product,
+        cache_dir=cache_dir,
+    )
+    left_components = tuple(
+        c for c in left_components if c.name not in ignore_component_names
```
ccwienk, line 201: I did not know there is no need for a second pair of parenthesis - thx; much more elegant this way!

product/util.py
```diff
@@ -154,15 +157,52 @@ def merge_products(left_product, right_product):
...
+    )
+    right_components = tuple(
+        c for c in right_components if c.name not in ignore_component_names
+    )
 
     return diff_components(
```
ccwienk, line 207: mhh..... so all `diff_products` does on top of `diff_components` is resolving the transitive closure of components? hmmmm :-/

product/util.py
```diff
@@ -171,164 +211,197 @@ def diff_products(left_product, right_product, ignore_component_names=()):
...
-    names_only_left: set = dataclasses.field(default_factory=set)
-    names_only_right: set = dataclasses.field(default_factory=set)
-    names_version_changed: set = dataclasses.field(default_factory=set)
-
+def diff_components(
+    left_components: typing.Tuple[cm.Component],
+    right_components: typing.Tuple[cm.Component],
+    ignore_component_names=(),
+) -> ComponentDiff:
+    left_identities = {
```
ccwienk, line 219: it might make the code easier to read if those were called component_identities (at the price of being veeeryyy long names..)

product/util.py
```diff
@@ -171,164 +211,197 @@ def diff_products(left_product, right_product, ignore_component_names=()):
...
+        components: typing.List[cm.Component],
+    ):
         for c in components:
-            if c.name() == changed_component.name():
+            if c.name == changed_component.name:
                 return (changed_component, c)
         return (changed_component, None) # no pair component found
 
-    components_with_changed_versions = FluentIterable(items=components_only_left) \
+    components_with_changed_versions = FluentIterable(items=left_components) \
```
ccwienk, line 248: nit: maybe we should replace this `FluentIterable` (which I stole from guava^^) with a plain list-comprehension, wdyt?

product/util.py
```diff
@@ -171,164 +211,197 @@ def diff_products(left_product, right_product, ignore_component_names=()):
...
-    right_component
-):
-    left_images = set(_effective_images(left_component_descriptor, left_component))
-    right_images = set(_effective_images(right_component_descriptor, right_component))
+class ResourceDiff:
+    left_component: cm.Component
+    right_component: cm.Component
+    resource_refs_only_left: typing.List[cm.Resource] = dataclasses.field(default_factory=list)
+    resource_refs_only_right: typing.List[cm.Resource] = dataclasses.field(default_factory=list)
+    resourcepairs_version_changed: typing.List[typing.Tuple[cm.Resource, cm.Resource]] = dataclasses.field(default_factory=list) # noqa:E501
```
ccwienk, line 278: I think the default value (list-factory..) could be re-used (which would save some line-leng..)

product/util.py
```diff
@@ -171,164 +211,197 @@ def diff_products(left_product, right_product, ignore_component_names=()):
...
+
+
+def _add_if_not_duplicate(list, res):
```
ccwienk, line 281: if only there were some datastructure that would feature only allowing unique elements ^^

product/util.py
```diff
@@ -171,164 +211,197 @@ def diff_products(left_product, right_product, ignore_component_names=()):
...
-
-    for name, img in left_names_to_imgs.items():
-        if not name in right_names_to_imgs:
-            img_diff.irefs_only_left.add(img)
+    if left_resource_identities.keys() == right_resource_identities_to_resource.keys():
+        return resource_diff
+
+    left_names_to_resource = {r.name: r for r in left_component.resources}
+    right_names_to_resource = {r.name: r for r in right_component.resources}
+    # get left exclisive resources
```
ccwienk, line 318: nit: typo

product/util.py
```diff
@@ -171,164 +211,197 @@ def diff_products(left_product, right_product, ignore_component_names=()):
...
+    for resource in left_resource_identities.values():
+        if not resource.name in right_names_to_resource:
+            _add_if_not_duplicate(resource_diff.resource_refs_only_left, resource)
+
+    # get right exclusive images
```
ccwienk, line 323: also nitty: why not be consistent w/ comments and symbol names (-> e.g. use left/right-only in both cases?)

product/util.py
```diff
@@ -171,164 +211,197 @@ def diff_products(left_product, right_product, ignore_component_names=()):
...
+    for resource in right_resource_identities_to_resource.values():
+        if not resource.name in left_names_to_resource:
+            _add_if_not_duplicate(resource_diff.resource_refs_only_right, resource)
+
+    def enumerate_group_pairs(
```
ccwienk, line 328: by now, i really feel bad about having implemented this with so few comments. e.g. a brief comment explaining why we build resource-group-pairs might help, wdyt (not help you, any more, since you already figured out the algorithm, obviously)?

product/util.py
```diff
@@ -171,164 +211,197 @@ def diff_products(left_product, right_product, ignore_component_names=()):
...
 
-    for name, img in right_names_to_imgs.items():
-        if not name in left_names_to_imgs:
-            img_diff.irefs_only_right.add(img)
+    for left_resource_group, right_resource_group in enumerate_group_pairs(
+        left_resources=left_component.resources,
+        right_resources=right_component.resources,
+    ):
+        if len(left_resource_group) == 1 and len(right_resource_group) == 1:
+            # if versions are equal resource will be ignored, resource is unchanged
```
ccwienk, line 348: actually, this really only holds true if the identities be equal. however, let us leave it like this for now, to not make things unnecessarily complicated..

ccwienk (commented):
some nits + let's discuss whether we should maybe leverage this incompatible change to also cleanup the signature some more.

also, there is a piece of code that checks for whether or not there has been a component-diff (need to check whether this was in cc-utils, or in lss/lssd)
-> those pieces of code we also need to change

product/util.py
jkrayl: no we don't :) I changed the Error type and shortened the message

product/util.py
jkrayl: done

product/util.py
jkrayl: you welcome

product/util.py
jkrayl: i changed this for the sake of readability

ccwienk (approved):

PR #521, opened 2021-01-11
Cd add git main repo

jkrayl:
labels git repo as main and auxiliary


ccwienk (commented):
content-wise lgtm

however, how about squashing the commits introducing and removing again the conflict-marker-lines?

PR #545, opened 2021-01-13
Adjust whitesource client to websockets

zkdev:
Adjust the WhiteSource client to latest changes to make use of websockets instead of plain HTTP.

Fixes #

**Release note**:

```improvement operator

```


cli/gardener_ci/whitesource_cli.py
```diff
@@ -8,15 +7,17 @@ def upload_and_scan_component(
     whitesource_cfg_name: str,
     component_descriptor_path: str,
     requester_mail: str,
-    component_name: str,
-    extra_whitesource_config: str = {},
-    notification_recipients: [str] = [],
     cve_threshold: float = 5.0,
+    notification_recipients: str = '[]',
```
ccwienk, line 11: in fact, cli.py does support lists, like so: `[str]` (admittedly, it would be more aligned w/ the `typing` module, if the implementation would instead rather accept `typing.List[str]`)

cli/gardener_ci/whitesource_cli.py
```diff
@@ -8,15 +7,17 @@ def upload_and_scan_component(
...
+    chunk_size: int = 1024,
+    ping_interval: int = 1000,
+    ping_timeout: int = 1000,
 ):
-    try:
-        extra_whitesource_config = json.loads(extra_whitesource_config)
-    except JSONDecodeError as e:
-        raise e
+
+    # parse sequence strings to actual sequences since when provided by cli only strings are possible
```
ccwienk, line 18: see above: not necessary

whitesource/client.py
```diff
@@ -64,41 +64,47 @@ def save_project_tag(self, projectToken: str, key: str, value: str):
...
-    def post_project(
+    async def ws_project(
         self,
         extra_whitesource_config: typing.Dict,
         file,
-        filename: str,
         project_name: str,
         requester_email: str,
+        length: int,
+        chunk_size: int,
```
ccwienk, line 74: I think this should be the _only_ place (for now) where where hard-code the chunk_size, ping_interval, and ping_timeout.

whitesource/client.py
```diff
@@ -64,41 +64,47 @@ def save_project_tag(self, projectToken: str, key: str, value: str):
             json=body,
         )
 
-    def post_project(
+    async def ws_project(
```
ccwienk, line 67: I would prefer a more expressive name. such as `upload_to_project` or sth. similar (the previous name was a bit better, but also not great, as it was named after the technical detail of the underlying http-post request. a good name should not have to be changed just because the communication method changes)

whitesource/client.py
```diff
@@ -64,41 +64,47 @@ def save_project_tag(self, projectToken: str, key: str, value: str):
...
-            'wsConfig': json.dumps({
-                'apiKey': self.api_key,
-                'projectName': project_name,
-                'productToken': self.product_token,
-                'requesterEmail': requester_email,
-                'userKey': self.creds.user_key(),
-                'wssUrl': self.wss_endpoint,
-                'extraWsConfig': extra_whitesource_config,
-            })
+        meta_data = {
```
ccwienk, line 79: a brief comment would be nice (explaining that this document will configure the interaction between client and backend).

also, I think this should be moved to the server-repository. somewhat like so (did not lookup the names):

```
import ws_api.protocol
communication_cfg = ws_api.protocol.communication_cfg(chunk_size=4096, length=<length>)
```

This way, it is a bit more obvious that this be a contract between server and client (+ the datastructure is only defined once)

whitesource/client.py
```diff
@@ -64,41 +64,47 @@ def save_project_tag(self, projectToken: str, key: str, value: str):
             json=body,
         )
 
-    def post_project(
+    async def ws_project(
         self,
         extra_whitesource_config: typing.Dict,
         file,
```
ccwienk, line 70: since file seems to be a filelike-object, you should indicate this via typehint

whitesource/client.py
```diff
@@ -64,41 +64,47 @@ def save_project_tag(self, projectToken: str, key: str, value: str):
...
-        )
+        async with websockets.connect(
+            uri=self.routes.ws_component(),
+            ping_interval=ping_interval,
+            ping_timeout=ping_timeout,
+            ) as websocket:
+            await websocket.send(json.dumps(meta_data))
+            await websocket.send(json.dumps(ws_config))
+            while True:
+                chunk = file.read(chunk_size)
```
ccwienk, line 102: the passer of the filelike-object might not want you to read the entire file (it could be larger than the specified size/leng)

--> rather keep track of sent octets, calculate octets to read (min(chunk_size, remainder)), abort once the deed is done
you might also want to handle the (error) case, where more octets are requested to be read than the filelikeobj has (this would be a programming error).

reading through the entire file and assuming everything be fine after the file has been entirely read is not a good idea, as it may shadow different kinds of errors (e.g. the filelike-object might not have been "seeked" to the beginning, so you would upload zero bytes, which most likely is not what the caller wanted..)

whitesource/client.py
```diff
@@ -64,41 +64,47 @@ def save_project_tag(self, projectToken: str, key: str, value: str):
...
+                if not chunk:
+                    break
+                await websocket.send(chunk)
+
+            return await websocket.recv()
```
ccwienk, line 107: what's the response (sorry, too lazy to lookup on server-side)? if it is again some machine-readable piece of json, it would be nice to deserialise it prior to returning

whitesource/client.py
```diff
@@ -171,7 +177,7 @@ def __init__(
         self.extension_endpoint = extension_endpoint
         self.wss_api_endpoint = wss_api_endpoint
 
-    def post_component(self):
+    def ws_component(self):
```
ccwienk, line 180: same here: please choose a name that expresses the (highlevel) operation, rather than the technical implementation-detail of the used http-"method" (ws is not actually a http-method, as you correctly noted yesterday)

whitesource/util.py
```diff
@@ -280,24 +284,30 @@ def scan_artifact_with_ws(
...
         clogger.info('downloading component for scan')
-        component_filename = whitesource.component.download_component(
+        whitesource.component.download_component(
             clogger=clogger,
             github_repo=github_repo,
             path_filter_func=path_filter_func,
             ref=commit_hash,
             target=tmp_file,
         )
+        file_size = tmp_file.tell()
```
ccwienk, line 294: this is somewhat hacky.. why not have download_component return the fsize?

whitesource/util.py
```diff
@@ -247,6 +248,9 @@ def scan_artifact_with_ws(
     requester_mail: str,
```
ccwienk, line 248: btw: with `ws` being used both for websocket, and whitesource, we now have a naming-clash.

--> maybe rather rename to whitesource and websocket (or some abbrevs, such as websock, white_src, .. it is ok to be brief, but we should avoid ambiguities)

ccwienk (changes requested):

whitesource/util.py
zkdev: Didn't even notice, will go for **ws** and **white_src** 👍

whitesource/client.py
zkdev: Sure, **upload_to_project** seems fine

whitesource/client.py
zkdev: Postponed

PR #552, opened 2021-01-20
Configure File-Logger with Logrotation for WHD

AndreasBurger:
Configures our WHD to _also_ log to a file (`/tmp/whd_log` by default). Log rotation is configured to take place as soon as this file has a size of 50 MiB with one backup (will then be named `whd_log.1`)

### DIFF
--- a/ctx.py
+++ b/ctx.py
@@ -15,6 +15,8 @@
 
 import enum
 import functools
+import logging
+import logging.config
 import os
 
 from pathlib import Path
@@ -203,13 +205,8 @@ def cfg_factory():
     return factory
 
 
-def configure_default_logging(stdout_level=None):
-    import logging
-    import logging.config
-    if not stdout_level:
-        stdout_level = logging.INFO
-
-    cfg = {
+def _default_logging_config(stdout_level):
+    return {
         'version': 1,
         'formatters': {
             'default': {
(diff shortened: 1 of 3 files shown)


ctx.py
```diff
@@ -203,11 +203,13 @@ def cfg_factory():
     return factory
 
 
-def configure_default_logging(stdout_level=None):
+def configure_default_logging(stdout_level=None, whd_log_file_path=None):
```
ccwienk, line 206: this is a general-purpose-function for all of cc-utils / gardener-cicd. I would not like to add a special-case-argument for whd, tbh

I understand you want to re-use the existing code. How about splitting this function into two parts:
1. one that will create / return a cfg-dict (identical to the one that is already created in this function)
2. one that will by default call the first function, and configure the logging frmwk

for WHD, you could then call the first function, enrich the default-dict with additional cfg and then "manually" configure the logging-module. Maybe the second function is actually not that much of a value-add, and might be omitted (need to check existing usages to tell)

ccwienk: as an alternative, I would also not be opposed to creating a dedicated logging-cfg just for WHD. WHD is a special, separate component (and should actually not necessarily live within cc-utils anyway), so why not have it define a custom logging-cfg?

ccwienk: as an alternative, thinking some more about it. if we maybe named this path "logfile" or sth similarly generic, it would make some sense to add it here

ctx.py
```diff
@@ -203,11 +203,13 @@ def cfg_factory():
...
     import logging
     import logging.config
     if not stdout_level:
         stdout_level = logging.INFO
+    if not whd_log_file_path:
+        whd_log_file_path = os.path.join('/', 'tmp', 'whd_log')
```
ccwienk, line 212: please do not do that.
if there is not logfile passed, there is not logging to a logfile

ccwienk (changes requested):

ccwienk (approved):

PR #557, opened 2021-01-29
Add dependabot scanner

zkdev:
Scans repositories from specified organization for `dependabot.yaml` to create a coverage report.

Fixes #

None

**Release note**:

```improvement operator

```


ccwienk (changes requested):

dso/dependabot.py
```diff
@@ -0,0 +1,187 @@
...
+    return github_org.repositories
+
+
+def _scan_repos_for_dependabot(
+    repos: typing.List[str],
+    host: str,
+) -> dso.model.DependabotCoverageReport:
+
+    clogger = dso.util.component_logger(__name__)
+    i = 1
```
ccwienk, line 78: nit: i is unused (de facto)

dso/dependabot.py
```diff
@@ -0,0 +1,187 @@
...
+        )
+        repo_report = dso.model.DependabotCoverageReportRepo(
+            dependabot=dependabot_found,
+            repo=repo
+        )
+        details.append(repo_report)
+        i += 1
+
+    return dso.model.DependabotCoverageReport(
+        coverage=0,
```
ccwienk, line 103: coverage=0? that cannot be what you intended. also: what is the semantics of this parameter?

dso/dependabot.py
```diff
@@ -0,0 +1,187 @@
...
+        details=details,
+        github=host,
+    )
+
+
+def _get_github_api(host):
```
ccwienk, line 109: nit: I would prefer to not name functions that are not simple getters (as in: return an attr (from an object), maybe doing some trivial operations on the attr) with a `get` prefix.

regardless if my nit, I think this convenience function should be moved to `ccc.github`, where it is obviously missing (in fact, I also needed such a function several times, but always was too lazy to implement one..)

+ please add a typehint for the return-value (and preferrably also the parameter - even though the latter is debatable, as it is just `str`..)

dso/dependabot.py
```diff
@@ -0,0 +1,187 @@
...
+    github_cfg = ccc.github.github_cfg_for_hostname(
+        cfg_factory=ci.util.ctx().cfg_factory(),
+        host_name=host,
+    )
+    return ccc.github.github_api(github_cfg)
+
+
+def _get_scoped_repos_for_org(
```
ccwienk, line 117: I do not think `scoped` is an appropriate term(*). In GitHub's terms, I "my" terms (as used elsewhere in cc-utils), I would tend to call `{org}/{repo}` a repo(sitory)-path. I think, github's API calls this the `full_name` of a repository. So repository-name might also be  adequate. or maybe qualified name.

(*) one could argue that the repo-name (w/o org) is only valid in the scope of its organisation. However, at least from my experience, I never saw the naming convention of calling a qualified name a scoped name (as opposed to canonical, qualified, or full name), and I find it unintuitive, at least (maybe it is not strictly "wrong").

dso/dependabot.py
```diff
@@ -0,0 +1,187 @@
...
+) -> typing.List[str]:
+    clogger = dso.util.component_logger(__name__)
+    repos: typing.List[str] = []
+
+    clogger.info(f'org specified, scanning {org}')
+    host, org = org.split('/')
+    github_api = _get_github_api(host)
+    github_org = github_api.organization(org)
+    for repo in github_org.repositories():
+        repos.append(repo.full_name)
```
ccwienk, line 128: nit: this might be done more elegant using a list-comprehension

dso/dependabot.py
```diff
@@ -0,0 +1,187 @@
...
+
+    return repos[:3]
```
ccwienk, line 130: why do you omit all but the first three repositories?

dso/dependabot.py
```diff
@@ -0,0 +1,187 @@
...
+
+
+def _generate_reporting_tables(
+    report: dso.model.DependabotCoverageReport,
+    tablefmt: str,
+):
+    # monkeypatch: disable html escaping
+    tabulate.htmlescape = lambda x: x
```
ccwienk, line 138: nit: only required for html (+ TODO: check whether this has not already been fixed upstream :))

dso/dependabot.py
```diff
@@ -0,0 +1,187 @@
...
+    tables.append(tabulate.tabulate(
+        headers=('Component', 'Dependabot'),
+        tabular_data=table_data,
+        tablefmt=tablefmt,
+        colalign=('left', 'center'),
+    ))
+
+    table_data = ((report.github, f'{report.coverage}%'),)
+    tables.append(tabulate.tabulate(
+        headers=('Overall Coverage', 'Percentage'),
```
ccwienk, line 156: nit: maybe express more explicit what "coverage" means (I assume it is the percentage of repositories that do use dependabot?)

dso/dependabot.py
```diff
@@ -0,0 +1,187 @@
...
+    print('\n')
+
+
+def dependabot_coverage(
+    org: str,
+    outfile_path: str,
+):
+    repos = _get_scoped_repos_for_org(org=org)
+    host, _ = org.split('/')
+    report = _scan_repos_for_dependabot(repos=repos, host=host)
```
ccwienk, line 185: so you first create repository-objects (`_get_scoped_repos_for_org'), to then discard those objects again and only use their full-names, which you then use to create repository.-objects _again_? this sounds a bit cumbersome to me (also, it uses two requests for each repository, where we in fact only would need one :-/

dso/model.py
```diff
@@ -17,3 +17,27 @@ class ScanArtifact:
...
+
+
+@dataclasses.dataclass(frozen=False)
+class DependabotCoverageReport:
+    coverage: float
+    github: str
+    details: typing.List[DependabotCoverageReportRepo]
+
+    def calculate_overall_percentage(self):
+        t = 0
```
ccwienk, line 35: stylistic nit: this is a very cheap operation. rather either already do it in the c'tor, or (pbly better) do the calculation upfront, and pass-in the calculated values (-> only use this dataclass as a "struct"). I think it is a bad pattern to have a class-definition leading to objects that are useless w/o calling a special method once

ccwienk (changes requested):
some more nits..

also, as a general remark, I suggest you either create a package `dependabot` on root-level of cc-utils, or a sub-package `sdo/dependabot`.

as for the repository-enumeration, I suggest so simply use the github3.py-api with less wrapping (both the org- and repository-objects it yields already have all the data you need).

roughly sketched, I would think we need one function that accepts a github3.py-api-repository-object, and returns a value indicating the dependabot-status (preferrably an enum), like so:

```
class DependebotStatus(enum.Enum):
 ENABLED = 'enabled'
 NOT_ENABLED = 'not_enabled'
 UNKNOWN = 'unknown' # not sure whether we need it. e.g. use it in case there was an error (say there is a dependabot-file, but it contained invalid yaml)

def dependabot_status(repository: github3.repos.repo.Repository):
 if not (files := retrieve_known_dependabot_files(repository)):
  return DependebotStatus.NOT_ENABLED
 # sanity-checks
 if len(files) > 1:
  # not sure - does dependabot allow this? - do something reasonable in this case..
  pass
 dependabot_file = files[0]
 if validate_dependabot_file(dependabot_file):
  return DependabotStatus.ENABLED
 else:
  # might also log a warning, ....
  return DependabotStatus.UNKNOWN

# using the function above, you might simply fill a mapping, like so:
repo_status = {
 repo: dependabot_status(repo)
 for repo in repositories
}
# where repositories is e.g. returned from a githu3.py-api-org-object

from such a dict, you can easily construct the reporting (read repo-full-name from repo (dict-key); dependabot-status is already the value..)

File: dependabot/status.py [hunk 11]
```diff
@@ -0,0 +1,142 @@
...
+
+
+def _validate_dependabot_file(
+    dependabot_file,
+) -> bool:
+
+    # TODO to be extended, e.g. scan for required attributes, see:
+    # https://docs.github.com/en/github/administering-a-repository/
+    # configuration-options-for-dependency-updates
+    content = dependabot_file.decoded.decode("utf-8")
```
ccwienk, line 36: nit: by convention, we tend to prefer to use single-quotes in cc-utils

dependabot/status.py
```diff
@@ -0,0 +1,142 @@
...
+        full_org_name=full_org_name,
+    )
+
+    with open(outfile_path, 'a') as f:
+        now = datetime.datetime.now()
+        f.write(f'{now.strftime("%d/%m/%Y %H:%M:%S")}\n')
+        for t in report_tables:
+            f.write(f'{t}\n\n')
+            print(f'\n{t}')
+        f.write('\n===================================================================\n\n')
```
ccwienk, line 113: hint: you do know you can multiply strings? :-)

ccwienk: also, I would like to advertise `textwrap.fill` (+ other functions^^)

dependabot/status.py
```diff
@@ -0,0 +1,142 @@
...
+    full_org_name: str,
+):
+
+    report_tables = _generate_report_tables_from_repo_status(
+        repo_status=repo_status,
+        full_org_name=full_org_name,
+    )
+
+    with open(outfile_path, 'a') as f:
+        now = datetime.datetime.now()
```
ccwienk, line 108: nit: how about isoformat()?

dependabot/status.py
```diff
@@ -0,0 +1,142 @@
...
+        f.write(f'{now.strftime("%d/%m/%Y %H:%M:%S")}\n')
+        for t in report_tables:
+            f.write(f'{t}\n\n')
+            print(f'\n{t}')
+        f.write('\n===================================================================\n\n')
+    print('\n')
+
+
+def _calculate_coverage_rate(
```
ccwienk, line 117: nit: rate should be between 0..1, should it not? what this function returns is probably rather called percentage, I think

dependabot/util.py
```diff
@@ -0,0 +1,16 @@
+import github3.repos
+import typing
+
+import ccc.github
+
+
+def repositories_for_org(
+    full_org_name: str,  # e.g. github.com/gardener
```
ccwienk, line 8: not sure I like `full_org_name` :-/
how about qualified_org, or github_org_url, or hostname_and_org?

dependabot/util.py
```diff
@@ -0,0 +1,16 @@
...
+) -> typing.List[github3.repos.repo.ShortRepository]:
+
+    host, org = full_org_name.split('/')
```
ccwienk, line 11: rather conditionally prepend a schema (if absent) + use urllib.parse.urlparse to parse into host and path?

ccwienk (changes requested):

dependabot/status.py
```diff
@@ -0,0 +1,142 @@
...
+    dependabot_file,
+) -> bool:
+
+    # TODO to be extended, e.g. scan for required attributes, see:
+    # https://docs.github.com/en/github/administering-a-repository/
+    # configuration-options-for-dependency-updates
+    content = dependabot_file.decoded.decode("utf-8")
+    try:
+        ci.util.load_yaml(content)
+    except (AttributeError,):
```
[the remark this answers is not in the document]
zkdev: Was meant to be extended to catch more parsing errors, will adjust

dependabot/status.py
```diff
@@ -0,0 +1,142 @@
...
+):
+
+    # According to https://docs.github.com/en/github/administering-a-repository/configuration-options
+    # -for-dependency-updates , there is only one valid path for the dependabot.yml.
+    try:
+        file = repository.file_contents(path='.github/dependabot.yml')
+        ci.util.info(f'dependabot in {repository.full_name}')
+        return file
+    except github3.exceptions.NotFoundError:
+        ci.util.info(f'dependabot not in {repository.full_name}')
```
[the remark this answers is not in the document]
zkdev: True, will adjust

dependabot/status.py
```diff
@@ -0,0 +1,142 @@
+import ci.util
```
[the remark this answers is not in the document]
zkdev: Yea that was not meant to be ordered like this, will rearrange.

dependabot/model.py
```diff
@@ -0,0 +1,15 @@
...
+
+class DependebotStatus(enum.Enum):
+    ENABLED = 'enabled'
+    NOT_ENABLED = 'not_enabled'
+    UNKNOWN = 'unknown'  # error case, e.g. dependabot.yaml found but faulty yaml
+
+
+@dataclasses.dataclass(frozen=True)
+class DependabotStatusForRepo:
+    repo: github3.repos.ShortRepository
```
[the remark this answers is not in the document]
zkdev: sure, see: https://github.com/sigmavirus24/github3.py/blob/6572d7556e925f45b713bea3fb93f791e2ac9074/src/github3/models.py#L123

dependabot/status.py
zkdev: Absolutely.
Maybe we should extend our linter to check for unnecessary double quotation 🤔?

see: https://github.com/zheller/flake8-quotes

dependabot/util.py
zkdev: `github_org_url` sounds good

dependabot/model.py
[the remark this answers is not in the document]
zkdev: the repo uri is hashed, see: https://github.com/sigmavirus24/github3.py/blob/6572d7556e925f45b713bea3fb93f791e2ac9074/src/github3/models.py#L43

dependabot/util.py
zkdev, line 8: here

zkdev:
Apparently dependabot is enabled for the whole gardener org.
Will put this PR on hold.

ccwienk:
closing, as discussed, as there is no need for dependabot-reportings (it is enabled for entire gh-orgs anyway)

@zkdev : sorry - my bad :-/

PR #558, opened 2021-01-29
Remove final dependencies on product.model

AndreasBurger:
Removes the remaining dependencies on product.model from our codebase.

Draft, as tests need to be adjusted and CTF still is not supported. Manual testing so far indicates that the generated release-notes are the same as before.

### DIFF
--- a/cnudie/util.py
+++ b/cnudie/util.py
@@ -296,7 +296,7 @@ def diff_resources(
     return resource_diff
 
 
-def component_descriptor_from_ctf_archive(
+def component_descriptors_from_ctf_archive(
     ctf_archive: typing.Union[str, typing.IO[bytes]],
 ) -> typing.Generator[cm.ComponentDescriptor, None, None]:
     if isinstance(ctf_archive, str):
--- a/concourse/steps/release.py
+++ b/concourse/steps/release.py
@@ -19,6 +19,8 @@ from ci.util import (
     warning,
 )
 import cnudie.retrieve
+import cnudie.util
+
 from gitutil import GitHelper
 from github.util import (
     GitHubRepositoryHelper,
@@ -555,7 +557,7 @@ class GitHubReleaseStep(TransactionalStep):
         githubrepobranch: GitHubRepoBranch,
         repo_dir: str,
         release_version: str,
-        component_descriptor_v2_path:str,
+        component_descriptor_v2_path: str,
         ctf_path:str,
     ):
         self.github_helper = not_none(github_helper)
@@ -662,28 +664,49 @@ class PublishReleaseNotesStep(TransactionalStep):
         self,
         githubrepobranch: GitHubRepoBranch,
         github_helper: GitHubRepositoryHelper,
+        component_descriptor_v2_path: str,
         release_version: str,
+        ctf_path: str,
         repo_dir: str,
     ):
         self.githubrepobranch = not_none(githubrepobranch)
         self.github_helper = not_none(github_helper)
         self.release_version = not_empty(release_version)
         self.repo_dir = os.path.abspath(not_empty(repo_dir))
+        self.component_descriptor_v2_path = component_descriptor_v2_path
+        self.ctf_path = ctf_path
 
     def validate(self):
         version.parse_to_semver(self.release_version)
         existing_dir(self.repo_dir)
+
+        have_ctf = os.path.exists(self.ctf_path)
+        have_cd = os.path.exists(self.component_descriptor_v2_path)
+        if not have_ctf ^ have_cd:
+            ci.util.fail('exactly one of component-descriptor, or ctf-archive must exist')
+        elif have_cd:
+            self.component_descriptor_v2 = cm.ComponentDescriptor.from_dict(
+            ci.util.parse_yaml_file(self.component_descriptor_v2_path),
+        )
+        elif have_ctf:
+            component_descriptors = list(cnudie.util.component_descriptors_from_ctf_archive(
+                self.ctf_path,
+            ))
+            if not component_descriptors:
+                ci.util.fail(f'No component descriptor found in CTF archive at {self.ctf_path}')
+            if len(component_descriptors) > 1:
+                ci.util.fail(
+                    f'More than one component descriptor found in CTF archive at {self.ctf_path}'
+                )
+            self.component_descriptor_v2 = component_descriptors[0]
 
     def apply(self):
         create_release_step_output = self.context().step_output('Create Release')
         release_tag = create_release_step_output['release_tag_name']
 
         release_notes = fetch_release_notes(
-            github_repository_owner=self.githubrepobranch.repo_owner(),
-            github_repository_name=self.githubrepobranch.repo_name(),
-            github_cfg=self.githubrepobranch.github_config(),
+            self.component_descriptor_v2.component,
             repo_dir=self.repo_dir,
-            github_helper=self.github_helper,
             repository_branch=self.githubrepobranch.branch(),
         )
         release_notes_md = release_notes.to_markdown()
@@ -930,6 +953,8 @@ def release_and_prepare_next_dev_cycle(
         githubrepobranch=githubrepobranch,
         github_helper=github_helper,
         release_version=release_version,
+        component_descriptor_v2_path=component_descriptor_v2_path,
+        ctf_path=ctf_path,
         repo_dir=repo_dir,
     )
 
--- a/github/release_notes/model.py
+++ b/github/release_notes/model.py
@@ -12,12 +12,12 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
-from dataclasses import dataclass
+from dataclasses import dataclass, field
+import gci
 from pydash import _
 
-from product.model import ComponentName
-from ci.util import check_type
+import ci.util
+import cnudie.util
 
 
 @dataclass
@@ -71,7 +71,7 @@ class ReleaseNote:
     user_login: str
     is_current_repo: bool
     from_same_github_instance: str
-    cn_source_repo: str
+    source_component: gci.componentmodel.Component = field(compare=False)
 
 
 class ReleaseNoteBlock(ReleaseNote):
@@ -83,17 +83,29 @@ class ReleaseNoteBlock(ReleaseNote):
         reference_type: ReferenceType,
         reference_id: str,
         user_login: str,
-        source_repo: str,
-        cn_current_repo: ComponentName,
+        current_component: gci.componentmodel.Component,
+        source_component: gci.componentmodel.Component,
     ):
         if reference_id:
-            check_type(reference_id, str)
+            ci.util.check_type(reference_id, str)
+
+        ci.util.not_none(source_component)
 
         reference = Reference(type=reference_type, identifier=reference_id)
 
-        cn_source_repo = ComponentName(name=source_repo)
-        is_current_repo = cn_current_repo == cn_source_repo
-        from_same_github_instance = cn_current_repo.github_host() == cn_source_repo.github_host()
+        source_component_access = cnudie.util.determine_main_source_for_component(
+            source_component
+        ).access
+        current_component_access = cnudie.util.determine_main_source_for_component(
+            current_component
+        ).access
+
+        is_current_repo = current_component.name == source_component.name
+
+        current_hostname = current_component_access.hostname()
+        source_hostname = source_component_access.hostname()
+        from_same_github_instance = current_hostname in source_hostname
+
         super().__init__(
             category_id,
             target_group_id,
@@ -102,7 +114,7 @@ class ReleaseNoteBlock(ReleaseNote):
             user_login,
             is_current_repo,
             from_same_github_instance,
-            cn_source_repo
+            source_component,
         )
 
     def ref(self):
@@ -124,7 +136,7 @@ class ReleaseNoteBlock(ReleaseNote):
             '```'.format(
                 cat=self.category_id,
                 t_grp=self.target_group_id,
-                src_repo=self.cn_source_repo.name(),
+                src_repo=self.source_component.name,
                 ref=self.ref(),
                 user=self.user(),
                 text=self.text
--- a/github/release_notes/renderer.py
+++ b/github/release_notes/renderer.py
@@ -15,7 +15,10 @@
 
 from abc import abstractmethod
 from collections import namedtuple
+import typing
 from pydash import _
+
+import cnudie.util
 
 from github.release_notes.model import ReleaseNote, REF_TYPE_COMMIT
 
@@ -117,23 +120,27 @@ CATEGORIES = [
 ]
 
 
+def _access(component):
+    return cnudie.util.determine_main_source_for_component(component, False).access
+
+
 class Renderer(object):
-    def __init__(self, release_note_objs: [ReleaseNote]):
+    def __init__(self, release_note_objs: typing.List[ReleaseNote]):
         self.rls_note_objs = _.uniq(release_note_objs)
 
     def render(self) -> str:
         origin_nodes = _\
             .chain(self.rls_note_objs)\
-            .sort_by(lambda rls_note_obj: rls_note_obj.cn_source_repo.github_repo())\
+            .sort_by(lambda rls_note_obj: _access(rls_note_obj.source_component).repository_name())\
             .sort_by(lambda rls_note_obj: rls_note_obj.is_current_repo, reverse=True)\
-            .uniq_by(lambda rls_note_obj: rls_note_obj.cn_source_repo.name())\
+            .uniq_by(lambda rls_note_obj: rls_note_obj.source_component.name)\
             .map(lambda rls_note_obj: TitleNode(
-                identifiers=[rls_note_obj.cn_source_repo.name()],
+                identifiers=[rls_note_obj.source_component.name],
                 title='[{origin_name}]'.format(
-                    origin_name=rls_note_obj.cn_source_repo.github_repo()
+                    origin_name=_access(rls_note_obj.source_component).repository_name(),
                 ),
                 nodes=CATEGORIES,
-                matches_rls_note_field_path='cn_source_repo.name' # path points to a function
+                matches_rls_note_field_path='source_component.name',
             ))\
             .value()
 
@@ -149,10 +156,10 @@ class Renderer(object):
 
     def _to_release_note_lines(
         self,
-        nodes: [TitleNode],
+        nodes: typing.List[TitleNode],
         level: int,
-        rls_note_objs: [ReleaseNote]
-    ) -> [str]:
+        rls_note_objs: typing.List[ReleaseNote],
+    ) -> typing.List[str]:
         lines = list()
         for node in nodes:
             filtered_rls_note_objects = _.filter(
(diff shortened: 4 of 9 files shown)


gitutil.py
```diff
@@ -19,10 +19,14 @@
 import subprocess
 import tempfile
 import urllib.parse
+import gci
```
ccwienk, line 22: nit: should be moved two "import-groups" below

github/util.py
```diff
@@ -373,6 +373,16 @@ def from_githubrepobranch(
             default_branch=githubrepobranch.branch(),
         )
 
+    @staticmethod
+    def from_github_access(
```
ccwienk, line 377: I would have preferred a util-function that would turn a github-access into a github-cfg (also below)

concourse/steps/release.py
```diff
@@ -678,12 +680,12 @@ def apply(self):
         create_release_step_output = self.context().step_output('Create Release')
         release_tag = create_release_step_output['release_tag_name']
 
+        component_descriptor_v2 = cm.ComponentDescriptor.from_dict(
```
ccwienk: note to self: we should definitely add an option to parse component-descriptor from file(-like object)

github/release_notes/model.py
```diff
@@ -12,12 +12,12 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
-from dataclasses import dataclass
+from dataclasses import dataclass, field
+import gci
```
ccwienk, line 16: nit: group imports (<stdlib>, <3rdparty>, <local>), in the special-case of component-spec (gci), consider a 4th group between 3rdp and local

github/release_notes/model.py
```diff
@@ -71,7 +71,9 @@ class ReleaseNote:
     user_login: str
     is_current_repo: bool
```
ccwienk, line 72: I have some questions w.r.t. this dataclass.

- what does `is_current_repo` mean? also: since there are now source_component_name and current_component_name, can this attr not be calculated (if it is at all required)?
- what is user_login used for?

github/release_notes/model.py
```diff
@@ -71,7 +71,9 @@ class ReleaseNote:
...
     from_same_github_instance: str
-    cn_source_repo: str
+    current_component_name: str
```
ccwienk, line 74: how about adding the full component (which would include the access), rather than merely the name?

github/release_notes/model.py
```diff
@@ -71,7 +71,9 @@ class ReleaseNote:
...
+    source_component_name: str
+    source_component_access: gci.componentmodel.GithubAccess = field(compare=False)
```
ccwienk, line 76: is it really required to omit it from comparison? however if it is: should we not also omit it from hash for reasons of symmetry?

github/release_notes/model.py
```diff
@@ -83,17 +85,29 @@ def __init__(
...
 
-        cn_source_repo = ComponentName(name=source_repo)
-        is_current_repo = cn_current_repo == cn_source_repo
-        from_same_github_instance = cn_current_repo.github_host() == cn_source_repo.github_host()
+        source_component_access = cnudie.util.determine_main_source_for_component(
+            source_component
+        ).access
+        current_component_access = cnudie.util.determine_main_source_for_component(
+            current_component
+        ).access
```
ccwienk, line 103: you should also check that it is a github-access, please (I know, currently this is the only one. however, cnudie allows different access-types)

github/release_notes/model.py
```diff
@@ -83,17 +85,29 @@ def __init__(
...
+
+        is_current_repo = current_component.name == source_component.name
```
ccwienk, line 105: nit: please talk about _components_ (I know, there is currently a 1:1-relationship between components and github-repositories. however this is not necessarily so in all cases in the future, thanks to CNUDIE's more generic model.

ccwienk: also, I would prefer to not persist a value that can be trivially calculated. rather calculate it where required (maybe add a method if it is required multiple times)

github/release_notes/model.py
```diff
@@ -83,17 +85,29 @@ def __init__(
...
+
+        current_hostname = current_component_access.hostname()
+        source_hostname = source_component_access.hostname()
+        from_same_github_instance = current_hostname in source_hostname
```
ccwienk, line 109: why is this even required? we do have a lookup-function for github-cfgs by hostname

github/release_notes/model.py
```diff
@@ -124,7 +140,7 @@ def to_block_str(self):
             '```'.format(
                 cat=self.category_id,
                 t_grp=self.target_group_id,
-                src_repo=self.cn_source_repo.name(),
+                src_repo=self.source_component_name,
```
ccwienk, line 143: if `src_repo` actually refers to a github-repository, please do not use the component-name / a variable named `component_name`. it is just a coincidence that most (ok, all..) of our component-names are in fact github-repository-URLs (w/o schema). preferrably pass the repository-access. or github-cfg. or the component. not just a str. this is the weakest kind of typing available, and I think we should leverage the benefits of the type-system

I know that the existing codebase is not ideal. however, I would like to at least improve it a bit if we have to touch it anyhow

github/release_notes/renderer.py
```diff
@@ -15,6 +15,7 @@
 
 from abc import abstractmethod
 from collections import namedtuple
+import typing
```
ccwienk, line 18: nit: blank line between import-blocks

github/release_notes/renderer.py
```diff
@@ -118,22 +119,22 @@ def get_or_call(obj, path):
...
             .map(lambda rls_note_obj: TitleNode(
-                identifiers=[rls_note_obj.cn_source_repo.name()],
+                identifiers=[rls_note_obj.source_component_name],
                 title='[{origin_name}]'.format(
-                    origin_name=rls_note_obj.cn_source_repo.github_repo()
+                    origin_name=rls_note_obj.source_component_access.repository_name(),
                 ),
                 nodes=CATEGORIES,
-                matches_rls_note_field_path='cn_source_repo.name' # path points to a function
+                matches_rls_note_field_path='source_component_name',
```
ccwienk, line 137: does it point to a function/callable?

github/release_notes/renderer.py
```diff
@@ -216,20 +217,16 @@ def _header_suffix_reference(
             if should_generate_link:
                 reference_id_text = rls_note_obj.reference.identifier[0:12] # short commit hash
 
-        reference = '{reference_prefix}{ref_id}'.format(
-            reference_prefix=reference_prefix,
-            ref_id=reference_id_text,
-        )
-
+        reference = f'{reference_prefix}{reference_id_text}'
         if is_reference_auto_linked:
```
ccwienk, line 221: nit: in general, it would make reviewing a lot easier if refactorings like this (there are also others) would be put into separate commits :-/

github/release_notes/renderer.py
```diff
@@ -216,20 +217,16 @@ def _header_suffix_reference(
...
             return reference
 
         if not should_generate_link:
             # returns e.g. gardener/cc-utils#42 or g. gardener/cc-utils@commit-hash
-            return '{repo_path}{reference}'.format(
-                    repo_path=rls_note_obj.cn_source_repo.github_repo_path(),
-                    reference=reference
-                )
+            return (
+                f'{rls_note_obj.source_component_access.org_name()}/'
```
ccwienk, line 227: this is a bit cumbersome. how about adding a convenience-method to the access-type? (e.g. repository_full_name, as github calls it). also, how about using urljoin? I think, the formatstrings in this case make things harder to read than if using simple str-concats tbh

github/release_notes/renderer.py
```diff
@@ -248,23 +245,23 @@ def _header_suffix_user(
 
         return self._github_user_profile_link(
             user=rls_note_obj.user_login,
-            github_url=rls_note_obj.cn_source_repo.github_url()
+            github_url=f'https://{rls_note_obj.source_component_access.hostname()}',
```
ccwienk, line 248: :-/

github/release_notes/renderer.py
```diff
@@ -248,23 +245,23 @@ def _header_suffix_user(
...
         self,
         rls_note_obj: ReleaseNote,
         reference: str
     ) -> str:
-        reference_link = '{source_repo_url}/{github_api_resource_type}/{ref_id}'.format(
-            source_repo_url=rls_note_obj.cn_source_repo.github_repo_url(),
-            ref_id=rls_note_obj.reference.identifier,
-            github_api_resource_type=rls_note_obj.reference.type.github_api_resource_type
+        reference_link = (
+            f'https://{rls_note_obj.source_component_name}/'
```
ccwienk, line 257: please use urljoin

github/release_notes/renderer.py
```diff
@@ -248,23 +245,23 @@ def _header_suffix_user(
...
+            f'{rls_note_obj.reference.type.github_api_resource_type}/'
+            f'{rls_note_obj.reference.identifier}'
         )
 
-        link_text = '{repo_path}{reference}'.format(
-            repo_path=rls_note_obj.cn_source_repo.github_repo_path(),
-            reference=reference
+        link_text = (
+            f'{rls_note_obj.source_component_access.org_name()}/'
```
ccwienk, line 263: urljoin

github/release_notes/util.py
```diff
@@ -13,50 +13,46 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import cnudie.util
```
ccwienk, line 16: nit: group imports

github/release_notes/util.py
```diff
@@ -13,50 +13,46 @@
...
 from github.util import GitHubRepositoryHelper
 from pydash import _
 import re
+import requests
+import tempfile
 import typing
 
 from github.release_notes.model import (
-    ReleaseNote,
+    REF_TYPE_PULL_REQUEST, ReleaseNote,
```
ccwienk, line 27: nit: two imports in one line

github/release_notes/util.py
```diff
@@ -13,50 +13,46 @@
...
 
 def fetch_release_notes(
-    github_repository_owner: str,
-    github_repository_name: str,
-    github_cfg: str,
-    repo_dir: str,
-    github_helper: GitHubRepositoryHelper,
+    component: gci.componentmodel.Component,
     repository_branch: str,
+    repo_dir: str,
```
ccwienk, line 51: why do we even need the repo-dir?

github/release_notes/util.py
```diff
@@ -125,29 +121,37 @@ def github_repo_path(owner, name):
...
-class ReleaseNotes(object):
+class ReleaseNotes:
     def __init__(
         self,
-        release_note_objs: [str]
+        component: gci.componentmodel.Component=None,
+        repo_dir: str=None,
     ):
-        self.release_note_objs = release_note_objs
+        if component:
```
ccwienk, line 130: in which case is it acceptable that no component is passed?

github/release_notes/util.py
```diff
@@ -125,29 +121,37 @@ def github_repo_path(owner, name):
...
+            source = cnudie.util.determine_main_source_for_component(component)
+            if repo_dir:
+                self.github_helper = GitHubRepositoryHelper.from_github_access(
+                    github_access=source.access,
+                )
+                self.git_helper = GitHelper.from_github_access(
+                    github_access=source.access,
+                    repo_path=repo_dir,
+                )
+        self.cd_cache_dir = tempfile.TemporaryDirectory()
```
ccwienk, line 141: tmpdir should be passed-in

github/release_notes/util.py
```diff
@@ -172,239 +176,187 @@ def release_note_blocks(self):
...
+            fetch_release_notes_from_commits(
+                commits=commits,
+                current_component=self.component,
             )
-        reachable_tags.append(root_commit.hexsha)
-
-    return reachable_tags
-
-
-def commits_in_range(
```
ccwienk, line 300: why is there this huge diff? judging from the signature, I would have thought that this function does not need to be touched at all?

github/release_notes/util.py
```diff
@@ -172,239 +176,187 @@ def release_note_blocks(self):
...
+        return commit_range
 
-        pr_number = str(pr_dict['number'])
-        if pr_number not in pr_numbers_in_range:
-            continue
+    def release_tags(
+        self,
+        release_tags: typing.List[str],
+    ) -> typing.List[str]:
+        # TODO: Seems unused.
```
ccwienk, line 239: then remove it

github/release_notes/util.py
```diff
@@ -472,12 +426,14 @@ def extract_release_notes(
...
 
-def release_note_block(
+def create_release_note_block(
     code_block: dict,
     reference_type: ReferenceType,
     user_login: str,
-    cn_current_repo: ComponentName,
-    reference_id: str=None
+    current_component: gci.componentmodel.Component,
+    source_component: gci.componentmodel.Component = None,
```
ccwienk, line 434: why do we now need the source-component, wheareas it was apparently not required before?

github/release_notes/util.py
```diff
@@ -472,12 +426,14 @@ def extract_release_notes(
...
+    reference_id: str=None,
+    cache_dir: str=None,
```
ccwienk, line 436: rather name it component_descriptor_cache_dir?
also, I tend to dislike that this needs to be passed all over the place

github/release_notes/util.py
```diff
@@ -486,14 +442,31 @@ def release_note_block(
...
+    if source_component:
+        reference_id = code_block.get('reference_id')
+        reference_type = reference_type_for_type_identifier(code_block.get('reference_type'))
+        user_login = code_block.get('user')
+    elif source_repo:
+        try:
+            # try to fetch cd for the parsed source repo. The actual version does not matter,
+            # we're only interested in the components GithubAccess (we assume it does not
+            # change).
+            source_cd = product.v2.greatest_component_version_by_name(
```
ccwienk, line 455: in general, it is preferrable to use `cnudie.retrieve`, as this delegates to delivery-svc, if possible (and also does in-memory caching, which removes the need to handling a cache-dir in our (transient) use-case)

github/release_notes/util.py
```diff
@@ -486,14 +442,31 @@ def release_note_block(
...
+                component_name=source_repo,
+                ctx_repo_base_url=current_component.current_repository_ctx().baseUrl,
+                cache_dir=cache_dir,
+            )
+            source_component = source_cd.component
+        except requests.exceptions.HTTPError:
+            print(f'Unable to retrieve component descriptor for source repository {source_repo}')
```
ccwienk, line 462: logger.warning / ci.util.warning?
also: is it really a good idea to let this pass (almost) silently (I mean, who looks into those logs..)

github/release_notes/util.py
```diff
@@ -486,14 +442,31 @@ def release_note_block(
...
+            return None
+
         reference_type = reference_type_for_type_identifier(code_block.get('reference_type'))
         reference_id = code_block.get('reference_id')
         user_login = code_block.get('user')
     else:
-        source_repo = cn_current_repo.name()
-        reference_type = reference_type
-        reference_id = reference_id
+        source_component = current_component
```
ccwienk, line 469: source is used quite often in this piece of code. what is the relation between source and current?

github/release_notes/util.py
```diff
@@ -502,10 +475,51 @@ def release_note_block(
...
-            source_repo=source_repo,
-            cn_current_repo=cn_current_repo
+            source_component=source_component,
+            current_component=current_component,
         )
 
 
+def commits_from_logs(
+    git_logs: typing.List[str]
+) -> typing.List[Commit]:
```
ccwienk, line 485: I suppose you moved this down from above?

ccwienk (changes requested):

cnudie/util.py
```diff
@@ -292,3 +294,29 @@ def enumerate_group_pairs(
             _add_if_not_duplicate(resource_diff.resource_refs_only_right, i)
 
     return resource_diff
+
+
+CTF_CD_FILENAME = 'component-descriptor.yaml'
```
ccwienk, line 299: not sure whether we should expose this that publicly, just yet

cnudie/util.py
```diff
@@ -292,3 +294,29 @@ def enumerate_group_pairs(
...
+
+
+def extract_component_descriptor_from_ctf_archive(
```
ccwienk, line 302: why not simply yield all component-descriptors? the caller may still decide to not allow / bail out if more than one cd is found. there is no reason not to support more than one cd in this function, methinks

ccwienk: naming-nit: I would suggest to simply omit the verb (extract), as it does not really add much semantics to the name

cnudie/util.py
```diff
@@ -292,3 +294,29 @@ def enumerate_group_pairs(
...
+    path_to_ctf_tarfile: str
```
ccwienk, line 303: I think you should call this ctf_path, and allow the path to also be a directory, and ideally also a filelike-object

cnudie/util.py
```diff
@@ -292,3 +294,29 @@ def enumerate_group_pairs(
...
+) -> cm.ComponentDescriptor:
+    '''Extract the component descriptor from a CTF-tarfile
```
ccwienk, line 305: also, it is not essential for the contract of this function that it "extracts" the component-descriptor. what is essential is that it returns it to the caller

cnudie/util.py
```diff
@@ -292,3 +294,29 @@ def enumerate_group_pairs(
...
+
+    Only supports simple CTF-tarfiles containing one component at the moment.
+    '''
+    with tarfile.open(path_to_ctf_tarfile) as ctf_tar:
```
ccwienk, line 309: as mentioned, please also support a directory + filelike-object. to support streaming, open it for streaming (mode='r|')

cnudie/util.py
```diff
@@ -292,3 +294,29 @@ def enumerate_group_pairs(
...
+        members = ctf_tar.getmembers()
```
ccwienk, line 310: simply yield all entries. also, getmembers() will consume the entire stream (in case of streaming) / read through the entire file

cnudie/util.py
```diff
@@ -292,3 +294,29 @@ def enumerate_group_pairs(
...
+        if len(members) > 1:
+            raise RuntimeError('Only CTF archives with one component are supported')
+        component_tar = members[0]
+
+        # manually check whether member is a file to be able to generate a more expressive
+        # error-msg
+        if not component_tar.isfile():
```
ccwienk, line 317: did we define ctc-archive(-tarfiles) to only contain files? if so, this check is of course valid. if not, we might want to skip entries that are not files

cnudie/util.py
```diff
@@ -292,3 +294,29 @@ def enumerate_group_pairs(
...
+            raise RuntimeError('Content of the CTF archive is not a file')
+
+        with tarfile.open(fileobj=ctf_tar.extractfile(component_tar)) as component_tar:
+            cd_dict = yaml.safe_load(component_tar.extractfile(CTF_CD_FILENAME))
```
ccwienk, line 321: nit: ci.util.load_yaml?
also, you might prefer to stream through the entries in case extractfile does not work when streaming

ccwienk (changes requested):
I would suggest to mv the ctf-implementation to another PR (since it is a lot smaller, we should be able to merge this one rather soon, thus also reducing the size of this (rather large) PR.

ccwienk (approved):

PR #563, opened 2021-02-08
Configure root logger in cli.py

zkdev:
https://github.com/gardener/cc-utils/commit/7ba48029c59f0a5623ba7e6908bdc6c9015043da configures the root logger for cc pipelines, this PR does the same for `cli.py` and ensures the root logger config is clear before applying custom configuration. Additionally the logger usage for the whitesource-client is updated and calls to the deprecated `ctx.configure_default_logging()` function are removed. Finally an unused `whd` import in `whdutil` is removed.

Fixes #

**Release note**:

```improvement operator

```

### DIFF
--- a/checkmarx/project.py
+++ b/checkmarx/project.py
@@ -1,13 +1,11 @@
 import logging
 import time
 
-import ctx
 import checkmarx.client
 import checkmarx.model as model
 import checkmarx.util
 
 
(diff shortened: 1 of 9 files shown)


ccwienk (approved):
lgtm, fix linter-issue(s)

PR #564, opened 2021-02-09
Update WHD logging usage and introduce `force` flag

zkdev:
The root logger configuration is loaded, a separate handler for file output initialized and attached to the `whd` logger. Additionally a `ci.log.get_default_fmt_string()` function returning the default logging format string is introduced. Logging configuration is moved from production only to top level of WHD initialization.

Apart from the WHD logging usage the root logger config now includes the `force` parameter which indicates whether all existing handlers shall be removed. Additionally the Falcon app is now initialized using `falcon.App` as `falcon.Api` is deprecated (see: https://github.com/falconry/falcon/blob/a5e72b287efb2b3da632cf6547ed3f07d8ec5380/falcon/app.py#L1060).

Fixes #

**Release note**:

```improvement operator

```

### DIFF
--- a/ci/log.py
+++ b/ci/log.py
@@ -46,14 +46,14 @@ class Bcolors:
 
 def configure_default_logging(
     stdout_level=None,
-    force=False,
+    force=True,
 ):
     if not stdout_level:
         stdout_level = logging.INFO
 
     # make sure to have a clean root logger (in case setup is called multiple times)
     if force:
-        for h in logging.root.handlers[:]:
+        for h in logging.root.handlers:
             logging.root.removeHandler(h)
             h.close()
 
@@ -63,7 +63,7 @@ def configure_default_logging(
     logging.root.addHandler(hdlr=sh)
     logging.root.setLevel(level=stdout_level)
 
-    # configure loggers
+    # both too verbose ...
     logging.getLogger('github3').setLevel(logging.WARNING)
     logging.getLogger('elasticsearch').setLevel(logging.WARNING)
 
--- a/ci/log.py
+++ b/ci/log.py
@@ -44,19 +44,29 @@ class Bcolors:
     BLUE = '\033[34m'
 
 
-def configure_default_logging(stdout_level=None):
+def configure_default_logging(
+    stdout_level=None,
+    force=False,
+):
     if not stdout_level:
         stdout_level = logging.INFO
 
     # make sure to have a clean root logger (in case setup is called multiple times)
-    logging.root.handlers.clear()
+    if force:
+        for h in logging.root.handlers[:]:
+            logging.root.removeHandler(h)
+            h.close()
 
     sh = logging.StreamHandler(stream=sys.stdout)
     sh.setLevel(stdout_level)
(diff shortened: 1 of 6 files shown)


ci/log.py
```diff
@@ -44,19 +44,29 @@ class Bcolors:
...
 
     sh = logging.StreamHandler(stream=sys.stdout)
     sh.setLevel(stdout_level)
-    sh.setFormatter(CCFormatter(fmt='%(asctime)s [%(levelprefix)s] %(name)s: %(message)s'))
+    sh.setFormatter(CCFormatter(fmt=get_default_fmt_string()))
     logging.root.addHandler(hdlr=sh)
     logging.root.setLevel(level=stdout_level)
 
-    # both too verbose ...
+    # configure loggers
```
jkrayl, line 66: I think it is better to describe why it is necessary to configure the loggers. I don't need that comment to see that the log level is set here.

jkrayl (changes requested):

ci/log.py
```diff
@@ -44,19 +44,29 @@ class Bcolors:
...
+    stdout_level=None,
+    force=False,
+):
     if not stdout_level:
         stdout_level = logging.INFO
 
     # make sure to have a clean root logger (in case setup is called multiple times)
-    logging.root.handlers.clear()
+    if force:
+        for h in logging.root.handlers[:]:
```
ccwienk, line 56: why not simply use `handlers.clear()`? also, you can omit the slicing `[:]`
edit: ok, I suppose you want to not only rm the handlers, but also close them - correct?
still, no slicing operator required

ci/log.py
ccwienk: yes - admittedly, my previous comment was not very elaborated, but it did give some hints about the motivation for the cfg-change..

ci/log.py
```diff
@@ -44,19 +44,29 @@ class Bcolors:
     BLUE = '\033[34m'
 
 
-def configure_default_logging(stdout_level=None):
+def configure_default_logging(
+    stdout_level=None,
+    force=False,
```
ccwienk, line 49: is it not rather the case we do by default want the force?

cli/gardener_ci/whdutil.py
```diff
@@ -50,6 +52,7 @@ def serve():
             p.start()
         serve()
     else:
+
```
ccwienk, line 55: why?

ccwienk (changes requested):
in general, I would prefer not to mix refactorings/reformattings w/ the actual "payload" (in thase case, logger-cfg)
at any rate, see inline comments. overall lgtm

ci/log.py
jkrayl: This is the way the python logging lib does it and the `logging.root.removeHandler()` method includes locking. I think you are right about the slicing tho it looks a bit redundant.

zkdev: Yes, its about closing them as well.
Sure, will remove the slicing.

ci/log.py
zkdev: True.

jkrayl (approved):

PR #565, opened 2021-02-10
Build kaniko-image from scratch using pypi

AndreasBurger:
Instead of using our job-image and installing kaniko within, use the job-image-base and install the latest version of our coding known to pypi.
This should prevent our kaniko-image from lagging behind one release version due to being based on the previous job-image.


Dockerfile.kaniko
```diff
@@ -1,3 +1,16 @@
...
+
+RUN pip3 install --upgrade \
+  pip \
+  wheel \
+&& pip3 install --upgrade \
+  --find-links /cc/utils/dist \
+  gardener-cicd-libs \
+  gardener-cicd-cli \
+  gardener-cicd-whd \
+  gardenlinux \
```
ccwienk, line 12: we do not need to refresh gardenlinux, whd and cli (just the libs)

PR #566, opened 2021-02-10
Support oAuth in Protecode Client

AndreasBurger:
Switches our Protecode client to use bearer-auth _only_.

Not sure if that's the right approach currently. Also, WIP

### DIFF
--- a/ccc/protecode.py
+++ b/ccc/protecode.py
@@ -5,7 +5,6 @@ import protecode.client
 
 def client(
     protecode_cfg: model.protecode.ProtecodeConfig,
-    parallel_jobs=12,
 ):
     if isinstance(protecode_cfg, str):
         import ci.util
@@ -15,8 +14,7 @@ def client(
     routes = protecode.client.ProtecodeApiRoutes(base_url=protecode_cfg.api_url())
     api = protecode.client.ProtecodeApi(
         api_routes=routes,
-        basic_credentials=protecode_cfg.credentials(),
-        tls_verify=protecode_cfg.tls_verify(),
+        protecode_cfg=protecode_cfg,
     )
     return api
 
--- a/model/protecode.py
+++ b/model/protecode.py
@@ -12,11 +12,19 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import enum
+import typing
 
 from model.base import (
     BasicCredentials,
     NamedModelElement,
+    TokenCredentials,
 )
+
+
+class ProtecodeAuthScheme(enum.Enum):
+    BASIC_AUTH = 'basic_auth'
+    BEARER_TOKEN = 'bearer_token'
 
 
 class ProtecodeConfig(NamedModelElement):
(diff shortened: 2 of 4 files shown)


ccc/protecode.py
```diff
@@ -5,7 +5,7 @@
 
 def client(
     protecode_cfg: model.protecode.ProtecodeConfig,
-    parallel_jobs=12,
+    parallel_jobs=12,  # FIXME: This is currently not used.
```
ccwienk, line 8: well, then rm it? :-)

model/protecode.py
```diff
@@ -12,13 +12,19 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import enum
 
 from model.base import (
-    BasicCredentials,
+    BasicCredentials, ModelValidationError,
```
ccwienk, line 18: nit: we only write one import per line, I think (see pep-8..)

model/protecode.py
```diff
@@ -12,13 +12,19 @@
...
     NamedModelElement,
 )
 
 
+class ProtecodeAuthScheme(enum.Enum):
+    BASIC_AUTH = 'basic_auth'
+    OAUTH = 'oauth'
```
ccwienk, line 25: nit: are you sure that the (new) auth-scheme we use against protecode is actually oauth? I think, oauth would include the actual token-retrieval (which I understood from our last discussion we do upfront); so maybe `token` or `token_auth`, or `bearer_token` might be more adequate names

model/protecode.py
```diff
@@ -35,4 +41,34 @@ def tls_verify(self):
 
 
 class ProtecodeCredentials(BasicCredentials):
-    pass
+    def auth_scheme(self):
```
ccwienk, line 44: nit: return-type-hint. also, the base-class is a bit misleading now, is it not?
-> rather define two credentials-classes (one for basic-auth, one for token-auth), check for configured auth-type, and return the appropriate class, rather than mixing those two together

model/protecode.py
```diff
@@ -35,4 +41,34 @@ def tls_verify(self):
...
+    def validate(self):
+        if (auth_scheme := self.auth_scheme()) is ProtecodeAuthScheme.BASIC_AUTH:
+            if not (self.passwd() and self.username()):
+                raise ModelValidationError(
+                    'If basic auth is configured, both username and password are required'
+                )
+        elif auth_scheme is ProtecodeAuthScheme.OAUTH:
+            if not self.token():
+                raise ModelValidationError(
+                    'If oAuth is configured, a token is required'
```
ccwienk, line 59: as indicated above. please do not mix those two credential-types into one class. (define one class per credential
consider defining the auth-type credential-class in the base.py-module, if this makes sense, but it is also OK for me to define it here.)

model/protecode.py
```diff
@@ -35,4 +41,34 @@ def tls_verify(self):
...
+                )
+        else:
+            raise NotImplementedError(auth_scheme)
+
+    def _required_attributes(self):
+        return ()
```
ccwienk, line 65: can be omitted, if it is the empty tuple

protecode/client.py
```diff
@@ -95,10 +96,24 @@ def scans(self, product_id: int):
...
+        api_routes,
+        credentials,
+        tls_verify=True,
+    ):
         self._routes = not_none(api_routes)
-        self._credentials = not_none(basic_credentials)
-        self._auth = (basic_credentials.username(), basic_credentials.passwd())
+        self._credentials = not_none(credentials)
+
+        # if self._auth is ProtecodeClientAuth.BEARER_TOKEN:
```
ccwienk, line 108: please do not check-in commented-out code-blocks, unless there is a good reason to do so

protecode/client.py
```diff
@@ -138,14 +153,22 @@ def _request(self, method, *args, **kwargs):
         else:
             cookies = None
 
-        auth = self._auth
+        method_kwargs = {
+            'verify': self._tls_verify,
```
ccwienk, line 157: if I see this properly, `verify`, `cookies`, and `headers` are always passed, and the only dynamic kwarg would be `auth`.

since this pattern of passing dynamic kwargs decreases readability, please do not do this unless actually required.

note that it is also not necessarily required to manually set the `Authorization`-header; requests also supports this (I think). But it is also ok for me to do this manually

protecode/client.py
```diff
@@ -336,6 +359,10 @@ def add_triage_raw(
     # --- "rest" routes (undocumented API)
 
     def login(self):
+
+        if self._credentials.auth_scheme is not ProtecodeAuthScheme.BASIC_AUTH:
```
ccwienk, line 363: nit: please check for all auth-types + add notimplemented in case we encounted a yet unknown one

ccwienk (changes requested):

protecode/client.py
```diff
@@ -138,15 +147,26 @@ def _request(self, method, *args, **kwargs):
         else:
             cookies = None
 
-        auth = self._auth
+        if isinstance(self._credentials, TokenCredentials):
```
ccwienk, line 150: meh... actually, I would have preferred to check against the schema-enum

protecode/client.py
```diff
@@ -138,15 +147,26 @@ def _request(self, method, *args, **kwargs):
...
+            headers['Authorization'] = f"Bearer {self._credentials.token()}"
+            return partial(
```
ccwienk, line 152: I do not get it - why use currying here at all?

ccwienk (changes requested):
remove (unnecessary) currying, please. this is unnecessarily cumbersome

ccwienk:
@AndreasBurger : please fix linter-err

PR #570, opened 2021-02-17
add diff_label algorithm

jkrayl:


cnudie/util.py
```diff
@@ -42,6 +42,13 @@ def determine_main_source_for_component(
     return component.sources[0]
 
 
+@dataclasses.dataclass
+class LabelDiff:
+    labels_only_left: typing.List[cm.Label] = dataclasses.field(default_factory=list)
+    labels_only_right: typing.List[cm.Label] = dataclasses.field(default_factory=list)
+    label_pairs_changed: typing.List[typing.Tuple[cm.Label, cm.Label]] = dataclasses.field(default_factory=list) # noqa:E501
```
ccwienk, line 49: stylistic nit: you might define a (reusable) module-variable e.g. `empty_list`, thus also achieving shorter lines

cnudie/util.py
```diff
@@ -159,6 +200,23 @@ def find_changed_component(
     )
 
 
+def _enumerate_group_pairs(
+    left_artifacts: typing.List[typing.Union[cm.Resource, cm.ComponentSource, cm.Label]],
```
ccwienk, line 204: nit: we only need `typing.Sequence`, I think (not that important..)

cnudie/util.py
```diff
@@ -159,6 +200,23 @@ def find_changed_component(
...
+    right_artifacts: typing.List[typing.Union[cm.Resource, cm.ComponentSource, cm.Label]],
+) -> typing.Tuple[typing.List[cm.Resource], typing.List[cm.Resource]]:
```
ccwienk, line 206: since you return a generator, the typehint is not correct

cnudie/util.py
```diff
@@ -159,6 +200,23 @@ def find_changed_component(
...
+    # group the resources with the same name on both sides
+    for artifact in left_artifacts:
+        right_resource_group = [r for r in right_artifacts if r.name == artifact.name]
+        # get resources for one resource via name
```
ccwienk, line 210: just to check I understand the semantics properly:

- we accept two sequences of name-bearing objects ("left" and "right")
- both sequences may contain elements w/ the same `name`
- not all names necessarily have to exist in both sequences
- if a name does not exist in the right group (but does in the left), it is dropped from the result-generator
- otherwise, a two-tuple of equally-named elements on left and right is yielded

ccwienk: -> is this asymmetry intended? one might also consider to drop elements that do not exist on both sides, as opposed to only dropping those that only exist on the left side

cnudie/util.py
```diff
@@ -101,6 +108,40 @@ def diff_component_descriptors(
...
+    ):
+        if len(left_group) == 1 and len(right_group) == 1:
+            if left_group[0].value == right_group[0].value:
+                continue
+            else:
+                label_diff.label_pairs_changed.append(
+                    (left_group[0], right_group[0]),
+                )
+        else:
+            raise ValueError('only one label with the same name is allowed')
```
ccwienk, line 140: would it maybe make sense to have the `enumerate_group_pairs`-function check for the maximum-allowed labels?

or even better, we should add such a check into the component-model (invalid component-descriptors should not even be allowed to be published in the first place).

for the special-case of unique names, it might be adequate to have a special codepath that will not return two-tuples of lists, but simple two-tuples of elements.

ccwienk (changes requested):

PR #571, opened 2021-02-24
Ignore dev component versions for update pull-request

schrodit:

Ignores dev versions for update pull requests.
Release versions are identified by the string `-dev-`.

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/version.py
+++ b/version.py
@@ -211,7 +211,7 @@ def process_version(
     return processed_version
 
 
-def find_latest_version(versions) -> str:
+def find_latest_version(versions, ignore_dev_versions=False) -> str:
     latest_candidate = None
     latest_candidate_str = None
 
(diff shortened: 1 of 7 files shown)


ccwienk:
I would prefer to make this configurable via pipeline-definition

version.py
```diff
@@ -211,11 +211,13 @@ def process_version(
     return processed_version
 
 
-def find_latest_version(versions) -> str:
+def find_latest_version(versions, ignore_dev_versions=False) -> str:
```
ccwienk, line 214: I think we should rather parse to semver (there is a util function for that in version.py), and check the parsing result for being "final"

ccwienk (changes requested):

version.py
schrodit: but we should still have the option to configure if dev versions should be ignored?

schrodit: should it be also possible to ignore dev versions in the other helper function like `find_latest_version_with_matching_minor` ?

AndreasBurger:
Closed in favor of #581 (contains everything here plus some small changes from me)

PR #573, opened 2021-02-24
add support for ctf in draft release notes

schrodit:

added support for ctf component descriptors in the draft release notes as it is already supported in the release notes.

Fixes #

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/concourse/steps/draft_release.mako
+++ b/concourse/steps/draft_release.mako
@@ -6,6 +6,7 @@ from makoutil import indent_func
 import os
 import concourse.steps.component_descriptor_util as cdu
 import gci.componentmodel
+import product.v2
 version_file = job_step.input('version_path') + '/version'
 repo = job_variant.main_repository()
 draft_release_trait = job_variant.trait('draft_release')
@@ -52,25 +53,13 @@ processed_version = version.process_version(
 
 repo_dir = ci.util.existing_dir('${repo.resource_name()}')
 
-have_ctf = os.path.exists('${ctf_path}')
-have_cd = os.path.exists('${component_descriptor_v2_path}')
-if not have_ctf ^ have_cd:
-    ci.util.fail('exactly one of component-descriptor, or ctf-archive must exist')
-elif have_cd:
-    component_descriptor_v2 = cm.ComponentDescriptor.from_dict(
-        ci.util.parse_yaml_file('${component_descriptor_v2_path}'),
-    )
-elif have_ctf:
-    component_descriptors = list(cnudie.util.component_descriptors_from_ctf_archive(
-        '${ctf_path}',
-    ))
-    if not component_descriptors:
-        ci.util.fail(f'No component descriptor found in CTF archive at ${ctf_path})
-    if len(component_descriptors) > 1:
-        ci.util.fail(
-            f'More than one component descriptor found in CTF archive at ${ctf_path}'
-        )
-    component_descriptor_v2 = component_descriptors[0]
+
+component_descriptor_v2 = cnudie.util.determine_main_component(
+    repository_hostname='${repo.repo_hostname()}',
+    repository_path='${repo.repo_path()}',
+    component_descriptor_v2_path='${component_descriptor_v2_path}',
+    ctf_path='${ctf_path}',
+)
 
 github_cfg = ccc.github.github_cfg_for_hostname('${repo.repo_hostname()}')
 
--- a/concourse/steps/draft_release.mako
+++ b/concourse/steps/draft_release.mako
@@ -14,12 +14,17 @@ component_descriptor_v2_path = os.path.join(
     job_step.input('component_descriptor_dir'),
     cdu.component_descriptor_fname(gci.componentmodel.SchemaVersion.V2),
 )
+ctf_path = os.path.join(
+  job_step.input('component_descriptor_dir'),
+  product.v2.CTF_OUT_DIR_NAME,
+)
 %>
 import version
 import pathlib
 
 import ci.util
 import ccc.github
+import cnudie.util
 import gci.componentmodel as cm
 
 from gitutil import GitHelper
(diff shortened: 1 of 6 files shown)


concourse/steps/draft_release.mako
```diff
@@ -14,13 +14,18 @@ component_descriptor_v2_path = os.path.join(
...
+  product.v2.CTF_OUT_DIR_NAME,
+)
 %>
 import version
 import pathlib
 
 import ci.util
 import ccc.github
 import gci.componentmodel as cm
+import cnudie.util
```
ccwienk, line 28: nit: according to pep8, imports should be ordered alphabetically

concourse/steps/draft_release.mako
```diff
@@ -46,9 +51,26 @@ processed_version = version.process_version(
...
+    component_descriptor_v2 = cm.ComponentDescriptor.from_dict(
+        ci.util.parse_yaml_file('${component_descriptor_v2_path}'),
+    )
+elif have_ctf:
+    component_descriptors = list(cnudie.util.component_descriptors_from_ctf_archive(
+        '${ctf_path}',
+    ))
+    if not component_descriptors:
+        ci.util.fail(f'No component descriptor found in CTF archive at ${ctf_path})
+    if len(component_descriptors) > 1:
```
ccwienk, line 69: can we not handle more than one?

schrodit: currently we also cannot handle more the one in the release trait.
I would keep it like this and open another PR with the support for multiple component descriptors in the release and draft release PR?

concourse/steps/draft_release.mako
```diff
@@ -46,9 +52,16 @@ processed_version = version.process_version(
...
+
+try:
+    component_descriptor_v2 = cnudie.util.determine_main_component(
+        repository_hostname='${repo.repo_hostname()}',
+        repository_path='${repo.repo_path()}',
+        component_descriptor_v2_path='${component_descriptor_v2_path}',
+        ctf_path='${ctf_path}',
+    )
+except ValueError as err:
+    ci.util.fail(str(err))
```
ccwienk, line 64: nit: since you fail anyway, you might omit the try-except (printing the stacktrace helps a lot w/ debugging)

concourse/steps/release.py
```diff
@@ -576,21 +576,25 @@ def name(self):
...
                 ci.util.parse_yaml_file(self.component_descriptor_v2_path),
-            )
-            # resolve all referenced components (thus ensure all dependencies are available)
-            tuple(cnudie.retrieve.components(component=component_descriptor_v2))
-            # TODO: more validation (e.g. check for uniqueness of names)
+            )]
         elif have_ctf:
-            # nothing to do, already uploaded in component_descriptor step.
-            pass
+            self.components = list(component_descriptors_from_ctf_archive(
```
ccwienk, line 590: nit: since it will not be mutated, `tuple` should be preferred (imo)

concourse/steps/release.py
```diff
@@ -576,21 +576,25 @@ def name(self):
...
+                self.ctf_path,
+            ))
+            if not component_descriptors:
+                ci.util.fail(f'No component descriptor found in CTF archive at {self.ctf_path}')
```
ccwienk, line 594: nit: `{self.ctf_path=}` will also print the variable-name (helpful for debugging)

concourse/steps/release.py
```diff
@@ -576,21 +576,25 @@ def name(self):
...
+
+        for component_descriptor_v2 in self.components:
+            tuple(cnudie.retrieve.components(component=component_descriptor_v2))
```
ccwienk, line 597: nit: since you merely want to exhaust the generator (at least this is what I assume), a nested `for` might be a bit more explicit

test/cnudie/cnudie_util_test.py
```diff
@@ -14,6 +14,12 @@
 # limitations under the License.
 
 import pytest
+import unittest
```
ccwienk, line 17: nit: sort alphabetically, please (pep-8)

test/cnudie/cnudie_util_test.py
```diff
@@ -33,7 +39,56 @@ def cid():
...
+def create_ctf(
+    output_filename: str,
+    component_descriptors: typing.List[cm.ComponentDescriptor],
+):
+    tempdir = tempfile.TemporaryDirectory()
+
+    with tarfile.open(output_filename, 'w') as ctf_tar:
+
+        for component_descriptor in component_descriptors:
+            cd_filename = component_descriptor.component.name.replace("/", "_")
```
ccwienk, line 70: nit: consistently use single or double-quotes (single-quotes are preferred in cc-utils)

test/cnudie/cnudie_util_test.py
```diff
@@ -33,7 +39,56 @@ def cid():
...
+            cd_temp_file = os.path.join(tempdir.name, cd_filename)
+
+            comp_bytes = io.StringIO()
+            component_descriptor.to_fobj(comp_bytes)
```
ccwienk, line 74: can this also be deduplicated w/ util function from gci.oci?

test/cnudie/cnudie_util_test.py
```diff
@@ -33,7 +39,56 @@ def cid():
...
+                )
+                comp_bytes.close()
+                tar.close()
+
+            ctf_tar.add(
+                name=cd_temp_file,
+            )
+
+        ctf_tar.close()
+    tempdir.cleanup()
```
ccwienk, line 91: you might want to use `with tempfile......` instead

test/cnudie/cnudie_util_test.py
```diff
@@ -338,3 +393,141 @@ def test_diff_label():
...
+    assert component_name == 'example.com/comp_a'
+
+    component_name = cnudie.util.determine_component_name(
+        repository_hostname='example.com',
+        repository_path='Comp_A',
+    )
+    assert component_name == 'example.com/comp_a'
+
+
+class DetermineMainComponentTest(unittest.TestCase):
```
ccwienk, line 412: I would prefer pytest-based tests; tmpdir might be injected via fixtures (https://docs.pytest.org/en/stable/fixture.html)

.. we are doing python, and not Java :-)
(I know, there is a legacy-base of junit-style tests in cc-utils. however, I migrate those to pytest, and write new tests excusively based on pytest..)

ccwienk (changes requested):
overall lgtm - some stylistic nits (see inline comments)

test/cnudie/cnudie_util_test.py
schrodit: forgot to add our changes yesterday

concourse/steps/release.py
schrodit: not sure about the nested for.
From my understanding the code just checks if all components can be resolved. so we do not even need a generator?

test/cnudie/cnudie_util_test.py
```diff
@@ -338,3 +383,134 @@ def test_diff_label():
...
+    assert component_name == 'example.com/comp_a'
+
+    component_name = cnudie.util.determine_component_name(
+        repository_hostname='example.com',
+        repository_path='Comp_A',
+    )
+    assert component_name == 'example.com/comp_a'
+
+
+@pytest.fixture
```
ccwienk, line 402: how about https://docs.pytest.org/en/stable/tmpdir.html#the-tmpdir-fixture? :-)

ccwienk (commented):
one last nit - otherwise lgtm

concourse/steps/release.py
ccwienk: the generator is returned from the `cnudie.retrieve.components` function (implicitly, because its definition contains `yield` stmts).

generators are evaluated lazily. that means, that if they are not consumed, no code will be executed at all. creating a `tuple` is a valid means to for the generator to be consumed (so your code is perfectly correct).

however, this has two disadvantages:

1. creation of the tuple will require memory (and thus destroys the potential benefit from the generator)
2. it may not be obvious to other readers that the tuple is created for the sole purpose of exhausting the generator

thus I think it might be a better "pattern" to use a nested for-loop, like so:

ccwienk (approved):

PR #579, opened 2021-03-15
refactor concourse-api-client

ccwienk:


concourse/client/api.py
```diff
@@ -18,6 +18,7 @@
 import typing
 
 from ensure import ensure_annotations
+import requests
```
AndreasBurger, line 21: nit: should come before `from ...`, looking at the other imports

AndreasBurger (approved):
lgtm

concourse/client/api.py
ccwienk: done using `importanize`

PR #580, opened 2021-03-18
Replace common stack-traces in WHD with proper log messages

AndreasBurger:


whd/dispatcher.py
```diff
@@ -279,18 +298,28 @@ def _set_pr_labels(self, pr_event, resources) -> bool:
             for resource in resources if resource.source.get('label') is not None
         }
         if not required_labels:
-            return False
+            return True
```
ccwienk, line 301: what's the semantics of the returned boolean value (and why was the previous value wrong)?

whd/dispatcher.py
```diff
@@ -279,18 +298,28 @@ def _set_pr_labels(self, pr_event, resources) -> bool:
...
         github_cfg = self.cfg_set.github()
         owner, name = repository_path.split('/')
-        github_helper = GitHubRepositoryHelper(
-            owner,
-            name,
-            github_cfg=github_cfg,
-        )
+
+        try:
+            github_helper = GitHubRepositoryHelper(
```
ccwienk, line 311: might be a candidate for a helper-function, wdyt?

ccwienk (approved):

whd/dispatcher.py
AndreasBurger: (already discussed elsewhere)
according to the docstring, this function is to return true if "if the required label was set"

PR #583, opened 2021-04-07
add secret_key model

jkrayl:


model/secret_key.py
```diff
@@ -0,0 +1,29 @@
...
+from model.base import (
+    NamedModelElement,
+)
+
+
+class SecretKey(NamedModelElement):
+    def key(self):
+        return self.raw.get('key')
+
+    def iv(self):
```
AndreasBurger, line 25: I'd prefer
```suggestion
    def initialisation_vector(self):
```
For the actual key I'm fine with `iv`, although consistency is always nice.

AndreasBurger (approved):

ccwienk (approved):

PR #585, opened 2021-04-08
add decryption to secrets server client

jkrayl:
depends https://github.com/gardener/cc-utils/pull/583 and https://github.wdf.sap.corp/kubernetes/cc-config/pull/132

### DIFF
--- a/ccc/secrets_server.py
+++ b/ccc/secrets_server.py
@@ -13,13 +13,30 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import base64
 import json
 import os
 import socket
+import typing
 
+import Crypto.Util.Padding
 import requests
 
 from ci.util import urljoin
+import model.secret
+
+
+def get_secret_cfg_from_env_if_available(
+    key_env_var='SECRET_KEY',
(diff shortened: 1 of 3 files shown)


ccc/secrets_server.py
```diff
@@ -39,38 +56,61 @@ def from_env(
...
-                cache_file=cache_file
+            endpoint_url=os.environ.get(endpoint_env_var),
+            concourse_secret_name=os.environ.get(concourse_secret_env_var),
+            cache_file=cache_file,
+            secret=secret,
         )
 
     @staticmethod
-    def default():
+    def default(encryption:bool=False):
```
ccwienk, line 69: hmmm..... actually, I would prefer to already prepare the incoming change to multiple secrets

--> please factor-out the calculation of the secrets-path and add this path as an argument (of course, keep previous behaviour, i.e.: if no path is given, fallback to default path

ccwienk (changes requested):
let's discuss a better option for the boolean flag (see inline comment)

also, please squash fixes for linter-errors

ccwienk (commented):
as just discussed: rm boolean flag -> use env-var only

ccwienk (approved):

PR #600, opened 2021-05-14
make cfg_factory serialiseable

jkrayl:


model/__init__.py
```diff
@@ -414,6 +414,16 @@ def __getattr__(self, cfg_type_name):
 
         return functools.partial(self._cfg_element, cfg_type_name)
 
+    def serialise(self):
```
ccwienk, line 417: rather prefix the method-name w/ an underscore to avoid potential conflicts w/ factory-method-names

PR #601, opened 2021-05-17
Fix breaking websockets 9.X change

zkdev:
websockets 9.X moved some modules to `legacy` sub-module even though they keep supporting them.
This PR adjusts the corresponding imports.
See: https://websockets.readthedocs.io/en/stable/changelog.html#id5

Fixes #

**Release note**:

```improvement operator

```


zkdev:
As the `websockets.client` import is only used for hinting, its replaced with a string literal.

PR #604, opened 2021-05-20
add job mapping to default template

jkrayl:


model/secrets_server.py
```diff
@@ -73,3 +73,15 @@ def concourse_cfg_name(self):
 
     def encrypted_concourse_cfg_name(self):
         return f'{self.encrypted_concourse_secret_name()}/{self.encrypted_concourse_attribute()}'
+
+    def secret_url_path(self, job_mapping=None, secret_cfg=None):
+        '''
+            used to set the retrieve the secret url path for given config in default template
```
AndreasBurger, line 79:
```suggestion
            used to retrieve the secret url path for given config in default template
```

model/secrets_server.py
```diff
@@ -73,3 +73,15 @@ def concourse_cfg_name(self):
 
     def encrypted_concourse_cfg_name(self):
         return f'{self.encrypted_concourse_secret_name()}/{self.encrypted_concourse_attribute()}'
+
+    def secret_url_path(self, job_mapping=None, secret_cfg=None):
```
AndreasBurger, line 77: the default for `job_mapping` seems strange to me:
- if secret_cfg is given, `None.secrets_repo()` will be accessed
- if secret_cfg is not given, `job_mapping` is not needed.

Also, a valid `secret_cfg` can effectively be anything truthy. Maybe something like
```
def secret_url_path(self, use_secret_cfg: bool = False, job_mapping = None):
  if use_secret_cfg and not job_mapping:
    raise RuntimeError('…')
…
```
That'd be a bit more precise in the signature and more expressive if it is used wrong.

The way this function is currently written I'd expect secrets_cfg to include the job_mapping and be accessed through it. Might be worth thinking about introducing an encapsulating type. I'll need to think about it some more.

AndreasBurger (commented):
generally lgtm

concourse/templates/default.mako
```diff
@@ -29,10 +30,10 @@ default_container_registry = config_set.container_registry()
 # expose secrets_server endpoint to all jobs
 secrets_server_cfg = config_set.secrets_server()
 
-if secret_cfg:
-  secrets_server_cc_cfg_name = secrets_server_cfg.secrets().encrypted_concourse_cfg_name()
-else:
-  secrets_server_cc_cfg_name = secrets_server_cfg.secrets().concourse_cfg_name()
+secrets_server_cc_cfg_name = secrets_server_cfg.secrets().secret_url_path(
```
AndreasBurger, line 33: late addition to the review, but I think somewhat related to my confusion:

`secrets_server_cc_cfg_name = secrets_server_cfg.secrets().secret_url_path(…)`

Is the config name actually a url? Also, is it so by design?

jkrayl: it's not a fully qualified url but a path to the secret

ccwienk: @AndreasBurger : `URL` would be the fully qualified endpoint (<schema>://<netloc>/<path>[<query>}), whereas url_path was intended to only refer to the `<path>` part

model/secrets_server.py
ccwienk: yeah, we should add typehints

ccwienk: @AndreasBurger (/cc @jkrayl ): I think the idea would be to _always_ require both cfg-elements as an input

AndreasBurger (approved):

PR #608, opened 2021-06-02
Add scan sources filter

zkdev:
This PR introduces filtering options for the `scan_sources` trait. `components`, `sources` and `resources` can be included / excluded, either with regex or the artifact type as a whole. Additionally `resources` can be matched which respect to their `resourceType`.
Sample trait:
```
scan_sources:
  filters:
  - type: component
    match:
      name: 'github.com/gardener/.*' # only include components from opensource-gardener-project
    action: 'include'
  - type: source
    match: false # reject all
    action: 'exclude'
  - type: resource
    match:
      type: ociImage # take all oci-images
    action: 'include'
```

Fixes #

**Release note**:

```improvement operator

```

### DIFF
--- a/concourse/model/traits/scan_sources.py
+++ b/concourse/model/traits/scan_sources.py
@@ -73,6 +73,31 @@ WHITESOURCE_ATTRIBUTES = (
 )
 
 
+FILTER_ATTRIBUTES = (
+    AttributeSpec.required(
+        name='type',
+        doc='defines type to apply filter on, (component|source|resource)',
+        type=str,
+    ),
+    AttributeSpec.required(
+        name='match',
+        doc='matches artifacts (list of regex|true|false)',
+        type=typing.Union[dict, bool],
+    ),
+    AttributeSpec.required(
+        name='action',
+        doc='defines action to matched artifacts (include|exclude)',
+        type=str,
+    ),
+)
+
+
+class FilterCfg(ModelBase):
+    @classmethod
+    def _attribute_specs(cls):
+        return FILTER_ATTRIBUTES
+
+
 class WhitesourceCfg(ModelBase):
     @classmethod
     def _attribute_specs(cls):
@@ -122,6 +147,12 @@ ATTRIBUTES = (
         doc='optional list of email recipients to be notified about critical scan results',
         type=typing.List[str],
     ),
+    AttributeSpec.required(
+        name='filters',
+        default=(),
+        doc='config to include and exclude sources, resources or whole components',
+        type=FilterCfg,
+    ),
     AttributeSpec.optional(
         name='checkmarx',
         type=CheckmarxCfg,
@@ -159,6 +190,10 @@ class SourceScanTrait(Trait):
     def whitesource(self):
         if whitesource := self.raw.get('whitesource'):
             return WhitesourceCfg(whitesource)
+
+    def filters(self):
+        if filters := self.raw.get('filters'):
+            return FilterCfg(filters)
 
     def transformer(self):
         return SourceScanTraitTransformer(trait=self)
--- a/concourse/model/traits/scan_sources.py
+++ b/concourse/model/traits/scan_sources.py
@@ -158,7 +158,7 @@ ATTRIBUTES = (
     ),
     AttributeSpec.required(
         name='filters',
-        default=(),
+        default={},
         doc='config to include and exclude sources, resources or whole components',
         type=FilterCfg,
     ),
--- a/concourse/model/traits/scan_sources.py
+++ b/concourse/model/traits/scan_sources.py
@@ -97,6 +97,15 @@ class FilterCfg(ModelBase):
     def _attribute_specs(cls):
         return FILTER_ATTRIBUTES
 
+    def type(self):
+        return self.raw['type']
+
+    def match(self):
+        return self.raw['match']
+
+    def action(self):
+        return self.raw['action']
+
 
 class WhitesourceCfg(ModelBase):
     @classmethod
--- a/whitesource/client.py
+++ b/whitesource/client.py
@@ -79,7 +79,7 @@ class WhitesourceClient:
 
     async def upload_to_project(
         self,
-        extra_whitesource_config: typing.Dict,
+        extra_whitesource_config: typing.Union[None, dict],
         file: typing.IO,
         project_name: str,
         length: int,
@@ -98,7 +98,7 @@ class WhitesourceClient:
             data_class=protocol.WhiteSourceApiExtensionWebsocketWSConfig,
             data={
                 'apiKey': self.api_key,
-                'extraWsConfig': extra_whitesource_config,
+                'extraWsConfig': extra_whitesource_config if extra_whitesource_config else {},
                 'productToken': self.product_token,
                 'projectName': project_name,
                 'requesterEmail': self.requester_mail,
--- a/whitesource/component.py
+++ b/whitesource/component.py
@@ -25,7 +25,7 @@ def _get_ws_label_from_artifact(source: cm.ComponentSource) -> dso.labels.Source
 
 
 def get_scan_artifacts_from_components(
-    components: typing.Generator[typing.Union[tuple, typing.Any], None, None],
+    components: typing.Generator[dso.model.ScanArtifact, None, None],
     filters: typing.List[whitesource.model.WhiteSourceFilterCfg],
 ) -> typing.Generator:
 
--- a/whitesource/component.py
+++ b/whitesource/component.py
@@ -1,4 +1,5 @@
 import logging
+import re
 import tarfile
 import typing
 
@@ -7,8 +8,8 @@ import dacite
 
 import dso.labels
 import dso.model
-
 import gci.componentmodel as cm
+import whitesource.model
 
 
 logger = logging.getLogger(__name__)
@@ -24,8 +25,12 @@ def _get_ws_label_from_artifact(source: cm.ComponentSource) -> dso.labels.Source
 
 
 def get_scan_artifacts_from_components(
-    components: typing.Sequence[cm.Component],
+    components: typing.Generator[typing.Union[tuple, typing.Any], typing.Any, typing.Any],
+    filters: typing.List[whitesource.model.WhiteSourceFilterCfg],
 ) -> typing.Generator:
+
+    components = apply_filters(components=components, filters=filters)
+
     for component in components:
         for artifact in component.sources + component.resources:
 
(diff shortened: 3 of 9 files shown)


whitesource/model.py
```diff
@@ -29,8 +28,15 @@ def max_cve(self) -> typing.Tuple[cve_name, cvss_score]:
...
     name: str
     highest_cve_name: str
     highest_cve_score: float
+
+
+@dataclass(frozen=True)
+class WhiteSourceFilterCfg:
+    type: str
+    match: typing.Union[bool, dict]
+    action: str
```
AndreasBurger, line 42: Both `type` and `action` might better be modeled as enums - wdyt?

whitesource/util.py
```diff
@@ -353,11 +357,13 @@ def print_scans(
...
-    notification_recipients: typing.List[str],
+    notification_recipients: typing.Union[None, typing.List[str]],
     cve_threshold: float,
     product_name: str,
     projects: typing.List[whitesource.model.WhiteSrcProject],
 ):
+    if not notification_recipients:
+        logger.warning('No recipients defined. No emails will be sent...')
 
     if len(notification_recipients) > 0:
```
AndreasBurger, line 368: That will not work for `None`. Also, I don't see the call being adjusted anywhere - can notification_recipients actually _be_ `None` (today)?

whitesource/component.py
```diff
@@ -84,3 +89,136 @@ def download_component(
...
+    action: str,
+    components,
+    match: bool,
+):
+    if action == 'include':
+        return components if match else ()
+    elif action == 'exclude':
+        return components if not match else ()
+    else:
+        raise RuntimeError
```
AndreasBurger, line 224: I think a `NotImplementedError(action)` would be more fitting/helpful, especially if enums are used.

whitesource/component.py
```diff
@@ -84,3 +89,136 @@ def download_component(
...
+    action: str,
+    artifacts,
+    match: bool,
+):
+    if action == 'include':
+        return artifacts if match else ()
+    elif action == 'exclude':
+        return artifacts if not match else ()
+    else:
+        raise RuntimeError
```
AndreasBurger, line 192: See below (above?) about using RuntimeError.

Also, these two functions behave pretty much the same. Consider somthing like e.g.:
```python3
def _apply_filter_action(
    action: str,
    obj: typing.Union[Foo, Bar], # !
    match: bool,
):
    if action == 'include':
        return obj if match else ()
    elif action == 'exclude':
        return obj if not match else ()
    else:
        raise RuntimeError
```

whitesource/component.py
```diff
@@ -84,3 +89,136 @@ def download_component(
...
+            raise NotImplementedError
+
+    if component_filter:
+        components = _apply_component_filter(filter=component_filter, components=components)
+
+    if source_filter:
+        components = _apply_artifact_filter(filter=source_filter, components=components)
+
+    if resource_filter:
+        components = _apply_artifact_filter(filter=resource_filter, components=components)
```
AndreasBurger, line 120: wouldn't later invocations override the results of previous ones, if more than one filter is defined?

Also, this function in general is doing this in quite a roundabout way. I'd suggest to call the appropriate function per filter type and aggregate all the results before moving on to the next filter type.

whitesource/component.py
```diff
@@ -24,8 +25,12 @@ def _get_ws_label_from_artifact(source: cm.ComponentSource) -> dso.labels.Source
 
 
 def get_scan_artifacts_from_components(
-    components: typing.Sequence[cm.Component],
+    components: typing.Generator[typing.Union[tuple, typing.Any], typing.Any, typing.Any],
```
AndreasBurger, line 28: I think `typing.Union[tuple, typing.Any], None, None` is the correct type, but please check

AndreasBurger (changes requested):

whitesource/filters.py
```diff
@@ -0,0 +1,160 @@
...
+            resource_type_name = artifact.type.value
+        elif isinstance(artifact.type, str):
+            resource_type_name = str(artifact.type)
+        else:
+            raise NotImplementedError
+
+        re_str = resource_type_name if filter_cfg.type is whitesource.model.FilterType.RESOURCE \
+            else artifact.name
+        re_pat = filter_cfg.match.get('type') if filter_cfg.type is \
+            whitesource.model.FilterType.RESOURCE else filter_cfg.match.get('name')
```
AndreasBurger, line 91:
```suggestion
        re_str = (
            resource_type_name if filter_cfg.type is whitesource.model.FilterType.RESOURCE
            else artifact.name
        )
        re_pat = (
            filter_cfg.match.get('type') if filter_cfg.type is whitesource.model.FilterType.RESOURCE
            else filter_cfg.match.get('name')
        )
```

whitesource/filters.py
```diff
@@ -0,0 +1,160 @@
...
+) -> bool:
+
+    if not artifact and not component:
+        raise NotImplementedError
+
+    if filter_cfg.action is whitesource.model.ActionType.INCLUDE:
+        if filter_cfg.match:
+            logger.debug(f'included {artifact.name if artifact else component.name},'
+                         f' reason: {filter_cfg.match=} and {filter_cfg.action=}')
+            return True
```
AndreasBurger, line 129:
```suggestion
        return filter_cfg.match
```
Not sure if the debug code is actually needed tbh

whitesource/filters.py
```diff
@@ -0,0 +1,160 @@
...
+    elif filter_cfg.action is whitesource.model.ActionType.EXCLUDE:
+        if not filter_cfg.match:
+            logger.debug(f'included {artifact.name if artifact else component.name},'
+                         f' reason: {filter_cfg.match=} and {filter_cfg.action=}')
+            return True
```
AndreasBurger, line 134:
```suggestion
        return not filter_cfg.match
```

whitesource/filters.py
```diff
@@ -0,0 +1,160 @@
...
+import gci.componentmodel as cm
+import whitesource.model
+
+
+logger = logging.getLogger(__name__)
+
+
+def _print_filter_stats(
+    components: typing.List[cm.Component],
+    filtered: bool
```
AndreasBurger, line 14:
```suggestion
    filtered: bool,
```

whitesource/component.py
```diff
@@ -24,8 +25,12 @@ def _get_ws_label_from_artifact(source: cm.ComponentSource) -> dso.labels.Source
 
 
 def get_scan_artifacts_from_components(
-    components: typing.Sequence[cm.Component],
+    components: typing.Generator[typing.Union[tuple, typing.Any], None, None],
```
AndreasBurger, line 28:
```suggestion
    components: typing.Generator[dso.model.ScanArtifact, None, None],
```
I think that's the correct annotation here

whitesource/client.py
```diff
@@ -79,7 +79,7 @@ def save_project_tag(self, projectToken: str, key: str, value: str):
 
     async def upload_to_project(
         self,
-        extra_whitesource_config: typing.Dict,
+        extra_whitesource_config: typing.Union[None, dict],
```
AndreasBurger, line 82: Didn't we agree on `typing.Optional[dict]` for these cases?

But in this case (a simple function def) I'd say `extra_whitesource_config: dict = None,` is sufficient.

In general (in our codebase at least), these hints are necessary for dataclasses to work correctly. For all other cases, they are merely hints to the developer. As they have to look up the function definition either way, I'd argue to use the "standard" way of declaring an optional variable. Even though the typehint should still technically be `typing.Optional[…]` I think readability is better with just the plain typehint+default.

Similar on other appearances of this kind of annotations on functions.

concourse/model/traits/scan_sources.py
```diff
@@ -122,6 +156,12 @@ def exclude_path_regexes(self):
         doc='optional list of email recipients to be notified about critical scan results',
         type=typing.List[str],
     ),
+    AttributeSpec.required(
+        name='filters',
+        default=(),
```
AndreasBurger, line 161:
```suggestion
        default={},
```

AndreasBurger (changes requested):

whitesource/filters.py
```diff
@@ -0,0 +1,160 @@
...
+    filters: typing.List[whitesource.model.WhiteSourceFilterCfg],
+) -> typing.Generator:
+
+    _print_filter_stats(components=components, filtered=False)
+    for f in filters:
+        if f.type is whitesource.model.FilterType.COMPONENT:
+            components = list(filter(lambda c: _component_filter(
+                filter_cfg=f,
+                component=c,
+            ), components))
```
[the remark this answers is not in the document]
ccwienk: please use list-comprehensions

whitesource/filters.py
```diff
@@ -0,0 +1,160 @@
...
+        else:
+            raise NotImplementedError(f.type)
+
+    _print_filter_stats(components=components, filtered=False)
+    yield from components
+
+
+def _apply_artifact_filter(
+    filter_cfg: whitesource.model.WhiteSourceFilterCfg,
+    components: typing.Generator[typing.Union[tuple, typing.Any], None, None],
```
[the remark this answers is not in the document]
ccwienk: also: a typehint stating `typing.Any` is kind of pointless 😅

whitesource/filters.py
```diff
@@ -0,0 +1,160 @@
...
+        artifacts_past = list(filter(lambda a: _artifact_filter(
+            filter_cfg=filter_cfg,
+            artifact=a,
+        ), artifacts_pre))
+        if filter_cfg.type is whitesource.model.FilterType.SOURCE:
+            c.sources = artifacts_past
+        elif filter_cfg.type is whitesource.model.FilterType.RESOURCE:
+            c.resources = artifacts_past
+        else:
+            raise NotImplementedError(filter_cfg.type)
```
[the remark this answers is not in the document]
ccwienk: please use list-comprehensions

whitesource/client.py
ccwienk: Union[xxx, None] is equivalent to Optional[xxx] @AndreasBurger 
I agree that the Optional-shorthand is the better choice, though

ccwienk:
fix unittests..

whitesource/filters.py
```diff
@@ -0,0 +1,160 @@
...
+
+
+def _match_bool_filter(
+    filter_cfg: whitesource.model.WhiteSourceFilterCfg,
+    artifact: typing.Union[cm.Resource, cm.ComponentSource] = None,
+    component: cm.Component = None,
+) -> bool:
+
+    if not artifact and not component:
+        raise NotImplementedError
```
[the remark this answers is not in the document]
zkdev: Not really, will change.

whitesource/filters.py
zkdev: I have the gut feeling this won't be the last time we're touching this module and since the filtering logic is not too trivial I think keeping the output may help.

whitesource/filters.py
zkdev: Regarding log, see https://github.com/gardener/cc-utils/pull/608#discussion_r646342666.
Why not explicitly return True? Its more straight forward to directly return the value, isn't it?

whitesource/client.py
zkdev: Yes, we did agree on `typing.optional` @AndreasBurger , these hints just sneaked through.
Will go for the default value approach since its more common in our codebase.

PR #611, opened 2021-06-09
Add replicate org secrets job

jkrayl:


ccc/github.py
```diff
@@ -180,6 +180,9 @@ def github_cfg_for_hostname(
         require_labels = tuple(require_labels)
 
     def has_required_labels(github_cfg):
+        if not require_labels:
+            return True
```
ccwienk, line 184: nit: it would be cleaner to explicitly return `False` (the way you implemented this getter, `None` will be returned. which will evaluate to `False` in a boolean ctx; however, if a caller would explicitly check for `is False`, this would yield an unexepected/unintuitive behaviour

ccwienk (commented):
one nit - otherwise lgtm

ccwienk (approved):

PR #613, opened 2021-06-15
move uploading of component descriptor to its own step

AndreasBurger:
Moves the upload of component descriptors to a separate step. This prevents issues with CDs being published before the corresponding release has been properly created.

This upload-step is always included for release-jobs, but it can also be configured for jobs containing the component-descriptor trait (to preserve existing functionality, if desired)

### DIFF
--- a/concourse/model/traits/component_descriptor.py
+++ b/concourse/model/traits/component_descriptor.py
@@ -95,6 +95,16 @@ ATTRIBUTES = (
         '''
     ),
     AttributeSpec.optional(
+        name='snapshot_ctx_repository',
+        type=str,
+        default=None,
+        doc='''
+            the component descriptor context repository cfg name for snapshot-component-descriptors,
+            i.e. component descriptors that are not for release-versions.
+            If not configured, no snapshot-component-descriptor will be published.
+        '''
+    ),
+    AttributeSpec.optional(
         name='component_labels',
         default=[],
         type=typing.List[Label],
@@ -138,6 +148,14 @@ class ComponentDescriptorTrait(Trait):
 
     def validation_policies(self):
         return ()
+
+    def snapshot_ctx_repository(self):
+        if snapshot_ctx_repo_name := self.raw['snapshot_ctx_repository']:
+            return self.cfg_set.ctx_repository(snapshot_ctx_repo_name)
+
+    def snapshot_ctx_repository_base_url(self):
+        if snapshot_repo_cfg := self.snapshot_ctx_repository():
+            return snapshot_repo_cfg.base_url()
 
     def ctx_repository(self) -> model.ctx_repository.CtxRepositoryCfg:
         ctx_repo_name = self.raw.get('ctx_repository')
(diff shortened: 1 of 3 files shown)


ccwienk:
@AndreasBurger : please fix linter error(s)

concourse/model/traits/component_descriptor.py
```diff
@@ -70,6 +70,14 @@ class StepInput:
         default={},
         doc='Specifies additional environment variables passed to .ci/component_descriptor script',
     ),
+    AttributeSpec.optional(
+        name='upload_component_descriptor',
```
ccwienk, line 74: please rm this (again)

concourse/model/traits/component_descriptor.py
```diff
@@ -139,6 +147,9 @@ def callback_env(self) -> dict:
     def validation_policies(self):
         return ()
 
+    def upload_component_descriptor(self):
```
ccwienk, line 150: also rm

concourse/model/traits/component_descriptor.py
```diff
@@ -139,6 +147,9 @@ def callback_env(self) -> dict:
...
+        return self.raw['upload_component_descriptor']
+
     def ctx_repository(self) -> model.ctx_repository.CtxRepositoryCfg:
```
ccwienk, line 153: how about adding a second attribute `snapshot_ctx_repository` (note: use cfg-element-name, as for this getter). if set, the following semantics applies:

1. component-descriptor-step _always_ uploads to configured (snapshot-)ctx-repository
2. release-trait should do a "promote"; second step can be omitted as a shortcut to get this change done a bit quicker

concourse/model/traits/component_descriptor.py
```diff
@@ -217,12 +228,38 @@ def inject_steps(self):
         yield self.descriptor_step
 
     def process_pipeline_args(self, pipeline_args: 'JobVariant'):
-        if pipeline_args.has_step('release'):
+        if pipeline_args.has_trait('release'):
```
ccwienk, line 231: I think this should be done the other way around.

component-descriptor step should upload to snapshot-repo (if available/configured)
however, it should be up to the release-step to react upon the component-descriptor trait's configuration, wdyt?

concourse/model/traits/component_descriptor.py
```diff
@@ -217,12 +228,38 @@ def inject_steps(self):
...
                 variable_name=ENV_VAR_NAME,
             )
+            upload_component_descriptor_step = pipeline_args.step('upload_component_descriptor')
+            upload_component_descriptor_step.add_input(
+                name=DIR_NAME,
+                variable_name=ENV_VAR_NAME,
+            )
+        else:
+            # if upload is configured, add the step here and configure it
+            if self.trait.upload_component_descriptor():
```
ccwienk, line 245: why do this in a separate step?

ccwienk (changes requested):

AndreasBurger:
Implemented the requested/discussed changes:

If configured, both CTF and CDv2 will now be uploaded during the CD-step to the desired ctx-repository. Otherwise, they will _only_ be published during the release-step

PR #620, opened 2021-06-21
Improve cfg-factory convenience a little bit

AndreasBurger:
Provide default getters (`ConfigFactory.<type name>()`) for cfg-types that are based on `NamedModelElement`.
Should make interacting with these simple types from within Python3 a little bit more convenient, as before it required going through the `_cfg_element` method

### DIFF
--- a/model/__init__.py
+++ b/model/__init__.py
@@ -404,16 +404,21 @@ class ConfigFactory:
             if (factory_method := cfg_type.factory_method()):
                 yield factory_method
 
+            # also include methods for NamedModelElement-derived types
+            elif cfg_type.cfg_type() == NamedModelElement.__name__:
+                yield cfg_type.cfg_type_name()
+
         yield from super().__dir__()
 
     def __getattr__(self, cfg_type_name):
         for cfg_type in self._cfg_types().values():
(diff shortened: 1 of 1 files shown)


model/__init__.py
```diff
@@ -85,6 +85,9 @@ def cfg_type(self):
         return self.model.type
 
 
+NAMED_MODEL_ELEMENT_TYPE_NAME = 'NamedModelElement'
```
ccwienk, line 88: rather reference the actual type from `model/__init__.py`

model/__init__.py
```diff
@@ -404,16 +407,21 @@ def __dir__(self):
...
+
         yield from super().__dir__()
 
     def __getattr__(self, cfg_type_name):
         for cfg_type in self._cfg_types().values():
             if cfg_type.factory_method() == cfg_type_name:
-                break
-        else:
-            raise AttributeError(cfg_type_name)
+                return functools.partial(self._cfg_element, cfg_type_name)
```
ccwienk, line 419: rather delegate checking to `self._cfg_element` (deduplicate lookup logic)

ccwienk: just re-considered. better leave it like this

PR #628, opened 2021-06-29
Fixes invalid oci artifact if old component-descriptor media typ and nexus content-length bug

enrico-kaack-comp:
This PR fixes two smaller bugs:
- If a component-descriptor uses an old mediaType, the replicate_blobs logic now only compares for digest and not the whole object (including media type and digest).
- The nexus OCI registry does not return a Content-Length Header on head_manifest. Therefore, it is set to None in this case.

Fixes #

**Release note**:

```bugfix developer
- component-descriptor patching now works correctly with old media type
- a missing content-length header from the nexus OCI registry is handled
```


oci/client.py
```diff
@@ -527,10 +527,13 @@ def head_manifest(
         # XXX Docker-Content-Digest header may be absent or incorrect
         # -> it would be preferrable to retrieve the manifest and calculate the hash manually
 
+        if (size := headers.get('Content-Length', None)):
```
AndreasBurger, line 530:
```suggestion
        if size := headers.get('Content-Length'):
```

oci/__init__.py
```diff
@@ -138,9 +138,11 @@ def replicate_blobs(
 
     Note that the uploaded artifact must be finalised after the upload by a "manifest-put".
     '''
+    blob_overwrites = {k.digest:v for (k,v) in blob_overwrites.items()}
+
     def replicate_blob(blob: om.OciBlobRef) -> om.OciBlobRef:
-        if (blob_overwrite_bytes := blob_overwrites.get(blob)):
-            logger.info(f'overwriting {blob=}')
+        if (blob_overwrite_bytes := blob_overwrites.get(blob.digest)):
```
AndreasBurger, line 144:
```suggestion
        if blob_overwrite_bytes := blob_overwrites.get(blob.digest):
```

oci/__init__.py
```diff
@@ -138,9 +138,11 @@ def replicate_blobs(
 
     Note that the uploaded artifact must be finalised after the upload by a "manifest-put".
     '''
+    blob_overwrites = {k.digest:v for (k,v) in blob_overwrites.items()}
```
AndreasBurger, line 141:
```suggestion
    blob_overwrites = {k.digest:v for k,v in blob_overwrites.items()}
```

AndreasBurger (approved):
Left a few suggestions if you want to save some parentheses. Feel free to disregard

ccwienk (approved):
lgtm - please squash

PR #630, opened 2021-07-01
Write scan results to compliance-db

zkdev:
Based on `compliance_db` trait, write compliance scan results to given DB cfg.

Steps:
- add compliance-db model (sqlalchemy orm)
- parse scan results to fit compliance-db model
- implement custom `insert_result()` for each scan tool
- trigger insertion with `delivery.client`

Misc:
- some logging updates

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/checkmarx/util.py
+++ b/checkmarx/util.py
@@ -1,4 +1,5 @@
 import concurrent.futures
+import dataclasses
 import functools
 import logging
 import shutil
@@ -12,45 +13,43 @@ import dacite
 import github3.exceptions
 import github3.repos
 
+import ccc.delivery
 import ccc.github
-import cnudie.retrieve
 import checkmarx.client
 import checkmarx.model as model
 import checkmarx.project
 import checkmarx.tablefmt
 import ci.util
+import cnudie.retrieve
+import dso.compliancedb.model
+import dso.labels
+import dso.model
+import dso.util
+import gci.componentmodel as cm
 import mailutil
 import product.util
 import reutil
-import dso.labels
-import dso.model
 
-import gci.componentmodel as cm
+
+logger = logging.getLogger(__name__)
 
 
 def scan_sources(
-    component_descriptor: cm.ComponentDescriptor,
+    artifacts: typing.Tuple[dso.model.ScanArtifact],
     cx_client: checkmarx.client.CheckmarxClient,
     team_id: str,
     threshold: int,
-    max_workers: int = 4, # only two scan will be run per user
     exclude_paths: typing.Sequence[str] = (),
     include_paths: typing.Sequence[str] = (),
+    max_workers: int = 4, # only two scan will be run per user
 ) -> model.FinishedScans:
-
-    components = tuple(cnudie.retrieve.components(component=component_descriptor))
-
-    # identify scan artifacts and collect them in a sequence
-    artifacts_gen = _get_scan_artifacts_from_components(
-        components=components,
-    )
 
     return scan_artifacts(
         cx_client=cx_client,
         exclude_paths=exclude_paths,
         include_paths=include_paths,
         max_workers=max_workers,
-        scan_artifacts=artifacts_gen,
+        scan_artifacts=artifacts,
         team_id=team_id,
         threshold=threshold,
     )
@@ -74,6 +73,11 @@ def _get_scan_artifacts_from_components(
                     access=source.access,
                     name=f'{component.name}_{source.identity(peers=component.sources)}',
                     label=cx_label,
+                    componentName=component.name,
+                    componentVersion=component.version,
+                    artifactName=source.name,
+                    artifactVersion=source.version,
+                    artifactType=dso.model.ArtifactType.SOURCE,
                 )
             elif cx_label.policy is dso.labels.ScanPolicy.SKIP:
                 continue
@@ -112,7 +116,7 @@ def scan_artifacts(
     artifact_count = len(artifacts)
     finished = 0
 
-    ci.util.info(f'will scan {artifact_count} artifacts')
+    logger.info(f'will scan {artifact_count} artifacts')
 
     scan_func = functools.partial(
         scan_gh_artifact,
@@ -162,7 +166,7 @@ def scan_artifacts(
                     finished_scans.scans_below_threshold.append(scan_result)
 
             finished += 1
-            ci.util.info(f'remaining: {artifact_count - finished}')
+            logger.info(f'remaining: {artifact_count - finished}')
 
     return finished_scans
 
@@ -279,7 +283,7 @@ def send_mail(
         cfg_factory = ci.util.ctx().cfg_factory()
         cfg_set = cfg_factory.cfg_set(default_cfg_set_name)
 
-        ci.util.info(f'sending notification emails to: {",".join(email_recipients)}')
+        logger.info(f'sending notification emails to: {",".join(email_recipients)}')
         mailutil._send_mail(
             email_cfg=cfg_set.email(),
             recipients=email_recipients,
@@ -287,11 +291,11 @@ def send_mail(
             subject='[Action Required] checkmarx vulnerability report',
             mimetype='html',
         )
-        ci.util.info('sent notification emails to: ' + ','.join(email_recipients))
+        logger.info('sent notification emails to: ' + ','.join(email_recipients))
 
     except Exception:
         traceback.print_exc()
-        ci.util.warning('error whilst trying to send notification-mail')
+        logger.warning('error whilst trying to send notification-mail')
 
 
 def print_scans(
@@ -300,23 +304,23 @@ def print_scans(
 ):
     if scans.scans_above_threshold:
         print('\n')
-        ci.util.info('critical scans above threshold')
+        logger.info('critical scans above threshold')
         checkmarx.tablefmt.print_scan_result(
             scan_results=scans.scans_above_threshold,
             routes=routes,
         )
     else:
-        ci.util.info('no critical artifacts above threshold')
+        logger.info('no critical artifacts above threshold')
 
     if scans.scans_below_threshold:
         print('\n')
-        ci.util.info('clean scans below threshold')
+        logger.info('clean scans below threshold')
         checkmarx.tablefmt.print_scan_result(
             scan_results=scans.scans_below_threshold,
             routes=routes,
         )
     else:
-        ci.util.info('no scans below threshold')
+        logger.info('no scans below threshold')
 
     if scans.failed_scans:
         print('\n')
@@ -325,7 +329,7 @@ def print_scans(
                 f'- {artifact_name}' for artifact_name in scans.failed_scans
             )
         )
-        ci.util.warning(f'failed scan artifacts:\n\n{failed_artifacts_str}\n')
+        logger.warning(f'failed scan artifacts:\n\n{failed_artifacts_str}\n')
 
 
 def _download_and_zip_repo(
@@ -411,3 +415,40 @@ def component_logger(artifact_name: str):
 def create_checkmarx_client(checkmarx_cfg_name: str):
     cfg_fac = ci.util.ctx().cfg_factory()
     return checkmarx.client.CheckmarxClient(cfg_fac.checkmarx(checkmarx_cfg_name))
+
+
+def insert_results(
+    scans: model.FinishedScans,
+    scan_artifacts,
+    compliancedb_cfg_name: str,
+):
+    if not compliancedb_cfg_name:
+        logger.warning('no compliance db cfg name found, skipping insertion')
+        return
+
+    client = ccc.delivery.default_client_if_available()
+
+    for scan_artifact in scan_artifacts:
+        for scan in scans.failed_scans + scans.scans_above_threshold + scans.scans_below_threshold:
+            if scan.artifact_name == scan_artifact.name:
+                client.post_compliance_scan(
+                    compliancedb_cfg_name=compliancedb_cfg_name,
+                    scan_data=dataclasses.asdict(scan),
+                    tool=dso.compliancedb.model.ScanTool.CHECKMARX,
+                    component_name=scan_artifact.componentName,
+                    component_version=scan_artifact.componentVersion,
+                    artifact_name=scan_artifact.artifactName,
+                    artifact_version=scan_artifact.artifactVersion,
+                    artifact_type=scan_artifact.artifactType,
+                )
+
+
+def scan_artifacts_from_component_descriptor(
+    component_descriptor: cm.ComponentDescriptor,
+) -> typing.Tuple[dso.model.ScanArtifact]:
+    components = tuple(cnudie.retrieve.components(component=component_descriptor))
+
+    # identify scan artifacts and collect them in a sequence
+    return tuple(_get_scan_artifacts_from_components(
+        components=components,
+    ))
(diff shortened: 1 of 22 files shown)


zkdev:
Tests fail due to missing dependency (sqlalchemy).

ccwienk (changes requested):
as partially already inlined:

1. please separate refactoring diffs into separate commit(s) (e.g. ci.util.<log> -> logger)
2. please separate tool-specific changes into separate commits (protecode, cx, ws, ..) - one per tool
3. common parts (if any) could be moved into preliminary commits

so roughly, I would expect commits matching this scheme and order:
```
refactoring commit(s) # might be pushed to master directly, especially if not relevant for this PR
preliminaries # preparation changes relevant for all scanning tools (if any)
changes for protecode # feel free to re-order the tools
changes for checkmarx
changes for whitesource
changes for clamav # if any
```

concourse/model/traits/image_scan.py
```diff
@@ -201,7 +222,14 @@ class Notify(enum.Enum):
         default=(),
         type=typing.Set[str],
         doc='if present, generated build steps depend on those generated from specified traits',
-    )
+    ),
+    # TODO required
```
ccwienk, line 226: well.. .change to required, then?

concourse/model/traits/image_scan.py
```diff
@@ -230,6 +258,10 @@ def clam_av(self):
         if self.raw['clam_av']:
             return ClamAVScanCfg(raw_dict=self.raw['clam_av'])
 
+    def compliance_db(self):
+        if self.raw['compliance_db']:
```
ccwienk, line 262: this check will not do what you intend.

rather use `.get('compliance_db')` (or `if 'compliance_db' in self.raw).

since python3.8, you may use assignment expressions, which could be considered somewhat more elegant (not every python developer seems to agree, though):

```
if compliance_db_dict := self.raw.get('compliance_db', None'): # None added for explicitness - can be omitted
  return ComplianceDBCfg(raw_dict=compliance_db_dict)
else:
 return None # also: added for explicitness - can be omitted
```

also, consider adding a return-type-hint (-> typing.Optional[ComplianceDBCfg])

ccwienk: also: since the attr is configured to be `required`, there would be no need to check for its existence. since we do not currently have a compliance-db-cfg everywhere, I recommend, to change the attr to optional..

concourse/model/traits/scan_sources.py
```diff
@@ -73,6 +73,15 @@ class Notify(enum.Enum):
 )
 
 
+COMPLIANCE_DB_ATTRIBUTES = (
+    AttributeSpec.required(
```
ccwienk, line 77: change to optional

concourse/model/traits/scan_sources.py
```diff
@@ -174,6 +192,13 @@ def exclude_path_regexes(self):
         default=(),
         doc='if present, perform whitesource scanning',
     ),
+    # TODO required
+    AttributeSpec.optional(
```
ccwienk, line 196: actually, I would prefer to mv the compliance-db-cfg one level up - this way, it only needs to be configured once, and will be honoured by all configured scans

concourse/steps/scan_container_images.mako
```diff
@@ -99,6 +101,20 @@ results_above_threshold, results_below_threshold, license_report = protecode.uti
   image_reference_filter=filter_function,
   cvss_version = CVSSVersion('${protecode_scan.cvss_version().value}'),
 )
+results_above_threshold = protecode_results.above
```
ccwienk, line 104: as discussed earlier: do not filter when writing to compliance-db. just collect everything - will filter later

concourse/steps/scan_sources.mako
```diff
@@ -10,11 +10,17 @@ repo_name = main_repo.logical_name().upper()
...
 checkmarx_cfg = source_scan_trait.checkmarx()
+compliance_db = source_scan_trait.compliance_db()
 whitesource_cfg = source_scan_trait.whitesource()
 email_recipients = source_scan_trait.email_recipients()
 component_trait = job_variant.trait('component_descriptor')
 component_descriptor_dir = job_step.input('component_descriptor_dir')
 scan_sources_filter = source_scan_trait.filters_raw()
+
+if not compliance_db:
+    compliancedb_cfg_name = None
```
ccwienk, line 21: just pass the compliance-db-cfg

concourse/steps/scan_sources.py
```diff
@@ -61,23 +72,36 @@ def scan_component_with_whitesource(
...
-        filters=filters,
+        scan_artifacts=scan_artifacts,
+    )
+
+    product = whitesource.util.product_for_component_descriptor(
+        whitesource_client=whitesource_client,
+        component_descriptor=component_descriptor,
+    )
+
+    whitesource.util.insert_results(
```
ccwienk, line 92: I think I did not yet reach the whitesource-changes.

generally speaking, I would have expected the util to upload to compliance-db w/ the compliance-db-related package 🤔

delivery/client.py
```diff
@@ -16,6 +18,13 @@ def component_descriptor(self):
             'component',
         )
 
+    def compliance_scan(self):
+        return 'http://' + ci.util.urljoin(
```
ccwienk, line 22: rather add schema to base_url, please

delivery/client.py
```diff
@@ -43,3 +52,31 @@ def component_descriptor(
             res.json(),
             validation_mode=validation_mode,
         )
+
+    def post_compliance_scan(
```
ccwienk, line 56: maybe slightly different than stated before: it is okay-ish to have this generic, tool-agnostic "lowlevel" posting-method.

however, as stated before, this method should accept component + artifact (rather than flattened attrs from those). of course, at this point, those attrs will have to be flattened into a json (but this should happen inside this method).

also, either add tool-specific (convenience) methods that will then delegate to this method, or implement conversion gear that will translate tool-specific scan-results into arg-vectors for this method

delivery/client.py
```diff
@@ -43,3 +52,31 @@ def component_descriptor(
...
+        scan_data: dict,
+        compliancedb_cfg_name: str,
+        tool: ScanTool,
+        component_name: str,
+        component_version: str,
+        artifact_name: str,
+        artifact_version: str,
+        artifact_type: ArtifactType,
+    ):
+        body = {
```
ccwienk, line 67: since you also need to deserialise this on the receiving endpoint, you should - as discussed before - add serialisation/deserialisation functions (this will come in handy for the PR w/ the delivery-service)

dso/compliancedb/db.py
```diff
@@ -0,0 +1,47 @@
...
+        )
+        self.Base = Base
+        # we configured our own root logger and use log propagation
+        # therefore pop streamhandler to not have duplicate output
+        logging.getLogger('sqlalchemy.engine.Engine').handlers.pop()
+        self.Base.metadata.create_all(self._engine)
+        self.Session = sqlalchemy.orm.Session(self._engine)
+
+
+class ComplianceDBFactory:
```
ccwienk, line 31: really..?

rather put it into `ccc.compliancedb`, please. use a plain function, please. this is not enterprise-java..

dso/compliancedb/model.py
```diff
@@ -0,0 +1,61 @@
...
+        sqlalchemy.TIMESTAMP(timezone=True),
+        default=datetime.datetime.utcnow(),
+    )
+    source = Column(sqlalchemy.types.Enum(
+        ScanTool,
+    ))
+    scan_data = Column(sqlalchemy.JSON)
+
+    def __repr__(self):
+        return f'ScanResult(id={self.id!r}, '\
```
ccwienk, line 58: is it actually necessary to define custom __repr__s? :-/
I thought sqalchemy would offer this (maybe through explicit cfg, which by default is not enabled)

dso/util.py
```diff
@@ -1,7 +1,55 @@
...
+    tool: dso.compliancedb.model.ScanTool,
+    component_name: str,
+    component_version: str,
+    artifact_name: str,
+    artifact_version: str,
+    artifact_type: dso.model.ArtifactType,
+    compliancedb_cfg_name: str,
+):
+    if not compliancedb_cfg_name:
+        logger.warning('no compliance db cfg name found, skipping insertion')
```
ccwienk, line 29: as also remarked before: check this at the caller. pass-in the cfg-element, or even better the initialised client here. if there is no compliance-db-cfg, do not create a client and do not call this function. otherwise, call the function, and regard it as an error if no usable client is passed-in

dso/util.py
```diff
@@ -1,7 +1,55 @@
...
+        return
+    logger.info('inserting results to database')
+
+    db_fac = dso.compliancedb.db.ComplianceDBFactory(cfg_name=compliancedb_cfg_name)
+    cdb = db_fac.make()
+
+    # TODO ensure uuids are actually unique
```
ccwienk, line 36: haha.. (https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_(random)). no need to be _that_ zealous ^^)

however, it might be better to use a different means to identify artifacts, like so:

```
component_id = f'{componentName/componentVersion}'
artifact_id = str(artifact.identity())

qualified_artifact_id = '/'.join(component_id, artifact_id)
```

Those IDs may become a bit lengthy, so we _might_ consider storing a hash-digest, rather than the full str (`hashlib.sha256(qualified_artifact_id).hexdigest()`)

protecode/model.py
```diff
@@ -85,6 +88,10 @@ def __repr__(self):
             f'{self.version() or "Version not detected"}'
         )
 
+    def __iter__(self):
```
ccwienk, line 91: why?

protecode/model.py
```diff
@@ -119,6 +126,11 @@ def __hash__(self):
             self.license_type(),
         ))
 
+    def __iter__(self):
+        yield 'name', self.name()
```
ccwienk, line 130: why?

protecode/model.py
```diff
@@ -156,6 +168,14 @@ def cve_major_severity(self, cvss_version) -> int:
     def __repr__(self):
         return f'{self.__class__.__name__}: {self.cve()}'
 
+    def __iter__(self):
+        yield 'historical', self.historical()
```
ccwienk, line 172: why?

whitesource/component.py
```diff
@@ -47,8 +47,15 @@ def get_scan_artifacts_from_components(
...
                     name=f'{component.name}:{component.version}/'
-                         f'{"source" if artifact in component.sources else "resources"}/'
+                         f'{"source" if artifact in component.sources else "resource"}/'
                          f'{artifact.name}:{artifact.version}',
+                    componentName=component.name,
+                    componentVersion=component.version,
+                    artifactName=artifact.name,
+                    artifactVersion=artifact.version,
+                    artifactType=dso.model.ArtifactType(
+                        'source' if artifact in component.sources else 'resource'
```
ccwienk, line 57: extract this into a function/method + check on the type. actually, since the type is already expressed in the artifact-instance itself, just pass the artifact and deal w/ the discrimination later

ccwienk (changes requested):

checkmarx/util.py
```diff
@@ -12,45 +13,43 @@
...
 import checkmarx.model as model
 import checkmarx.project
 import checkmarx.tablefmt
 import ci.util
+import cnudie.retrieve
+import dso.compliancedb.model
+import dso.labels
+import dso.model
+import dso.util
+import gci.componentmodel as cm
```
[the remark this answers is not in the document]
zkdev: Didn't know about the forth block, will adjust.

checkmarx/util.py
```diff
@@ -411,3 +415,40 @@ def component_logger(artifact_name: str):
...
+    compliancedb_cfg_name: str,
+):
+    if not compliancedb_cfg_name:
+        logger.warning('no compliance db cfg name found, skipping insertion')
+        return
+
+    client = ccc.delivery.default_client_if_available()
+
+    for scan_artifact in scan_artifacts:
+        for scan in scans.failed_scans + scans.scans_above_threshold + scans.scans_below_threshold:
```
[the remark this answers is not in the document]
zkdev: The threshold is not really considered here since in summary, those three lists contain all scan results for checkmarx.
For email / log printing this separation is required. Thus either checkmarx needs some refactoring to extract the separation logic from `checkmarx.util.scan_sources`, call `insert_results()` and split into those three lists afterwards, or, as suggested, use the concatenation.

checkmarx/util.py
```diff
@@ -411,3 +415,40 @@ def component_logger(artifact_name: str):
...
+            if scan.artifact_name == scan_artifact.name:
```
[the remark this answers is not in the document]
zkdev: A `scan_artifact`, type `dso.model.ScanArtifact` contains the component context and artifact information (name, version, type). The actual `scan` however lacks these information (only artifact.name). I kind of agree, this filtering is by far not optimal, but the cleaner solution to add component context to scans requires refactoring.

concourse/model/traits/image_scan.py
zkdev: Wanted to discuss in first place since some pipelines might break.

concourse/model/traits/image_scan.py
zkdev: Wanted to keep consistency within `image_scan.py` as this is the current pattern (I've seen other traits use your mentioned pattern).
Will apply pattern and change attr to optional.

concourse/model/traits/scan_sources.py
zkdev: Thought about the same, will do.

concourse/steps/scan_container_images.mako
zkdev: These filtered sub lists aren't taken for result insertion.
```python
results_above_threshold = protecode_results.above
results_below_threshold = protecode_results.below
license_report = protecode_results.licenseReport

protecode.scanning_util.insert_results(
    protecode_results=protecode_results,
    compliancedb_cfg_name=compliancedb_cfg_name,
)
```
Dropping those completely would require refactoring. 
I'm aware of that being not pretty at all, but it is by far the simplest solution.

concourse/steps/scan_sources.mako
zkdev: Yea, you're totally right. Will do.

dso/util.py
zkdev: Yea, sure thing, but we have to discuss what is considered as `artifact.identity`.

protecode/model.py
zkdev: The `dict()` constructor accepts an iterable of `(key, value)` pairs to construct a dictionary.
Implementing the `__iter__` enables simple dict parsing from class, in order to store scan_result in compliance db.

protecode/model.py
zkdev: see https://github.com/gardener/cc-utils/pull/630#discussion_r662890854

protecode/model.py
zkdev: see https://github.com/gardener/cc-utils/pull/630#discussion_r662890854

delivery/client.py
zkdev: Will do.

PR #632, opened 2021-07-12
enable plaintext as encryption mode

ccwienk:


jkrayl:
lgtm

jkrayl (approved):

jensh007:
Needs additional logic: env vars are already processed even if plaintext algorithm is set. 
get_secret_cfg_from_env_if_available
is anyway confusing: "if_available" suggests that env vars are optional. Instead this function raises an exception.
Suggestion: Use plaintext mode by default if no ennvironment variable is set at all.

ccwienk:
@jensh007 : I changed the behaviour to return `None` if the env-var is not set (/cc @jkrayl ), which now does fit better to the name.

We might also default to `PLAINTEXT`. I think this has even been the behaviour during migration to encrypted secrets. There is, however, one disadvantage. Namely, in cases where there actually _is_ encryption, and the env-var is absent, we would run into rather confusing error messages (json.load against some garbage-octetts), so for the time being, I would rather prefer the current (explicit) approach.

PR #633, opened 2021-07-12
Workaround ghcr registry and media type distribution

enrico-kaack-comp:

Fixes:
- workaround for gchr registry not providing a jwt token during oauth flow but a base64(username:password) token
- NotImplemented for docker media type distribution list

**Release note**:

```other operator

```


container/util.py
```diff
@@ -57,7 +57,11 @@ def filter_image(
             oci_client=oci_client,
         )
 
-    manifest = oci_client.manifest(image_reference=source_ref)
+    try:
+        manifest = oci_client.manifest(image_reference=source_ref)
+    except NotImplementedError as e:
+        logger.warning(f'could not filter image {e}. Skipping...')
```
ccwienk, line 63: do not (almost) silently skip. let the exception propagate. it should be up to the caller do decide whether or not to ignore this

oci/convert.py
```diff
@@ -11,16 +13,35 @@
...
+    oci_client =  ccc.oci.oci_client()
+    uncompressed_layers_digests = []
+    for layer in manifest.layers:
+        cfg_hash = hashlib.sha256() # we need to write "non-gzipped" hash to cfg-blob
+
+        src_content = oci_client.blob(
+            image_reference=src_image_reference,
+            digest=layer.digest,
+            stream=False,
+        ).content
```
ccwienk, line 34: please do not read full blob into memory. this is almost never a good idea

ccwienk (changes requested):

oci/client.py
```diff
@@ -453,6 +458,15 @@ def manifest(
 
         manifest_dict = res.json()
 
+        distribution_list_mediatype = 'application/vnd.docker.distribution.manifest.list.v2+json'
+        if (
+                'mediaType' in manifest_dict and
+                (mediaType := manifest_dict['mediaType']) == distribution_list_mediatype
+            ):
+            #XXX: ignore for now, need to be fixed later
+            raise NotImplementedError(f'OCI Manifest media type {mediaType}'
```
ccwienk, line 467: nit: just pass the mediaType?
also: the comment "ignore for now" is a bit misleading. raising an exception is not quite the same as ignoring..

enrico-kaack-comp: mediaType: passing it directly would displease the linter and would destroy the fromatting.
Comment fixed.

ccwienk (approved):

PR #643, opened 2021-08-11
require cc instance url in default template

jkrayl:


AndreasBurger (approved):

concourse/templates/default.mako
```diff
@@ -20,6 +20,7 @@ target_team = pipeline.get('target_team')
 background_image = pipeline.get('background_image', 'https://i.imgur.com/raPlg21.png')
 job_mapping = pipeline.get('job_mapping')
 secret_cfg = pipeline.get('secret_cfg')
+concourse_instance_url = pipeline.get('concourse_instance_url')
```
ccwienk, line 23: is it actually acceptable if the cc-instance-url is absent?

PR #644, opened 2021-08-13
Support large release-notes

AndreasBurger:
Github has a limit of 25.000 characters for release-bodies. Work around that by attaching release-notes as release-asset if they are too long.

n.b.: since release-notes are computed from scratch every time they are needed, any component that is downstream will also have release-notes that are too long (and thus are abbreviated here).

Does not affect release-notes posted to Slack.

### DIFF
--- a/github/util.py
+++ b/github/util.py
@@ -417,22 +417,73 @@ class GitHubRepositoryHelper(RepositoryHelperBase):
             tagger=author
         )
 
+    MAXIMUM_GITHUB_RELEASE_BODY_LENGTH = 25000
+    '''The limit is not documented explicitly in the GitHub docs. To see it, the error returned by
+    GitHub when creating a release with more then the allowed number of characters must be
+    looked at.'''
+
+    def _replacement_release_notes(
+        self,
+        asset_url: str,
+        component_name: str,
+        component_version: str,
+    ):
+        return (
+            f'The release-notes for component **{component_name}** in version '
+            f'**{component_version}** exceeded the maximum length of '
+            f'{self.MAXIMUM_GITHUB_RELEASE_BODY_LENGTH} characters allowed by GitHub for '
+            'release-bodies.\n'
+            f'They have been uploaded as release-asset and can be found at {asset_url}.'
+        )
+
+    RELEASE_NOTES_ASSET_NAME = 'release_notes.md'
+
     def create_release(
         self,
         tag_name: str,
         body: str,
         draft: bool=False,
         prerelease: bool=False,
-        name: str=None
+        name: str=None,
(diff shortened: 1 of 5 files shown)


ccwienk:
@AndreasBurger : please fix test-errs

github/util.py
```diff
@@ -42,6 +42,12 @@
 from model.github import GithubConfig
 
 
+# The limit is not documented explicitly in the GitHub docs. To see it, the error returned by
+# GitHub when creating a release with more then the allowed number of characters must be
+# looked at.
+MAXIMUM_GITHUB_RELEASE_BODY_LENGTH = 25000
```
ccwienk, line 48: nits:
- rather use a docstr ('''.....''')?
- at least in comment mention unit (is it actually octets, or is it rather characters (not necessarily identical e.g. w/ utf-8)?)
- if it is just used _once_, preferrably either make it "private" (+ possibly lc), or inline it?

github/util.py
```diff
@@ -417,22 +423,62 @@ def create_tag(
...
 
+    def _replacement_release_notes(
+        self,
+        asset_url: str,
+        component_name: str,
+        component_version: str,
+    ):
+        return (
+            f'The release-notes for component **{component_name}** in version '
+            f'**{component_version}** exceeded the maximum length allowed by GitHub for '
```
ccwienk, line 434: nit: print max-allowed leng vs actual leng

github/util.py
```diff
@@ -417,22 +423,62 @@ def create_tag(
...
+            'release-bodies.\n'
+            f'They have been uploaded as release-asset and can be found at {asset_url}.'
+        )
+
+    RELEASE_NOTES_ASSET_NAME = 'release_notes.md'
```
ccwienk, line 439: if used only once, consider inlineing, or at least define it later

github/util.py
```diff
@@ -417,22 +423,62 @@ def create_tag(
...
+        if len(body) < MAXIMUM_GITHUB_RELEASE_BODY_LENGTH:
+            return self.repository.create_release(
+                tag_name=tag_name,
+                body=body,
+                draft=draft,
+                prerelease=prerelease,
+                name=name
+            )
+        else:
+            release = self.repository.create_release(
```
ccwienk, line 460: add a brief comment as to why we create the release w/ empty body, to edit it afterwards (reason being we need to know the asset-url, which we cannot know (can we not actually?) w/o previously uploading the asset (which we cannot do w/o an existing release).

to be a bit more on the safe side, we might consider to firstly create the release as a draft, and then promote it to final after asset-upload (or purge it upon errs) - wdyt?

PR #647, opened 2021-08-19
Rework cc client factory

jkrayl:
adds new client factory methods for concourse

- [x] ensure ConcourseTeamConfig and ConcourseEndpoint model is present in all secrets
- [ ] ensure to adapt all consumers
- [x] ensure that tm, ls-tools and cc-job image on alpine 3.14 and runc 1.0.0~rc93

### DIFF
--- a/ccc/concourse.py
+++ b/ccc/concourse.py
@@ -22,25 +22,6 @@ import model.concourse
 import ensure
 
 
-@functools.lru_cache()
-@ensure.ensure_annotations
-def client_from_cfg_name(
-    concourse_cfg_name: str,
-    team_name: str
-):
-    cfg_factory = ci.util.ctx().cfg_factory()
-
-    cc_cfg = cfg_factory.concourse(concourse_cfg_name)
-    uam_cfg = cfg_factory.concourse_uam(cc_cfg.concourse_uam_config())
-
-    return concourse.client.from_cfg(
-        concourse_cfg=cc_cfg,
-        concourse_uam_cfg=uam_cfg,
-        team_name=team_name,
-        verify_ssl=True,
-    )
-
-
 def lookup_cc_team_cfg(cfg_set, team_name) -> model.concourse.ConcourseTeamConfig:
     for cc_uam in cfg_set._cfg_elements('concourse_uam'):
         if cc_uam.team_name() == team_name:
--- a/ccc/concourse.py
+++ b/ccc/concourse.py
@@ -80,6 +80,7 @@ def client_from_cfg(
     )
 
 
+@functools.cache
 def client_from_env(
     team_name: str=None,
 ) -> concourse.client.api.ConcourseApiBase:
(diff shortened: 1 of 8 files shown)


ccc/concourse.py
```diff
@@ -16,50 +16,95 @@
...
+    password: str,
+    team_name: str,
+    username: str,
+    verify_ssl: bool = True,
+    concourse_api_version=None,
+) -> concourse.client.api.ConcourseApiBase:
+    """
+    returns a concourse-client w/ the credentials valid for the current execution environment.
+    The returned client is authorised to perform operations in the same concourse-team as the
+    credentials provided calling this function.
```
ccwienk, line 46: add a brief comment explaining the difference / intention between `client_from_parameters` vs `client_from_cfg`.

also, on an additional note, how about merging them into one factory-function, as done in https://github.com/gardener/cc-utils/blob/master/ccc/clamav.py#L24 ?

team_name + cfg_set ==> no other arguments allowed
team_name + no cfg_set ==> require all other arguments

ccc/concourse.py
```diff
@@ -16,50 +16,95 @@
...
+    return concourse_api
+
+
+@functools.cache
+def client_from_cfg(
+    cfg_set,
+    team_name: str,
+) -> concourse.client.api.ConcourseApiBase:
+    """
+        returns a concourse-client w/ the credentials valid for the current execution environment.
```
ccwienk, line 69: nit: consitently indent doc-strs within one module

ccc/concourse.py
```diff
@@ -16,50 +16,95 @@
...
+        username=concourse_team_config.username(),
+    )
+
+
+@functools.cache
 def client_from_env(
     team_name: str=None,
-):
-    '''
+) -> concourse.client.api.ConcourseApiBase:
```
ccwienk, line 87: nit: is it really ConcourseApiBase? 🤔

concourse/steps/scan_container_images.py
```diff
@@ -440,14 +441,9 @@ def print_protecode_info_table(
...
-    concourse_cfg = concourse.util._current_concourse_config()
-    uam_cfg_name = concourse_cfg.concourse_uam_cfg()
-    concourse_uam_cfg = ctx.cfg_factory().concourse_uam(uam_cfg_name)
-
     pipeline_metadata = concourse.util.get_pipeline_metadata()
-    client = concourse.client.from_cfg(
-        concourse_cfg=concourse_cfg,
-        concourse_uam_cfg=concourse_uam_cfg,
+    client = ccc.concourse.client_from_cfg(
+        cfg_set=ctx.cfg_factory(),
```
ccwienk, line 446: are you sure this will work (I did not check, so it might)? cfg_factory does not define default values (as opposed to cfg-sets)

ccwienk (changes requested):

concourse/steps/scan_container_images.py
jkrayl: that's no problem the factory doesn't need any defaults

PR #648, opened 2021-08-31
Add new information to meta-step

AndreasBurger:
Adds additional information that should help debugging to the meta-step. Example :
```
{
  "uuid": "f0d9b165-d707-44a9-ad4c-6ab018f83142",
  "render origin": "local", # possible other values: webhook dispatcher, pipeline replication
  "cc-utils version used for rendering": "1.1425.0",
  "pipeline commit sha": "d854c83d5dd7f7ea0b54e925e65722b528d6111b",
  "cc-utils-version": "75342bd606da0714dce5ca2856afb3e2513831f3",
  "replication pipeline name": __foo__, # only shows up for centrally replicated pipelines
  "render-timestamp": "2021-08-31T13:39:03.959443"
}
```

### DIFF
--- a/concourse/enumerator.py
+++ b/concourse/enumerator.py
@@ -88,6 +88,7 @@ class DefinitionEnumerator:
         override_definitions={},
         job_mapping: model.concourse.JobMapping = None,
         target_team: str=None,
+        pipeline_definition_committish: str = None,
     ) -> 'DefinitionDescriptor':
         if not target_team:
             target_team = self.job_mapping.team_name()
@@ -103,6 +104,7 @@ class DefinitionEnumerator:
                 override_definitions=[override_definitions.get(name,{}),],
                 secret_cfg=secret_cfg,
                 job_mapping=job_mapping,
+                pipeline_definition_committish=pipeline_definition_committish,
             )
 
 
@@ -285,6 +287,9 @@ class GithubDefinitionEnumeratorBase(DefinitionEnumerator):
                     path='.ci/pipeline_definitions',
                     ref=branch_name
                 )
+                pipeline_definition_committish = repository.ref(
+                    ref=f'heads/{branch_name}'
+                ).object.sha
             except NotFoundError:
                 continue # no pipeline definition for this branch
 
@@ -323,6 +328,7 @@ class GithubDefinitionEnumeratorBase(DefinitionEnumerator):
                 target_team=target_team,
                 secret_cfg=secret_cfg,
                 job_mapping=job_mapping,
+                pipeline_definition_committish=pipeline_definition_committish,
             )
 
 
@@ -459,6 +465,7 @@ class DefinitionDescriptor:
         job_mapping: model.concourse.JobMapping = None,
         override_definitions=[{},],
         exception=None,
+        pipeline_definition_committish: str = None,
     ):
         try:
             self.pipeline_name = not_empty(pipeline_name)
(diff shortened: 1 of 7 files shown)


concourse/replicator.py
```diff
@@ -91,22 +98,36 @@ def replicate_pipelines(
...
+        return None
+
+
 class Renderer:
     def __init__(
         self,
         cfg_set,
         template_retriever: TemplateRetriever=TemplateRetriever(),
         template_include_dir=None,
+        render_origin: RenderOrigin = RenderOrigin.PIPELINE_REPLICATION,
```
ccwienk, line 116: nit: rather use a value such as "unknown" as default

concourse/replicator.py
```diff
@@ -38,6 +39,12 @@
 logger = logging.getLogger(__name__)
 
 
+class RenderOrigin(enum.Enum):
```
ccwienk, line 42: tbh, I think in this case, a `str` might actually a better choice (allows for more flexibility); the actual value is in fact only intended to be understood/used by humans, is it not? there is of course not hard reason against using an enum

concourse/replicator.py
```diff
@@ -157,8 +178,19 @@ def _render(self, definition_descriptor):
             'target_team': definition_descriptor.concourse_target_team,
             'secret_cfg': definition_descriptor.secret_cfg,
             'job_mapping': definition_descriptor.job_mapping,
+            'render_origin': self.render_origin,
```
ccwienk, line 181: `.value`..?

concourse/steps/meta.mako
```diff
@@ -8,6 +8,7 @@ from concourse.steps import step_lib
 import concourse.paths
 
 extra_attrs = {}
+extra_attrs.update(additional_meta_data)
```
ccwienk, line 11: stylistic nit: `|=` :-)

concourse/enumerator.py
```diff
@@ -285,6 +287,8 @@ def _scan_repository_for_definitions(
                     path='.ci/pipeline_definitions',
                     ref=branch_name
                 )
+                # it's necessary to prefix 'heads/' as per github3.py doc
```
ccwienk, line 290: of course we should use canonical ref-names - no need to document this imo..

concourse/enumerator.py
```diff
@@ -88,6 +88,7 @@ def _wrap_into_descriptors(
         override_definitions={},
         job_mapping: model.concourse.JobMapping = None,
         target_team: str=None,
+        ref_sha: str = None,
```
ccwienk, line 91: I think this name is a bit too short. how about `pipeline_definition_committish` (since this is to my understanding what it is - it is the commit hash pointing to the commit from which the pipelinedefinition was read..)

concourse/replicator.py
```diff
@@ -157,8 +178,19 @@ def _render(self, definition_descriptor):
...
+            'cc_utils_version': _cc_utils_version(),
+            'ref_sha': definition_descriptor.ref_sha,
         }
 
+        # also pass pipeline name if this was rendered by a replication job. Will be printed
+        # in the meta-step later
+        if (
```
ccwienk, line 188: why not simply always add PIPELINE_NAME or `<none>`?

ccwienk (changes requested):

concourse/replicator.py
AndreasBurger: Handling enums is a bit less error prone than using strings directly, I think.
I'll keep the enum (with an added `UNKNOWN`

concourse/steps/meta.mako
AndreasBurger: new in 3.9. TIL

concourse/enumerator.py
AndreasBurger: It's in contrast to the usage two lines above. But that's a github.py detail, you're right. Removed.

concourse/replicator.py
AndreasBurger: vanity

ccwienk (approved):

PR #649, opened 2021-09-03
Do not publish snapshot descriptors for release jobs

AndreasBurger:
This PR prevents release jobs from publishing snapshot component descriptors. These snapshot descriptors could sometimes be published a long time ahead of the actual release, which causes problems with our release-note transport downstream.
Since we configured all our pipelines to publish snapshot descriptors into the same repository as their release descriptors there should be no noticeable change in behaviour for most (if not all) of our components. Components with custom snapshot-repos, however, will not have their release descriptors published to the snapshot-repo - this will be fixed at a future point in time.

Local testing was successful, will be merged on Monday


ccwienk:
@AndreasBurger : please fix tests

PR #663, opened 2021-11-11
make target team for secret repl configurable

jkrayl:


ccwienk (approved):

ccwienk:
@jkrayl : would it make sense to already include the change reacting upon this new optional attr?

jkrayl:
The changes reacting upon this are in cicd-outbound and in the cfg-repo

PR #664, opened 2021-11-18
add model for alertmanager

neo-liang-sap:

add model for alertmanager
Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/model/alertmanager.py
+++ b/model/alertmanager.py
@@ -1,0 +1,30 @@
+# Copyright (c) 2019-2021 SAP SE or an SAP affiliate company. All rights reserved. This file is
+# licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


model/alertmanager.py
```diff
@@ -0,0 +1,30 @@
...
+        return self.raw.get('username')
+
+    def password(self):
+        return self.raw.get('password')
+
+    def _required_attributes(self):
+        return (
+            'username',
+            'password',
+        )
```
AndreasBurger, line 30: We do already have a base class one could use here.
```suggestion
from model.base import BasicCredentials


class AlertManagerConfig(BasicCredentials):
    def password(self):
        return self.raw.get('password')
```
By default, the provided password-getter would be `passwd()`.

Then again, instead of introducing this boilerplate-class, you _should_ be able to use the BasicCredential class directly in the `config_types.yaml`. I haven't tried that though.

Let's see whether @ccwienk has some thoughts on this :).

ccwienk: @neo-liang-sap : as @AndreasBurger indicated, you might configure this type in `config_types.yaml`, like so:

```
alert_manager:
 src:
  file: alertmanager.yaml
 model:
  factory_method: alertmanager
  cfg_type_name: alertmanager
  type: BasicCredentials
```

(in which case you would not need to add the type here). I am however also fine with merging this PR, as is

neo-liang-sap: OK i will close this PR and use `BasicCredentials `, thanks!

PR #665, opened 2021-11-19
adapt siging cfg model to new service auth

jkrayl:


model/delivery.py
```diff
@@ -101,19 +101,25 @@ def scope(self):
...
+    def id(self) -> str:
+        return self.raw.get('id')
+
     def algorithm(self):
         return self.raw.get('algorithm', 'HS256')
 
     def secret(self):
         return self.raw.get('secret')
 
+    def purpose_label(self):
```
ccwienk, line 113: maybe it would be more "future-proof" to make this a `list` (label_s_), even if we will only have one single label to start with?

model/delivery.py
```diff
@@ -101,19 +101,25 @@ def scope(self):
...
+        return self.raw.get('purpose_label')
+
 
 class DeliverySvcCfg(ModelBase):
     def external_host(self):
         return self.raw.get('external_host')
 
-    def signing_cfg(self):
-        return SigningCfg(self.raw.get('signing'))
+    def signing_cfgs(self) -> list[SigningCfg]:
```
ccwienk, line 121: you might want to add an optional formal parameter of purpose-labels to filter for

ccwienk (changes requested):

PR #666, opened 2021-11-19
DeliveryDB Client no ORM

zkdev:
This PR adds a `DeliveryDB` client without ORM.
Instead of `SQLAlchemy`, the `psycopg` driver is accessed directly.

Fixes #

**Release note**:

```other operator
NONE
```


ccc/deliverydb.py
```diff
@@ -1,18 +1,32 @@
...
+
+from deliverydb.db import DeliveryDB, delivery_db_no_orm
 import model.compliancedb
 
 
 def default_with_db_cfg(
     db_cfg: model.compliancedb.ComplianceDbConfig,
     hostname: str = None,
-) -> DeliveryDB:
+    orm: bool = True,
```
ccwienk, line 12: as a general stmt, it is very rarely a good idea to use boolean flags to switch between different flavours or variants (there are more ORMs besides SQLAlchemy, also there even are IIRC more postgresql-drivers than psycopg).

So - if at all - I would prefer to use an enum. However, I would not actually suppose this to be a good idea either way.

I would rather recommend to factor-out the (supposedly) common functionality between creating a SQLAlchemy session, and a psycopq-connection. Which I assume would be to create a db-connection-URL from a db-cfg object.

Also (yes, this is not part of this PR), hostname should by default be read from db-cfg. I understand you introduced this for local development. at the very least, this ought rather be called sth. like `overwrite_hostname`, and only be used if passed. even better: accept a full db-connection-URL. then everything can be overwritten.

ccc/deliverydb.py
```diff
@@ -1,18 +1,32 @@
...
         hostname = db_cfg.hostname()
 
-    return DeliveryDB(
-        username=db_cfg.credentials().username(),
-        password=db_cfg.credentials().password(),
-        hostname=hostname,
-        port=db_cfg.port(),
-    )
+    if orm:
+        return DeliveryDB(
```
ccwienk, line 19: I know, not part of this PR: I think, DeliveryDB should not deal w/ setting up sqlalchemy. It should rather be passed an initialised instance of a sqlalchemy-session(-factory), and expose (as it already does) methods that _use_ said session / session-factory

requirements.dso.txt
```diff
@@ -1,3 +1,4 @@
+psycopg[binary]
 psycopg2-binary
```
ccwienk, line 2: I am curious: why do we need both psycopg[binary] and psycopg2-binary?

ccwienk (changes requested):

requirements.dso.txt
zkdev: Like written in 0b03230397735973f159b95b98fa7f58b9064436, SQLAlchemy (ORM) still relies on `psycopg2`, while our None-ORM approach is implemented in `psycopg3` (packages is called `psycopg`)...

ccc/deliverydb.py
```diff
@@ -1,18 +1,46 @@
...
-        port=db_cfg.port(),
+def psycopg_with_db_cfg(
+    db_cfg: model.compliancedb.ComplianceDbConfig,
+    overwrite_hostname: str = None,
+    overwrite_username: str = None,
+    overwrite_password: str = None,
+    overwrite_port: int = None,
+) -> deliverydb.db.DeliveryDB:
+
+    db_url = deliverydb.db.database_conncetion_url_from_cfg(
```
ccwienk, line 36: nit: typo (conncetion)

ccc/deliverydb.py
```diff
@@ -1,18 +1,46 @@
...
-        password=db_cfg.credentials().password(),
-        hostname=hostname,
-        port=db_cfg.port(),
+def psycopg_with_db_cfg(
+    db_cfg: model.compliancedb.ComplianceDbConfig,
+    overwrite_hostname: str = None,
+    overwrite_username: str = None,
+    overwrite_password: str = None,
+    overwrite_port: int = None,
+) -> deliverydb.db.DeliveryDB:
```
ccwienk, line 34: return typehint is not correct, I think

deliverydb/db.py
```diff
@@ -3,19 +3,16 @@
...
 
 class DeliveryDB:
     def __init__(
         self,
-        username: str,
-        password: str,
-        hostname: str,
-        port: int,
-        dialect: str = 'postgresql',
+        db_conn_url: str,
```
ccwienk, line 12: that's more like it. how about moving the util-function we use to create this connection-url either to the `ccc.deliverydb` module, or (e.g. as a method) to the `model.deliverydb` - oh wait: ...compliancedb cfg class? then, the deliverydb-client would not need to have a direct dependency towards those.

on an additional note: I just noticed we use (in the module / package names) both deliverydb and compliancedb. we should use consistent naming (-> please rename compliancedb -> deliverydb)

deliverydb/db.py
```diff
@@ -40,3 +37,21 @@ def insert_compliance_issue(
         )
 
         self.Session.add(scan)
+
+
+def database_conncetion_url_from_cfg(
```
ccwienk, line 42: typo

deliverydb/db.py
```diff
@@ -40,3 +37,21 @@ def insert_compliance_issue(
...
+    db_cfg: model.compliancedb.ComplianceDbConfig,
+    dialect: str = 'postgresql',
+    overwrite_hostname: str = None,
+    overwrite_username: str = None,
+    overwrite_password: str = None,
+    overwrite_port: int = None,
+) -> str:
+    username = overwrite_username if overwrite_username else db_cfg.credentials().username()
```
ccwienk, line 50: stylistic nit: you might use `username = overwrite_username or db_cfg.credentials().username()` as a slightly shorter style

requirements.dso.txt
ccwienk: understood. as discussed earlier, please add a _brief_ comment (inline, into this file, after the respective line) stating this (like: # required by sqlalchemy, or sth the like)

ccwienk (changes requested):

deliverydb/__init__.py
```diff
@@ -0,0 +1,4 @@
+from psycopg.rows import dict_row as psycopg_dict_row
+
+
+dict_row = psycopg_dict_row
```
AndreasBurger, line 4: `dict_row` is imported as `psycopg_dict_row` but also made available as `dict_row`. But I cannot find a use for either outside this module.

If this is due to an existing use somewhere else, I'd rather handle this function-namechange there using `try ... except ImportError`

Either way, the same with a few characters less:
```suggestion
from psycopg.rows import dict_row


psycopg_dict_row = dict_row
```

zkdev: The idea is to keep the hard dependency on `psycopg` in the `deliverydb` package.

deliverydb/db.py
zkdev: Will move to `ccc.deliverydb` for now and follow up on renaming the cfg class.

deliverydb/__init__.py
zkdev: A consumer of `deliverydb` can access the `psycopg_dict_row` without hard depending on `psycopg` on code level.

zkdev: The dependency remains inside the `deliverydb` package.

PR #667, opened 2021-11-22
Fix erroneous webhook-urls being applied

AndreasBurger:
The previous logic did not take into account that we have job-mappings configured for our internal concourse that reference orgs in github.com. These job-mappings cause the previous logic to set an unreachable webhook-url on the orgs on github.com. Since the order in which we set webhooks is not deterministic, this may result in webhook-deliveries being lost if such a faulty update is performed after the correct one was set.

This change introduces a heuristic to determine undeliverable webhook- urls and avoids setting them at all.

### DIFF
--- a/concourse/util.py
+++ b/concourse/util.py
@@ -109,9 +109,11 @@ def _enumerate_required_org_webhooks(
         job_mapping_set = cfg_factory.job_mapping(concourse_cfg.job_mapping_cfg_name())
 
         for github_orgname, github_cfg_name in _enumerate_github_org_configs(job_mapping_set):
-            github_api = ccc.github.github_api(
-                github_cfg=cfg_factory.github(github_cfg_name),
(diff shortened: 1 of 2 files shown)


ccwienk (changes requested):
@AndreasBurger : interesting - I did not even know/remember there were such an attribute (github_enterprise_host) on our concourse-cfg (the commit that introduces this is from 2018)

how about we make this slightly more generic, like so:
```
# concourse.yaml
external:
 domain_rules:
  - domain_regex: '.*\.sap\.?$'
    accessible: False # assumption: any domain that is not explicitly unaccessible is accessible
```

concourse-cfg could then offer a new method `url_accessible(self, url)`. Note that in canonical form, domain names end with a `.` character (which is omitted in most cases in practise) - which I reflected in the example regex above.

PR #668, opened 2021-11-29
paginate through apps-list

ccwienk:
as of version 2021.06, Protecode (BDBA) incompatibly introduces
pagination (default page size being 1000) for /apps route. Collect all
available data.

**Release note**:
```other
prepare for upgrade to Protecode/BDBA >= 2021.06
```


jkrayl (approved):
lgtm

PR #669, opened 2021-11-29
purge concourse-image-resource-based oci-builder

ccwienk:
Concourse's oci image(-builder) resource has shown to be deficient. In
particular, it does not allow for authentication credentials to be
passed-in for "image-pulls" to be done during image-builds.

Within our pipeline-template, we by now feature both `kaniko` (with some
quirks) and a docker(d)-based oci-builder as alternatives, which are
both run as regular build-steps (which fits better into the existing
pipeline-template).


AndreasBurger (approved):
lgtm

PR #670, opened 2021-12-02
Add connection pool client for DeliveryDB

zkdev:
This PR introduces a function to retrieve not only a psycopg deliverydb connection, but a connection pool.

Fixes #

**Release note**:

```other operator
NONE
```

### DIFF
--- a/ccc/deliverydb.py
+++ b/ccc/deliverydb.py
@@ -5,7 +5,7 @@ import deliverydb.db
 import model.compliancedb
 
 
-def sqlalchemy_with_db_cfg(
+def sqlalchemy_from_db_cfg(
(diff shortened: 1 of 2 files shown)


ccc/deliverydb.py
```diff
@@ -25,7 +26,7 @@ def sqlalchemy_with_db_cfg(
     )
 
 
-def psycopg_with_db_cfg(
+def psycopg_connection_with_db_cfg(
```
jkrayl, line 29: nit: I think our convention at this point is to use the the word `from` instead of `with` like `psycopg_connection_from_db_cfg`

jkrayl (changes requested):
general lgtm except for one small nit

PR #676, opened 2022-01-17
add cfg element compliance checking gear

ccwienk:
We use several cfg elements, containing secrets that might e.g. be
subject to "rotation policies", distributed across several secrets
stores.

Add some reporting gear to continuously generate compliance status
reports, based on (new) machine-redable metadata.

New metadata:

- config_policies(.yaml): defines rules, and maps them to cfg-elements
- config_responsibles(.yaml): assigns humans or teams to cfg-elements
- config_status(.yaml): collect cfg element metadata, such as last
  credential update timestamp; expected (but not required) to be updated
  automatically (as opposed to the first two files that are expected to
  be maintained manually)

This first version of said reporting gear will simply categorise all
discovered cfg-elements (e.g. "fully compliant", "no responsible
assigned", ..) and print the qualified cfg-element-names.

Of course, this can be extended w/ additional automation scenarios.


ccwienk:
already pushed (by accident..)

PR #677, opened 2022-01-17
Fix protecode comp version bug

jkrayl:


protecode/scanning_util.py
```diff
@@ -290,15 +294,18 @@ def enumerate_reference_triages():
 
         if self._processing_mode is ProcessingMode.FORCE_UPLOAD:
             logger.info('force-upload - will re-upload all images')
-            images_to_upload += list(resource_group.resources())
+            for upload_element in upload_elements:
+                elements_to_upload.append(upload_element)
```
zkdev, line 298: this loop seems a bit inconvenient

protecode/scanning_util.py
```diff
@@ -338,7 +346,8 @@ def enumerate_reference_triages():
                     'Marked existing product(s) with ID(s) '
                     f"'{','.join([str(p.product_id()) for p in protecode_apps_to_remove])}' "
                     'that had no match in the current group '
-                    f"'{resource_group.component().name}, {resource_group.image_name()}' "
+                    # TODO don't access first element
```
zkdev, line 349: TODO can be removed

protecode/scanning_util.py
```diff
@@ -377,41 +387,45 @@ def enumerate_reference_triages():
...
+                logger.info(f'finished waiting for {protecode_app.product_id()}')
 
+        scan_results = list(
+            wait_for_scan_results(
+                protecode_apps_to_consider=protecode_apps_to_consider,
+            )
+        )
         # apply imported triages for all protecode apps
-        for protecode_app in protecode_apps_to_consider:
+        for protecode_app in scan_results:
```
zkdev, line 428: we should rename many variable and refactor some parts to simplify maintenance, but not part of this PR :)

protecode/util.py
```diff
@@ -137,17 +138,17 @@ def _filter_resources_to_scan(component: cm.Component, resource: cm.Resource):
                 return True
 
         for _, components in component_groups.items():
```
zkdev, line 140: we should use `.values()` instead

zkdev (changes requested):

protecode/scanning_util.py
ccwienk: true - it does not improve readability compared to the previous version (might be slightly more efficient, though). also, since this does not change the semantics, I think this ought to be part of a refactoring-commit

protecode/scanning_util.py
ccwienk: +1 - please keep the semantical focus of this PR on corecting the bug, and add a refactoring commit afterwards.

protecode/util.py
ccwienk: please also do this in a separate refactoring commit

protecode/model.py
```diff
@@ -257,7 +265,7 @@ class UploadResult:
     def __init__(
             self,
             status: UploadStatus,
-            component: Component,
+            component,
```
ccwienk, line 268: what's the expected type?

protecode/model.py
```diff
@@ -32,6 +34,12 @@ class ProcessingStatus(Enum):
     FAILED = 'F'
 
 
+@dataclasses.dataclass(frozen=True)
+class UploadElement:
```
ccwienk, line 38: I think this does not reasonably belong to protecode's model module. Might be a better fit for either `gci.componentmodel`, or our local `cnudie` package. The name is also too strongly influenced by the use-case for which it was added. It is in fact a two-tuple of a resource and its resource. Why not simply call it ComponentResource.

for the time being, at least mv it to protecode.scanning_util

protecode/scanning_util.py
```diff
@@ -234,30 +235,33 @@ def _metadata(
 
     def upload_container_image_group(
         self,
-        resource_group: ResourceGroup,
+        upload_elements: typing.List[UploadElement],
```
ccwienk, line 238: do you actually need a list? if so, you could use the shorthand form `list[T]`. if not, rather use `typing.Iterable`

protecode/scanning_util.py
```diff
@@ -368,7 +378,7 @@ def enumerate_reference_triages():
                             f'{resource.access.imageReference=} no longer available to protecode '
                             f'- will upload. Corresponding product: {protecode_app.product_id()}'
                         )
-                        images_to_upload.append(resource)
+                        elements_to_upload.append(upload_element)
```
ccwienk, line 381: naming-nit: the term we use in cnudie would be either `resources`, or - covering also `sources` - `artefacts`. Please choose either `artefact` or `resource`, if doing a rename.

note that currently, the artefacts will always be `OciImages`, as we did so far not implement this for any other artefact types.

protecode/scanning_util.py
```diff
@@ -377,41 +387,45 @@ def enumerate_reference_triages():
                 self._api.rescan(protecode_app.product_id())
 
         # upload new images
-        for resource in images_to_upload:
+        for upload_element in elements_to_upload:
             try:
                 scan_result = self._upload_image(
```
ccwienk, line 392: nit: if thinking about renaming the variable names (image -> _artefact_), the method names ought also be renamed. to have a chance of eventually finalising this PR, please revert those name-changes back for the time being, so we can keep the focus on correcting the bug. do refactoring later.

ccwienk (changes requested):

ccwienk:
obsolete (was split into two PRs)

PR #679, opened 2022-01-19
refactor protecode upload coding

jkrayl:


protecode/scanning_util.py
```diff
@@ -231,40 +227,45 @@ def _metadata(
...
+        product_id_to_resource: dict[str, cnudie.util.ComponentResource] = dict()
+        protecode_products_to_consider = list() # consider to rescan; return results
+        protecode_products_to_remove = set()
+        component_resources_to_upload = list()
         triages_to_import = set()
-        product_id_to_resource = dict()
+
+        # HACK since the component and resource name are the same for all elements 
+        # (only the component and resource version differ) we can use the names for all elements
+        first_component_resource = next(component_resources.__iter__())
```
ccwienk, line 247: according to typehint, `component_resources` might be any iterable (including e.g. a generator - or, more generically speaking - an iterable that will always return the same iterator object). therfore, using this pattern (and declaring any iterable will be okay) will lead to a subtle bug if such a "non-repeatable" iterable is passed-in: if this happens, the first element will be removed from it, and it will not be included later in this function.

therefore, I deem it preferable to either:
- change typehint to `list[T]` and use `component_resources[0]` to get the first element (this will at least fail during runtime if e.g. a generator is passed in)
or:
- convert iterable into a list (or tuple). also in the second case, I would deem using `[0]` to get the first element more idiosyncratic (+ easier to read / shorter)

protecode/scanning_util.py
```diff
@@ -284,104 +285,108 @@ def enumerate_reference_triages():
...
-        # upload new images
-        for resource in images_to_upload:
+        # upload new resources
+        for component_resource in component_resources_to_upload:
             try:
+                logger.info(
+                    f'uploading resource with name {component_resource.resource.name} '
+                    f'and version {component_resource.resource.version}'
+                )
                 scan_result = self._upload_image(
```
ccwienk, line 387: also rename method name (image -> resource)?

protecode/util.py
```diff
@@ -94,9 +93,15 @@ def _upload_tasks():
...
+        def group_by_resource_name(
+            components
+        ) -> typing.Generator[
+            list[cnudie.util.ComponentResource],
+            None,
+            None
+        ]:
             # groups resources of components by resource name
-            resource_groups = collections.defaultdict(list)
+            resource_groups = collections.defaultdict(list[cnudie.util.ComponentResource])
```
ccwienk, line 104: interesting. I would not have guessed that this works (considering the argument of defaultdict is used to construct new instances for absent keys). Seems like a good pattern - did you come up w/ that?

protecode/util.py
```diff
@@ -111,8 +116,8 @@ def group_resources(components):
                         )
                     )
 
-            for resource_name, resources in resource_groups.items():
-                yield resources
+            for component_resource in resource_groups.values():
```
ccwienk, line 119: nit: you might use `yield from` (since you are already doing a refactoring..)

ccwienk (changes requested):

protecode/model.py
```diff
@@ -257,7 +257,7 @@ class UploadResult:
     def __init__(
             self,
             status: UploadStatus,
-            component: Component,
+            component,
```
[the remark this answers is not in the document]
jkrayl: we are not entirely sure which type is correct here. In this method we only use cnudie components but @zkdev was sure that it was used as a protecode component in another function

protecode/util.py
jkrayl: well I just added the type hint and it worked 😅

ccwienk (approved):

PR #681, opened 2022-01-24
Upload protecode product url to db

zkdev:
This PR extends the uploaded metadata from protecode scans by a protecode product url taken from protecode api response.
Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/protecode/util.py
+++ b/protecode/util.py
@@ -170,7 +170,6 @@ def upload_grouped_images(
         _upload_results_to_db(
             results=relevant_results + results_below_threshold,
             client=delivery_client,
-            api_url=protecode_cfg.api_url(),
         )
     else:
         logger.warning('not uploading results to deliverydb, client not available')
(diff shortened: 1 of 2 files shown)


protecode/util.py
```diff
@@ -343,6 +347,14 @@ def upload_result_to_compliance_issue(
     issue = dso.model.ComplianceIssue(
         artefact=artefact_ref,
         meta=meta,
-        data={'greatestCvss3Score': greatest_cvss3_score},
+        data={
+            'greatestCvss3Score': greatest_cvss3_score,
+            'protecodeProductUrl': ci.util.urljoin(
```
ccwienk, line 352: I think it should be possible to read this url from either the protecode-results, or using the protecode-client, should it not?

ccwienk (changes requested):
see inline comment

protecode/util.py
zkdev: had to extend the model, product url is now taken from protecode api response :)

ccwienk (approved):

PR #684, opened 2022-02-14
Add support for github ssh-key autorotation

AndreasBurger:
Adds rotation of our technical users' ssh keys.

__Note__: This will remove all comments from the files affected. We should check whether we need anything stored in there before merging.

### DIFF
--- a/cfg_mgmt/github.py
+++ b/cfg_mgmt/github.py
@@ -8,6 +8,7 @@ import yaml
 from Crypto.PublicKey import RSA
 
 import ccc.github
+import cfg_mgmt
 import cfg_mgmt.util
 import ci.util
 import model
@@ -35,7 +36,7 @@ def create_secret_and_persist_in_cfg_repo(
     cfg_dir: str,
     cfg_element: GithubConfig,
     cfg_metadata: CfgMetadata,
-):
+) -> cfg_mgmt.revert_function:
     cfg_factory = model.ConfigFactory.from_cfg_dir(cfg_dir, disable_cfg_element_lookup=True)
 
     local_sources = cfg_mgmt.util.local_cfg_type_sources(
--- a/cfg_mgmt/github.py
+++ b/cfg_mgmt/github.py
@@ -1,0 +1,286 @@
+import dataclasses
+import datetime
+import logging
+import os
+import typing
+import yaml
+
+from Crypto.PublicKey import RSA
+
+import ccc.github
+import cfg_mgmt.util
+import ci.util
+import model
+import model.github
+
+from cfg_mgmt.model import (
+    CfgQueueEntry,
+    CfgTarget,
+    CfgStatus,
+    CfgMetadata,
+    cfg_queue_fname,
+    cfg_status_fname,
+)
+from model.github import (
+    GithubConfig,
+    GithubCredentials,
+)
+
+
+ci.log.configure_default_logging()
+logger = logging.getLogger(__name__)
+
+
+def create_secret_and_persist_in_cfg_repo(
+    cfg_dir: str,
+    cfg_element: GithubConfig,
+    cfg_metadata: CfgMetadata,
+):
+    cfg_factory = model.ConfigFactory.from_cfg_dir(cfg_dir, disable_cfg_element_lookup=True)
+
+    local_sources = cfg_mgmt.util.local_cfg_type_sources(
+        cfg_element=cfg_element,
+        cfg_factory=cfg_factory,
+    )
+
+    if len(local_sources) != 1:
+        raise NotImplementedError(
+            'Do not know how to rotate github configs sourced from more than one file.'
+        )
+
+    source_file = next((s for s in local_sources))
+
+    known_github_configs: typing.Iterable[GithubConfig] = [
+        c for c in cfg_factory._cfg_elements(
+            cfg_type_name=cfg_element._type_name,
+        )
+    ]
+
+    cfg_name = cfg_element.name()
+    if not (cfg_to_rotate := next(
+        (c for c in known_github_configs if c.name() == cfg_name),
+        None,
+    )):
+        raise RuntimeError(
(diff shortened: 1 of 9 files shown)


cfg_mgmt/github.py
```diff
@@ -0,0 +1,326 @@
...
+    GithubConfig,
+)
+
+
+ci.log.configure_default_logging()
+logger = logging.getLogger(__name__)
+
+
+def rotate_github_config(
+    cfg_dir: str,
```
ccwienk, line 37: we added a new parameter `cfg_factory` to `rotate_cfg_element` yesterday; -> you might want to follow suit and also pass-in cfg-factory to this function

cfg_mgmt/github.py
```diff
@@ -0,0 +1,326 @@
...
+
+    known_github_configs: typing.Iterable[GithubConfig] = [
+        c for c in cfg_factory._cfg_elements(
+            cfg_type_name=cfg_element._type_name,
+        )
+    ]
+
+    cfg_name = cfg_element.name()
+    if not (cfg_to_rotate := next(
+        filter(lambda c: c.name() == cfg_name, known_github_configs),
```
ccwienk, line 68: nit: rather use generator- or list-comprehension?

cfg_mgmt/github.py
```diff
@@ -0,0 +1,326 @@
...
+        'public_key': update_user(cfg_to_rotate, u.username()),
+    } for u in cfg_to_rotate._technical_users()]
+
+    _write_github_configs(
+        cfg_dir=cfg_dir,
+        cfg_file_name=source_file,
+        github_configs=known_github_configs,
+    )
+
+    _update_config_queue(
```
ccwienk, line 86: naming-nit: `_add_to` (or append, if you prefer - choose a verb matching the semantics more specifically than update, please)

cfg_mgmt/github.py
```diff
@@ -0,0 +1,326 @@
...
+
+        f'changed element: {config_element.name()}'
+    )
+    if ci.util._running_on_ci():
+        commit_message += f'\nJob url: {concourse.util.own_running_build_url}'
+
+    return commit_message
+
+
+def _push_to_github(
```
ccwienk, line 143: the name does not match the semantics -> better: create_commit_and_push?
also, in theory, the push-target would not have to be a github-hosted git repository (just omit gh?)

cfg_mgmt/github.py
```diff
@@ -0,0 +1,326 @@
...
+    username: str,
+) -> str:
+    credential = github_config.credentials(technical_user_name=username)
+    gh_api = ccc.github.github_api(github_config, username=username)
+    old_private_key = credential.private_key()
+
+    # current private key uses EdDSA which is not supported by Pycryptodome
+    try:
+        old_public_key = _corresponding_public_key(old_private_key)
+    except ValueError:
```
ccwienk, line 244: is there a more specific execption available? also, I would at least log a warning in this case. also: let us add a note that we plan to remove this try-execpt after first rotation

cfg_mgmt/github.py
```diff
@@ -0,0 +1,326 @@
...
+    private_key, public_key = _create_key_pair()
+    gh_api.create_key(
+        title='Auto-generated key',
+        key=public_key,
+    )
+    credential.set_key(private_key)
+    return old_public_key
+
+
+def undo_update(
```
ccwienk, line 256: actually, I would prefer to keep the newly registered public-key in memory, and use it "directly" for reverting.

probably, this can simply be inlined.

cfg_mgmt/github.py
```diff
@@ -0,0 +1,326 @@
...
+        # rendered types we create our own Dumper.
+        if isinstance(data, str) and '\n' in data:
+            return self.represent_scalar(u'tag:yaml.org,2002:str', data, style='|')
+        # Also, don't include keys with None/null values.
+        if data is None:
+            return self.represent_scalar('tag:yaml.org,2002:null', '')
+        return super().represent_data(data)
+
+
+def test():
```
ccwienk, line 309: rm test code....

cfg_mgmt/rotate.py
```diff
@@ -43,5 +44,14 @@ def rotate_cfg_element(
...
         return False
 
+    if type_name == 'github':
+        cmgh.rotate_github_config(
+            cfg_dir=cfg_dir,
+            cfg_element=cfg_element,
+            git_helper=git_helper,
+            target_ref=target_ref,
+            cfg_metadata=cfg_metadata
+        )
```
ccwienk, line 54: return True, please

model/github.py
```diff
@@ -169,6 +172,9 @@ def set_auth_token(self, auth_token):
     def private_key(self):
         return self.raw.get('privateKey')
 
+    def set_key(self, key: str):
+        self.raw['privateKey'] = key
```
ccwienk, line 176: hmmm..... I am not too sure about that tbh :-/
-> all other model-classes only have readonly-methods - this would be the very first one that deviates from that :-/

ccwienk (changes requested):

cfg_mgmt/github.py
AndreasBurger: Added a TODO, but unfortunately that is the error raised by Pycryptodome.

cfg_mgmt/github.py
AndreasBurger: While I think that this makes it, as a whole, less readable and provides no benefit that I can see. The public key has to be deterministic for a given key as far as I understand it.
Still, I've changed it to pass the generated key through.

cfg_mgmt/github.py
AndreasBurger: Done :(

model/github.py
AndreasBurger: Done, albeit at the price of direct access to the raw config in the added module instead. I'd have expected us to add more setters if/when we add more rotation code, but we can also revisit the issue then.

PR #688, opened 2022-02-21
Refactor cfg-element-secret deletion

AndreasBurger:
Also enables it for github

### DIFF
--- a/cfg_mgmt/gcp.py
+++ b/cfg_mgmt/gcp.py
@@ -124,3 +124,18 @@ def create_secret_and_persist_in_cfg_repo(
         )
 
     return revert
+
+
+def delete_config_secret(
+    cfg_element: model.container_registry.ContainerRegistryConfig,
+    cfg_dir: str,
+    cfg_queue_entry: cmm.CfgQueueEntry,
+):
+    logger.info('deleting old gcr secret')
+    iam_client = ccc.gcp.create_iam_client(
+        cfg_element=cfg_element,
+    )
+    delete_service_account_key(
+        iam_client=iam_client,
+        service_account_key_name=cfg_queue_entry.secretId['gcp_secret_key'],
+    )
--- a/cfg_mgmt/rotate.py
+++ b/cfg_mgmt/rotate.py
@@ -4,9 +4,6 @@ import os
 import typing
 import yaml
 
-import traceback
-
-import ccc.gcp
 import cfg_mgmt.gcp as cmg
 import cfg_mgmt.github as cmgh
 import cfg_mgmt.model as cmm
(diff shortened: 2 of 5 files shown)


cfg_mgmt/gcp.py
```diff
@@ -124,3 +124,19 @@ def revert():
         )
 
     return revert
+
+
+def delete_config_secret(
+    cfg_factory: model.ConfigFactory,
+    cfg_queue_entry: cmm.CfgQueueEntry,
```
ccwienk, line 131: how about passing-in the cfg-element, rather than the cfg-factory? 🤔

cfg_mgmt/gcp.py
```diff
@@ -124,3 +124,19 @@ def revert():
...
+    cfg_element = cfg_factory.container_registry(cfg_queue_entry.target.name)
+    logger.info('deleting old gcr secret')
+    iam_client = ccc.gcp.create_iam_client(
+        cfg_element=cfg_element,
+    )
+    delete_service_account_key(
+        iam_client=iam_client,
+        service_account_key_name=cfg_queue_entry.secretId['gcp_secret_key'],
+    )
+    return True
```
ccwienk, line 142: what should be the semenatics of returning `False`? I think we should rather rely on exceptions (as this code does).

-> omit return-value

cfg_mgmt/rotate.py
```diff
@@ -43,29 +42,35 @@ def delete_cfg_element(
     git_helper: gitutil.GitHelper,
     target_ref: str,
 ) -> bool:
-    did_remove = False
-    if cfg_queue_entry.target.type == 'container_registry':
+
+    delete_func: typing.Callable[[model.ConfigFactory, cmm.CfgQueueEntry], bool] = None
```
ccwienk, line 46: as commented above: let us simply omit return value / return None

cfg_mgmt/rotate.py
```diff
@@ -43,29 +42,35 @@ def delete_cfg_element(
...
+
+    try:
+        did_remove = delete_func(cfg_factory=cfg_fac, cfg_queue_entry=cfg_queue_entry)
+    except Exception as e:
+        logger.error(
+            f"error deleting secret for cfg-type '{type_name}' with name "
+            f"'{cfg_queue_entry.target.name}': {e}."
+        )
+        traceback.print_exc()
+        did_remove = False
```
ccwienk, line 73: let us rather early-exit upon errors, wdyt (thus reducing indentation level for subsequent lines of code by one)?

ccwienk (changes requested):

PR #689, opened 2022-02-21
Factor out config analysis

zkdev:
`create_report` used to analyse a config status and print a report.
With the introduction of multiple elasticsearch metrics, the analysis
was extended by several semantics to calculate those metrics.
This commit separates the analysis and reporting to have clear concerns.

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/cfg_mgmt/model.py
+++ b/cfg_mgmt/model.py
@@ -180,15 +180,15 @@ class CfgMetadata:
     queue: list[CfgQueueEntry]
 
 
-@dataclasses.dataclass
+@dataclasses.dataclass(frozen=True)
 class CfgStatusAnalysis:
     fullyCompliant: bool
     hasResponsible: bool
     hasRule: bool
     assignedRuleRefersToUndefinedPolicy: bool
     hasStatus: bool
-    requiresStatus: bool = None
-    credentialsOutdated: bool = None
+    requiresStatus: typing.Optional[bool]
+    credentialsOutdated: typing.Optional[bool]
 
 
 @dataclasses.dataclass
--- a/cfg_mgmt/model.py
+++ b/cfg_mgmt/model.py
@@ -181,7 +181,7 @@ class CfgMetadata:
 
 
 @dataclasses.dataclass(frozen=True)
-class CfgStatusAnalysis:
+class CfgStatusEvaluationResult:
     fullyCompliant: bool
     hasResponsible: bool
     hasRule: bool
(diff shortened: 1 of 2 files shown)


cfg_mgmt/reporting.py
```diff
@@ -30,6 +30,60 @@ class CfgElementStatusReport:
...
+    if not cfg_element_status.rule:
+        analysis.fullyCompliant = False
+        analysis.hasRule = False
+
+    elif not cfg_element_status.policy:
+        analysis.fullyCompliant = False
+        analysis.assignedRuleRefersToUndefinedPolicy = True
+
+    else:
+        # have rule w/ policy
```
ccwienk, line 58: instead of writing this comment, you might express this condition with an additional `elif` (as opposed to using an else w/ implicit conditions). also add `else: raise NotImplemented # or RuntimeError` to guard against programming errors

cfg_mgmt/reporting.py
```diff
@@ -30,6 +30,60 @@ class CfgElementStatusReport:
     status: typing.Optional[cmm.CfgStatus]
 
 
+def analyse_cfg_element_status(
+    cfg_element_status: CfgElementStatusReport,
+) -> cmm.CfgStatusAnalysis:
+
+    analysis = cmm.CfgStatusAnalysis(
```
ccwienk, line 37: as a general pattern, I would rather collect the state in variables, and create the result object at the end (rather than mutating its state)

cfg_mgmt/reporting.py
```diff
@@ -30,6 +30,60 @@ class CfgElementStatusReport:
     status: typing.Optional[cmm.CfgStatus]
 
 
+def analyse_cfg_element_status(
```
ccwienk, line 33: naming-nit: I would consider this operation to be more adequately described by either `determine`, or `calculate`

cfg_mgmt/model.py
```diff
@@ -180,6 +180,17 @@ class CfgMetadata:
     queue: list[CfgQueueEntry]
 
 
+@dataclasses.dataclass
+class CfgStatusAnalysis:
```
ccwienk, line 184: I think the name is misleading. This is the _result_ of an analysis, maybe, but does not represent the analysis itself. Naming seems a bit difficult here, since we already have a `CfgReportingSummary`. How about `CfgStatusEvaluation(Result)`? Also, a brief docstr explaining the interrelationship between CfgStatus and the (derived) CfgStatusEvaluation(Result) might help.

ccwienk (changes requested):
while in general the idea seems nice - is there actually a second usage? 🤔

PR #691, opened 2022-02-23
Refactor config-rotation

AndreasBurger:
Changes to the config-repository, the config-file of the targeted config-element and the config-metadata are now separated from the code performing the actual rotation/deletion. These tasks are handled the same for all config-elements.

__Note:__ Not yet tested.


cfg_mgmt/rotate.py
```diff
@@ -57,53 +53,44 @@ def delete_cfg_element(
...
 def rotate_cfg_element(
-    cfg_dir: str,
     cfg_element: model.NamedModelElement,
-    target_ref: str,
-    github_cfg: model.github.GithubConfig,
-    cfg_metadata: cmm.CfgMetadata,
-    github_repo_path: str,
-) -> bool:
+    cfg_factory: model.ConfigFactory,
+) -> typing.Union[typing.Tuple[cfg_mgmt.revert_function, dict, model.NamedModelElement], None]:
```
AndreasBurger, line 77: This differs from what we discussed. Since the values returned by this functions are actually needed, I've opted to switch from using `True`/`False` (and Errors) to using truthy/falsy values.
Not sure if `None` or an empty tuple are the best choice to signal noop.

ccwienk (approved):

PR #693, opened 2022-02-28
implement multiarch build and publishing

ccwienk:
If OciBuilder is set to the (new) `docker-buildx`, and one or more
target-platforms are defined, run each platform-specific build in a
separate job. Add platform-identifier as suffix to target-tag.

As last operation, collect all variant-images (or, more specifically,
their oci-manifests' digests), and publish an "manifest-list" the
configured push-target.

Configuring build to target multiple platforms should be
backwards-compatible for all OCI-Clients that can properly work w/
manifest-lists (this is known to be true for docker and containerd).

Enables multiarch-builds, including (but not limited to) arm64.

### DIFF
--- a/concourse/model/traits/publish.py
+++ b/concourse/model/traits/publish.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 
 import enum
+import textwrap
 import typing
 
 import dacite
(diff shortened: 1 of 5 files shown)


jensh007 (approved):

concourse/model/traits/publish.py
```diff
@@ -231,21 +287,44 @@ def _children(self):
...
+                for name, args
+                in image_dict.items()
+            )
+
+    def platforms(self) -> typing.Optional[list[str]]:
+        '''
+        the list of explicitly configured build platforms
+        guaranteed to be either non-empty, or None
+        '''
+        return self.raw.get('platforms', None)
```
zkdev, line 309: Think about omitting `None` since its the default value anyways.

zkdev (approved):
/lgtm
just two nits

PR #694, opened 2022-02-28
Sanitise ES logs (reduce tracebacks to only one time print)

zkdev:
This PR sanitises ES logging.
If no concourse metadata is present, each `emit` printed a traceback.

On `OciRequestHandler` initialisation, cc-metadata availability is checked and if absent a warning is printed.
The availability status is stored in state and an early-exit returns `emit`.

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/ccc/oci.py
+++ b/ccc/oci.py
@@ -3,7 +3,9 @@ import logging
 import traceback
 import typing
 
+import ccc.concourse
 import ccc.elasticsearch
+import ctx
 import oci.auth as oa
 import oci.client as oc
 import model.container_registry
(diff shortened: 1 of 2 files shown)


ccc/oci.py
```diff
@@ -72,14 +73,19 @@ class _OciRequestHandler(logging.Handler):
     def __init__(
         self,
         level,
-        es_client,
+        es_client: ccc.elasticsearch.ElasticSearchClient,
```
ccwienk, line 76: since it is very unlikely that the checked condition will change, rather than adding an early-exit, I would deem it more reasonable to simply not register this request-handler in the first place.

ccwienk (changes requested):
factoring out the lookup of "metadata-file" certainly is okay. however, the `meta` step (and thus said metadata file) ought to be present in all of our pipelines, as it is always injected (see concourse/factory.py).

So in this case, I would rather find out why said meta-step's metadata file does not seem to be present, rather than silencing resulting warnings/error output.

concourse/util.py
```diff
@@ -179,6 +179,23 @@ def own_running_build_url(cfg_factory=None):
     )
 
 
+def meta_info_file_from_env() -> str:
+    meta_dir = os.path.join(
+        os.path.abspath(check_env('CC_ROOT_DIR')),
+        concourse.model.traits.meta.DIR_NAME
+    )
+    return os.path.join(
+        meta_dir,
```
ccwienk, line 188: os.path.join accepts an arbitrary amount of arguments (so you might combine those two calls)

also, rather invoke os.path.abspath for the resulting path, rather just the root-dir)

ccwienk (approved):

PR #695, opened 2022-03-07
Add support for auto-rotation of Azure service principals

AndreasBurger:
Note: not yet fully tested as there seems to be some issues with the user I've used for testing so far.


cfg_mgmt/azure.py
```diff
@@ -0,0 +1,183 @@
...
+import msal
+import requests
+
+import cfg_mgmt
+import model
+import model.azure
+
+from cfg_mgmt.model import CfgQueueEntry
+
+AZURE_APP_ROOT_URL = 'https://graph.microsoft.com/v1.0/applications'
```
ccwienk, line 15: inline as parameter w/ default value, please

cfg_mgmt/azure.py
```diff
@@ -0,0 +1,183 @@
...
+    displayName: str
+    keyId: str
+    endDateTime: str
+    startDateTime: typing.Optional[str]
+    customKeyIdentifier: typing.Optional[str]
+    hint: typing.Optional[str]
+    secretText: typing.Optional[str]  # the actual password, if present
+
+    def __post_init__(self):
+        self.endDateTime = _to_datetime(self.endDateTime)
```
ccwienk, line 29: while this works, I would prefer to mv `_to_datetime`'s code to before those usages
also: how about adding a "getter-method" doing the conversion (rather than altering the parsed attrs)?

cfg_mgmt/azure.py
```diff
@@ -0,0 +1,183 @@
...
+    else:
+        raise RuntimeError(f'Unexpected response: {response}')
+
+    return access_token
+
+
+def _to_datetime(
+    azureDateString:str,
+):
+    # Azure ISO date format has to be converted manually to datetime
```
ccwienk, line 60: rather use `dateutil.parser.isoparse` - this should make this function obsolete.

cfg_mgmt/azure.py
```diff
@@ -0,0 +1,183 @@
...
+    service_principal: model.azure.AzureServicePrincipal,
+    access_token: str,
+) -> typing.Iterable[PasswordCredential]:
+    response = requests.get(
+        url=f'{AZURE_APP_ROOT_URL}/{service_principal.object_id()}',
+        headers={'Authorization': f'Bearer {access_token}'},
+    )
+    response_dict = response.json()
+    if not 'passwordCredentials' in response_dict:
+        # no keys
```
ccwienk, line 78: log a warning?

cfg_mgmt/azure.py
```diff
@@ -0,0 +1,183 @@
...
+def _get_passwords(
+    service_principal: model.azure.AzureServicePrincipal,
+    access_token: str,
+) -> typing.Iterable[PasswordCredential]:
+    response = requests.get(
+        url=f'{AZURE_APP_ROOT_URL}/{service_principal.object_id()}',
+        headers={'Authorization': f'Bearer {access_token}'},
+    )
+    response_dict = response.json()
+    if not 'passwordCredentials' in response_dict:
```
ccwienk, line 77: stylistic nit: you could deduplicate dereferentiation of `passwordCredentials` by using an assignment expression (aka walrus operator)

cfg_mgmt/azure.py
```diff
@@ -0,0 +1,183 @@
...
+        # no keys
+        return []
+
+    return [PasswordCredential(**e) for e in response_dict['passwordCredentials']]
```
ccwienk, line 81: preferably use `dacite` - this tends to be more robust (e.g. will not break upon extra attrs)

cfg_mgmt/azure.py
```diff
@@ -0,0 +1,183 @@
...
+):
+    # Azure ISO date format has to be converted manually to datetime
+    pos = azureDateString.find('.')
+    if pos > 0:
+        azureDateString = azureDateString[:pos] + 'Z'
+
+    return datetime.datetime.strptime(azureDateString, '%Y-%m-%dT%H:%M:%SZ')
+
+
+def _get_passwords(
```
ccwienk, line 68: naming-nit: since this function is not a simple "getter" (as in: returns an attr from a "POJO"), a verb such as `retrieve` would feel more adequate to me.
also, how about naming it `retrieve_password_credentials`, considering you also thus named the type (and I assume there be more credential types..)

cfg_mgmt/azure.py
```diff
@@ -0,0 +1,183 @@
...
+    response = requests.post(
+        f'{AZURE_APP_ROOT_URL}/{service_principal.object_id()}/addPassword',
+        json=body,
+        headers={'Authorization': f'Bearer {access_token}'},
+    )
+    response.raise_for_status()
+    response_dict = response.json()
+
+    # filter metadata fields from the returned response (by filtering unexpected fields).
+    dataclass_fields = PasswordCredential.__annotations__.keys()
```
ccwienk, line 106: use dacite..

cfg_mgmt/azure.py
```diff
@@ -0,0 +1,183 @@
...
+    return PasswordCredential(**filtered_response)
+
+
+def _remove_password(
+    service_principal: model.azure.AzureServicePrincipal,
+    key_id: str,
+    access_token: str,
+):
+    response = requests.post(
+        f'{AZURE_APP_ROOT_URL}/{service_principal.object_id()}/removePassword',
```
ccwienk, line 118: considering the amount of routes, I would suggest to use a routes-class, as we do in other api-clients - wdyt?

as a general remark, please use dependency-injection. base-url should be a function parameter. initialise it w/ a default value for convenience, fine w/ me. but avoid consuming from global/module scope, please.

cfg_mgmt/azure.py
```diff
@@ -0,0 +1,183 @@
...
+        json={"keyId": key_id},
+        headers={'Authorization': f'Bearer {access_token}'},
```
ccwienk, line 120: again, there seems to be quite some redundancy. so a client-object, offering authenticated get/post might be better

cfg_mgmt/azure.py
```diff
@@ -0,0 +1,183 @@
...
+    )
+    new_element.raw['client_secret'] = new_password_credential.secretText
+
+    secret_id = {'keyId': current_password_credential.keyId}
+
+    def revert_function():
+        _remove_password(
+            service_principal=cfg_element,
+            key_id=new_password_credential.keyId,
+            access_token=access_token,  # token is good for 20 minutes by default,
```
ccwienk, line 165: can we read validity from the token?

cfg_mgmt/azure.py
```diff
@@ -0,0 +1,183 @@
...
+
+
+def delete_config_secret(
+    cfg_element: model.azure.AzureServicePrincipal,
+    cfg_factory: model.ConfigFactory,
+    cfg_queue_entry: CfgQueueEntry,
+):
+    key_id = cfg_queue_entry.secretId['keyId']
+    access_token = _get_access_token_for_principal(service_principal=cfg_element)
+    _remove_password(
```
ccwienk, line 179: ..password_credential?

also, rather than add/remove, how about `register`/`unregister` instead?

ccwienk (changes requested):
see inline commments

PR #700, opened 2022-03-23
Rm concourse externalUrl

zkdev:
https://github.com/gardener/cc-utils/commit/fbec124ac26c5f432a3e20b356518e0f4b7c3cda introduced a `concourse endpoint` config attribute. We should rather use the referenced `concourse_endpoint cfg` to retrieve the `externalUrl`.

This PR removes the `externalUrl` attribute and replaces its usages with `concourse_endpoint cfg`.

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/ccc/elasticsearch.py
+++ b/ccc/elasticsearch.py
@@ -65,10 +65,16 @@ def _metadata_dict():
     if not ci.util._running_on_ci():
         return {}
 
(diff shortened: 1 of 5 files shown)


AndreasBurger (approved):
lgtm

PR #702, opened 2022-03-24
Introduce trait to omit github releases

zkdev:
This PR adds the `release_on_github` trait (default `True`) to controller whether a github release is published on release.
As some steps rely on `GitHubReleaseStep`, the `release_on_github` condition is used multiple times.

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/concourse/steps/release.py
+++ b/concourse/steps/release.py
@@ -720,7 +720,7 @@ class UploadComponentDescriptorStep(TransactionalStep):
         else:
             upload_ctf(ctf_path=self.ctf_path)
 
-        def attach_cd_to_release():
+        def upload_component_descriptor_as_release_asset():
             try:
                 release = self.github_helper.repository.release_from_tag(release_tag_name)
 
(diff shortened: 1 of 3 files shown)


concourse/steps/release.py
```diff
@@ -717,27 +720,31 @@ def upload_ctf(ctf_path: str):
         else:
             upload_ctf(ctf_path=self.ctf_path)
 
-        try:
-            release = self.github_helper.repository.release_from_tag(release_tag_name)
+        def attach_cd_to_release():
```
ccwienk, line 723: naming nits:
- rather not use `cd` as an abbreviation for component-descriptor
- how about: `upload_component_descriptor_as_release_asset`?

concourse/steps/release.py
```diff
@@ -1064,19 +1072,21 @@ def release_and_prepare_next_dev_cycle(
...
-    step_list.append(github_release_step)
+    if release_on_github:
+        github_release_step = GitHubReleaseStep(
+            github_helper=github_helper,
+            githubrepobranch=githubrepobranch,
+            repo_dir=repo_dir,
+            component_name=component_name,
+            release_version=release_version,
+        )
+        step_list.append(github_release_step)
```
ccwienk, line 1083: nit: step could be inlined (which is a refactoring, not necessarily part of this PR)

PR #704, opened 2022-03-31
Added model and rotation of BTP service bindings

MartinWeindel:
In the context of the enterprise policy filter (gardener-extension-shoot-networking-filter) we need to rotate certificates and service bindings.
This first PR is about rotating service bindings.
Using a service binding to a BTP service manager instance for authentication, service bindings in the same subaccount can be created and deleted. "Rotating" a service binding therefore means here to create a new one and delete the old one.

Fixes #

**Release note**:

```feature operator
Added model and rotation of BTP service bindings
```

/invite @AndreasBurger

### DIFF
--- a/cfg_mgmt/btp_service_binding.py
+++ b/cfg_mgmt/btp_service_binding.py
@@ -1,8 +1,8 @@
 import copy
-import json
 import logging
 import typing
 import requests
+from dataclasses import dataclass
 
 import cfg_mgmt
 import cfg_mgmt.model as cmm
@@ -16,27 +16,31 @@ ci.log.configure_default_logging()
 logger = logging.getLogger(__name__)
 
 
-def _filter_service_bindings(
-    result: dict,
+@dataclass
+class ServiceBindingInfo:
+    id: str
+    name: str
+    service_instance_id: str
+
+
+def _find_next_service_binding_serial_number(
+    bindings: list[ServiceBindingInfo],
     binding_name_prefix: str,
     instance_id: str,
-) -> tuple[list[str], int]:
-    outdated_ids = []
-    next = 1
-    for item in result["items"]:
-        name = item["name"]
-        if name.startswith(binding_name_prefix) and item["service_instance_id"] == instance_id:
+) -> int:
+    next_sn = 1
+    for item in bindings:
+        if item.name.startswith(binding_name_prefix) and item.service_instance_id == instance_id:
             try:
-                n = int(name[len(binding_name_prefix):])
-                outdated_ids.append(item["id"])
-                if n >= next:
-                    next = n + 1
+                n = int(item.name[len(binding_name_prefix):])
+                if n >= next_sn:
+                    next_sn = n + 1
             except ValueError:
-                logger.warn("ignored {}".format(name))
-    return outdated_ids, next
+                logger.warn(f'ignored {item.name}')
+    return next_sn
 
 
-class sbClient:
+class SBClient:
     def __init__(
         self,
         sm_url: str,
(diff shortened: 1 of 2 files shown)


cfg_mgmt/btp_service_binding.py
```diff
@@ -0,0 +1,159 @@
...
+        result = resp.json()
+        id = result["id"]
+        logger.info('Creating service binding {} ({})'.format(binding_name, id))
+        return id, result["credentials"]
+
+    def get_service_bindings(self) -> dict:
+        url = "{}/v1/service_bindings".format(self.sm_url)
+        headers = {
+            'Accept': 'application/json',
+            'Authorization': 'Bearer {}'.format(self.access_token),
```
ccwienk, line 81: since there is a lot of redundancy w.r.t. setting those headers, how about either defining a wrapper method for requests.request|get|post that will patch-in those?

example: https://github.com/gardener/cc-utils/blob/master/oci/client.py#L356-L421 (most relevantly consider at last line; requests.Session is (almost) identical to requests w.r.t. signatures)

cfg_mgmt/btp_service_binding.py
```diff
@@ -0,0 +1,159 @@
...
+
+
+ci.log.configure_default_logging()
+logger = logging.getLogger(__name__)
+
+
+def _filter_service_bindings(result: dict, binding_name_prefix: str, instance_id: str) -> tuple[list[str], int]:
+    outdated_ids = []
+    next = 1
+    for item in result["items"]:
```
ccwienk, line 22: rather pass-in sequence of items, instead of a dict w/ expected key `items`?

it might be worth considering to define a dataclass (convenient deserialisation is offered by `dacite`), e.g.:

```
@dataclasses.dataclass
class ServiceBinding:
 name: str
 service_instance_id: str

 
 # you may optionally add methods..
 def idx(self):
   # add code for parsing idx here?
   return <parsed-idx>

bindings = [dacite.from_dict(data_class=ServiceBinding, data=binding_dict) for binding_dict in bindings]

for binding in bindings:
  if binding.name .....
```

cfg_mgmt/btp_service_binding.py
```diff
@@ -0,0 +1,159 @@
...
+import model.container_registry
+
+
+ci.log.configure_default_logging()
+logger = logging.getLogger(__name__)
+
+
+def _filter_service_bindings(result: dict, binding_name_prefix: str, instance_id: str) -> tuple[list[str], int]:
+    outdated_ids = []
+    next = 1
```
ccwienk, line 21: nit: `next` shadows built-in function `next`

cfg_mgmt/btp_service_binding.py
```diff
@@ -0,0 +1,159 @@
...
+    for item in result["items"]:
+        name = item["name"]
+        if name.startswith(binding_name_prefix) and item["service_instance_id"] == instance_id:
+            try:
+                n = int(name[len(binding_name_prefix):])
+                outdated_ids.append(item["id"])
+                if n >= next:
+                    next = n + 1
+            except ValueError:
+                logger.warn("ignored {}".format(name))
```
ccwienk, line 31: nit: we consistently use single-quotes in this repository - at least, please do not mix within the same module

cfg_mgmt/btp_service_binding.py
```diff
@@ -0,0 +1,159 @@
...
+    return outdated_ids, next
```
ccwienk, line 32: you never seem to consume the outdated_ids - so why bother to return them?

ccwienk (changes requested):

cfg_mgmt/btp_service_binding.py
```diff
@@ -0,0 +1,159 @@
...
+        if resp.status_code != 201:
+            msg = 'create_service_binding failed: {} {}'.format(resp.status_code, resp.text)
+            logger.error(msg)
+            raise requests.HTTPError(msg)
+        result = resp.json()
+        id = result["id"]
+        logger.info('Creating service binding {} ({})'.format(binding_name, id))
+        return id, result["credentials"]
+
+    def get_service_bindings(self) -> dict:
```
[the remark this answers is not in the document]
MartinWeindel: It's returning a dict. You are right, naming the variable list is confusing.

cfg_mgmt/btp_service_binding.py
```diff
@@ -0,0 +1,159 @@
...
+    data = {
+        'grant_type': 'client_credentials', 
+        'token_format': 'bearer',
+        'client_id': credentials['clientid'],
+        'client_secret': credentials['clientsecret'],
+    }
+    headers = {
+        'Accept': 'application/json',
+    }
+    resp = requests.post(url, data=data, headers=headers)
```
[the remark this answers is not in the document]
MartinWeindel: Here it is expecting x-www-form-urlencoded format, so `data=data` is correct.

cfg_mgmt/btp_service_binding.py
MartinWeindel: In general I agree with removing redundancy, but here I don't see that it would improve the code. There would be more code lines and there will be no reuse.

cfg_mgmt/btp_service_binding.py
MartinWeindel: It's the only place where the results are used, therefore no worth the effort.

MartinWeindel:
@ccwienk 
I appreciate all your hints to improve the python code, thanks a lot!
Most of your change requests have been applied. Only for a few I have a different view and have commented so explicitly.

cfg_mgmt/btp_service_binding.py
```diff
@@ -0,0 +1,159 @@
...
+    def delete_service_binding(self, name: str, id: str):
+        headers = {
+            'Authorization': 'Bearer {}'.format(self.access_token),
+        }
+        url = "{}/v1/service_bindings/{}".format(self.sm_url, id)
+        resp = requests.delete(url, headers=headers)
+        if resp.status_code != 200:
+            msg = 'delete_service_binding failed: {} {}'.format(resp.status_code, resp.text)
+            logger.error(msg)
+            raise requests.HTTPError(msg)
```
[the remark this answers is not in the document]
MartinWeindel: need response body to see error details

ccwienk (approved):

PR #705, opened 2022-04-01
CODEOWNER resolving utils

zkdev:
Add utils to resolve codeowners not only by email, but an extensible set of attributes.

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/github/codeowners.py
+++ b/github/codeowners.py
@@ -12,8 +12,11 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-from pathlib import Path
+import dataclasses
+import enum
 import logging
+from pathlib import Path
+import typing
 
 from github3 import GitHub
 from github3.exceptions import NotFoundError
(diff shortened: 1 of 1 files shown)


github/codeowners.py
```diff
@@ -84,6 +88,107 @@ def _first(iterable):
...
 
 
+@dataclasses.dataclass(frozen=True)
+class CodeOwnerGithubUser:
+    user: str
+    source: str
+
+
+@dataclasses.dataclass(frozen=True)
+class CodeOwnerPersonalName:
```
ccwienk, line 98: nit: it is just a PersonalName

github/codeowners.py
```diff
@@ -84,6 +88,107 @@ def _first(iterable):
         return None
 
 
+@dataclasses.dataclass(frozen=True)
+class CodeOwnerGithubUser:
```
ccwienk, line 92: GithubUser

github/codeowners.py
```diff
@@ -84,6 +88,107 @@ def _first(iterable):
...
+
+@dataclasses.dataclass(frozen=True)
+class CodeOwnerPersonalName:
+    firstName: str
+    lastName: str
+    source: str
+
+
+@dataclasses.dataclass(frozen=True)
+class CodeOwnerEmail:
```
ccwienk, line 105: EmailAddress

github/codeowners.py
```diff
@@ -84,6 +88,107 @@ def _first(iterable):
...
+    email: str
+    source: str
+
```
ccwienk, line 108: define a base class `UserInfo(Base?)` that defines the `source` attr, which is common to all DCs you defined in this diff

github/codeowners.py
```diff
@@ -84,6 +88,107 @@ def _first(iterable):
...
+
+@dataclasses.dataclass
+class CodeOwner:
```
ccwienk, line 111: that's not exactly how I imagined this :-/

also, I think this code actually belongs into the delivery-service repository

I would like to have a more abstract object representing a UserIdentity (maybe call it like this?)

A UserIdentity consists of a non-empty collection of (typed) UserInfo objects:
- email-address
- github-user
- personal-name
- ... (for now, there are no more, but there might be. e.g. "d-user")

github/codeowners.py
```diff
@@ -84,6 +88,107 @@ def _first(iterable):
...
+    github: typing.Optional[CodeOwnerGithubUser]
+    personalName: typing.Optional[CodeOwnerPersonalName]
+    email: typing.Optional[CodeOwnerEmail]
+
+    @staticmethod
+    def create(
```
ccwienk, line 117: as mentioned above, this should be a sequence..

github/codeowners.py
```diff
@@ -84,6 +88,107 @@ def _first(iterable):
...
+    PERSONAL_NAME = 'personalName'
+    GITHUB_USER = 'githubUser'
+    EMAIL = 'email'
+    UNKNOWN = 'unknown'
+
+    @staticmethod
+    def for_codeowner_attribute(
+        attribute: typing.Union[CodeOwnerGithubUser, CodeOwnerEmail, CodeOwnerPersonalName],
+    ) -> 'CodeOwnerMetadataTypes':
+        if isinstance(attribute, CodeOwnerGithubUser):
```
ccwienk, line 171: to convert into enums, simply use the enum's c'tor

--> ```
CodeOwnerMetadataTypes('personalName') is CodeOwnerMetadataTypes(CodeOwnerMetadataTypes.PERSONAL_NAME)
```

also, by convention (in this repository), use singular for enum-type-names, please

File: github/codeowners.py [hunk 8]
```diff
@@ -84,6 +88,107 @@ def _first(iterable):
...
+        elif isinstance(attribute, CodeOwnerEmail):
+            return CodeOwnerMetadataTypes.EMAIL
+        elif isinstance(attribute, CodeOwnerPersonalName):
+            return CodeOwnerMetadataTypes.PERSONAL_NAME
+        else:
+            return CodeOwnerMetadataTypes.UNKNOWN
+
+
+def iter_codeowner_attribute_dict(
+    codeowner: CodeOwner,
```
ccwienk, line 182: not needed, i think

ccwienk (changes requested):

zkdev:
https://github.com/gardener/cc-utils/pull/710

PR #706, opened 2022-04-04
Add model and rotation for BTP application certificates

MartinWeindel:
**What this PR does / why we need it**:
In the context of the enterprise policy filter (gardener-extension-shoot-networking-filter) we need to rotate certificates and service bindings.
This is the second PR for rotating application certificates.
Using a service binding for a certificate service instance, a new certificate is added to the application. "Rotating" a certificate means here creating a new one, keeping the old one and only deleting certificates older than the old one.

**Special notes for your reviewer**:
first PR was #704 

**Release note**:

```other operator
Add model and rotation for BTP application certificates
```

/invite @AndreasBurger

PR #708, opened 2022-04-04
Add model and rotation for BTP application certificates

ccwienk:
In the context of the enterprise policy filter (gardener-extension-shoot-networking-filter) we need to rotate certificates and service bindings.
This is the second PR for rotating application certificates.
Using a service binding for a certificate service instance, a new certificate is added to the application. "Rotating" a certificate means here creating a new one, keeping the old one and only deleting certificates older than the old one.

first PR was #704 

**Release note**:

```other operator
Add model and rotation for BTP application certificates
```

/invite @AndreasBurger 

copied from https://github.com/gardener/cc-utils/pull/706 (which I merged by accident)

### DIFF
--- a/cfg_mgmt/btp_application_certificate.py
+++ b/cfg_mgmt/btp_application_certificate.py
@@ -1,0 +1,250 @@
+import copy
+import logging
+import typing
+import requests
+import tempfile
+import os
+from contextlib import contextmanager
+from dataclasses import dataclass
+from cryptography.hazmat.primitives import serialization
+from cryptography.hazmat.primitives.asymmetric import rsa
+from cryptography import x509
+from cryptography.x509.oid import NameOID
+from cryptography.hazmat.primitives import hashes
+from cryptography.hazmat.primitives.serialization import pkcs7
+
+import cfg_mgmt
+import cfg_mgmt.model as cmm
+import ci.log
+import ci.util
+import model
+import model.container_registry
+from model.btp_application_certificate import BtpApplicationCertificate
+
+
+ci.log.configure_default_logging()
+logger = logging.getLogger(__name__)
+
+
+class CertServiceClient:
+    def __init__(
+        self,
+        credentials: dict,
+    ):
+        self.credentials = credentials
+        self._setup_oauth_token()
+
+    def _setup_oauth_token(self):
+        uaa = self.credentials['uaa']
+        data = {
+            'grant_type': 'client_credentials',
+            'token_format': 'bearer',
+            'client_id': uaa['clientid'],
+            'client_secret': uaa['clientsecret'],
+        }
+        headers = {
+            'Accept': 'application/json',
+        }
+        resp = requests.post(f'{uaa["url"]}/oauth/token', data=data, headers=headers)
+        resp.raise_for_status()
+        self.access_token = resp.json()['access_token']
+
+    def create_client_certificate_chain(self, csr_pem: str, validity_in_days: int) -> dict:
+        headers = {
+            'Accept': 'application/json',
+            'Authorization': f'Bearer {self.access_token}',
+        }
+        data = {'certificate-signing-request':{
+            'value': csr_pem,
+            'type': 'pkcs10-pem',
+            'validity': {'type': 'DAYS', 'value': validity_in_days},
+        }}
+        url = self.credentials['certificateservice']['profileurl']
+        resp = requests.post(url, json=data, headers=headers)
+        resp.raise_for_status()
+        logger.info('Created certificate')
+        return resp.json()['certificate-response']
+
+
+def _write_temp_file(temp_dir: str, fname: str, content: str) -> str:
+    fpath = os.path.join(temp_dir, fname)
+    with open(fpath, 'w', encoding='utf-8') as f:
+        f.write(content)
+    return fpath
+
+
+@dataclass
+class CertInfo:
+    id: str
+    dn: str
+    cn: str
+    serial_no: int
+
+
+class GBaasAppClient:
+    def __init__(self, auth: BtpApplicationCertificate):
+        endpoint = auth.application_endpoint()
+        self.url = f'{endpoint}/service/sps/{auth.application_id()}/apiCertificate'
+        self.clienturl = f'{endpoint}/service/sps/{auth.client_id()}/apiCertificate'
+        self.certificate_pem = auth.certificate_pem()
+        self.private_key_pem = auth.private_key_pem()
+
+    @contextmanager
+    def _session_with_cert(self):
+        with tempfile.TemporaryDirectory() as temp_dir:
+            session = requests.Session()
+            crt_fname = _write_temp_file(temp_dir, 'cert.pem', self.certificate_pem)
+            key_fname = _write_temp_file(temp_dir, 'key.pem', self.private_key_pem)
+            session.cert = (crt_fname, key_fname)
+            yield session
+
+    def put_certificate(self, cert_pem: str, desc: str, scopes: list[str]) -> str:
+        cert_pem = cert_pem.strip()
+        cert_pem = cert_pem.removeprefix("-----BEGIN CERTIFICATE-----\n")
+        cert_pem = cert_pem.removesuffix("-----END CERTIFICATE-----")
+        data = {
+            'description': desc,
+            'scopes': scopes,
+            'base64': cert_pem,
+        }
+        with self._session_with_cert() as session:
+            resp = session.put(self.url, json=data)
+        resp.raise_for_status()
+        id = resp.json()['certificateId']
+        logger.info(f'Added certificate {id}')
+        return id
+
+    def delete_certificate(self, common_name: str, cert_id: str):
+        data = {
+            'certificateId': cert_id,
+        }
+        with self._session_with_cert() as session:
+            resp = session.delete(self.url, json=data)
+        resp.raise_for_status()
+        logger.info(f'Deleted certificate {common_name} ({cert_id})')
+
+    @staticmethod
+    def _find_cn(dn: str) -> str:
+        for part in dn.split(','):
+            key, value = part.split('=')
+            if key == 'CN':
+                return value
+        return None
+
+    def list_certificates_by_base(
+        self,
+        common_name_base: str
+    ) -> typing.Generator[CertInfo, None, None]:
+        with self._session_with_cert() as session:
+            resp = session.get(self.clienturl)
+        resp.raise_for_status()
+        for item in resp.json():
+            id = item['dnId']
+            dn = item['dn']
+            cn = GBaasAppClient._find_cn(dn)
+            if cn:
+                try:
+                    serial_no, base = BtpApplicationCertificate.parse_serial_no_from_common_name(cn)
+                    if common_name_base == base:
+                        yield CertInfo(id=id, dn=dn, cn=cn, serial_no=serial_no)
+                except ValueError:
+                    pass
+
+
+_str_to_names = {
+    'C': NameOID.COUNTRY_NAME,
+    'ST': NameOID.STATE_OR_PROVINCE_NAME,
+    'O': NameOID.ORGANIZATION_NAME,
+    'OU': NameOID.ORGANIZATIONAL_UNIT_NAME,
+    'L': NameOID.LOCALITY_NAME,
+    'CN': NameOID.COMMON_NAME,
+    'EMAIL': NameOID.EMAIL_ADDRESS,
+}
+
+
+def _create_csr(subject: str) -> tuple[str, str]:
+    logger.info(f'Creating CSR for {subject}')
+    key = rsa.generate_private_key(public_exponent=65537, key_size=2048)
+    key_pem = key.private_bytes(
+        encoding=serialization.Encoding.PEM,
+        format=serialization.PrivateFormat.TraditionalOpenSSL,
+        encryption_algorithm=serialization.NoEncryption(),
+    ).decode('utf-8')
+    attributes = []
+    for part in subject.split(', '):
+        name, value = part.split('=')
+        oid = _str_to_names[name]
+        attributes.append(x509.NameAttribute(oid, value))
+    csr = x509.CertificateSigningRequestBuilder().subject_name(
+        x509.Name(attributes)
+    ).sign(key, hashes.SHA256())
+    csr_pem = csr.public_bytes(serialization.Encoding.PEM).decode('utf-8')
+    return csr_pem, key_pem
+
+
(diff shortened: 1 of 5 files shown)


cfg_mgmt/btp_application_certificate.py
```diff
@@ -0,0 +1,256 @@
+import copy
+import logging
+import typing
+import requests
+import tempfile
+import os
+from contextlib import contextmanager
+from dataclasses import dataclass
+from cryptography.hazmat.primitives import serialization
```
ccwienk, line 9: `cryptography` package is referenced indirectly - please rather use `pycryptodome` (see requirements.txt)

```
pip show cryptography
Name: cryptography
Version: 36.0.2
Summary: cryptography is a package which provides cryptographic recipes and primitives to Python developers.
Home-page: https://github.com/pyca/cryptography
Author: The Python Cryptographic Authority and individual contributors
Author-email: cryptography-dev@python.org
License: BSD or Apache License, Version 2.0
Location: /usr/lib/python3.10/site-packages
Requires: cffi
Required-by: pdfminer.six, SecretStorage, aliyun-python-sdk-core-v3, paramiko, endesive, jwcrypto, homeassistant, pyOpenSSL
```

cfg_mgmt/btp_application_certificate.py
```diff
@@ -0,0 +1,256 @@
...
+    def __init__(self, auth: BtpApplicationCertificate):
+        endpoint = auth.application_endpoint()
+        self.url = f'{endpoint}/service/sps/{auth.application_id()}/apiCertificate'
+        self.clienturl = f'{endpoint}/service/sps/{auth.client_id()}/apiCertificate'
+        self.certificate_pem = auth.certificate_pem()
+        self.private_key_pem = auth.private_key_pem()
+
+    @contextmanager
+    def _session_with_cert(self):
+        temp_dir = tempfile.TemporaryDirectory()
```
ccwienk, line 94: how about `tempfile.NamedTemporaryFile()` instead?

cfg_mgmt/btp_application_certificate.py
```diff
@@ -0,0 +1,256 @@
...
+        key_fname = _write_temp_file(temp_dir, 'key.pem', self.private_key_pem)
+        session.cert = (crt_fname, key_fname)
+        try:
+            yield session
+        finally:
+            temp_dir.cleanup()
+
+    def put_certificate(self, cert_pem: str, desc: str, scopes: list[str]) -> str:
+        idx = cert_pem.find('-----\n')
+        if idx != -1:
```
ccwienk, line 106: nit: how about `str.removeprefix(prefix)`?
-> https://docs.python.org/3.9/library/stdtypes.html#str.removeprefix

cfg_mgmt/btp_application_certificate.py
```diff
@@ -0,0 +1,256 @@
...
+        }
+        with self._session_with_cert() as session:
+            resp = session.delete(self.url, json=data)
+        resp.raise_for_status()
+        logger.info(f'Deleted certificate {common_name} ({cert_id})')
+
+    @staticmethod
+    def _find_cn(dn: str) -> str:
+        for part in dn.split(','):
+            (key,value) = part.split('=')
```
ccwienk, line 135: nit: parentheses on left side are optional

cfg_mgmt/btp_application_certificate.py
```diff
@@ -0,0 +1,256 @@
...
+        certs = []
+        for item in resp.json():
+            id = item['dnId']
+            dn = item['dn']
+            cn = GBaasAppClient._find_cn(dn)
+            if cn:
+                serial_no, base = BtpApplicationCertificate.get_serial_no_from_common_name(cn)
+                if serial_no > 0 and common_name_base == base:
+                    certs.append(CertInfo(id=id, dn=dn, cn=cn, serial_no=serial_no))
+        return certs
```
ccwienk, line 153: stylistic nit: you might `yield` the values instead of creating a list (this would change the function's return type to `typing.Generator[CertInfo, None, None]`, which I consider slightly more elegant.

cfg_mgmt/btp_application_certificate.py
```diff
@@ -0,0 +1,256 @@
...
+    logger.info(f'Creating CSR for {subject}')
+    key = rsa.generate_private_key(public_exponent=65537, key_size=2048)
+    key_pem = key.private_bytes(
+        encoding=serialization.Encoding.PEM,
+        format=serialization.PrivateFormat.TraditionalOpenSSL,
+        encryption_algorithm=serialization.NoEncryption(),
+    ).decode('utf-8')
+    attributes = []
+    for part in subject.split(', '):
+        (name, value) = part.split('=')
```
ccwienk, line 177: parantheses are optional

cfg_mgmt/btp_application_certificate.py
```diff
@@ -0,0 +1,256 @@
...
+        x509.Name(attributes)
+    ).sign(key, hashes.SHA256())
+    csr_pem = csr.public_bytes(serialization.Encoding.PEM).decode('utf-8')
+    return csr_pem, key_pem
+
+
+def _extract_client_certificate(cert_response: dict) -> str:
+    pkcs7_pem = cert_response['value']
+    certs = pkcs7.load_pem_pkcs7_certificates(pkcs7_pem.encode('utf-8'))
+    if len(certs) == 0:
```
ccwienk, line 190: nit: it is considered to be more "pythonic" to use falseness of empty collections, like so:

```
if not certs:
  raise VE..
```

cfg_mgmt/btp_application_certificate.py
```diff
@@ -0,0 +1,256 @@
...
+        raise ValueError('no certificates found in response')
+    return certs[0].public_bytes(serialization.Encoding.PEM).decode('utf-8')
```
ccwienk, line 192: is it okay if there were more than one? if not, I would suggest to validate for leng == 1, rather than > 0

cfg_mgmt/btp_application_certificate.py
```diff
@@ -0,0 +1,256 @@
...
+    cfg_element: BtpApplicationCertificate,
+    cfg_factory: model.ConfigFactory,
+) -> typing.Tuple[cfg_mgmt.revert_function, dict, model.NamedModelElement]:
+    gbaas_auth = cfg_factory.btp_application_certificate(cfg_element.auth_application_certificate())
+    gbaas_client = GBaasAppClient(gbaas_auth)
+
+    # calc next serial no
+    cn = cfg_element.common_name()
+    serial_no, base = BtpApplicationCertificate.get_serial_no_from_common_name(cn)
+    if serial_no == 0:
```
ccwienk, line 205: nit: you might just write:
```
if not serial_no:
 raise VE
```

numbers evaluate to False iff the equal zero (otherwise True, including negative values). A (small) benefit would be that this check will also detect other unexpected values, such as `None`, of `False`

cfg_mgmt/btp_application_certificate.py
```diff
@@ -0,0 +1,256 @@
...
+        raise ValueError(f'unexpected cn: {cn}')
+    next_sn = serial_no + 1
+    for info in gbaas_client.list_certificates_by_base(base):
+        if info.serial_no >= next_sn:
+            next_sn = info.serial_no + 1
+    next_cn = f'{next_sn}.{base}'
+
+    # create certificate
+    csr_pem, key_pem = _create_csr(cfg_element.subject(next_cn))
```
ccwienk, line 214: I am too lazy to google: what does `csr` mean? might the function maybe rather be named `_create_certificate` (or similar), as suggested by the comment above?

cfg_mgmt/btp_application_certificate.py
```diff
@@ -0,0 +1,256 @@
...
+    sb_auth = cfg_factory.btp_service_binding(cfg_element.cert_service_binding())
+    cs_client = CertServiceClient(sb_auth.credentials())
+    response = cs_client.get_client_certificate_chain(csr_pem, cfg_element.validity_in_days())
+    cert_pem = _extract_client_certificate(response)
+
+    # add certificate to GBaas application
+    id = gbaas_client.put_certificate(
```
ccwienk, line 221: the comment says `add`, the method says `put` (the latter implies that the existing value would be replaced). which one is it? :-)

cfg_mgmt/btp_application_certificate.py
```diff
@@ -0,0 +1,256 @@
...
+    cfg_element: BtpApplicationCertificate,
+    cfg_queue_entry: cmm.CfgQueueEntry,
+    cfg_factory: model.ConfigFactory,
+):
+    logger.info('Deleting old certificates')
+    gbaas_auth = cfg_factory.btp_application_certificate(cfg_element.auth_application_certificate())
+    gbaas_client = GBaasAppClient(gbaas_auth)
+    cn = cfg_queue_entry.secretId['common_name']
+    serial_no, base = BtpApplicationCertificate.get_serial_no_from_common_name(cn)
+    if serial_no == 0:
```
ccwienk, line 252: if not serial_no?

ccwienk: also: would it make sense to already validate/raise in `get_serial_no....` method?

model/btp_application_certificate.py
```diff
@@ -0,0 +1,90 @@
...
+        scopes for certificate
+        '''
+        return self.raw['scopes']
+
+    def subject(self, cn: str) -> str:
+        return self.subject_template().format(cn=cn)
+
+    @staticmethod
+    def get_serial_no_from_common_name(cn: str) -> tuple[int, str]:
+        idx = cn.find(".")
```
ccwienk, line 30: nit: please consistently use single-quotes (or double-quotes) within the same module

model/btp_application_certificate.py
```diff
@@ -0,0 +1,90 @@
...
+        '''
+        scopes for certificate
+        '''
+        return self.raw['scopes']
+
+    def subject(self, cn: str) -> str:
+        return self.subject_template().format(cn=cn)
+
+    @staticmethod
+    def get_serial_no_from_common_name(cn: str) -> tuple[int, str]:
```
ccwienk, line 29: naming-nit: since this method is not a simple getter, but a simple parser, I would rather choose `parse_..` as a verb

it would be nice to briefly explain the parsed/expected format in method's docstr

model/btp_application_certificate.py
```diff
@@ -0,0 +1,90 @@
...
+        idx = cn.find(".")
+        if idx <= 0:
+            return 0, cn
+        try:
+            return int(cn[:idx]), cn[idx+1:]
+        except ValueError:
+            return 0, cn
```
ccwienk, line 36: why not simply let the exception propagate, and have the callers deal w/ it? currently, callers all seem to check for `0`, and then raise themselves..

cfg_mgmt/btp_application_certificate.py
MartinWeindel: yes, it may return a certificate chain with multiple certs

cfg_mgmt/btp_application_certificate.py
MartinWeindel: CSR is a "certificate signing request". The comment `create certificate` is for the code block.
I have renamed `get_client_certificate_chain` to `create_client_certificate_chain` to make the crucial call clearer.

cfg_mgmt/btp_application_certificate.py
MartinWeindel: The API uses a `PUT` request to add a certificate :-)

cfg_mgmt/btp_application_certificate.py
MartinWeindel: There are two temporary files needed. I don't think it would be shorter or easier with `tempfile.NamedTemporaryFile`

cfg_mgmt/btp_application_certificate.py
MartinWeindel: There is no support for creating CSR in `pycryptodome`. I also don't find support for pkcs7.
Should I add `cryptography` as explicit requirement?

cfg_mgmt/btp_application_certificate.py
ccwienk: I see. That's a bad choice of HTTP-Methods on API-side, then (but of course nothing we can influence on our end)

ccwienk:
@MartinWeindel : there seems to be a bug in your last code (maybe you could also include the fix in this PR):
```
2022-04-05 07:01:26,360 [ERROR] cfg_mgmt.rotate: error deleting secret for cfg-type 'btp_service_binding' with name 'dev-i503479-sm-binding': 'id'.
  File "<string>", line 256, in <module>
  File "<string>", line 56, in process_config_queue
  File "/usr/lib/python3.9/site-packages/cfg_mgmt/util.py", line 451, in process_cfg_queue_and_persist_repo
    if not cmro.delete_expired_secret(
  File "/usr/lib/python3.9/site-packages/cfg_mgmt/rotate.py", line 72, in delete_expired_secret
    delete_func(
  File "/usr/lib/python3.9/site-packages/cfg_mgmt/btp_service_binding.py", line 171, in delete_config_secret
    id = cfg_queue_entry.secretId['id']
KeyError: 'id'
```

cfg_mgmt/btp_application_certificate.py
MartinWeindel: done

cfg_mgmt/btp_application_certificate.py
MartinWeindel: done

model/btp_application_certificate.py
MartinWeindel: done

MartinWeindel:
> @MartinWeindel : there seems to be a bug in your last code (maybe you could also include the fix in this PR):
> 
> ```
> 2022-04-05 07:01:26,360 [ERROR] cfg_mgmt.rotate: error deleting secret for cfg-type 'btp_service_binding' with name 'dev-i503479-sm-binding': 'id'.
> ```

A[ bug fix](https://github.com/gardener/cc-utils/pull/708/files#diff-4bd5a42e3debc7c8d5d28301d89c5b0406140114c8d2f397280814f0ebcc0d9dR171-R175) is already included in this PR.

MartinWeindel:
@ccwienk @AndreasBurger 
All requested changes are addressed, please take another look.

cfg_mgmt/btp_application_certificate.py
ccwienk: yes, please :-)

ccwienk (commented):
/lgtm

PR #709, opened 2022-04-07
Add support for on_abort-hook

AndreasBurger:
Steps may now define an executable and, optionally, additional args that will be called if a job is aborted. The expected form (and the resulting behaviour) is similar to the existing `execute`-parameter. The resulting step will reuse the configuration of its defining parent-step.

### DIFF
--- a/concourse/model/step.py
+++ b/concourse/model/step.py
@@ -202,12 +202,38 @@ def attrs(pipeline_step):
             counted as failed
             ''',
         ),
+        AttributeSpec.optional(
+            name='on_abort',
+            default=None,
+            doc='''
+            The executable (with optional additional arguments) to run in case the job is aborted.
+            The executable path is calculated relative to `<main_repo>/.ci`.
+
+            Just like `execute` this has two forms:
+
+            - scalar value (str in most cases) --> no shell-escaping is done
+            - list of scalar values -> used verbatim as ARGV
+
+            The resulting `on_abort` step will use the same container image reference as the job
+            that defines it. Also, it will have the same in- and outputs available and access to
+            the same env-vars.
+
+            .. note::
+                `on_abort`-steps themselves cannot be aborted. Beware.
+
+            ''',
+        ),
     )
 
 
 class StepNotificationPolicy(enum.Enum):
     NO_NOTIFICATION = enum.auto()
     NOTIFY_PULL_REQUESTS = enum.auto()
+
+
+class TaskHook(enum.Enum):
+    NONE = enum.auto()
+    ON_ABORT = enum.auto()
 
 
 class PipelineStep(ModelBase):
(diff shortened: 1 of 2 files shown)


concourse/model/step.py
```diff
@@ -405,6 +426,35 @@ def retries(self):
     def notification_policy(self):
         return self._notification_policy
 
+    def _on_abort(self):
+        return self.raw.get('on_abort', None)
+
+    def _on_abort_argv(self):
```
ccwienk, line 432: factor-out redundant argv-calculation?

concourse/model/step.py
```diff
@@ -405,6 +426,35 @@ def retries(self):
...
+        if not self._on_abort_argv():
+            raise RuntimeError("'on_abort' not configured")
+
+        if isinstance(prefix, str):
+            prefix = [prefix]
+        return os.path.join(*prefix, self._on_abort_argv()[0])
+
+    def on_abort_execute(self, prefix=''):
+        if not self._on_abort_argv():
+            raise RuntimeError("'on_abort' not configured")
```
ccwienk, line 452: if this happened, this would imply there is a bug in our template, correct? if I am correct, a brief stmt about this would be helpful (so we know it's not a cfg error potentially stemming from user-cfg)

concourse/templates/default.mako
```diff
@@ -213,6 +213,70 @@ if secret_cfg:
 % endif
 </%def>
 
+<%def name="on_abort(indent, job_variant, job_step)" filter="indent_func(indent),trim">
+<%
+if job_variant.has_main_repository():
```
ccwienk, line 218: factor-out source_repo-handling (I assume this is copy-pasta from regular build step?)?

concourse/templates/default.mako
```diff
@@ -213,6 +213,70 @@ if secret_cfg:
...
+  source_repo = job_variant.main_repository()
+  source_repo_github_cfg_name = source_repo.cfg_name() or github.name()
+else:
+  source_repo = None
+  source_repo_github_cfg_name = github.name()
+%>
+on_abort:
```
ccwienk, line 225: can this be deduplicated from the existing step-code? I assume the only difference should be indentation level + `on_abort` toplevel attr, right?

concourse/templates/default.mako
```diff
@@ -213,6 +213,70 @@ if secret_cfg:
...
+
+if cloned_main_repo_name:
+  prefix = (cloned_main_repo_name, '.ci')
+else:
+  prefix = (source_repo.resource_name(), '.ci')
+
+executable_file = job_step.on_abort_executable(prefix=prefix).rstrip()
+executable_cmd = job_step.on_abort_execute(prefix=prefix).rstrip()
+%>
+        if readlink -f .>/dev/null 2>&1; then
```
ccwienk, line 263: as mentioned above, this is a bit too much code to be duplicated, too my taste

concourse/templates/default.mako
```diff
@@ -231,7 +295,7 @@ else:
 % if job_step.retries():
   attempts: ${job_step.retries()}
 % endif
-  config:
+  config:${f' &{job_variant.variant_name}_{job_step.name}_config' if job_step._on_abort() else ''}
```
ccwienk, line 298: stylistic nit: I think the following style would be easier to read:
```
% if job_step._on_abort():
  # ... 
% else:
  config:
% endif

also, I think it would really be nicer to refactor the step-coding

**Review by ccwienk — CHANGES_REQUESTED** [review]

File: concourse/templates/default.mako [hunk 7]
```diff
@@ -231,7 +320,11 @@ else:
 % if job_step.retries():
   attempts: ${job_step.retries()}
 % endif
+% if job_step._on_abort():
+  config: ${f'&{job_variant.variant_name}_{job_step.name}_config'}
```
ccwienk, line 324: it would be nice to factor-out the calculation of yaml-anchor-name (might add a brief comment explaining the idea, too)

ccwienk (changes requested):
w.r.t avoiding code duplication, lgtm

PR #710, opened 2022-04-07
Eliminate codeowners class

zkdev:
Eliminate oop pattern by factoring out the actual methods.
One small change is also made, the remote enumaration was expecting a
`GithubRepoHelper` which in a nutshell holds two information, the repo
and its api. As only the repo is needed (read file contents, separate
api object not necessary), the function now only expects the repo.

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/concourse/replicator.py
+++ b/concourse/replicator.py
@@ -23,7 +23,8 @@ from ci.util import (
     urljoin,
 )
 from mailutil import _send_mail
-from github.codeowners import CodeownersEnumerator, CodeOwnerEntryResolver
+from github.codeowners import CodeOwnerEntryResolver
+import github.codeowners
 
 from concourse.factory import DefinitionFactory, RawPipelineDefinitionDescriptor
 from concourse.enumerator import (
@@ -524,10 +525,9 @@ class ReplicationResultProcessor:
             branch=main_repo['branch'],
         )
 
-        codeowners_enumerator = CodeownersEnumerator()
         codeowners_resolver = CodeOwnerEntryResolver(github_api=github_api)
         recipients = set(codeowners_resolver.resolve_email_addresses(
-            codeowners_enumerator.enumerate_remote_repo(github_repo_helper=repo_helper)
+            github.codeowners.enumerate_remote_repo(repo=repo_helper.repository)
         ))
 
         # in case no codeowners are available, resort to using the committer
--- a/concourse/replicator.py
+++ b/concourse/replicator.py
@@ -23,7 +23,6 @@ from ci.util import (
     urljoin,
 )
 from mailutil import _send_mail
-from github.codeowners import CodeOwnerEntryResolver
 import github.codeowners
 
 from concourse.factory import DefinitionFactory, RawPipelineDefinitionDescriptor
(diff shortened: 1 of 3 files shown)


github/codeowners.py
```diff
@@ -27,54 +27,56 @@
...
+):
+    file_path = existing_file(file_path)
+    with open(file_path) as f:
+        yield from filter_codeowners_entries(f.readlines())
+
+
+def enumerate_local_repo(
+    repo_dir: str,
+    paths: typing.Sequence[str] = ('CODEOWNERS', '.github/CODEOWNERS', 'docs/CODEOWNERS'),
+):
```
AndreasBurger, line 54:
```suggestion
) -> typing.Generator[str, None, None]:
```

AndreasBurger (approved):
lgtm

concourse/replicator.py
```diff
@@ -524,10 +525,9 @@ def _notify_broken_definition_owners(self, failed_descriptor):
             branch=main_repo['branch'],
         )
 
-        codeowners_enumerator = CodeownersEnumerator()
         codeowners_resolver = CodeOwnerEntryResolver(github_api=github_api)
```
ccwienk, line 528: for the sake of consistent API, I think COER should also be turned into functions, wdyt?

github/codeowners.py
```diff
@@ -27,54 +27,56 @@
 ci.log.configure_default_logging()
 
 
-class CodeownersEnumerator:
-    '''
-    Parses GitHub CODEOWNSERS files [0] from the documented default locations for a given
-    (git) repository work tree into a stream of codeowners entries.
+def enumerate_remote_repo(
```
ccwienk, line 30: `enumerate_codeowners` or `enumerate_repo(sitory)_codeowners` or `enumerate_codeowners_from_repo`?

github/codeowners.py
```diff
@@ -27,54 +27,56 @@
...
+    repo: github3.repos.repo.Repository,
+    paths: typing.Sequence[str] = ('CODEOWNERS', '.github/CODEOWNERS', 'docs/CODEOWNERS'),
```
ccwienk, line 32: nit: it is enough to require `paths` to be an `typing.Iterable[str]`

github/codeowners.py
```diff
@@ -27,54 +27,56 @@
...
+        try:
+            yield from filter_codeowners_entries(
+                repo.file_contents(path=path).decoded.decode('utf-8').split('\n')
+            )
+        except NotFoundError:
+            pass # ignore absent files
 
-    [0] https://help.github.com/articles/about-codeowners/
+
+def enumerate_single_file(
```
ccwienk, line 43: `enumerate_codeowners_file` or `enumerate_codeowners_from_file`?

github/codeowners.py
```diff
@@ -27,54 +27,56 @@
...
+        raise ValueError(f'not a git root directory: {repo_dir}')
+
+    for path in paths:
+        codeowners_file = repo_dir.joinpath(path)
+        if codeowners_file.is_file():
+            with open(codeowners_file) as f:
+                yield from filter_codeowners_entries(f.readlines())
+
+
+def filter_codeowners_entries(lines):
```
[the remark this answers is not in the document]
ccwienk: generally speaking, I would prefer typehints to be concise. It will be very rarely be the case that we do actually require a `sequence` (I think most notably, sequences define a fixed order of elements. by stating we need a sequence, we would e.g. "forbid" sets to be passed)

ccwienk (changes requested):
some naming nits + could also also resolve the COER into function(s)?

concourse/replicator.py
zkdev: will do

PR #713, opened 2022-04-19
Make Merge-Method Configurable

AndreasBurger:
This allows for the merge-method (`merge`, `rebase`, or `squash`) to be configured in our update-component-deps trait.

This can be necessary for repositories, where merge-commits are not allowed (such as this very repository).

### DIFF
--- a/concourse/model/traits/update_component_deps.py
+++ b/concourse/model/traits/update_component_deps.py
@@ -44,6 +44,12 @@ class MergePolicy(enum.Enum):
     AUTO_MERGE = 'auto_merge'
 
 
+class MergeMethod(enum.Enum):
+    MERGE = 'merge'
+    REBASE = 'rebase'
+    SQUASH = 'squash'
+
+
 class UpstreamUpdatePolicy(enum.Enum):
     STRICTLY_FOLLOW = 'strictly_follow'
     ACCEPT_HOTFIXES = 'accept_hotfixes'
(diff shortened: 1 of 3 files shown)


concourse/model/traits/update_component_deps.py
```diff
@@ -65,6 +71,15 @@ class UpstreamUpdatePolicy(enum.Enum):
         type=MergePolicy,
         doc='whether or not created PRs should be automatically merged',
     ),
+    AttributeSpec.optional(
+        name='merge_method',
+        default='merge',
```
ccwienk, line 76: why not use `rebase` as default? 🤔

concourse/model/traits/update_component_deps.py
```diff
@@ -65,6 +71,15 @@ class UpstreamUpdatePolicy(enum.Enum):
...
+        type=MergeMethod,
+        doc=(
+            'The method to use when merging PRs automatically. The default is `merge`, i.e. to '
+            'create a merge commit.'
```
ccwienk, line 80: nit: might link to GitHub's documentation (https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/about-merge-methods-on-github)

concourse/steps/update_component_deps.py
```diff
@@ -398,8 +399,16 @@ def create_upgrade_pr(
 
     if merge_policy is MergePolicy.MANUAL:
         return
-    # auto-merge - todo: make configurable (e.g. merge method)
-    pull_request.merge()
```
ccwienk, line 402: what's the default behaviour? if not setting a value will default to repository-default, then we should also add this as fourth value (e.g. `repository-default`)

concourse/steps/update_component_deps.py
```diff
@@ -398,8 +399,16 @@ def create_upgrade_pr(
...
+
+    if merge_method is MergeMethod.MERGE:
```
ccwienk, line 403: just use `merge_method.value`? checking for unknown / unimplemented values is fine, of course

ccwienk (changes requested):
addendum: please check my last comment about default behaviour

concourse/model/traits/update_component_deps.py
AndreasBurger: 👍

concourse/model/traits/update_component_deps.py
AndreasBurger: Since we currently use github3.py's default, we use merge. I've opted to keep it that way - can of course be easily changed later.

PR #717, opened 2022-05-20
Add support for kubeconfig-autorotation

AndreasBurger:
Enables automatic rotation of service-account kubeconfigs using our tried and tested approach.

It's probably not a good Idea to use this with kubeconfigs that have multiple context, as the writing of the kubeconfig simply replaces the first context. Also, with this PR our kubernetes-config supports the explicit configuration of the service-account-logation, e.g.:
```yaml
example-config:
  kubeconfig:
    …
  service_account:
    name: foo-serviceaccount
    namespace: default
```
It is **required** by the code that this configuration is present for the rotation to work, as there no 100% reliable way to retrieve the necessary information from the kubeconfig alone.
It goes without saying that the service-account that is backing the kubeconfig needs to have the necessary permissions to edit Serviceaccounts and create/delete Secrets in the Cluster.


AndreasBurger:
@hendrikKahl fyi. I'll get back to you once this is merged and deployed (and proven to work)

AndreasBurger:
tests fail due to missing `ruamel.yaml` in image. Tested locally and ran unit-tests, both is fine. Will merge soon

PR #718, opened 2022-05-23
Protecode noskip

jensh007:

* Refactor protecode code to make it more readable
* Adjust a few clip parameters to make it easier to test
* scan images with skip label and auto-triage all vulnerabilities instead of ignoring them entirely

**Release note**:

```other operator

```

### DIFF
--- a/cli/gardener_ci/protecode_cli.py
+++ b/cli/gardener_ci/protecode_cli.py
@@ -88,5 +88,6 @@ def scan_without_notification(
     results_above_threshold = [r for r in results if r.greatest_cve_score >= cve_threshold]
     results_below_threshold = [r for r in results if r.greatest_cve_score < cve_threshold]
 
+    logger.info('Summary of found vulnerabilities:')
     logger.info(f'{len(results_above_threshold)=}; {results_above_threshold=}')
     logger.info(f'{len(results_below_threshold)=}; {results_below_threshold=}')
--- a/cli/gardener_ci/protecode_cli.py
+++ b/cli/gardener_ci/protecode_cli.py
@@ -30,12 +30,19 @@ def scan_without_notification(
     exclude_image_names: typing.List[str] = [],
     include_component_names: typing.List[str] = [],
     exclude_component_names: typing.List[str] = [],
+    no_license_report: bool = False,
 ):
+    cfg_factory = ci.util.ctx().cfg_factory()
+    protecode_cfg = cfg_factory.protecode(protecode_cfg_name)
+
+    if not protecode_api_url:
+        protecode_api_url = protecode_cfg.api_url()
+        logger.info(f'Using Protecode at: {protecode_api_url}')
+
+    protecode_group_url = f'{protecode_api_url}/group/{protecode_group_id}/'
     cd = component_descriptor_util.component_descriptor_from_component_descriptor_path(
         cd_path=component_descriptor_path,
     )
-    cfg_factory = ci.util.ctx().cfg_factory()
-    protecode_cfg = cfg_factory.protecode(protecode_cfg_name)
 
     protecode_api_url = protecode_cfg.api_url()
     protecode_group_url = ci.util.urljoin(protecode_api_url, 'group', str(protecode_group_id))
@@ -83,4 +90,3 @@ def scan_without_notification(
 
     logger.info(f'{len(results_above_threshold)=}; {results_above_threshold=}')
     logger.info(f'{len(results_below_threshold)=}; {results_below_threshold=}')
-    logger.info('finished')
--- a/cli/gardener_ci/protecode_cli.py
+++ b/cli/gardener_ci/protecode_cli.py
@@ -21,6 +21,7 @@ def scan_without_notification(
     parallel_jobs: int=2,
     processing_mode: str='rescan',
     cve_threshold: float=7.0,
+    protecode_api_url=None,
     allowed_licenses: typing.List[str] = [],
     prohibited_licenses: typing.List[str] = [],
     reference_protecode_group_ids: typing.List[int] = [],
--- a/protecode/client.py
+++ b/protecode/client.py
@@ -295,6 +295,28 @@ class ProtecodeApi:
         )
         return result.json().get('custom_data', {})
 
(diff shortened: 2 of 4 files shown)


protecode/scanning_util.py
```diff
@@ -132,6 +134,75 @@ def _attribute_specs(cls):
         )
 
 
+class Shared:
```
ccwienk, line 137: stylistic nit: instead of defining a class w/ exactly one static method, why not just define a function?

ccwienk (commented):
overall lgtm

some remarks considering you were doing some refactoring (in addition to inline-comment).

- it would be nice to also add return-typehints (for functions/methods that were subject to refactoring), considering we tend to use typehints more and more (which I very much like)
- talking about typehints: as of python3.9, `typing.List` (+ tuple, set, ..) can be written using the built-ins (e.g. `list[str]`). it is not wrong to use `typing.*`, but there is a shortcut by now, so I would tend to use it (if touching / refactoring code anyhow)

cli/gardener_ci/protecode_cli.py
```diff
@@ -75,17 +81,19 @@ def scan_without_notification(
...
+                prohibited_licenses,
+            )
         )
-    )
 
     logger.info(f'{len(results_above_threshold)=}; {results_above_threshold=}')
     logger.info(f'{len(results_below_threshold)=}; {results_below_threshold=}')
-    logger.info(f'{len(updated_license_report)=}; {updated_license_report=}')
+    if not no_license_report:
+        logger.info(f'{len(updated_license_report)=}; {updated_license_report=}')
```
AndreasBurger, line 98: From reading the code: Shouldn't this be called `rejected_licenses_report` or something similar? Also, if that's really only the rejected licenses that are found, we could roll it in the first `if not no_license_report` block

protecode/client.py
```diff
@@ -295,6 +295,28 @@ def metadata(self, product_id: int):
...
+    ):
+        url = self._routes.triage()
+        result = self._get(
+            url=url,
+            params={
+                'component': component_name,
+                'vuln_id': vulnerability_id,
+                'scope': scope,
+                'version': component_version,
+                'description': description,
```
AndreasBurger, line 314: Out of curiosity: Is `None` allowed here?

AndreasBurger: Ah, I see - this is for the paranoid double-check. 👍

protecode/scanning_util.py
```diff
@@ -221,129 +304,100 @@ def _metadata(
...
-            resource_name=resource_name,
-        )
-
-        existing_protecode_products = self._api.list_apps(
-            group_id=self._group_id,
-            custom_attribs=metadata,
-        )
-
+    def _get_existing_protecode_apps(self) -> typing.Tuple[AnalysisResult]:
         # import triages from local group
```
AndreasBurger, line 313: I think this comment is no longer valid

protecode/scanning_util.py
```diff
@@ -431,27 +539,14 @@ def wait_for_scan_to_finish(protecode_products_to_consider):
...
-
-            yield pm.BDBA_ScanResult(
-                component=component_resource.component,
-                status=UploadStatus.DONE, # XXX remove this
-                result=analysis_result,
-                resource=component_resource.resource,
-                greatest_cve_score=analysis_result.greatest_cve_score(),
-            )
-
+    def _delete_outdatet_protecode_apps(self):
```
AndreasBurger, line 542:
```suggestion
    def _delete_outdated_protecode_apps(self):
```

protecode/scanning_util.py
```diff
@@ -463,6 +558,129 @@ def wait_for_scan_to_finish(protecode_products_to_consider):
...
+                    )
+
+    def process_component_resources(self) -> typing.Iterable[pm.BDBA_ScanResult]:
+        # depending on upload-mode, determine an upload-action for each related image
+        # - resources to upload
+        # - protecode-apps to remove
+        # - triages to import
+
+        logger.info(f'Processing component resource group for {self.component_name=} and '
+            + f'{self.resource_name=}')
```
AndreasBurger, line 608:
```suggestion
        logger.info(
            f'Processing component resource group for {self.component_name=} and '
            f'{self.resource_name=}'
        )
```

protecode/scanning_util.py
```diff
@@ -463,6 +558,129 @@ def wait_for_scan_to_finish(protecode_products_to_consider):
...
+
+        # apply imported triages for all protecode products
+        self._apply_triages(analysis_results, triages_to_import)
+
+        # apply auto triages for resources labeled with skip binary scan
+        for protecode_product in self.protecode_products_to_consider:
+            component_resource = self.product_id_to_resource[protecode_product.product_id()]
+            has_skip_label = self._has_skip_label(component_resource.resource)
+            logger.info(f'{component_resource.component.name}  has skip-scan label {has_skip_label}')
+            if has_skip_label:
```
AndreasBurger, line 662:
```suggestion
            if self._has_skip_label(component_resource.resource):
```
Logging a bool here does not add much. Or are we interested that much in components without skip label? If so, I'd probably log in with an `else:` instead of logging twice in the positive case.

AndreasBurger (approved):

protecode/scanning_util.py
jensh007: valid comment, was not sure if there will remain only one function. Might be a bit more emphasizing that this is shared between the other classes.

cli/gardener_ci/protecode_cli.py
jensh007: change head-lines of logs to make it more transparent, only one if block, keep the command-line option as is to keep it short.

protecode/scanning_util.py
jensh007: Done!

protecode/scanning_util.py
jensh007: Done

protecode/scanning_util.py
jensh007: Agreed, only logging if label is present.

PR #720, opened 2022-06-01
[WHD] Validate .ci/pipeline_definitions on PR creation

AndreasBurger:
With this PR, the Webhook-Dispatcher validates `.ci/pipeline_definitions` when a new PR is opened and comments the findings on the PR.

### DIFF
--- a/concourse/enumerator.py
+++ b/concourse/enumerator.py
@@ -270,7 +270,6 @@ class GithubDefinitionEnumeratorBase(DefinitionEnumerator):
         github_cfg,
         org_name,
         job_mapping: model.concourse.JobMapping = None,
-        repository_filter: callable=None,
         target_team: str=None,
         secret_cfg=None,
     ) -> RawPipelineDefinitionDescriptor:
--- a/concourse/enumerator.py
+++ b/concourse/enumerator.py
@@ -87,9 +87,9 @@ class DefinitionEnumerator:
         raw_definitions,
         secret_cfg,
         override_definitions={},
-        job_mapping: model.concourse.JobMapping = None,
+        job_mapping: model.concourse.JobMapping=None,
         target_team: str=None,
-        pipeline_definition_committish: str = None,
+        pipeline_definition_committish: str=None,
     ) -> 'DefinitionDescriptor':
         if not target_team:
             target_team = self.job_mapping.team_name()
@@ -117,8 +117,8 @@ class SimpleFileDefinitionEnumerator(DefinitionEnumerator):
         repo_path,
         repo_branch,
         repo_host='github.com',
-        job_mapping: model.concourse.JobMapping = None,
-        secret_cfg: model.secret.Secret = None,
+        job_mapping: model.concourse.JobMapping=None,
+        secret_cfg: model.secret.Secret=None,
     ):
         self.definition_file = definition_file
         self.repo_path = repo_path
@@ -280,7 +280,7 @@ class GithubDefinitionEnumeratorBase(DefinitionEnumerator):
         repository,
         github_cfg,
         org_name,
-        job_mapping: model.concourse.JobMapping = None,
+        job_mapping: model.concourse.JobMapping=None,
         target_team: str=None,
         secret_cfg=None,
         branch: str=None,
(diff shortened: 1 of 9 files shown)


concourse/enumerator.py
```diff
@@ -243,36 +243,54 @@ def _branch_cfg_or_none(
...
-            return
 
-        for branch in repository.branches():
-            cfg_entry = branch_cfg.cfg_entry_for_branch(branch.name)
-            if cfg_entry:
-                yield (branch.name, cfg_entry)
+        if branch:
+            if not branch_cfg:
+                # use only pipeline def as found on the requested branch
+                yield (branch, None)
```
ccwienk, line 258: nit: parentheses are not required

concourse/enumerator.py
```diff
@@ -243,36 +243,54 @@ def _branch_cfg_or_none(
...
+                if cfg_entry:
+                    yield (branch.name, cfg_entry)
 
     def _scan_repository_for_definitions(
         self,
         repository,
         github_cfg,
         org_name,
-        job_mapping: model.concourse.JobMapping = None,
-        repository_filter: callable=None,
```
ccwienk, line 273: good catch!

concourse/replicator.py
```diff
@@ -266,10 +266,18 @@ def ok(self):
...
 
 class DefinitionDeployer:
-    def deploy(self, definition_descriptor, pipeline):
+    def deploy(self, definition_descriptor):
         raise NotImplementedError('subclasses must overwrite')
 
 
+class NoOpDeployer:
+    def deploy(self, definition_descriptor):
+        return DeployResult(
```
ccwienk, line 275: nit: maybe we should at least let no-op deployed write a log entry, stating that it successfully ignored the request to do sth.?

github/util.py
```diff
@@ -624,6 +624,9 @@ def retrieve_asset_contents(self, release_tag: str, asset_label: str):
         asset.download(buffer)
         return buffer.getvalue().decode()
 
+    def files_changed_in_pr(self, pr_number: str) -> typing.Generator[str, None, None]:
+        yield from (f.filename for f in self.repository.pull_request(pr_number).files())
```
ccwienk, line 628: it should be equivalent to simply returning `.files()`, should it not? also: why not simply inline? apart from wrapping a generator in another generator, this method does not seem to do much, but pass-in pr_number

github/util.py
```diff
@@ -650,6 +653,10 @@ def search_issues_in_repo(self, query: str):
         search_result = self.github.search_issues(query)
         return search_result
 
+    def pr_labels(self, pull_request_number):
+        pull_request = self.repository.pull_request(pull_request_number)
+        return [l.name for l in pull_request.issue().labels]
```
ccwienk, line 658: inline?

github/util.py
```diff
@@ -659,6 +666,10 @@ def add_labels_to_pull_request(self, pull_request_number, *labels):
         pull_request = self.repository.pull_request(pull_request_number)
         pull_request.issue().add_labels(*labels)
 
+    def remove_label_from_pull_request(self, pull_request_number, label):
+        pull_request = self.repository.pull_request(pull_request_number)
+        pull_request.issue().remove_label(label)
```
ccwienk, line 671: inline (unless used more than once)

otherwise, maybe mv as util-function to github/util.py?

whd/dispatcher.py
```diff
@@ -395,6 +402,24 @@ def dispatch_pullrequest_event(
...
+                        f'repository {pr_event.repository().repository_path()}.'
+                    )
+                return
+
+            if (
+                self._pr_modified_pipeline_definitions(pr_event)
+                and pr_event.action() in [PullRequestAction.OPENED, PullRequestAction.SYNCHRONIZE]
+            ):
+                logger.info(f'Validating .ci/pipeline-definition for PR #{pr_event.number()}')
+                self._validate_pipeline_definitions(
```
ccwienk, line 417: generally speaking, how about moving the whole code that will check a PR for pipeline-definition-errs into a separate module (e.g. whd/pr_validator.py or whd/pr.py or whd/validate.py), so dispatcher.py can focus on, well, dispatching..

whd/pipelines.py
```diff
@@ -17,22 +17,24 @@
...
 
-def update_repository_pipelines(
-    repo_url,
-    cfg_set,
-    whd_cfg,
+def replicate_repository_pipelines(
+    repo_url: str,
+    cfg_set: model.ConfigurationSet,
+    whd_cfg: model.webhook_dispatcher.WebhookDispatcherConfig,
+    validate_only: bool=False,
```
ccwienk, line 32: rather pass-in (optionally) the deployed to use. or define a second function. I do not actually think adding such a switch will be beneficial, or save a lot of redundancies (but: will make code more complex, and more difficult to understand/maintain)

ccwienk (changes requested):
most notably:
- consider moving out pr-validation to separate module
- try to avoid large if/else based on switches
(see inline comments)

PR #722, opened 2022-06-02
Write detected libraries to deliverydb

zkdev:
Extract detected libraries (name + version) from protecode scan results and send to delivery-db.

Fixes #

**Release note**:

```other operator

```


ccwienk (changes requested):
genereally speaking, I think there is some refactoring we should do, instead of extending codebase by repeating those improvement-worthy patterns.

- define dataclasses for each specific metadata-type
- reconsider ComplianceMetadata (see below)
- factor out redundancies and make convenience methods actually convenient (or rm them)

# ComplianceMetadata

Currently, we just have a single attr `datasource`, which implies the data-type (semantically and structurally). However, already in the "protecode-case", we have three different types of data:

- license data
- ((strongly) aggregated) vulnerability data
- (lowlevel) component version data (you call them "libaries" in this PR, however, the name is not a perfect match)

I think we need to explicitly express the type of data, and make the datasource an additional attribute. Maybe like so:

```
- `vulnerabilities/aggregated` # for the existing protecode-originated cve-reports
- `vulnerabilities/raw` # not yet there - full data dump (may be added later)
- `licenses/aggregated` # deduplicated list of licenses (w/o mapping to lowlevel-components)
- `components/bdba` # components/libs as they are reported by bdba; in case this adheres to a standardised scheme, it should be thus-named
```

As a concrete example, the existing cve-data-entries might look like so:

```
meta:
 datasource: 'bdba' # let us rename this, while we are at it
 type: 'vulnerabilities/aggregated`
data: .. # keep as is
artefact: .. # keep as is
```

Obviously, this will need to be prepared on delivery-dashboard + delivery-service + delivery-db

/cc @mandelsoft

dso/model.py
```diff
@@ -22,56 +24,94 @@ class ScanArtifact:
...
+@dataclasses.dataclass(frozen=True)
+class Artefact:
+    artefact_name: str
+    artefact_version: str
+    artefact_type: str
+    artefact_extra_id: dict
+
+
+@dataclasses.dataclass(frozen=True)
 class ArtefactReference:
```
ccwienk, line 42: having both `Artefact` and `ArtefactReference` is rather misleading, considering `Artefact` is, in fact, an artefact reference, whereas `ArtefactReference` is a "qualified" artefact reference.

Rather call them, maybe: LocalArtefactId, and ComponentArtefactId (or -Reference, if you so prefer)

dso/model.py
```diff
@@ -22,56 +24,94 @@ class ScanArtifact:
...
-    creationDate: typing.Union[datetime.datetime, str]
+    type: str
+    creation_date: datetime.datetime
 
 
-@dataclasses.dataclass
-class ComplianceData:
+@dataclasses.dataclass(frozen=True)
+class GreatestCVE:
+    greatestCvss3Score: typing.Optional[float]
```
ccwienk, line 83: is this actually optional? please check - I remember we store `-1` as special value for "scanned, but nothing found"

dso/model.py
```diff
@@ -22,56 +24,94 @@ class ScanArtifact:
...
+    reportUrl: str
+
+
+@dataclasses.dataclass(frozen=True)
+class OsID:
+    os_info: unixutil.model.OperatingSystemId
+
+
+@dataclasses.dataclass(frozen=True)
+class License:
```
ccwienk, line 93: the name is rather misleading. why would a class named `License` (singular) have a field `licenses` (plural)? maybe rather call it `LicenseSummary`, or sth. the like? adding a short docstr could also help to make it easier to understand what this type represents

dso/model.py
```diff
@@ -22,56 +24,94 @@ class ScanArtifact:
...
+
+
+@dataclasses.dataclass(frozen=True)
+class ComponentVersion:
+    name: str
+    version: str
+
+
+@dataclasses.dataclass(frozen=True)
+class Component:
```
ccwienk, line 104: same as for `License`. where/for what is this used, btw (I did not check)?

dso/model.py
```diff
@@ -22,56 +24,94 @@ class ScanArtifact:
...
+    components: list[ComponentVersion]
+
+
+@dataclasses.dataclass(frozen=True)
+class Malware:
```
ccwienk, line 109: rather call this `MalwareSummary` or `MalwareStatus`. also add a docstr (in particular, state that an empty list of findings (if this is true..) means that there was a successful scan that yielded no findings)

protecode/util.py
```diff
@@ -242,58 +231,69 @@ def to_result(result):
...
+            meta=meta,
+            data=cve,
+        )
+
+        meta = dso.model.Metadata(
+            datasource=dso.model.Datasource.BDBA,
+            type=dso.model.Datatype.LICENSES_AGGREGATED,
+            creation_date=datetime.datetime.now()
+        )
+        licenses = list(dict.fromkeys(
```
ccwienk, line 262: nit: rather call this `license_names`

ccwienk (changes requested):

dso/model.py
zkdev: componentversions in resource identified by protecode

ccwienk (approved):

PR #729, opened 2022-06-20
Lazy load component-cli and helm

zkdev:

Fixes #

**Release note**:

```other operator

```


ci/util.py
```diff
@@ -646,3 +649,39 @@ def represent_data(self, data):
         if data is None:
             return self.represent_scalar('tag:yaml.org,2002:null', '')
         return super().represent_data(data)
+
+
+def github_release_download_url(
```
ccwienk, line 654: mv to github.util, or inline in "loader-script"

ci/util.py
```diff
@@ -646,3 +649,39 @@ def represent_data(self, data):
...
+    '''
+    returns latest github release asset download url with given name
+    '''
+    release = repo.latest_release()
+    for asset in release.assets():
+        if asset.name == asset_name:
+            return asset.browser_download_url
+
+
+def download_file(url, dest):
```
ccwienk, line 667: inline in "downloader-script"

ci/util.py
```diff
@@ -646,3 +649,39 @@ def represent_data(self, data):
...
+
+
+def call_executable_with_argv(
+    executable: str,
+    argv: list[str],
+):
+    '''
+    call given executable with argv via subprocess
+    '''
+    # replace python filename
```
ccwienk, line 685: ?????????????????????????????????????????????????????????

component-cli-loader.py
```diff
@@ -0,0 +1,59 @@
...
+import tempfile
+import shutil
+import stat
+import sys
+import util
+
+import ccc.github
+
+
+ASSET_NAME = 'componentcli-linux-amd64.gz'
```
ccwienk, line 15: ????????????????????????????????

ccwienk (changes requested):
just place two shell-scripts:

bin/helm
bin/component-cli

implement them in such a way that they appear to have the same behaviour as the original executables

one possible implementation approach might be:

- check if <own-name>.elf exists
- if so: exec <own-name>.elf
- if not: retrieve file from github; be sure to authenticate against github-api (this might be a candidate for "cli.py")
- while retrieving, be sure to retrieve to a temporary file (mktemp); use `mv` after download finished successfully (-> avoid concurrency issues)

bin/component-cli
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+set -eu
+set -o pipefail
+
+DOWNLOAD_URL=$(curl -s https://api.github.com/repos/gardener/component-cli/releases/latest | jq -r ".assets[] | select(.name | test(\"${componentcli-linux-amd64.gz}\")) | .browser_download_url")
```
ccwienk, line 6: hint: for readability, it is a good idea to wrap to multiple lines (this works also inside expressions), like so:

```
DOWNLOAD_URL=$(curl -s "https......." \
  | jq -r ".assets[]" \
  | ....
)
```

ccwienk (approved):
lgtm

I would personally prefer to add linebreaks so only one command is written per line to improve readability. however, considering those scripts will likely not see a lot of subsequent changes, and the used commands are somewhat straight-fwd, it is okay for me to not do so

PR #730, opened 2022-06-22
issue-reporting: use artifact instead of resource

jensh007:
Refactor issue reporting code to use artifacts instead of resources as preparation for checkmarx integration

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/github/compliance/report.py
+++ b/github/compliance/report.py
@@ -69,22 +69,22 @@ def _delivery_dashboard_url(
 
 def _compliance_status_summary(
     component: cm.Component,
-    resources: cm.Resource,
+    artifacts: typing.Sequence[cm.Artifact],
     report_urls: str,
     issue_description: str,
     issue_value: str,
 ):
-    if isinstance(resources[0].type, enum.Enum):
-        resource_type = resources[0].type.value
+    if isinstance(artifacts[0].type, enum.Enum):
+        resource_type = artifacts[0].type.value
     else:
-        resource_type = resources[0].type
+        resource_type = artifacts[0].type
 
     def pluralise(prefix: str, count: int):
         if count == 1:
             return prefix
         return f'{prefix}s'
 
-    resource_versions = ', '.join((r.version for r in resources))
+    artifact_versions = ', '.join((r.version for r in artifacts))
 
     report_urls = '\n- '.join(report_urls)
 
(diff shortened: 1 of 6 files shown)


github/compliance/report.py
```diff
@@ -116,14 +116,14 @@ def _template_vars(
     delivery_dashboard_url: str='',
 ):
     component = result_group.component
-    resource_name = result_group.resource_name
-    resources = [res.resource for res in result_group.results]
+    resource_name = result_group.name
```
zkdev, line 119: I'm confused, `<ScanResultGroup>.resource_name()` did return the `cm.Resource`.
Later in this PR, you replace it with `<ScanResultGroup>.artifact()` which returns `cm.Artifact`.

`<ScanResultGroup>.name` however contains `{component.name}:{resource.name}` which is not equal to `<ScanResultGroup>.resource_name()`.

github/compliance/result.py
```diff
@@ -57,8 +57,8 @@ def component(self) -> cm.Component:
         return self.results[0].component
 
     @property
-    def resource_name(self) -> cm.Resource:
-        return self.results[0].resource
+    def artifact(self) -> cm.Artifact:
```
zkdev, line 60: This is the mentioned property change.

zkdev (changes requested):

github/compliance/report.py
```diff
@@ -431,7 +430,7 @@ def component_resource_label(issue: github3.issues.Issue) -> str:
     for result_group in result_groups:
         resource_label = github.compliance.issue.resource_digest_label(
             component=result_group.component,
-            resource=result_group.resource_name,
+            artifact=result_group.name,
```
zkdev, line 433: Same here, I would rather expect the newly introduced `<ScanResultGroup>.artifact()`.

github/compliance/report.py
jensh007: fixed in next commit

github/compliance/result.py
jensh007: fixed in next commit

github/compliance/report.py
jensh007: fixed in next commit

github/compliance/report.py
```diff
@@ -95,16 +95,16 @@ def pluralise(prefix: str, count: int):
         | -- | -- |
         | Component | {component.name} |
         | Component-Version | {component.version} |
-        | Resource  | {resources[0].name} |
-        | {pluralise('Resource-Version', len(resources))}  | {resource_versions} |
+        | Resource  | {artifacts[0].name} |
```
ccwienk, line 98: either rename Resource -> Art(e|i)fact, or choose depending on type

ccwienk (changes requested):

PR #733, opened 2022-06-23
Introduce separate whd image (+build)

zkdev:

Fixes #

**Release note**:

```other operator

```


ccwienk (changes requested):
please mv this to a separate repository / create a dedicated component; let it consume only qualified versions of cc-utils (cicd-outbound)

PR #736, opened 2022-07-07
oci ref annotations

jschicktanz:
adds the "annotations" attribute to the classes OciBlobRef and OciImageManifestListEntry.
this was previously introduced in #732 and later reverted because multi-arch manifests with "annotations" being null weren't accepted by GCR.
therefore, now a more "defensive" approach is used, where the "annotations" and "platform" attributes are omitted from being serialised if they are null.

Fixes #

**Release note**:

```feature developer
- adds the "annotations" attribute to the classes OciBlobRef and OciImageManifestListEntry
```

### DIFF
--- a/cli/gardener_ci/_oci.py
+++ b/cli/gardener_ci/_oci.py
@@ -1,4 +1,3 @@
-import dataclasses
 import enum
 import hashlib
 import pprint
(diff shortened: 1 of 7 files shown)


AndreasBurger (approved):
lgtm

PR #737, opened 2022-07-07
Refactor Protecode Scanning

AndreasBurger:
This refactors our protecode scanning. It is now fully driven by our concept of artifact groups (a set of similar components with similar resources/artifacts) and uses these to create and track scans in protecode.

A top level overview would be as follows:
- The component-descriptor is used to generate all artifact groups. Currently, we distinguish between two types here, based on their usage:
  1. Artifact Groups for OCI image scanning: These group a component and an OCI image - both component and oci image may be present in multiple versions.
  2. Artifact groups for scanning gardenlinux-rootfs tars: Here, a group consists of a single (gardenlinux) component version, but it contains all rootfs-resources of this version.
- A processor creates scan requests for each artifact group. For OCI image artifact groups this will result in one scan request for each image version, matching previous behaviour. For gardenlinux-rootfs artifact groups this will result in one scan request per gardenlinux version.
- The processor handles uploading, triage-transport and auto-triages for these scan requests, also keeping track of previous scans (via custom data we attach to the scans, as before). Finally, it returns a sequence of scan results for each artifact group.

Once all scans for all artifact groups are finished, a sequence of scan results is returned (as before)

In order to understand whats happening, I think starting with `protecode.util.upload_grouped_images` and following the flow from there is a good idea.
Also, this PR contains some minor refactoring.


jensh007 (approved):
I appreciate documentation about the idea behind the abstraction. We lack this at many places.

protecode/model.py
```diff
@@ -235,48 +279,222 @@ def __hash__(self):
...
+class OciResourceBinary(Binary):
+    artifact: gci.componentmodel.Resource
+
+    def display_name(self):
+        image_reference = self.artifact.access.imageReference
+        _, image_tag = image_reference.split(':')
+        return f'{self.artifact.name}_{image_tag}'
+
+    def upload_data(self) -> typing.Iterable[bytes]:
+        # XXX need to check whether resource is actually a oci-resource
```
zkdev, line 327: Should the check be part of this PR?

protecode/scanning.py
```diff
@@ -0,0 +1,290 @@
...
+                f'Group {id}'
+            )
+            yield from products
+
+    def scan_requests(
+        self,
+        artifact_group: pm.ArtifactGroup,
+        known_artifact_scans: typing.Dict[str, typing.Iterable[pm.Product]]
+    ) -> typing.Iterable[pm.ScanRequest]:
+        match artifact_group:
```
zkdev, line 59: First time I see python 3.10 match syntax, nice !

protecode/model.py
AndreasBurger: Thinking about it, that case is now handled by Scan-Request-creation - no check necessary, I'll remove the comment

protecode/model.py
```diff
@@ -235,48 +279,222 @@ def __hash__(self):
...
+            src_tf=src_tarfiles,
+            filter_func=process_tarinfo,
+        )
+
+    def auto_triage(self) -> bool:
+        return False
+
+
+@dataclasses.dataclass(frozen=True)
+class ComponentArtifacts:
```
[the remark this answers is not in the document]
AndreasBurger: It couples components and ~~resources~~ artifacts. ArtifactGroups consist of a number of these ComponentArtifacts, signifying a concrete artifact for a concrete component that is to be scanned.

But I agree, the plural should be removed from this class, it's a leftover.

protecode/model.py
```diff
@@ -235,48 +279,222 @@ def __hash__(self):
...
+    def resource_type(self):
+        return self.component_artifacts[0].artifact.type
+
+    def __str__(self):
+        return f'{self.name}[{len(self.component_artifacts)}]'
+
+
+@dataclasses.dataclass(frozen=True)
+class OciArtifactGroup(ArtifactGroup):
+    pass
```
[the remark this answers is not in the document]
AndreasBurger: This class is used to distinguish between the different kinds of `ArtifactGroup`. Consider `ArtifactGroup` the (abstract) base class and this and `TarRootfsArtifactGroup` its concrete implementations.

PR #740, opened 2022-07-13
Fallback PR action to unknown

zkdev:
Adds an `UNKNOWN` fallback for PR actions.

Fixes #

**Release note**:

```other operator

```


whd/model.py
```diff
@@ -114,11 +114,21 @@ class PullRequestAction(enum.Enum):
...
+        action = self.raw['action']
+
+        if not action in [
+            pr_action.value
+            for pr_action
+            in PullRequestAction
+        ]:
+            return PullRequestAction.UNKNOWN
+
+        return PullRequestAction(action)
```
AndreasBurger, line 131:
```suggestion
        try:
            return PullRequestAction(action)
        except ValueError:
            return PullRequestAction.Unknown
```
wdyt?

zkdev: We implement `UNKNOWN` fallback because we expect more PR actions to be introduced in future.
Therefore, I prefer to not use `try ... expect` pattern for what seems to be regular workflow control.

PR #745, opened 2022-08-04
Rework Filter for Image Scan

AndreasBurger:
This PR generalises filtering by adding support for paths in the component's (or resources') descriptor that can be filtered using a regular expression.

An example for a filter config would be:

```yaml
  name: component_name_starts_with_foo # arbitrary
  rules:
  - target: component.name
    expression: "^foo.*"
    matching_semantics: include # include|exclude
```

The first part of the `target` will be used to determine what to check against. If the target starts with `resource`, the expression will be checked against resources in the component descriptor, e.g.:

```yaml
  name: resource_name_starts_with_foo
  rules:
  - target: resource.name
    expression: "^foo.*"
    matching_semantics: include
```

A single config may contain multiple rules. In that case, the rules will be combined using a logical `or` when filtering. The following example would include components whose name starts with `foo` or `bar`:

```yaml
  name: foo_or_bar
  rules:
  - target: component.name
    expression: "^foo.*"
    matching_semantics: include
  - target: component.name
    expression: "^bar.*"
    matching_semantics: include
```

If multiple configs are given, they will be combined using the logical `and`, i.e. a component or resource matches iff it matches all configs.

```yaml
matching_config:
- name: component_version_does_not_start_with_1
  rules:
  - target: component.version
    expression: '^1\..*'
    matching_semantics: exclude
- name: resource_name_starts_with_bar
  rules:
  - target: resource.name
    expression: "^bar.*"
    matching_semantics: include
```

No backwards compatibility included - we will need to adjust the image-scanning jobs when we merge this.

### DIFF
--- a/concourse/model/traits/filter.py
+++ b/concourse/model/traits/filter.py
@@ -1,0 +1,136 @@
+# Copyright (c) 2019-2020 SAP SE or an SAP affiliate company. All rights reserved. This file is
+# licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import dataclasses
+import enum
+import typing
+
+import dacite
+import pydash
+
+from concourse.model.base import (
+    AttributeSpec,
+    ModelBase,
+)
+import gci.componentmodel as cm
+import reutil
+
+
+class ImageFilterMixin(ModelBase):
+    def matching_config(self):
+        return self.raw['matching_config']
+
+
+FILTER_ATTRS = (
+    AttributeSpec.optional(
+        name='matching_config',
+        default=[],
+        doc='''
+        a list of configs to use for matching
+        ''',
+        type=list,
+    ),
+)
+
+
+class ComponentFilterSemantics(enum.Enum):
+    INCLUDE = 'include'
+    EXCLUDE = 'exclude'
+
+
+@dataclasses.dataclass
+class ConfigRule:
+    target: str
+    expression: str
+    matching_semantics: ComponentFilterSemantics
+
+
+@dataclasses.dataclass
+class MatchingConfig:
+    name: str
+    rules: typing.List[ConfigRule]
+
+
+def filter_for_matching_configs(
+    configs: typing.Collection[MatchingConfig]
+) -> typing.Callable[[cm.Component, cm.Resource], bool]:
+    # A filter for several matching configs is the combination of its constituent filters joined
+    # with a boolean AND
+    filters_from_configs = [
+        filter_for_matching_config(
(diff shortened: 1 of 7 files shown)


concourse/model/traits/filter.py
```diff
@@ -0,0 +1,136 @@
...
+                return re_filter(pydash.get(resource, tail))
+            case _:
+                raise ValueError(f"Unable to parse matching rule '{rule.target}'")
+
+    return filter_func
+
+
+def matching_configs_from_dicts(
+    dicts: typing.Iterable[dict],
+) -> typing.Iterable[MatchingConfig]:
```
ccwienk, line 127: nit: actually, this function will always return a `list`

test/concourse/model/traits/filter_test.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+# Dummy dataclass for testing
+@dataclasses.dataclass
+class Dummy:
+    name: str
+
+
+@pytest.fixture
+def dummy():
```
ccwienk, line 16: would it not be easier / less "boilerplaty" to simply add the default value to the dataclass-definition, and use its c'tor (/initialiser)? To me, a fixture makes more sense of it either offers some parametrisation (e.g. add additional attrs), or construction of returned object(s) is "cumbersome" (as in: requires multiple calls / operations)

ccwienk: then again: we do we even need this dummy class?

test/concourse/model/traits/filter_test.py
```diff
@@ -0,0 +1,140 @@
...
+        return Dummy(
+            name=name,
+        )
+
+    return _dummy
+
+
+class TestMatchingFilter:
+    @pytest.fixture
+    def rule(self):
```
ccwienk, line 30: why define the first fixture (which we might not actually need..) on module-level, and the second one as a method?

test/concourse/model/traits/filter_test.py
```diff
@@ -0,0 +1,140 @@
...
+        ):
+            return examinee.MatchingConfig(
+                name=name,
+                rules=rules,
+            )
+
+        return _matching_config
+
+    def test_unspecific_target_fails(self, dummy, matching_config, rule):
+        config = matching_config(
```
ccwienk, line 60: hmmm.... is this actually that much shorter than:
```
MatchingConfig(
 name='some-name',
 rules=(
  MatchingConfigRule(
   target='tgt',
  expression='some-expression',
  matching_semantics='include',
 ),
)
```

by line-count, I think this is not the case; it would make the test a tiny bit easier to understand, because there would be less abstraction from the actual code, though.

test/concourse/model/traits/filter_test.py
```diff
@@ -0,0 +1,140 @@
...
+                rule(
+                    target='component.name',
+                    expression='TestComponent',
+                    matching_semantics='include',
+                ),
+            ],
+        )
+        filter = examinee.filter_for_matching_config(config)
+        component_dummy = dummy(name='TestComponent')
+        assert filter(component_dummy, None) == True
```
ccwienk, line 86: nit: I think it is a bad practise to "shadow" builtin names (filter, map, ..)
next nit: boolean values should not be used in comparisons, see https://peps.python.org/pep-0008/
-> `assert filter_for_matching_config(cfg)`

ccwienk (changes requested):
overall lgtm.

suggestion: how about renaming MatchingConfigRule to just `ConfigRule`?

some inside comments w.r.t. unittests + how about simply using module-level-functions? thus far, there is only one single class, so there is not much benefit from potential grouping that could be achieved from using a class

PR #746, opened 2022-08-05
Factor out `find_branch_info`

zkdev:
Mv `find_branch_info` from closure to top level.
Alter shadowing `os_infos: list[delivery.model.OsReleaseInfo]` to function parameter.

Small nit:
Added `import delivery.model` as required for typehint integrity.

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/concourse/steps/os_id.py
+++ b/concourse/steps/os_id.py
@@ -11,6 +11,7 @@ import gci.componentmodel as cm
 import ci.log
 import ci.util
 import delivery.client
+import delivery.model
 import dso.model as dm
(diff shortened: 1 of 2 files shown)


ccwienk (changes requested):
as discussed "offline":
- functions defined below `concourse/steps/.*` are not intended to be re-used outside of pipeline-template
- -> mv function-definition to `delivery/`, either into a new module `util.py`, or the existing `model.py`
- rm commit fixing linter warning from this PR - it does not belong to this PR (+ I already fixed this in `master`)

PR #748, opened 2022-08-08
Enable manual retriggering of pipeline-definition-validation for pull-requests

AndreasBurger:
Allows a user to manually trigger another pr-validation by removing the label. This will cause some redundant checks but should be safe in respect to false-positives since automatically unlabeled prs already passed validation once before.


AndreasBurger:
Closing in favor of a change that causes the WHD to report the broken state for every update (already on master). See b1832dc8aa6e1c560c2ddd8e9bb2a63a163d8af6 for more details.

PR #751, opened 2022-08-16
Improve image scan for Tar Images

AndreasBurger:
Removes the hardcoded restriction to our supported platforms for the image scan and also changes the uploaded archives so that it can be determined from which original tarfile the uploaded file originated.


tarutil.py
```diff
@@ -25,19 +25,28 @@ def filtered_tarfile_generator(
...
+    modifications like renaming.
 
     This function is particularly useful for streaming. Note that `_FilelikeProxy` can be used
     to wrap a generator yielding an (input-) tarfile-stream.
 
     In combination, this can be used to - in a streaming manner - retrieve a tarfile-stream, e.g.
     using a http-request (e.g. with requests), and upload the filtered tarfile-stream (e.g. again
     with a http-request send e.g. with requests).
+
+    Finally, the `finalise` parameter controls whether the end-of-file marker (two 512-byte blocks
```
ccwienk, line 46: nit: it is not actually an end-of-file marker, as it might be known from ASCII (eof), rather than a marker denoting the end of a tar stream (which needs not be the end of the underlying file(system))

AndreasBurger: That is specified by the part in parenthesis. The general terminology was taken from https://www.gnu.org/software/tar/manual/html_node/Standard.html

PR #753, opened 2022-09-19
add oci platform filter

jschicktanz:
add oci platform filter

Fixes #

**Release note**:

```other operator
- add oci platform filter
```

### DIFF
--- a/oci/platform.py
+++ b/oci/platform.py
@@ -1,10 +1,61 @@
 import dataclasses
+import enum
 import typing
 
 import dacite
 
 import oci.client as oc
 import oci.model as om
+
+
+class OperatingSystem(enum.Enum):
+    '''
+    OperatingSystem contains the values for the 'os' property in an oci multiarch image.
+    See https://go.dev/doc/install/source#environment.
+    '''
+    AIX = 'aix'
+    ANDROID = 'android'
+    DARWIN = 'darwin'
+    DRAGONFLY = 'dragonfly'
+    FREEBSD = 'freebsd'
+    ILLUMOS = 'illumos'
+    IOS = 'ios'
+    JS = 'js'
+    LINUX = 'linux'
+    NETBSD = 'netbsd'
+    OPENBSD = 'openbsd'
+    PLAN9 = 'plan9'
+    SOLARIS = 'solaris'
+    WINDOWS = 'windows'
+
+    @classmethod
+    def contains_value(cls, value: str):
+        return value in [v.value for v in OperatingSystem]
+
+
+class Architecture(enum.Enum):
+    '''
+    Architecture contains the values for the 'architecture' property in an oci multiarch image.
+    See https://go.dev/doc/install/source#environment.
+    '''
+    PPC64 = 'ppc64'
+    _386 = '386'
+    AMD64 = 'amd64'
+    ARM = 'arm'
+    ARM64 = 'arm64'
+    WASM = 'wasm'
+    LOONG64 = 'loong64'
+    MIPS = 'mips'
+    MIPSLE = 'mipsle'
+    MIPS64 = 'mips64'
+    MIPS64LE = 'mips64le'
+    PPC64le = 'ppc64le'
+    RISCV64 = 'riscv64'
+    S390X = 's390x'
+
+    @classmethod
+    def contains_value(cls, value: str):
+        return value in [v.value for v in Architecture]
 
 
 def from_manifest(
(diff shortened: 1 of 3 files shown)


oci/platform.py
```diff
@@ -7,6 +8,56 @@
...
+    JS = 'js'
+    LINUX = 'linux'
+    NETBSD = 'netbsd'
+    OPENBSD = 'openbsd'
+    PLAN9 = 'plan9'
+    SOLARIS = 'solaris'
+    WINDOWS = 'windows'
+
+    @classmethod
+    def contains_value(cls, value: str):
```
ccwienk, line 32: rather overwrite __contains__
as an (more pythonic) alternative, one could resort to leveraging enum.Enum's autogenerated initialiser's semantics

-> `OperatingSystem('does-not-exist')` will raise `ValueError`

oci/platform.py
```diff
@@ -7,6 +8,56 @@
...
+    MIPS = 'mips'
+    MIPSLE = 'mipsle'
+    MIPS64 = 'mips64'
+    MIPS64LE = 'mips64le'
+    PPC64le = 'ppc64le'
+    RISCV64 = 'riscv64'
+    S390X = 's390x'
+
+    @classmethod
+    def contains_value(cls, value: str):
```
ccwienk, line 57: ditto

oci/platform.py
```diff
@@ -160,3 +211,57 @@ def iter_platforms(
...
 
         yield (sub_img_ref, platform)
+
+
+class PlatformFilter:
+    @staticmethod
+    def create(
+        included_platforms: typing.List[str],
+    ) -> typing.Callable[[om.OciPlatform], bool]:
+        matchers = []
```
ccwienk, line 221: nit: use list-comprehension

oci/platform.py
```diff
@@ -160,3 +211,57 @@ def iter_platforms(
...
+                if PlatformFilter._match(m, platform_to_match):
+                    return True
+
+            return False
+
+        return filter
+
+    @staticmethod
+    def _parse_expr(platform_expr: str) -> dict:
+        splitted = platform_expr.split('/')
```
ccwienk, line 236: nit: split, split, split

oci/platform.py
```diff
@@ -160,3 +211,57 @@ def iter_platforms(
...
+        if len(splitted) < 2 or len(splitted) > 3:
```
ccwienk, line 237: rather use separate if-clauses for both variants - or at least output actual length in error-msg to make debugging easier

ccwienk: you might want to consider using structural pattern matching (https://peps.python.org/pep-0622/)

oci/platform.py
```diff
@@ -160,3 +211,57 @@ def iter_platforms(
...
+            raise ValueError(f'invalid oci platform expression {platform_expr=}.'
+                              ' expression must have the format os/architecture[/variant]')
+
+        os = splitted[0]
+        if os != '*' and not OperatingSystem.contains_value(os):
+            raise ValueError(f'invalid os in oci platform expression {platform_expr=}.'
+                             f' allowed values are {["*"] + [o.value for o in OperatingSystem]}')
+
+        architecture = splitted[1]
+        if architecture != '*' and not Architecture.contains_value(architecture):
```
ccwienk, line 247: as mentioned above: simply use Archecture(value) and let ValueError propagate

oci/platform.py
```diff
@@ -160,3 +211,57 @@ def iter_platforms(
...
+            variant = splitted[2]
+
+        return {
+            'os': os,
+            'architecture': architecture,
+            'variant': variant,
+        }
+
+    @staticmethod
+    def _match(m: dict, p: om.OciPlatform) -> bool:
```
ccwienk, line 262: there is __eq__ for that

ccwienk (changes requested):
generally speaking, I think I would prefer to keep most of the platform-handling code w/ the using code, and keep semantic focus of `oci` package on both model classes for (de-)serialisation of oci-manifests (+ sub-datastructures), as well as interactions w/ oci-api. Normalisation seems like a good idea (however, I am uncertain whether we should actually be strict and specify allowed platforms/architectures - this is something we do not need to restrict, and could resort to just see thos IDs as strings). As for filtering, I think this would fit better into user's code.

this will allow us to keep oci-package and using "ctt" more losely coupled

jschicktanz:
coding was moved to https://github.com/gardener/cnudie-transport-tool/pull/488

PR #759, opened 2022-10-06
`product.v2` refactoring

8R0WNI3:
Moves and adjusts component version retrievals from `product.v2` to `cnudie.retrieve`.
Also, replaces `product.v2` occurrences in CC-Utils with their respective counterpart from `cnudie.retrieve` where it is possible.

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/cnudie/retrieve.py
+++ b/cnudie/retrieve.py
@@ -1,5 +1,6 @@
 import dataclasses
 import functools
+import hashlib
 import io
 import json
 import logging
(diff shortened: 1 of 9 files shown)


cnudie/retrieve.py
```diff
@@ -634,3 +647,88 @@ def greatest_component_versions(
...
 
     return versions[-max_versions:]
+
+
+def greatest_version_before(
+    component_name: str,
+    component_version: str,
+    ctx_repo: cm.RepositoryContext,
+    oci_client: oc.Client=None,
+):
```
AndreasBurger, line 657:
```suggestion
) -> str | None:
```

cnudie/retrieve.py
```diff
@@ -634,3 +647,88 @@ def greatest_component_versions(
...
+    )
+
+
+def greatest_component_version_by_name(
+    component_name: str,
+    ctx_repo: cm.RepositoryContext,
+    ignore_prerelease_versions: bool=False,
+    oci_client: oc.Client=None,
+    component_descriptor_lookup: ComponentDescriptorLookupById=None,
+):
```
AndreasBurger, line 711:
```suggestion
) -> cm.component:
```
not entirely sure if `cm.component` is the correct alias, please check :). But given the name of the function one might expect a `str`, so I think this hint might be useful

AndreasBurger (approved):
lgtm

PR #765, opened 2022-10-20
Make WHD Auto-Labeling more configurable

AndreasBurger:
in addition to org-membership, the webhook-dispatcher may now check for membership in special teams to determine whether required labels should be set automatically.
The teams that should be considered trustworthy can be configured on a per-job-mapping basis.

### DIFF
--- a/whd/pull_request.py
+++ b/whd/pull_request.py
@@ -15,6 +15,7 @@ import concourse.client.model
 import concourse.enumerator
 import concourse.replicator
 import model
+import model.concourse
 import model.webhook_dispatcher
 import whd.dispatcher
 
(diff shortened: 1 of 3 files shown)


whd/pull_request.py
```diff
@@ -261,18 +263,40 @@ def set_pr_labels(
...
+    job_mapping_set: model.concourse.JobMappingSet = cfg_set.job_mapping()
+    job_mapping = job_mapping_set.job_mapping_for_repo_url(
+        repo.repository_url(), cfg_set
+    )
+    trusted_teams = job_mapping.trusted_teams()
 
     sender_login = pr_event.sender()['login']
-    if pr_event.action() is PullRequestAction.OPENED:
-        if github_helper.is_pr_created_by_org_member(pr_number):
+    if pr_event.action() in PullRequestAction.OPENED:
```
ccwienk, line 279: factor-out the checking-code into a separate function (check_if_label_is_required, or similarly-named)? this should (I think?) deduplicate (again) the new code introduced in this PR, should it not?

ccwienk (changes requested):

PR #770, opened 2022-10-28
Add func to resolve codeowner entries to gh usernames

zkdev:

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/github/codeowners.py
+++ b/github/codeowners.py
@@ -156,13 +156,10 @@ def determine_email_address(
     return EmailAddress(user.email)
 
 
-def usernames_with_email_address(
+def usernames_from_email_address(
     email_address: EmailAddress,
     gh_api: GitHub | GitHubEnterprise,
 ) -> typing.Generator[Username, None, None]:
-    '''
-    Return generator yielding usernames found for email addresse.
-    '''
     yield from (
         Username(res.user.login)
         for res in gh_api.search_users(query=f'{email_address} in:email')
@@ -245,7 +242,7 @@ def resolve_usernames(
             continue
 
         if isinstance(codeowner_entry, EmailAddress):
-            for username in usernames_with_email_address(
+            for username in usernames_from_email_address(
                 email_address=codeowner_entry,
                 gh_api=github_api,
             ):
--- a/github/codeowners.py
+++ b/github/codeowners.py
@@ -16,10 +16,11 @@ from pathlib import Path
 import logging
 import typing
 
-from github3 import GitHub
+from github3 import GitHub, GitHubEnterprise
 from github3.exceptions import NotFoundError
 import github3.orgs
 import github3.repos.repo
+import github3.search.user
 import github3.users
 
 from ci.util import existing_dir, existing_file, not_none
@@ -155,6 +156,19 @@ def determine_email_address(
     return EmailAddress(user.email)
 
 
+def find_users_with_email_address(
+    email_address: EmailAddress,
+    gh_api: GitHub | GitHubEnterprise,
+) -> typing.Generator[Username, None, None]:
+    '''
+    Return generator yielding usernames found for email addresse.
+    '''
+    yield from (
+        res.user
+        for res in gh_api.search_users(query=f'{email_address} in:email')
+    )
+
+
 def resolve_team_members(
     team: Team,
     github_api: GitHub,
@@ -186,7 +200,7 @@ def resolve_email_addresses(
     github codeowners entries.
     Teams are resolved to Users recursively.
     Users are resolved to exposed email addresses.
-    If not email address is exposed the User is skipped.
+    If no email address is exposed the User is skipped.
     '''
     for codeowner_entry in codeowners_entries:
         if isinstance(codeowner_entry, EmailAddress):
@@ -203,6 +217,40 @@ def resolve_email_addresses(
 
         if isinstance(codeowner_entry, Team):
             yield from resolve_email_addresses(
+                codeowners_entries=resolve_team_members(
+                    team=codeowner_entry,
+                    github_api=github_api,
+                ),
+                github_api=github_api,
+            )
+            continue
+
+
+def resolve_usernames(
+    codeowners_entries: typing.Iterable[Username | EmailAddress | Team],
+    github_api: GitHub,
+) -> typing.Generator[Username, None, None]:
+    '''
+    Returns a generator yielding the resolved usernames for the given iterable of
+    github codeowners entries.
+    Teams are resolved to Users recursively.
+    Emails are resolved to users.
+    If no username is found for given email address, its skipped.
+    '''
+    for codeowner_entry in codeowners_entries:
+        if isinstance(codeowner_entry, Username):
+            yield codeowner_entry
+            continue
+
+        if isinstance(codeowner_entry, EmailAddress):
+            yield from find_users_with_email_address(
+                email_address=codeowner_entry,
+                gh_api=github_api,
+            )
+            continue
+
+        if isinstance(codeowner_entry, Team):
+            yield from resolve_usernames(
                 codeowners_entries=resolve_team_members(
                     team=codeowner_entry,
                     github_api=github_api,
--- a/github/codeowners.py
+++ b/github/codeowners.py
@@ -156,7 +156,7 @@ def determine_email_address(
     return EmailAddress(user.email)
 
 
-def find_users_with_email_address(
+def usernames_with_email_address(
     email_address: EmailAddress,
     gh_api: GitHub | GitHubEnterprise,
 ) -> typing.Generator[Username, None, None]:
@@ -164,7 +164,7 @@ def find_users_with_email_address(
     Return generator yielding usernames found for email addresse.
     '''
     yield from (
-        res.user
+        Username(res.user.login)
         for res in gh_api.search_users(query=f'{email_address} in:email')
     )
 
@@ -237,24 +237,32 @@ def resolve_usernames(
     Emails are resolved to users.
     If no username is found for given email address, its skipped.
     '''
+    unique_usernames = set()
+
     for codeowner_entry in codeowners_entries:
         if isinstance(codeowner_entry, Username):
-            yield codeowner_entry
+            unique_usernames.add(codeowner_entry)
             continue
 
         if isinstance(codeowner_entry, EmailAddress):
-            yield from find_users_with_email_address(
+            for username in usernames_with_email_address(
                 email_address=codeowner_entry,
                 gh_api=github_api,
-            )
+            ):
+                unique_usernames.add(username)
+
             continue
 
         if isinstance(codeowner_entry, Team):
-            yield from resolve_usernames(
+            for username in resolve_usernames(
                 codeowners_entries=resolve_team_members(
                     team=codeowner_entry,
                     github_api=github_api,
                 ),
                 github_api=github_api,
-            )
+            ):
+                unique_usernames.add(username)
+
             continue
+
+    yield from unique_usernames


github/codeowners.py
```diff
@@ -146,8 +145,50 @@ def resolve_email_addresses(
...
                 github_api=github_api,
-            )
+            ):
+                if member.email:
+                    yield member.email
+                else:
+                    logger.warning(f'no email found for GitHub user {member}')
+
+
+def resolve_gh_usernames(
```
ccwienk, line 160: this duplicates "resolve_email_addresses", does it not? rather rename "resolve_email_addresses"  to "iter_codeowners", have iter_codeowners return either of email-addr, gh-username, gh-team, and have functions for filtering by type, and resolving, e.g. username -> email-addr (or none), team -> members

ccwienk: nit: since this resides within the github.codeowners module, I think it is rather redundant to have the `gh`-infix in the function's name.

ccwienk (changes requested):

github/codeowners.py
```diff
@@ -116,38 +118,146 @@ def resolve_team_members(
     if not team_or_none:
         logger.warning('failed to lookup team {t}'.format(t=team_name))
         return []
-    for member in map(github_api.user, team_or_none.members()):
-        if member.email:
-            yield member.email
+    yield from map(github_api.user, team_or_none.members())
```
ccwienk, line 121: nit: rather use generator-comprehension (`return (github_api.user(member) for member in team_or_non.members()`)

github/codeowners.py
```diff
@@ -116,38 +118,146 @@ def resolve_team_members(
...
+
+
+@dataclasses.dataclass
+class CodeownerEntry:
+    value: str
+
+
+@dataclasses.dataclass
+class UsernameCodeownerEntry(CodeownerEntry):
```
ccwienk, line 130: how about just.. username?

consider implementing __str__ (or just inheriting from str, even..- not much value-add from dataclasses in this case..)

github/codeowners.py
```diff
@@ -116,38 +118,146 @@ def resolve_team_members(
...
+    pass
+
+
+@dataclasses.dataclass
+class EmailCodeownerEntry(CodeownerEntry):
```
ccwienk, line 135: less smurf-typing, please - it is an emailaddress

github/codeowners.py
```diff
@@ -116,38 +118,146 @@ def resolve_team_members(
...
+    pass
+
+
+@dataclasses.dataclass
+class TeamnameCodeownerEntry(CodeownerEntry):
```
ccwienk, line 140: it is a teamname

github/codeowners.py
```diff
@@ -116,38 +118,146 @@ def resolve_team_members(
...
+
+    @staticmethod
+    def teamnames(codeowner_entry: CodeownerEntry) -> bool:
+        return isinstance(codeowner_entry, TeamnameCodeownerEntry)
+
+
+def iter_codeowners(
+    codeowners_entries,
+    node_filter: typing.Callable[[CodeownerEntry], bool]=None,
+) -> typing.Generator[str, None, None]:
```
ccwienk, line 161: I think the generator yields CodeownerEntry-instances, does it not? we might also omit inheritance and enumerate the three different types - it is not that many.

github/codeowners.py
```diff
@@ -116,38 +118,146 @@ def resolve_team_members(
...
+    '''
+    returns a generator yielding the codeowner entries.
+
+    @param codeowner_entries:  root for iteration
+    @node_filter:              use to filter emitted nodes (see Filter for predefined filters)
```
ccwienk, line 166: you never call the emitted objects nodes, neither in code, not in doc. why call them nodes here? rather name this just filter, or entry_filter.

github/codeowners.py
```diff
@@ -116,38 +118,146 @@ def resolve_team_members(
...
+            node = UsernameCodeownerEntry(value=codeowners_entry[1:])
+
         else:
-            logger.warning(f'no email found for GitHub user {member}')
+            node = TeamnameCodeownerEntry(value=codeowners_entry[1:])
+
+        if node_filter and not node_filter(node):
+            continue
+
+        yield node.value
```
ccwienk, line 186: hm, okay, so you _do_ just return the strs.. but then, we lose the information about what it actually is. that's inconvenient.

github/codeowners.py
```diff
@@ -116,38 +118,146 @@ def resolve_team_members(
...
 
 
 def resolve_email_addresses(
-    codeowners_entries,
+    codeowners_entries: typing.Iterable[str],
```
ccwienk, line 190: - pass-in typed codeowner-entries
- define a resolver to resolve teams into team-members (this might already be done by iter_codeowners)
- define a resolver to resolve users into email-addresses

github/codeowners.py
```diff
@@ -116,38 +118,146 @@ def resolve_team_members(
...
-        elif '/' not in codeowner_entry:
-            email_addr = determine_email_address(
-                github_user_name=codeowner_entry[1:],
-                github_api=github_api,
-            )
-            if email_addr:
-                yield email_addr
+
+    # iter multiple times
+    codeowners_entries = list(codeowners_entries)
```
ccwienk, line 198: if mutability is not required, prefer tuple
why do we need to iterate multiple times?

ccwienk (changes requested):

github/codeowners.py
```diff
@@ -12,21 +12,51 @@
...
 
-from ci.util import existing_dir, existing_file, not_none
+from ci.util import existing_dir, existing_file
 import ci.log
 
 logger = logging.getLogger(__name__)
 ci.log.configure_default_logging()
 
 
+@dataclasses.dataclass(frozen=True)
```
ccwienk, line 36: as already discussed earlier, I do not see any point in using a dataclass here, and also do not see a lot of benefits from having a common base class

- inherit from str
- consider removing base-class

if keeping base-class, we might consider to add parsing to it (not sure whether this would result in nice code, though)

github/codeowners.py
```diff
@@ -12,21 +12,51 @@
...
+
+
+@dataclasses.dataclass(frozen=True)
+class EmailAddress(Codeowner):
+    pass
+
+
+@dataclasses.dataclass(frozen=True)
+class Teamname(Codeowner):
+    # must be in format org/team
```
ccwienk, line 56: well... validate in initialiser? also, the (raw) format also demands a leading `@` character.

github/codeowners.py
```diff
@@ -81,73 +111,173 @@ def filter_codeowners_entries(
...
-    iterable: typing.Iterable,
-):
+def resolve_team_members(
+    team_name: str,
+    org: github3.orgs.Organization,
+    absent_ok: bool=False,
+) -> typing.Generator[github3.users.ShortUser, None, None]:
+    '''
+    Returns generator yielding members from team.
+    Absent teams passed silently with `absent_ok`.
```
ccwienk, line 121: what does "passed silently" mean?

github/codeowners.py
```diff
@@ -81,73 +111,173 @@ def filter_codeowners_entries(
...
+    Subteams are resolved as well.
```
ccwienk, line 122: a more common (and accurate) wording would be to mention nested teams being resolved recursively; also, since this is the core of this function's logic, I think it should be included int the first sentence, rather than being added as an additional hint after documenting the semantics of the `absent_ok` parameter.

I think we also should allow `Teamname` as input as well (-> typehint)

github/codeowners.py
ccwienk: also, add properties for parsing, like so:
```
@property
def org_name(self): <return org, parsed from raw teamname>

@property
def team_name(self): <return non-qualified teamname, parsed from raw teamname>
```

considering there seem to be two flavours of teamnames ("qualified", i.e. including org-prefix and "unqualified" (name only)), the name seems to be chose a bit unfortunately..

github/codeowners.py
```diff
@@ -81,73 +111,173 @@ def filter_codeowners_entries(
...
+    yield from team.members()
 
 
-def determine_email_address(
-    github_user_name: str,
-    github_api: GitHub,
-) -> typing.Optional[str]:
-    not_none(github_user_name)
+def mail_address_for_user(
+    user_name: str,
```
ccwienk, line 135: also allow `Username`

github/codeowners.py
```diff
@@ -81,73 +111,173 @@ def filter_codeowners_entries(
...
+    gh_api: GitHub | GitHubEnterprise,
+    absent_ok: bool=True,
+) -> str | None:
+    '''
+    Returns email associated with github user for username.
+    `None` returned if user not found and `absent_ok`.
+    Users without associated email return `None`.
```
ccwienk, line 142: please be accurate w/ comments. it is not true users do not have an "assiciated email [address]". instead, users may choose to not expose a public email address; they always have at least one email-address (it is possible to have multiple email-addresses, with one of them being the "primary" one which is used e.g. for passwd-resets).

github/codeowners.py
```diff
@@ -81,73 +111,173 @@ def filter_codeowners_entries(
...
-    if not team_or_none:
-        logger.warning('failed to lookup team {t}'.format(t=team_name))
-        return []
-    for member in map(github_api.user, team_or_none.members()):
-        if member.email:
-            yield member.email
-        else:
-            logger.warning(f'no email found for GitHub user {member}')
+def user_for_mail_address(
+    mail_address: str,
```
ccwienk, line 156: nit: consistently use `email_address`, at least within function and its docstr, preferably also within same module

github/codeowners.py
```diff
@@ -81,73 +111,173 @@ def filter_codeowners_entries(
...
+    gh_api: GitHub | GitHubEnterprise,
+) -> github3.users.ShortUser | None:
+    '''
+    Find user for email address.
+    Exactly one user for `None` will be returned.
```
ccwienk, line 161: > Exactly one user for `None` will be returned.

1. I think this should actually read _or_ rather than _for_?
2. this just repeats the typehint -> no added value -> can be removed

github/codeowners.py
```diff
@@ -81,73 +111,173 @@ def filter_codeowners_entries(
...
 
+    if not results:
+        return None
 
-def resolve_email_addresses(
-    codeowners_entries,
-    github_api: GitHub,
-) -> typing.Generator[str, None, None]:
+    if len(results) > 1:
+        logger.warning(f'more than one user found for {mail_address=}')
```
ccwienk, line 170: why return None in this case? how about just returning all?

github/codeowners.py
```diff
@@ -81,73 +111,173 @@ def filter_codeowners_entries(
...
+        return None
+
+    return results[0].user
+
+
+def iter_codeowners(
+    codeowners_entries: collections.abc.Collection[str],
```
ccwienk, line 177: nit: if we do not need to iterate more than once, it is more adequate to expect an typing.Iterable
also, why not pass-in `Codeowner`-objects?

the function's name is misleading. while it does return a generator, what it actually does is to convert `str` into typed objects (Username, Teamname, Emailaddress).

This should reasonably already be done at the function(s) we have that currently emit entries from CODEOWNERS files

github/codeowners.py
```diff
@@ -81,73 +111,173 @@ def filter_codeowners_entries(
...
+                        continue
+
+                    yield mail
+                    seen_email_addresses.append(mail)
+
+
+def resolve_usernames(
+    codeowners_entries: collections.abc.Collection[str],
+    github_api: GitHub | GitHubEnterprise,
+    prune_unique: bool=False,
```
ccwienk, line 247: naming: deduplicate?

github/codeowners.py
```diff
@@ -81,73 +111,173 @@ def filter_codeowners_entries(
...
+) -> typing.Generator[str, None, None]:
+    seen_usernames = []
```
ccwienk, line 249: use `set`

github/codeowners.py
```diff
@@ -81,73 +111,173 @@ def filter_codeowners_entries(
...
+            )):
+                username = user.login
+                if prune_unique and username in seen_usernames:
+                    continue
+
+                yield username
+                seen_usernames.append(username)
+
+        elif isinstance(codeowner, Teamname):
+            org_name, team_slug = str(codeowner).split('/') # always of form 'org/name'
```
ccwienk, line 272: as written above, mv to model-class; this is one of the few good object-oriented usages imo

ccwienk (changes requested):

ccwienk:
looking again at the existing code, remarks from top to bottom:
- change `enumerate_codeowners_from_remote_repo` so that it returns a generator yielding Teamname | EmailAddress | Username
- change `enumerate_codeowners_from_file` as `enumerate_codeowners_from_remote_repo`
- change `enumerate_codeowners_from_local_repo` as `enumerate_codeowners_from_remote_repo`
- since this makes three usages, a single (private) function accepting a `str` (line from CODEOWNERS file) as input, and returning either of Teamname | EmailAddress | Username will be adequate

- _first: consider removal / inlining / replace w/ simple try-except
- determine_email_address: (almost) keep as is | also accept (typehint-change only) Username
- `resolve_team_members`: yield `Username` rather than `str`; should not return []
- `resolve_email_addresses`: pass-in iterable of `Username` | `Emailaddress` | `Teamname` (in other words: do not parse any longer at this point). change type of yielded elements to just `Emailaddress`; for each processed element: if type is Username, resolve to EmailAddress (if succeeded, yield EmailAddress, otherwise skip), if type is EmailAddress, yield EmailAddress, as is, if type is TeamName, recurse (resolve to team-members, yield from resolve_email_addresses, passing resolved team-members to `resolve_email_addresses` function

Implement those changes, and those changes only. let us review those separately. then let us extend the functionality for the new use-case.

zkdev:
Model introduction, codeowner parsing:
https://github.com/gardener/cc-utils/pull/771

github/codeowners.py
```diff
@@ -155,6 +156,19 @@ def determine_email_address(
     return EmailAddress(user.email)
 
 
+def usernames_with_email_address(
```
ccwienk, line 159: this function determines gh-usernames from email-addresses - this should be reflected in the name

examples that come to my mind:
- iter_usernames
- find_github_usernames
- [github_]usernames_from_email_address
- to_github_usernames

github/codeowners.py
```diff
@@ -155,6 +156,19 @@ def determine_email_address(
...
+    email_address: EmailAddress,
+    gh_api: GitHub | GitHubEnterprise,
+) -> typing.Generator[Username, None, None]:
+    '''
+    Return generator yielding usernames found for email addresse.
```
ccwienk, line 164: this comment merely repeats the typehints - just omit it
also: address does not end with an `e` in english

github/codeowners.py
```diff
@@ -186,7 +200,7 @@ def resolve_email_addresses(
     github codeowners entries.
     Teams are resolved to Users recursively.
     Users are resolved to exposed email addresses.
-    If not email address is exposed the User is skipped.
+    If no email address is exposed the User is skipped.
```
ccwienk, line 203: is this even still true?

github/codeowners.py
```diff
@@ -210,3 +224,65 @@ def resolve_email_addresses(
                 github_api=github_api,
             )
             continue
+
+
+def resolve_usernames(
+    codeowners_entries: typing.Iterable[Username | EmailAddress | Team],
+    github_api: GitHub,
+    deduplicate: bool=False,
```
ccwienk, line 232: thinking of it: is there any case where we would not want to deduplicate? I cannot think of any
-> just always deduplicate; store all users in a set, and leverage its implicit deduplication

ccwienk (changes requested):

github/codeowners.py
zkdev: yes

PR #773, opened 2022-11-08
Support email-configs referring to aws-configs

AndreasBurger:
Provides drop-in support for using aws-configs instead of username/password credentials in our email-configs.

For this iteration, I've kept the observable behaviour the same where possible.
- When `credentials()` is called with absent credentials, exactly the same error is raised as before.
- The returned EmailCredentials' `passwd()` and `username()` functions will use the default ctx-factory when called without args so that no consumer needs to be adjusted. Explicit cfg-factories can be passed, if necessary.

I haven't yet found the license under which the AWS-code-snippet I've used here is published, so we still need to check if there are any problems here. Also, I am not _too_ happy with putting it under `mail.util` but it might have uses outside of our config and we don't have `awsutils` or something like that currently.

### DIFF
--- a/.ci/lint
+++ b/.ci/lint
@@ -10,6 +10,10 @@ error=0
 
 export PYTHONPATH="${src_dir}"
 
+# ensure that all required packages are installed (useful for PRs)
+echo "Making sure that packages are installed ..."
+pip3 install -r "${src_dir}/requirements.txt" --quiet
+
 echo 'running pylama for all modules (errors only)'
 (
     pushd "${src_dir}"
--- a/.ci/test
+++ b/.ci/test
@@ -7,14 +7,15 @@ if ! which pytest &>/dev/null; then
   exit 1
 fi
 
-# HACK: install awesomeversion to fix unittests until next release -> remove again
-pip3 install awesomeversion
-
 # usage: $1: <src-dir>
 
 src_dir="${1:-"$(readlink -f "$(dirname "${0}")/..")"}"
 
 export PYTHONPATH="${src_dir}:${src_dir}/cli/gardener_ci:${PYTHONPATH:-}"
+
+# ensure that all required packages are installed (useful for PRs)
+echo "Making sure that packages are installed ..."
+pip3 install -r "${src_dir}/requirements.txt" --quiet
 
 if pytest "${src_dir}" "${@}"; then
     echo 'Unittest executions succeeded'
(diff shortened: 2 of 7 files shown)


mail/util.py
```diff
@@ -0,0 +1,62 @@
...
+    'ca-central-1',    # Canada (Central)
+    'eu-central-1',    # Europe (Frankfurt)
+    'eu-west-1',       # Europe (Ireland)
+    'eu-west-2',       # Europe (London)
+    'sa-east-1',       # South America (Sao Paulo)
+    'us-gov-west-1',   # AWS GovCloud (US)
+]
+
+# These values are required to calculate the signature. Do not change them.
+DATE = "11111111"
```
ccwienk, line 28: inline, please. also: where did you find this? it would be nice to reference the documentation (from which I assume) this was taken

mail/util.py
```diff
@@ -0,0 +1,62 @@
...
+VERSION = 0x04
+
+
+def _sign(key, msg):
+    return hmac.new(key, msg.encode('utf-8'), hashlib.sha256).digest()
+
+
+def calculate_ses_key(secret_access_key: str, region:str) -> str:
+    if region not in SMTP_REGIONS:
+        raise ValueError(f"The {region} Region doesn't have an SMTP endpoint.")
```
ccwienk, line 41: I would rather either calculate this dynamically, or mv this to cfg, or drop this check (it will fail during runtime + I do not think we will see a lot of additional usage, so making it more convenient to learn about supported regions - which should be documented by ses anyhow) is not much of a concern)

adding this as a comment is fine. however, I do not at all like hardcoding those regions, as I assume they might change in the future

mail/util.py
```diff
@@ -0,0 +1,62 @@
...
+    return smtp_password.decode('utf-8')
+
+
+def ses_key_for_aws_config(aws_config_name: str, cfg_factory=None) -> str:
+    if not cfg_factory:
+        cfg_factory = ctx.cfg_factory()
+
+    aws_config = cfg_factory.aws(aws_config_name)
+
+    return calculate_ses_key(
```
ccwienk, line 59: inline. although, thinking of it, I think this module is not a good place to add this code anyhow.

how about adding this to `model/email.py` instead? this code is not really related to smtp/email-sending, but rather to AWS/SES (which might also disqualify model/email.py as a good fit; still, I think it is more adequate than mail-package

ccwienk (changes requested):

PR #774, opened 2022-11-09
Pass labels rather than OCM component / artifact

zkdev:
Labels are used to find issues, thus updating, deleting, and so on. As preparation for cfg-element policy violation issues, this hard contract based on OCM has to be resovled as cfg-elements do not belong to an OCM component. Instead, pass down labels directly.

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/github/compliance/issue.py
+++ b/github/compliance/issue.py
@@ -1,5 +1,4 @@
 import hashlib
-import itertools
 import logging
 import re
 import typing
@@ -11,6 +10,7 @@ import github3.issues.issue
 import github3.issues.milestone
 import github3.repos
 
+import cnudie.iter
 import gci.componentmodel as cm
 
 import ci.log
@@ -88,15 +88,48 @@ def repository_labels(
         yield from extra_labels
 
 
+def labels_from_target(
+    target: cnudie.iter.SourceNode | cnudie.iter.ResourceNode | None,
+    issue_type: str | None,
+    extra_labels: typing.Iterable[str]=None,
+) -> typing.Generator[str, None, None]:
+
+    if not target:
+        yield from repository_labels(
+            component=None,
+            artifact=None,
+            issue_type=issue_type,
+            extra_labels=extra_labels,
+        )
+        return
+
+    if isinstance(target, (cnudie.iter.SourceNode, cnudie.iter.ResourceNode)):
+        yield from repository_labels(
+            component=target.component,
+            artifact=target_artifact(target),
+            issue_type=issue_type,
+            extra_labels=extra_labels,
+        )
+        return
+
+    raise NotImplementedError(f'{target=}')
+
+
 @github.retry.retry_and_throttle
 def enumerate_issues(
-    labels: frozenset[str],
+    target: cnudie.iter.SourceNode | cnudie.iter.ResourceNode | None,
     known_issues: typing.Sequence[github3.issues.issue.ShortIssue],
+    issue_type: str | None,
     state: str | None = None, # 'open' | 'closed'
 ) -> typing.Generator[github3.issues.ShortIssue, None, None]:
     '''Return an iterator iterating over those issues from `known_issues` that match the given
     parameters.
     '''
+    labels = frozenset(labels_from_target(
+        target=target,
+        issue_type=issue_type,
+    ))
+
     def filter_relevant_issues(issue):
         if issue.state != state:
             return False
@@ -110,14 +143,22 @@ def enumerate_issues(
 
 @github.retry.retry_and_throttle
 def _create_issue(
-    labels: frozenset[str],
+    target: cnudie.iter.ResourceNode | cnudie.iter.SourceNode,
+    issue_type: str,
     repository: github3.repos.Repository,
     body: str,
     title: str,
+    extra_labels: typing.Iterable[str]=None,
     assignees: typing.Iterable[str]=(),
     milestone: github3.issues.milestone.Milestone=None,
 ) -> github3.issues.issue.ShortIssue:
     assignees = tuple(assignees)
+
+    labels = frozenset(labels_from_target(
+        target=target,
+        issue_type=issue_type,
+        extra_labels=extra_labels,
+    ))
 
     try:
         return repository.create_issue(
@@ -136,10 +177,12 @@ def _create_issue(
 
 @github.retry.retry_and_throttle
 def _update_issue(
-    labels: frozenset[str],
+    target: cnudie.iter.ResourceNode | cnudie.iter.SourceNode,
+    issue_type: str,
     body:str,
     title:typing.Optional[str],
     issue: github3.issues.Issue,
+    extra_labels: typing.Iterable[str]=None,
     milestone: github3.issues.milestone.Milestone=None,
     assignees: typing.Iterable[str]=(),
 ) -> github3.issues.issue.ShortIssue:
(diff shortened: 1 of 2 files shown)


github/compliance/issue.py
```diff
@@ -85,23 +90,13 @@ def repository_labels(
 
 @github.retry.retry_and_throttle
 def enumerate_issues(
-    component: cm.Component | None,
-    artifact: cm.Artifact | None,
+    labels: frozenset[str],
```
ccwienk, line 93: hmm, actually I do not think we need a frozenset - any iterable should do

github/compliance/issue.py
```diff
@@ -85,23 +90,13 @@ def repository_labels(
...
     known_issues: typing.Sequence[github3.issues.issue.ShortIssue],
-    issue_type: str | None,
```
ccwienk, line 91: why remove the issue_type?

github/compliance/issue.py
```diff
@@ -66,7 +71,7 @@ def repository_labels(
     artifact: cm.Artifact | None,
     issue_type: str | None=_label_bdba,
     extra_labels: typing.Iterable[str]=None
-):
+) -> typing.Generator[str, None, None]:
```
ccwienk, line 74: is `repository_labels` still used?

github/compliance/issue.py
```diff
@@ -115,54 +110,38 @@ def filter_relevant_issues(issue):
...
-    artifact: cm.Artifact,
+    labels: frozenset[str],
     repository: github3.repos.Repository,
-    body:str,
-    title:typing.Optional[str],
+    body: str,
+    title: str,
     assignees: typing.Iterable[str]=(),
     milestone: github3.issues.milestone.Milestone=None,
-    issue_type: str=_label_bdba,
```
ccwienk, line 125: ditto

github/compliance/issue.py
```diff
@@ -115,54 +110,38 @@ def filter_relevant_issues(issue):
...
 def _update_issue(
-    component: cm.Component,
-    artifact: cm.Artifact,
+    labels: frozenset[str],
     body:str,
     title:typing.Optional[str],
     issue: github3.issues.Issue,
     milestone: github3.issues.milestone.Milestone=None,
     assignees: typing.Iterable[str]=(),
-    issue_type: str=_label_bdba,
```
ccwienk, line 164: ditto

github/compliance/issue.py
```diff
@@ -192,42 +164,39 @@ def _update_issue(
...
+    labels: frozenset[str],
     repository: github3.repos.Repository,
-    body:str,
+    body: str,
     known_issues: typing.Iterable[github3.issues.issue.ShortIssue],
-    title:str=None,
+    title: str,
     assignees: typing.Iterable[str]=(),
     milestone: github3.issues.milestone.Milestone=None,
-    issue_type: str=_label_bdba,
```
ccwienk, line 203: ditto

github/compliance/issue.py
```diff
@@ -192,42 +164,39 @@ def _update_issue(
...
     extra_labels: typing.Iterable[str]=None,
```
ccwienk, line 174: why did you remove extra_labels from the other functions, but keep it here?

github/compliance/issue.py
```diff
@@ -192,42 +164,39 @@ def _update_issue(
...
-            artifact=artifact,
+            labels=labels,
             known_issues=known_issues,
-            issue_type=issue_type,
             state='open',
         )
     )
     if (issues_count := len(open_issues)) > 1:
-        raise RuntimeError(f'more than one open issue found for {component.name=}{artifact.name=}')
+        raise RuntimeError(f'more than one open issue found for {labels=}')
```
ccwienk, line 185: hmm, that's not nice - those labels are quite cumbersome to read, considering the component/artefact-names are only there as hash-digests

Thus, I think it should be better to change the passed-in "target-identifier", like so:
- instead of component and resource, use `cnudie.iter.Node` (or for a first iteration, we might only accept ResourceNode)
- for cfg-elements, we might accept cfg-element

I am considering something like so:

```
def enumerate_issues(
 target: cnudie.iter.ResourceNode | model.base.NamedModelElement
 .. <other parameters, as they were defined before>
):
 labels = labels_from_target(target=target)
```

def labels_from_target(target: cnudie.iter.ResourceNode | model.base.NamedModelElement):
 if isinstance(target, RN):
   <code from before this change>
 elif isinstance(NME):
  <read element's name and typename to create labels>
 else:
  raise NotImplementedError(target)
```

**Review by ccwienk — CHANGES_REQUESTED** [review]

File: github/compliance/issue.py (hunk 2 above)
Comment (zkdev) [reply github/compliance/issue.py:91]: `issue_type` is only used to get repository labels
> yield f'cicd/{issue_type}'

as we pass in labels already, `issue_type` is not required in this func anymore

File: github/compliance/issue.py (hunk 3 above)
Comment (zkdev) [reply github/compliance/issue.py:74]: yes, `repository_labels` yields _all_ labels for a repo, e.g. `cicd/auto-generated` AND the actual artifact digest label.

Comment (zkdev) [reply github/compliance/issue.py:74]: I separated the digest calculation from repo label yielding as a preparation

File: github/compliance/issue.py (hunk 1 above)
Comment (zkdev) [reply github/compliance/issue.py:93]: We at least have to use collections, see line 103

File: github/compliance/issue.py (hunk 7 above)
Comment (zkdev) [reply github/compliance/issue.py:174]: The other func retrieved repo labels based off component+artifact and the passed extra labels.
As we now pass in labels directly, I merged those to only pass one `labels` arg.

This func however has a special semantic for extra labels, that is considering the passed `preserve_labels_regexes`. Therefore I kept the separation here.

File: github/compliance/issue.py (hunk 8 above)
Comment (zkdev) [reply github/compliance/issue.py:185]: We have to at least implement `cnudie.iter.ResourceNode` and `cnudie.iter.SourceNode` targets as checkmarx scans are in place already.

How about introducing a dataclass?

```python
@dc
class Target:
    target: ResourceNode | SourceNode | ...

    # could be used for logging
    @property
    def name(self) -> str:
        if isinstance(cnudie.iter.ResourceNode):
            return f'{self.target.component.name}:{self.target.resource.name}'

        if isinstance(cnudie.iter.SourceNode):
            return f'{self.target.component.name}:{self.target.source.name}'

    # determining target digest label here, so repository_labels could take an optional
    # digest_label callback and yield "common" labels (cicd/auto-generated, ...)
    def digest_label(self) -> str:
        if isinstance(self.target, cnudie.iter.ResourceNode):
            return artifact_digest_label(
                component=self.target.component,
                artifact=self.target.resource,
            )
        if isinstance(self.target, cnudie.iter.SourceNode):
            return artifact_digest_label(
                component=self.target.component,
                artifact=self.target.source,
            )
        # can be extended for efg-element
```

ccwienk: well, let's support ResourceNode and SourceNode in parallel, then - why introduce a new class for model elements that already have one?

zkdev: > why introduce a new class for model elements that already have one?

to group functions for "targets" like `name()` or `digest_label()`.

ccwienk: I consider this avoidable boilerplate.

just define a function that accepts a target and returns labels for the target, and - if required multiple times - a second function returning the display-name.

the structure as suggested above mixes different concerns, for which even dedicated functions already exist.

if at all, in order to reduce the amount of functions that need to dispatch between different target-types (from two to one), one could consider to define a dataclass holding name and labels. tbh I do not see much value here, and such an approach will not lead to a better codebase

ccwienk (approved):

PR #775, opened 2022-11-10
GitHub issues for cfg-element policy violations

zkdev:

Fixes #

**Release note**:

```other operator

```


cfg_mgmt/reporting.py
```diff
@@ -201,6 +201,10 @@ def storage_summary(element_storage: str) -> cmm.CfgStorageSummary:
     yield from storage_summaries.values()
 
 
+def cfg_element_status_name(status: CfgElementStatusReport):
```
ccwienk, line 204: this one I would have considered a good candidate for a method/@property; however, I am also fine w/ having separate functions, generally speaking.

cli/gardener_ci/_cfg_mgmt.py
```diff
@@ -135,3 +141,63 @@ def rotate(
...
+        name: str | None,
+        type: str | None,
+    ) -> bool:
+        if not name and not type: return True
+        if name != cfg_element.element_name: return False
+        if type != cfg_element.element_type: return False
+        return True
+
+    if any((cfg_element_type, cfg_element_name)) and not all((cfg_element_name, cfg_element_type)):
+        raise RuntimeError('either both cfg_element type and name, none must be given!')
```
ccwienk, line 165: use ValueError

cli/gardener_ci/_cfg_mgmt.py
```diff
@@ -135,3 +141,63 @@ def rotate(
...
+
+def create_or_update_tickets(
+    cfg_dir: str,
+    issue_repo: str, # must be of format hostname/owner/repo
+    delivery_endpoints_cfg_name: str,
+    overwrite_assignee: [str]=[],
+    cfg_element_type: str = None,
+    cfg_element_name: str = None,
+):
+    def is_target(
```
ccwienk, line 154: this name is greatly misleading. from its name, I would have expected the function to determine whether or not the passed arguments _are_ a target. what the function actually does is comparing whether or not the given cfg_element has name and type(-name) of passed-in name and type(-name) arguments!

ccwienk (changes requested):

github/compliance/issue.py
```diff
@@ -38,6 +40,21 @@ def _issue_labels(
...
 
 
+def digest_label(
+    label_str: str,
+    length: int=18,
+) -> str:
+    # pylint does not know `length` parameter (it is even required, though!)
+    label_dig =  hashlib.shake_128(label_str.encode('utf-8')).hexdigest(length=length) # noqa
+
+    label = f'ocm/resource/{label_dig}'
```
ccwienk, line 50: also: use ocm/resource | ocm/source, depending on actual type

ccwienk (changes requested):

github/compliance/report.py
```diff
@@ -311,6 +311,23 @@ class PROCESSING_ACTION(enum.Enum):
     REPORT = 'report'
 
 
+def milestone_for_max_days(
+    max_days: float,
+    delivery_svc_client: delivery.client.DeliveryServiceClient,
+    repository: github3.repos.repo.Repository,
+) -> github3.issues.milestone.Milestone:
+    latest_processing_date = datetime.date.today() + datetime.timedelta(days=max_days)
```
ccwienk, line 321: pass-in latest_processing_date
name the function to express this semantics (e.g. latest_milestone_before)

ccwienk: also: is this really worth factoring out?

ccwienk (changes requested):

concourse/model/traits/image_scan.py
```diff
@@ -299,6 +299,9 @@ def for_cve(self, cve_score: float):
         else:
             return self.low
 
+    def for_cfg_element(self):
+        return self.high
```
ccwienk, line 303: this is not a good location for that - consider the module name..

just hard-code - as also done, e.g. for clamav-scans - the severities, based on cfg-element-policy-violation. also, we likely will want to use different priorities (no-owner might have different priority than overdue rotation)

ccwienk (changes requested):

github/compliance/report.py
```diff
@@ -331,6 +333,224 @@ def milestone_for_max_days(
...
+        'cfg_element_name': status.element_name,
+        'cfg_element_type':status.element_type,
+        'cfg_element_storage': status.element_storage,
+        'policy_violations': policy_violations,
+    }
+
+    return template_variables
+
+
+def create_or_update_github_issues_for_cfg(
```
ccwienk, line 375: hmmm... this turns out to be more (partially redundant) LOCs than I hoped for :-/

-> that makes it more reasonable to do unify this w/ ResultGroupCollection, does it not?

ccwienk (changes requested):

github/compliance/issue.py
```diff
@@ -31,6 +31,9 @@
 _label_os_outdated = 'os/outdated'
 _label_malware = 'malware/clamav'
 
+_label_prefix_ocm_resource = 'ocm/resource/'
+_label_prefix_ocm_source = 'ocm/source/'
```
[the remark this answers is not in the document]
ccwienk: prefixes should not have trailing slashes, though (rationale: at least if considering those as "paths", the last respective path ends with the last name)

github/compliance/issue.py
```diff
@@ -220,6 +216,7 @@ def _update_issue(
 
 def create_or_update_issue(
     target: cnudie.iter.ResourceNode | cnudie.iter.SourceNode,
+    digest_label_prefix: str,
```
ccwienk, line 225: ditto

github/compliance/issue.py
```diff
@@ -237,14 +234,16 @@ def create_or_update_issue(
             issue_type=issue_type,
             known_issues=known_issues,
             state='open',
+            digest_label_prefix=digest_label_prefix,
```
ccwienk, line 243: ditto

github/compliance/issue.py
```diff
@@ -237,14 +234,16 @@ def create_or_update_issue(
...
         )
     )
     if (issues_count := len(open_issues)) > 1:
-        raise RuntimeError(f'more than one open issue found for {_target_name(target)=}')
+        raise RuntimeError(f'more than one open issue found for {target_name(target)=}')
```
ccwienk, line 247: should remain "private" (target_name)

github/compliance/issue.py
```diff
@@ -237,14 +234,16 @@ def create_or_update_issue(
...
     elif issues_count == 0:
         return _create_issue(
             target=target,
             issue_type=issue_type,
+            digest_label_prefix=digest_label_prefix,
```
ccwienk, line 252: ditto

github/compliance/issue.py
```diff
@@ -275,6 +274,7 @@ def labels_to_preserve():
         return _update_issue(
             target=target,
             issue_type=issue_type,
+            digest_label_prefix=digest_label_prefix,
```
ccwienk, line 283: ditto

github/compliance/issue.py
```diff
@@ -289,6 +289,7 @@ def labels_to_preserve():
 @github.retry.retry_and_throttle
 def close_issue_if_present(
     target: cnudie.iter.SourceNode | cnudie.iter.ResourceNode,
+    digest_label_prefix: str,
```
ccwienk, line 298: ditto

github/compliance/issue.py
```diff
@@ -299,16 +300,17 @@ def close_issue_if_present(
             issue_type=issue_type,
             known_issues=known_issues,
             state='open',
+            digest_label_prefix=digest_label_prefix,
```
ccwienk, line 309: ditto

github/compliance/issue.py
```diff
@@ -299,16 +300,17 @@ def close_issue_if_present(
...
         )
     )
     open_issues: tuple[github3.issues.ShortIssue]
 
     logger.info(f'{len(open_issues)=} found for closing {open_issues=}')
 
     if (issues_count := len(open_issues)) > 1:
-        logger.warning(f'more than one open issue found for {_target_name(target)=}')
+        logger.warning(f'more than one open issue found for {target_name(target)=}')
```
ccwienk, line 317: should remain private

github/compliance/issue.py
```diff
@@ -299,16 +300,17 @@ def close_issue_if_present(
...
     elif issues_count == 0:
-        logger.info(f'no open issue found for {_target_name(target)=}')
+        logger.info(f'no open issue found for {target_name(target)=}')
```
ccwienk, line 319: should remain private

github/compliance/model.py
```diff
@@ -94,6 +94,7 @@ class ScanResultGroup:
     name: str # {component.name}:{artifact.name}
     results: tuple[ScanResult]
     issue_type: str
+    digest_label_prefix: str
```
ccwienk, line 105: drop

github/compliance/model.py
```diff
@@ -163,6 +164,7 @@ def results_without_findings(self) -> tuple[ScanResult]:
 class ScanResultGroupCollection:
     results: tuple[ScanResult]
     issue_type: str
+    digest_label_prefix: str
```
ccwienk, line 186: drop

github/compliance/model.py
```diff
@@ -185,6 +187,7 @@ def result_groups(self) -> tuple[ScanResultGroup]:
                 name=group_name,
                 results=results,
                 issue_type=self.issue_type,
+                digest_label_prefix=self.digest_label_prefix,
```
ccwienk, line 214: drop

github/compliance/report.py
```diff
@@ -10,6 +10,7 @@
 import cachetools
 import github3
 import github3.issues
+import github3.issues.label
```
ccwienk, line 13: drop

github/compliance/report.py
```diff
@@ -356,6 +357,7 @@ def process_result(
         nonlocal gh_api
         nonlocal err_count
         issue_type = result_group.issue_type
+        digest_label_prefix = result_group.digest_label_prefix
```
ccwienk, line 553: drop

github/compliance/report.py
```diff
@@ -395,6 +397,7 @@ def process_result(
             github.compliance.issue.close_issue_if_present(
                 target=target,
                 issue_type=issue_type,
+                digest_label_prefix=digest_label_prefix,
```
ccwienk, line 578: drop

github/compliance/report.py
```diff
@@ -478,6 +481,7 @@ def process_result(
                 issue = github.compliance.issue.create_or_update_issue(
                     target=target,
                     issue_type=issue_type,
+                    digest_label_prefix=digest_label_prefix,
```
ccwienk, line 648: drop

github/compliance/report.py
```diff
@@ -549,12 +553,14 @@ def sortable_result_str(result: gcm.ScanResult):
...
         close_issues_for_absent_labels(
             issue_type=result_group_collection.issue_type,
+            digest_label_prefix=result_group_collection.digest_label_prefix,
             present_digest_labels=[
-                github.compliance.issue.artifact_digest_label(
-                    component=result_group.component,
-                    artifact=result_group.artifact,
+                github.compliance.issue.digest_label(
+                    label_str=github.compliance.issue.target_name(result_group.target),
+                    prefix=result_group_collection.digest_label_prefix,
```
ccwienk, line 752: pass-in target.

github/compliance/issue.py
```diff
@@ -93,32 +97,19 @@ def _search_labels(
...
-    else:
-        raise NotImplementedError(target)
 
 
 @github.retry.retry_and_throttle
 def enumerate_issues(
     target: cnudie.iter.SourceNode | cnudie.iter.ResourceNode | None,
     known_issues: typing.Sequence[github3.issues.issue.ShortIssue],
     issue_type: str,
+    digest_label_prefix: str='',
```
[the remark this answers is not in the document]
zkdev: Given a 1:1 relation between issue type and digest label, what about scanners capable of yielding scan results for multiple sources?

Let's take Protecode / BDBA for example, if this tool could scan source code for vulnerabilities as well, we would have to introduce a dedicated issue type. We could not recycle `'vulnerabilities/bdba'`.

github/compliance/issue.py
zkdev: What about the logging in `report.py`?

github/compliance/issue.py
zkdev: lets discuss here https://github.com/gardener/cc-utils/pull/775/files#r1025169735

github/compliance/issue.py
zkdev: lets discuss here https://github.com/gardener/cc-utils/pull/775/files#r1025169735

github/compliance/report.py
zkdev: issue_type is required as well

github/compliance/issue.py
```diff
@@ -39,47 +43,70 @@ def _issue_labels(
...
 
     this is useful as GitHub labels are limited to 50 characters
     '''
-    if isinstance(artifact, cm.Resource):
-        name = artifact.name
-    else:
-        name = artifact
 
-    label_str = f'{component.name}:{name}'
+    name = target_name(target=target)
```
zkdev, line 57: Note:
Previously, only for `cm.Resource` the `.name` attribute was used, in `cm.Source` case (e.g. CheckMarx) the whole object was used instead.

ccwienk: I see - that means all cx-issues will likely be re-created after this change, as digest should change, right?
I think that's okay, considering those issues are short-lived anyhow

github/compliance/issue.py
```diff
@@ -39,43 +41,63 @@ def _issue_labels(
...
+    if isinstance(target, (cnudie.iter.ResourceNode, cnudie.iter.SourceNode)):
+        artifact = cnudie.iter.artifact_from_node(target)
+        return f'{target.component.name}:{artifact.name}'
+
     else:
-        raise NotImplementedError(target)
+        raise TypeError(f'{type(target)} not a valid target')
+
+
+def issue_type_digest_label_prefix(
```
ccwienk, line 92: some thoughts:

- this function should be marked as "private" (or rather be inlined..)
- label-prefix should depend on the _target_, not on the issue-type in `digest_label` function

-> remove this function, and simply determine the appropriate prefix - inline -

github/compliance/report.py
```diff
@@ -586,21 +587,24 @@ def close_issues_for_absent_labels(
         state='open',
     )
 
-    def ocm_resource_label(issue: github3.issues.Issue) -> str:
+    digest_label_prefix = github.compliance.issue.issue_type_digest_label_prefix(issue_type)
```
ccwienk, line 779: prefix should depend on target (type). if we do not know the target type, then we will have to search for all known prefixes instead

ccwienk (changes requested):

github/compliance/issue.py
zkdev: Exactly, we won't be able to find existing CX issues -> they will be re-created.

github/compliance/model.py
```diff
@@ -46,9 +46,15 @@ class ScanState(enum.Enum):
     FAILED = 'failed'
 
 
+Target = cnudie.iter.ResourceNode | cnudie.iter.SourceNode
+'''
+type of an element to reference the target of a scan result
```
ccwienk, line 52: this docstr is too short, and gives too little context to actually be useful. I suggest to simply remove it for now, and maybe add a comprehensive documentation, e.g. on module-level, later.

github/compliance/model.py
```diff
@@ -101,14 +107,25 @@ class ScanResultGroup:
     def results_with_successful_scans(self):
         return tuple(r for r in self.results if r.scan_succeeded)
 
+    @property
+    def target(self) -> Target:
```
ccwienk, line 113: consistently call it `scanned_element`, as in the underlying results. is this property actually required?

github/compliance/model.py
```diff
@@ -101,14 +107,25 @@ class ScanResultGroup:
...
+        return self.results[0].scanned_element
+
     @property
     def component(self) -> cm.Component:
-        return self.results[0].scanned_element.component
+        element = self.results[0].scanned_element
+        if not isinstance(element, (cnudie.iter.ResourceNode, cnudie.iter.SourceNode)):
+            raise TypeError(f'unable to get component from {type(element)=}')
```
ccwienk, line 120: generally speaking, just raise TE w/ offending value as parameter
in this specific case, I would, however, use RuntimeError (since no parameter is explicitly passed here; TE is also okay, I think). the exception's message should state that the used property is not allowed to be used if the scanned_element is not either a RN or a SN (basically repeating the condition above).

if just stating "unable to get component from", this is not very helpful, as it does not explain what the expected condition (in which the property/method would work) are.

github/compliance/model.py
```diff
@@ -101,14 +107,25 @@ class ScanResultGroup:
...
+
+        return element.component
 
     @property
     def artifact(self) -> cm.ComponentSource | cm.Resource:
-        result = self.results[0]
-        return cnudie.iter.artifact_from_node(result.scanned_element)
+        element = self.results[0].scanned_element
+        if not isinstance(element, (cnudie.iter.ResourceNode, cnudie.iter.SourceNode)):
+            raise TypeError(f'unable to get artifact from {type(element)=}')
```
ccwienk, line 128: ditto (see comment to component-property)

github/compliance/model.py
zkdev: It's a simple shortcut to the very first result (`results[0].scanned_element`) but its used multiple times.
It follows the same semantics as `@property component` or `@property artifact`.

github/compliance/model.py
```diff
@@ -174,8 +191,11 @@ def result_groups(self) -> tuple[ScanResultGroup]:
             return ()
 
         for result in self.results:
-            artifact = cnudie.iter.artifact_from_node(result.scanned_element)
-            group_name = f'{result.scanned_element.component.name}:{artifact.name}'
+            if isinstance(result.scanned_element, (cnudie.iter.ResourceNode, cnudie.iter.SourceNode)): # noqa: E501
```
ccwienk, line 196: since this is now checked for the third time, why not make this a property? e.g. `scanned_element_is_ocm_artefact`

github/compliance/model.py
```diff
@@ -174,8 +191,11 @@ def result_groups(self) -> tuple[ScanResultGroup]:
...
+                artifact = cnudie.iter.artifact_from_node(result.scanned_element)
```
ccwienk, line 197: mv artifact_from_node from cnudie.iter, e.g. to this module. it is not a good match for cnudie.iter, I think

github/compliance/report.py
```diff
@@ -300,31 +397,25 @@ def _latest_processing_date(
...
-
-    raise RuntimeError(f'{artifact=}')
-
-
 class PROCESSING_ACTION(enum.Enum):
     DISCARD = 'discard'
     REPORT = 'report'
 
 
+def _issue_type_close_comment(issue_type: str) -> str:
```
ccwienk, line 504: this needs to depend on target-type. also, why not simply rephrase to make it generic. the semantics will always be the same regardless the type, will it not?

github/compliance/report.py
```diff
@@ -300,31 +397,25 @@ def _latest_processing_date(
...
+        _compliance_label_vulnerabilities,
+        _compliance_label_licenses,
+        _compliance_label_os_outdated,
+        _compliance_label_checkmarx,
+        _compliance_label_malware,
+    ):
+        return 'closing, because component/resource no longer present in BoM'
+
+    else:
+        raise NotImplementedError(f'{issue_type=}')
```
ccwienk, line 518: I think, during our discussion on builtin exceptions, you found out (and convinced me) that NIE should not be used here (instead, VE should be more adequate)..

github/compliance/report.py
```diff
@@ -363,31 +454,16 @@ def process_result(
...
-            raise ValueError('not all component names are identical')
-
-        component = result_group.component
-        artifact = result_group.artifact
-
-        target = target_from_component_artifact(
-            component=component,
-            artifact=artifact,
-        )
+        if isinstance(target, (cnudie.iter.ResourceNode, cnudie.iter.SourceNode)):
```
ccwienk, line 561: see above - factor-out.

github/compliance/report.py
```diff
@@ -488,19 +550,38 @@ def process_result(
...
 
                         return header + result_group.comment_callback(result)
 
                     def sortable_result_str(result: gcm.ScanResult):
-                        c = result.scanned_element.component
-                        a = cnudie.iter.artifact_from_node(result.scanned_element)
-                        return f'{c.name}:{c.version}/{a.name}:{a.version}'
+                        if isinstance(
+                            result.scanned_element,
+                            (cnudie.iter.ResourceNode, cnudie.iter.SourceNode),
```
ccwienk, line 680: factor-out

github/compliance/report.py
```diff
@@ -561,7 +642,7 @@ def sortable_result_str(result: gcm.ScanResult):
                 for result_group in result_groups
             ],
             known_issues=known_issues,
-            close_comment='closing, because component/resource no longer present in BoM',
+            close_comment=_issue_type_close_comment(result_group_collection.issue_type),
```
ccwienk, line 753: rephrase

ccwienk (changes requested):

github/compliance/issue.py
zkdev: Its used two times
- in `issue.py` to get label prefix when calculating the digest label
- in `report.py` to find all issues for an `issue_type` (could be target ofc)

I would rather have a separate non-private function than inline this twice.

github/compliance/issue.py
```diff
@@ -81,6 +82,9 @@ def target_name(
         artifact = cnudie.iter.artifact_from_node(target)
         return f'{target.component.name}:{artifact.name}'
 
+    elif isinstance(target, cmr.CfgElementStatusReport):
+        return target.name
+
     else:
         raise TypeError(f'{type(target)} not a valid target')
```
ccwienk, line 89: the exception-message is not quite accurate. most likely, this exception will be raised if we add (sometime in the future) another "target-type", and forget to add a codepath for it here. in which case the value would very well be valid, but not yet "supported".

at any rate: just raise TE w/ the offending value as sole parameter. always. be specific if and only if you are sure to always be correct. otherwise, be less specific. this will also save some typing

github/compliance/report.py
zkdev: We did agree on NIE for iterating over Enum. Of course, this is not an Enum but it feels quite similar. But no strong opinion of mine.

github/compliance/report.py
zkdev: What exactly, the typing tuple

github/compliance/model.py
zkdev: What about the `_target_*` functions, such as `_target_repository` or `_target_assignees`?
Should we still pass target and manually check for `isinstance(target, (SN, RN))` or alter to:
```python
def _scan_result_repository(result: ScanResult):
  if result.scanned_element_is_ocm_resource:
    ...
```
?

github/compliance/issue.py
ccwienk: agree w.r.t. not duplicating the function. change to calculate prefix depending on target's type, though.

github/compliance/model.py
ccwienk: yeah, this is pretty hacky (has been before already). which operations are done when it is used? maybe it would be better to offer shortcuts to those functions, instead.

github/compliance/model.py
ccwienk: exactly, this deduplication was what I had in mind.

github/compliance/report.py
ccwienk: I am fine w/ NIE, considering we use it all over our codebase already.

however, please do not unnecessarily wrap values. in this case, we will even lose information as `__str__`  will be called, which most likely will yield less output.

I cannot stress this (or so it seems) often enough: please just pass the offensive values _unaltered_ to builtin exceptions, such as ValueError, TypeError, NotImpementedError,. ..

If - and _only if_ - there is more specific (and accurate (!) information to be added that will be helpful (!) for someone reading the exception-message - then, and only then, add an error-text.

in this specific case (we use this pattern a lot), the only reason why this exception will ever be raised will be because either we did a programming error and passed-in a completely wrong value, or because we extended to allowed set of values (/types), but did forgot this one location. in which case the stacktrace + full dump about the value is exactly what we need. anything else will actually make it harder to quickly identify what the underlying issue is.

github/compliance/model.py
zkdev: A `ScanResultGroup.target` is used as an representative `ScanResult.scanned_element` to get common things across `ScanResult`s in a `ScanResultGroup`.
Its used e.g. to determine responsibles or reporting repository for a `ScanResultGroup`.

We could consider adding those functions directly to the model class or add shortcuts, but in the end the underlaying hacky part with all its implications remains.

ccwienk (approved):

PR #776, opened 2022-11-11
Log gh session requests to ES

zkdev:

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/ccc/github.py
+++ b/ccc/github.py
@@ -47,7 +47,6 @@ class SessionAdapter(enum.Enum):
 
 def github_api_ctor(
     github_url: str,
-    # can't use bitwise-or for forward reference
     es_client: typing.Optional['ccc.elasticsearch.ElasticSearchClient']=None,
     verify_ssl: bool=True,
     session_adapter: SessionAdapter=SessionAdapter.RETRY,
@@ -58,7 +57,7 @@ def github_api_ctor(
     returned with the url argument preset, thus disburdening users to differentiate
     between github.com and non-github.com cases.
 
-    If es_client is provided, session requests are logged to ES.
+    If es_client is passed, all requests against github-api will be logged to Elasticsearch.
     '''
     parsed = urllib.parse.urlparse(github_url)
     if parsed.scheme:
@@ -69,11 +68,14 @@ def github_api_ctor(
     session = github3.session.GitHubSession()
     original_request = session.request
 
-    def intercepted_request(*args, **kwargs):
-        req = original_request(*args, **kwargs)
+    def intercepted_request(
+        method: str,
+        url: str,
+        **kwargs,
+    ):
+        req = original_request(method, url, **kwargs)
 
         try:
-            method, url = args
             es_client.store_document(
                 index='github_request',
                 body={
@@ -84,8 +86,7 @@ def github_api_ctor(
                 },
             )
         except:
-            # dont raise due to logging
-            logger.warning('unable to log session requests to ES')
+            logger.warning('unable to log github api request to Elasticsearch')
 
         return req
 
--- a/ccc/github.py
+++ b/ccc/github.py
@@ -47,6 +47,7 @@ class SessionAdapter(enum.Enum):
 
 def github_api_ctor(
     github_url: str,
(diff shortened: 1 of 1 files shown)


ccc/github.py
```diff
@@ -47,6 +47,8 @@ class SessionAdapter(enum.Enum):
 
 def github_api_ctor(
     github_url: str,
+    # can't use bitwise-or for forward reference
+    es_client: typing.Optional['ccc.elasticsearch.ElasticSearchClient']=None,
```
ccwienk, line 51: while this comment may be helpful to understand this specific diff / why you chose not to use the  shorthand `TypeA | None` in this case, I think this comment will be confusing if read later. also, `typing.Optional` is a fine choice - there is no reason to "justify" using it.

you should be able to write the entire typehint as a str-literal, btw (`'ESC | None'`); considering there is a default-value defined, one might also argue it would be okay to simply not declare the argument to be optional (this is fairly obvious from the presence of a default-value).

in short: please rm the comment - the code itself is fine w/ me, of course

ccc/github.py
```diff
@@ -55,6 +57,8 @@ def github_api_ctor(
     In case github_url does not refer to github.com, the c'tor for GithubEnterprise is
     returned with the url argument preset, thus disburdening users to differentiate
     between github.com and non-github.com cases.
+
+    If es_client is provided, session requests are logged to ES.
```
ccwienk, line 61: it is okay to abbreviate elastisearch-client to es_client, esp. considering there is a typehint. However, please do not abreviate Elasticsearch to just ES in the docstr.

some more nits:
- this is a plain function; es_client is a formal parameter of this function. I think it is more conventional to speak of "passing an argument", or "giving" (as in: if es_client is passed / given / non-none)
- while it is technically correct that "session requests" will be logged, this is not the relevant semantics, but an implementation detail. what happens is an elasticsearch-client is passed, is any api-requests issues by github-api-clients created by the returned function will be logged (to elasticsearch). we might shorten this to just "all requests against github-api will be logged", for example (document WHAT/WHY, not HOW - the latter is already written down in the actual code; documentation should almost never repeat the code)

ccc/github.py
```diff
@@ -63,6 +67,31 @@ def github_api_ctor(
         raise ValueError('failed to parse url: ' + str(github_url))
 
     session = github3.session.GitHubSession()
+    original_request = session.request
+
+    def intercepted_request(*args, **kwargs):
+        req = original_request(*args, **kwargs)
+
+        try:
+            method, url = args
```
ccwienk, line 76: .. how about moving positional parameters to the signature of `intercepted_request`?

ccc/github.py
```diff
@@ -63,6 +67,31 @@ def github_api_ctor(
...
+                index='github_request',
+                body={
+                    'method': method,
+                    'url': url,
+                    'data': kwargs,
+                    'creation_date': datetime.datetime.now().isoformat(),
+                },
+            )
+        except:
+            # dont raise due to logging
```
ccwienk, line 87: nit: it is don't (conjunction form for do not)
also: It should already be obvious from try-except w/ warning as single instruction that we do not want to raise. so I would argue the comment could simply be dropped

however, if keeping it, I would rather - as always - see comments that explain WHAT and WHY, rather than repeating the code. we _ignore_ errors from logging. this is different from the comment above (which just states "due to logging").

as a general recommendation, just consider omitting comments in cases where there is no need to explain the underlying idea. but if adding comments, please always be sure to not just repeat the code, and focus on the "higherlevel" semantics, as understood by / relevant for a user | maintainer of the documented piece of code

ccc/github.py
```diff
@@ -63,6 +67,31 @@ def github_api_ctor(
...
+            logger.warning('unable to log session requests to ES')
```
ccwienk, line 88: please do not abbreviate elasticsearch to just ES - this will make the log output very hard to understand

also: please _do_ state that we were unable to log a Github-API-Request! just calling this a "session request" if meaningless w/o knowing the code. Please consider there might be readers of the warnings that do not know our code-base in detail, and just see the warning-messages.

ccwienk (changes requested):

PR #778, opened 2022-11-23
Add optional ctx label(s)

zkdev:
Enables issue subsets with all issues sharing the same issue_type.

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/github/compliance/issue.py
+++ b/github/compliance/issue.py
@@ -33,6 +33,7 @@ _label_cfg_policy_violation = 'cfg-element/policy-violation'
 
 _label_prefix_ocm_artefact = 'ocm/artefact'
 _label_prefix_cicd_cfg_element = 'cicd-cfg-element'
+_label_prefix_ctx = 'ctx'
 
 
 @cachetools.cached(cache={})
@@ -123,6 +124,7 @@ def enumerate_issues(
     scanned_element: gcm.Target | None,
     known_issues: typing.Sequence[github3.issues.issue.ShortIssue],
     issue_type: str,
+    extra_labels: typing.Iterable[str]=(),
     state: str | None = None, # 'open' | 'closed'
 ) -> typing.Generator[github3.issues.ShortIssue, None, None]:
     '''Return an iterator iterating over those issues from `known_issues` that match the given
@@ -131,6 +133,7 @@ def enumerate_issues(
     labels = frozenset(_search_labels(
         scanned_element=scanned_element,
         issue_type=issue_type,
+        extra_labels=extra_labels,
     ))
 
     def filter_relevant_issues(issue):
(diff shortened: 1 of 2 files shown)


github/compliance/report.py
```diff
@@ -770,10 +790,23 @@ def sortable_result_str(result: gcm.ScanResult):
     if overwrite_repository:
         known_issues = _all_issues(overwrite_repository)
         issue_type = result_group_collection.issue_type
+
+        all_ctx_labels = set()
+
+        if not result_groups:
+            all_ctx_labels = result_group_collection.ctx_labels
```
zkdev, line 797: This is a potential error source resulting in unintended closed issues.

Given following example:
`_scanned_element_ctx_label` yields two ctx labels, `ctx/A` and `ctx/B`.
The passed `fallback` ctx labels (for no scan result case) is just `ctx/A`.

As the fallback is included, the issue filtering will yield the issue, thus "ignoring" `ctx/B`.
We check whether all `search_labels` are included in `issue_labels` (intersection, see [here](https://github.com/gardener/cc-utils/blob/bd5c0abac87b7ab3b2a609bd1f57176c33a23b8f/github/compliance/issue.py#L143)).

zkdev: We could consider dropping the "dynmaic" ctx label calculation and use the current "fallback" as single source of ctx labels, or tweak our issue filtering to ensure all ctx labels match (e.g. by defining a fixed ctx label prefix).

github/compliance/issue.py
```diff
@@ -234,20 +236,27 @@ def create_or_update_issue(
     latest_processing_date: datetime.date|datetime.datetime=None,
     extra_labels: typing.Iterable[str]=None,
     preserve_labels_regexes: typing.Iterable[str]=(),
+    ctx_labels: typing.Iterable[str]=(),
```
ccwienk, line 239: a docstr - _briefly_ explaining the semantics of non-obvious parameters (such as ctx_labels) would be nice.

in particular, there are both extra_labels and ctx_labels - one might wonder where we have the two, rather than e.g. just extra_labels (I know: we discussed this, and I am not saying we should not have two parameters; but there should be some explanation as to _why_).

something like so:

creates or updates a github issue for the given `scanned_element`. If no issue exists, yet, it will be created, otherwise, it is checked that at most one matching issue exists; which will then be updated.

Issues are found by labels. Some of those are derived from the scanned_element. If given, ctx_labels are also included in search-query (i.e. an issue must have all of the given ctx_labels to be considered a match).

All of those labels + any given `extra_labels` are assigned to the issue, both in case it is created anew, or updated. Extra labels that are not ignored via the `preserve_labels_regexes` argument are dropped.

ccwienk (changes requested):

github/compliance/issue.py
```diff
@@ -234,20 +237,38 @@ def create_or_update_issue(
...
 ) -> github3.issues.issue.ShortIssue:
+    '''
+    Creates or updates a github issue for the given scanned_element. If no issue exists, yet, it will
+    be created, otherwise, it is checked that at most one matching issue exists; which will then be
+    updated.
+    Issues are found by labels. Some of those are derived from the scanned_element. If given,
+    ctx_labels are also included in search-query (i.e. an issue must have all of the given ctx_labels
+    to be considered a match). All of those labels + any given extra_labels are assigned to the
+    issue, both in case it is created anew, or updated. Extra labels that are not ignored via the
+    preserve_labels_regexes argument are dropped.
```
ccwienk, line 250: need to add all labels starting w/ ctx-prefix to "labels to preserve", do we not?

ccwienk (approved):

PR #784, opened 2022-11-29
Dedicated label for cfg policy violations

zkdev:
Policy violations have different severities, thus each violation requires a dedicated issue.

Create result_group_collections for passed policy violations with dedicated label. Adjust ticket text to only expect a single policy violation. Have a common prefix for policy violation labels (issue types). Also remove comment callback, as the only update a policy violation ticket can have is to close it (cfg element dropped or violation resolved).

Fixes #

**Release note**:

```other operator

```


zkdev:
/hold
until extra args are passed properly

concourse/steps/replicate_secrets.py
```diff
@@ -176,13 +212,24 @@ def report_cfg_policy_status(
         for status_report in status_reports
     ]
 
-    result_group_collection = cmu.scan_result_group_collection(results)
+    def iter_scan_result_group_collection() -> typing.Generator[gcm.ScanResultGroupCollection, None, None]: # noqa: E501
```
ccwienk, line 215: break line rather than silencing linter

concourse/steps/replicate_secrets.py
```diff
@@ -153,13 +154,48 @@ def replicate_secrets(
     logger.info(f'deployed encrypted secret for team: {team_name}')
 
 
+def _issue_type_for_policy_violation(
+    policy_violation: cmm.CfgStatusEvaluationAspects,
```
ccwienk, line 158: I seem to have missed this in previous PRs. please choose a more adequate name for this enum. do we even need this enum? just use the type-labels. then you even save some code (this function, ..)

as for the name: those are compliance _violations_. An "evaluation aspect" could be the inspection of whether or not a certain criterion is met. This is not identical to the _result_ of such a check. Please, as mentioned before, take care to choose names that match the semantics. Use simple terms, do not add verbal boilerplate.

concourse/steps/replicate_secrets.mako
```diff
@@ -13,8 +13,9 @@ job_mapping_name = extra_args['job_mapping_name']
 secrets_repo_url = extra_args['secrets_repo_url']
 compliance_reporting_repo_url = extra_args['compliance_reporting_repo_url']
 delivery_endpoints_cfg_name = extra_args['delivery_endpoints_cfg_name']
-github_issue_template = extra_args['github_issue_template']
+github_issue_template_cfgs = extra_args['github_issue_template_cfgs']
```
ccwienk, line 16: since we are back to passing-in a mapping, revert the change previously introduced for passing-in a single template.

cfg_mgmt/util.py
```diff
@@ -496,23 +495,23 @@ def process_cfg_queue_and_persist_in_repo(
...
 ) -> gcm.ScanResultGroupCollection:
     def classification_callback(result: gcm.CfgScanResult) -> gcm.Severity:
-        return gcm.Severity.HIGH
+        return severity
 
     def findings_callback(result: gcm.CfgScanResult) -> bool:
-        if not result.evaluation_result.nonCompliantReasons:
-            return False
+        policy_violations = result.evaluation_result.nonCompliantReasons
+        if policy_violation in policy_violations:
```
ccwienk, line 507: see comment w.r.t. issue-type - merge enum w/ labels, please.

ccwienk (changes requested):

cfg_mgmt/util.py
zkdev: As discussed, keep labels and compliance violations duplicated (due to circular dependency) and have a consistent mapping (rather than passing around).

concourse/steps/replicate_secrets.py
zkdev: I agree, the naming is not matching the semantics. Let's handle renaming `CfgStatusEvaluationAspects` -> `CfgStatusComplianceViolations` in a separate PR.

github/compliance/model.py
```diff
@@ -37,6 +37,30 @@ def parse(value: int | str):
         return Severity(value)
 
 
+@dataclasses.dataclass(frozen=True)
+class MaxProcessingTimesDays:
+    '''
+    defines maximum processing time in days, based on issue "criticality"
```
ccwienk, line 43: add hint about this dataclass being used for deserialisation from pipeline-definitions into docstr

github/compliance/report.py
```diff
@@ -596,23 +596,14 @@ def _target_sprint(
     return target_sprint
 
 
-def _latest_processing_date(
```
ccwienk, line 567: was this function unused? if not, this commit (eccdf62ac9ec9d0079cf726eeec232a5af1a19b5) will lead to a broken state

cfg_mgmt/util.py
```diff
@@ -494,25 +494,46 @@ def process_cfg_queue_and_persist_in_repo(
     return True
 
 
+def _policy_violation_for_issue_type(issue_type: str) -> cmm.CfgStatusEvaluationAspects:
+    if issue_type == gci._label_no_responsible:
+        return cmm.CfgStatusEvaluationAspects.NO_RESPONSIBLE
```
ccwienk, line 499: as stated before, please rename enum. for example to just "CfgElementPolicyViolation"

cfg_mgmt/util.py
```diff
@@ -494,25 +494,46 @@ def process_cfg_queue_and_persist_in_repo(
...
+
+
 def scan_result_group_collection(
     results: tuple[gcm.CfgScanResult],
+    severity: gcm.Severity,
+    issue_type: str,
 ) -> gcm.ScanResultGroupCollection:
     def classification_callback(result: gcm.CfgScanResult) -> gcm.Severity:
-        return gcm.Severity.HIGH
+        return severity
```
ccwienk, line 523: hardcode severity here; there is not need to pass it in for now

github/compliance/report.py
```diff
@@ -47,11 +47,35 @@
...
+_compliance_label_credentials_outdated = github.compliance.issue._label_outdated_credentials
+_compliance_label_no_responsible = github.compliance.issue._label_no_responsible
+_compliance_label_no_rule = github.compliance.issue._label_no_rule
+_compliance_label_no_status = github.compliance.issue._label_no_status
+_compliance_label_undefined_policy = github.compliance.issue._label_undefined_policy
 
 _ctx_label_prefix = github.compliance.issue._label_prefix_ctx
 
 
+def _friendly_label_name(label: str) -> str:
```
ccwienk, line 59: why? just include this in the issue-template(s), as done for the other compliance-type-labels

github/compliance/report.py
```diff
@@ -169,7 +191,7 @@ def _cfg_policy_violation_status_summary(
         | Element Type | {scanned_element.element_type} |
         | Element Name | {scanned_element.element_name} |
         | {_pluralise("Responsible", responsibles_len)} | {responsibles} |
-        | {_pluralise(issue_description, policy_violation_count)} | {issue_value} |
+        | {issue_description} | {issue_value} |
```
ccwienk, line 194: place description below summary table

github/compliance/report.py
```diff
@@ -393,8 +416,17 @@ def _template_vars(
     elif issue_type == _compliance_label_malware:
         template_variables |= _malware_template_vars(result_group)
 
-    elif issue_type == _compliance_label_cfg_policy_violation:
-        template_variables |= _cfg_policy_violation_template_vars(result_group)
+    elif issue_type in (
```
ccwienk, line 419: place instrumentation to using pipeline-template

ccwienk (changes requested):

github/compliance/report.py
zkdev: Yes, it was unused across cc-utils and our other repos.

cfg_mgmt/util.py
zkdev: Please see https://github.com/gardener/cc-utils/pull/784#discussion_r1038124370.
I'd prefer to do this in a separate PR.

cfg_mgmt/util.py
```diff
@@ -494,25 +494,46 @@ def process_cfg_queue_and_persist_in_repo(
     return True
 
 
+def _policy_violation_for_issue_type(issue_type: str) -> cmm.CfgStatusEvaluationAspects:
```
ccwienk, line 497: I just checked the `cfg_mgmt` package and found there is a `reporting.py` submodule. please move reporting-related code from `cfg_mgmt/util.py` to `cfg_mgmt/reporting.py`

considering the existing reporting-functions (e.g. [this one](https://github.com/gardener/cc-utils/blob/master/concourse/steps/scan_container_images.py#L46), I think it is not (fully) adequate (or at least not symmetrical/inconsistent) to add the instrumenting code into our reporting-module. a better place might be step-template (which you already touched in this PR).

ccwienk (changes requested):

github/compliance/report.py
```diff
@@ -393,8 +385,17 @@ def _template_vars(
...
+        _compliance_label_credentials_outdated,
+        _compliance_label_no_responsible,
+        _compliance_label_no_rule,
+        _compliance_label_no_status,
+        _compliance_label_undefined_policy,
+    ):
+        template_variables |= _cfg_policy_violation_template_vars(
```
ccwienk, line 395: if this group of issue-types shares commit template-vars, define a dedicated function for those, rather than passing-ing issue-type. or just add the one additional template-var at the beginning. do we actually need this as template-var? why not simply add to the type-specific template?

ccwienk (changes requested):

PR #786, opened 2022-12-14
Also consider failed checkmarx scans for issue reporting

zkdev:
GitHub issue reporting closes issues for absent artefacts, therefore all artefacts must be known..
In case of checkmarx scan failure, no scan result was provided resulting in issue reporting assuming artefact was absenti, thus closing issue(s).

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/checkmarx/util.py
+++ b/checkmarx/util.py
@@ -6,7 +6,6 @@ import logging
 import shutil
 import tarfile
 import tempfile
-import traceback
 import typing
 import zipfile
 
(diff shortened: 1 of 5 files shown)


checkmarx/util.py
```diff
@@ -306,6 +303,9 @@ def scan_gh_artifact(
 
 
 def greatest_severity(result: model.ScanResult) -> model.Severity | None:
+    if not result.scan_statistic:
```
ccwienk, line 306: raise (e.g. runtime-error); explicitly check for error-state? if there is not scan-statistics, result must be erronrous (scan failed). in that case, it is an error to read severity.

PR #788, opened 2023-01-11
Log github-api requests to Elasticsearch (if default cfg is found)

zkdev:

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/ccc/github.py
+++ b/ccc/github.py
@@ -49,12 +49,15 @@ def github_api_ctor(
     github_url: str,
     verify_ssl: bool=True,
     session_adapter: SessionAdapter=SessionAdapter.RETRY,
+    cfg_factory: model.ConfigFactory=None,
 ):
     '''returns the appropriate github3.GitHub constructor for the given github URL
 
     In case github_url does not refer to github.com, the c'tor for GithubEnterprise is
     returned with the url argument preset, thus disburdening users to differentiate
     between github.com and non-github.com cases.
+
+    github-api requests will be logged to Elasticsearch if a default Elasticsearch config is found.
     '''
     parsed = urllib.parse.urlparse(github_url)
     if parsed.scheme:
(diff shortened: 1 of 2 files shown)


ccc/github.py
```diff
@@ -63,6 +66,46 @@ def github_api_ctor(
         raise ValueError('failed to parse url: ' + str(github_url))
 
     session = github3.session.GitHubSession()
+    original_request = session.request
+
+    short = False
```
AndreasBurger, line 71: consider naming this `skip_logging` or even `skip_logging_to_els` or something.

ccc/github.py
```diff
@@ -63,6 +66,46 @@ def github_api_ctor(
...
+                body={
+                    'method': method,
+                    'url': url,
+                    'data': kwargs,
+                    'creation_date': datetime.datetime.now().isoformat(),
+                },
+            )
+        except:
+            logger.warning('unable to log github api request to Elasticsearch')
+            short = True
```
AndreasBurger, line 96: We won't log for retries if the original request failed to log, right? Not an issue, I think, but is this really necessary?

AndreasBurger (approved):

ccc/github.py
ccwienk: +1
as an alternative, one might also consider `shortcut_logging` (as for naming)

however, if we are persistently shortcutting logging (which is a valid approach), why not rather patch-back to the original method, thus saving a little overhead?

also, in addition to logging a warning, it would be nice to also state the future requests in same session will not even be attempted to be logged

ccwienk (changes requested):
generally okay - as @AndreasBurger pointed out, please choose more expressive names for shortcutting-variable

ccc/github.py
zkdev: > We won't log for retries if the original request failed to log, right?

Correct.

However, it is not about single requests with retries but multiple requests.
In case logging fails (elasticsearch down, ...) we would write to stderr for every request, resulting in spam.
Therefore, shortcut logging for github-api object on first logging failure.

ccc/github.py
zkdev: do agree, patching back instead of shortcutting is better 👍

ccwienk (approved):

PR #789, opened 2023-01-13
Adjust GCP ServiceAccount key rotation to understand GCP config

zkdev:

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/cfg_mgmt/gcp.py
+++ b/cfg_mgmt/gcp.py
@@ -14,6 +14,7 @@ import ci.log
 import ci.util
 import model
 import model.container_registry
+import model.gcp
 
 
 ci.log.configure_default_logging()
@@ -53,7 +54,7 @@ def delete_service_account_key(
 
 
 def rotate_cfg_element(
-    cfg_element: model.container_registry.ContainerRegistryConfig,
+    cfg_element: model.container_registry.ContainerRegistryConfig | model.gcp.GcpServiceAccount,
     cfg_factory: model.ConfigFactory,
 ) ->  typing.Tuple[cfg_mgmt.revert_function, dict, model.NamedModelElement]:
     client_email = cfg_element.client_email()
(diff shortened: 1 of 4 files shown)


model/gcp.py
```diff
@@ -10,6 +10,12 @@ def service_account_key(self):
         '''
         return self.raw['service_account_key']
 
+    def service_account_credentials(self): # -> 'google.oauth2.service_account.Credentials':
```
ccwienk, line 13: nit: why did you comment-out the typehint?

cfg_mgmt/gcp.py
```diff
@@ -53,10 +54,19 @@ def delete_service_account_key(
...
 ) ->  typing.Tuple[cfg_mgmt.revert_function, dict, model.NamedModelElement]:
-    client_email = json.loads(cfg_element.password())['client_email']
+    if not type(cfg_element) in (
+        model.container_registry.ContainerRegistryConfig,
+        model.gcp.GcpServiceAccount,
+    ):
+        raise TypeError(cfg_element)
+
+    if isinstance(cfg_element, model.container_registry.ContainerRegistryConfig):
+        client_email = json.loads(cfg_element.password())['client_email']
```
ccwienk, line 67: it would be nicer, and more consistent to remaining codebase to implement the retrieval of client_email and private_key_id in model-classes at `model` package

cfg_mgmt/gcp.py
```diff
@@ -53,10 +54,19 @@ def delete_service_account_key(
...
+    elif isinstance(cfg_element, model.gcp.GcpServiceAccount):
+        client_email = cfg_element.service_account_key()['client_email']
```
ccwienk, line 69: rather add `else: raise ValueError` here (if you do, you might drop the initial value-checking at the beginning of the function.
rationale: there are multiple places where additional, typespecific handling is required, which all need to be considered. should a third option be added, there is a change extension of typespecific handling might be missed, resulting, potentially in a non-obvious follow-up error later

cfg_mgmt/gcp.py
```diff
@@ -66,7 +76,11 @@ def rotate_cfg_element(
         client_email,
     )
 
-    old_key_id = json.loads(cfg_element.password())['private_key_id']
+    if isinstance(cfg_element, model.container_registry.ContainerRegistryConfig):
+        old_key_id = json.loads(cfg_element.password())['private_key_id']
+    elif isinstance(cfg_element, model.gcp.GcpServiceAccount):
+        old_key_id = cfg_element.service_account_key()['private_key_id']
```
ccwienk, line 82: ditto

ccwienk (changes requested):

cfg_mgmt/gcp.py
zkdev: Agree, [adjusted ContainerRegistryConfig](https://github.com/gardener/cc-utils/commit/be471e4fd08402cf01a292d66878b0d4f1a500df) and will also do for GcpServiceAccount.

ccwienk (approved):

PR #793, opened 2023-01-18
Upload checkmarx scans to deliverydb

zkdev:

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/checkmarx/util.py
+++ b/checkmarx/util.py
@@ -12,6 +12,7 @@ import zipfile
 import github3.exceptions
 import github3.repos
 
+import ccc.delivery
 import ccc.github
 import cnudie.retrieve
 import checkmarx.client
@@ -19,6 +20,7 @@ import checkmarx.model as model
 import checkmarx.project
 import checkmarx.tablefmt
 import ci.util
+import github.compliance.model
 import model.checkmarx as cmmmodel
 import product.util
 import reutil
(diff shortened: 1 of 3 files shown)


dso/model.py
```diff
@@ -118,8 +119,18 @@ class MalwareSummary:
     findings: list[clamav.client.ScanResult]
 
 
+@dataclasses.dataclass(frozen=True)
+class Codecheck:
```
ccwienk, line 123: naming: rather call this CodecheckResult, or CodecheckSummary (similar to MalwareSummary), or CodecheckFindings

as for the attributes, rather name those, e.g. high_findings. or add an attribut `findings`, like so:
```
findings:
 high
 medium
 low
 info
```

dso/model.py
```diff
@@ -118,8 +119,18 @@ class MalwareSummary:
...
+    high: int
+    medium: int
+    low: int
+    info: int
+    scanRisk: int
```
ccwienk: this name seems odd. should this not rather be named sth. like `risk_rating` (or just `risk`)?

dso/model.py
```diff
@@ -118,8 +119,18 @@ class MalwareSummary:
...
+    scanRiskSeverity: int
```
ccwienk: ditto

checkmarx/util.py
```diff
@@ -176,6 +179,15 @@ def init_scan(
             finished += 1
             logger.info(f'remaining: {artifact_count - finished}')
 
+    try:
+        upload_results_to_deliverydb(
```
ccwienk, line 183: please mv this to the step-implementation. doing such operations implicitly I consider a bad pattern (admittedly one that we also have in other places)

consider the name of the method you added this code to - it reads `init_scan`. this is strongly misleading.

instead, please rework the code like so (pseudo-pseudo-code):

```
scan_results = scan_with_checkmarx(source_nodes_to_scan)
update_github_issues(scan_results)
upload_to_delivery_service(scan_results)
```

checkmarx/util.py
```diff
@@ -473,3 +485,37 @@ def get_checkmarx_cfg(checkmarx_cfg_name: str) -> cmmmodel.CheckmarxConfig:
 @functools.lru_cache()
 def create_checkmarx_client(checkmarx_cfg: cmmmodel.CheckmarxConfig):
     return checkmarx.client.CheckmarxClient(checkmarx_cfg)
+
+
+def upload_results_to_deliverydb(
```
ccwienk, line 490: naming-nit: while it is true that the results end up in delivery-db, we actually upload them to delivery-service.

ccwienk (changes requested):

dso/model.py
ccwienk: also, I just saw you already did change those names in the second commit from this pull-request.. please mv the diffs between commits accordingly

checkmarx/util.py
```diff
@@ -481,3 +484,39 @@ def get_checkmarx_cfg(checkmarx_cfg_name: str) -> cmmmodel.CheckmarxConfig:
 @functools.lru_cache()
 def create_checkmarx_client(checkmarx_cfg: cmmmodel.CheckmarxConfig):
     return checkmarx.client.CheckmarxClient(checkmarx_cfg)
+
+
+def upload_results_to_delivery_service(
```
ccwienk, line 489: code organisation could still be improved:

- have a function that will create artefactmetadata from scan-result
- call this function in template, where scan-results are present

instrumenting delivery-service-client is not specific to the type of uploaded data, thus such code should not (redundantly) be placed in packages specific to certain metadata types.

concourse/steps/scan_sources.mako
```diff
@@ -129,4 +131,18 @@ github.compliance.report.create_or_update_github_issues(
 )
 % endif
 
+delivery_service_client = ccc.delivery.default_client_if_available()
+if delivery_service_client:
+    logger.info('Uploading result to delivery-service')
+    try:
+        checkmarx.util.upload_results_to_delivery_service(
```
ccwienk, line 138: I think there is no need to be thus defensive. the data-upload happens as last step - so there is nothing bad happening if we let an error propagate.

ccwienk (changes requested):

ccwienk (approved):

PR #804, opened 2023-02-01
Close issues with recursive retry

zkdev:

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/github/compliance/issue.py
+++ b/github/compliance/issue.py
@@ -14,6 +14,7 @@ import cfg_mgmt.reporting as cmr
 import ci.log
 import github.compliance.model as gcm
 import github.retry
+import github.util
 
 '''
 functionality for creating and maintaining github-issues for tracking compliance issues
@@ -351,7 +352,7 @@ def close_issue_if_present(
     for issue in open_issues:
         issue: github3.issues.ShortIssue
         issue.create_comment('closing ticket, because there are no longer unassessed findings')
-        if not issue.close():
+        if not github.util.close_issue(issue):
             logger.warning(f'failed to close {issue.id=}, {repository.url=}')
 
     return issue
--- a/github/compliance/report.py
+++ b/github/compliance/report.py
@@ -887,7 +887,7 @@ def close_issues_for_absent_resources(
                 "scanned element matching its digest."
             )
             issue.create_comment('closing, because scanned element no longer present in BoM')
-            issue.close()
+            github.util.close_issue(issue)
 
     def has_ctx_label(
         issue: github3.issues.Issue,
--- a/github/util.py
+++ b/github/util.py
@@ -28,6 +28,7 @@ from pydash import _
 import requests
 
 import github3
+import github3.issues
 from github3.exceptions import NotFoundError
 from github3.github import GitHub
 from github3.orgs import Team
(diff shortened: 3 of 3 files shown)


github/compliance/issue.py
```diff
@@ -316,6 +316,27 @@ def labels_to_preserve():
...
+    issue: github3.issues.ShortIssue,
+) -> bool:
+    '''
+    close issue with recursive retries
+    '''
+    try:
+        return issue.close()
+
+    except github3.exceptions.UnprocessableEntity:
+        # assignee was suspended from github
```
ccwienk, line 329: do we know this for sure?  if it is just _one_ (likely) cause, we should say so

zkdev: will re-phrase 👍🏻

ccwienk (changes requested):
generally speaking, functions that implement retry-semantics should always have a strict limit as to the amount of retries.

considering naming, `retry` is not a good match. if we call something "retry", my expectation is that we try the exact same operation (that is prone to sporadically occuring transient errors) again; this function, however, is different in that it has a special-case handling for a certain type of systematic (and thus persistent) error.

a better name for such a function might be, e.g. `purge_issue` (my personal wording-style - rationale: purging is more radical; downside: the issues is actually only closed, and never removed), or `stubbornly_close_issue`, or `tenaciously_close_issue`; or just `close_issue` (I mean, why not; if someone wonders why the regular issue.close is not good enough, this can be learnt from docstr).

as for the actual code, I think it is not a good idea to "retry" more than once.

quite the contrary: try once (regularly), then upon "expected" error, remove assignees, and try exactly once more. publish comment if this fails. (re-)raise in that case.

ccwienk:
additional remark: there are cases where it is convenient to not raise exceptions, and report success/failure via return-values (e.g. when implementing some lookup function, where it is a usual case that no result was found); however, as a general rule-of-thumb, errors/exceptions should be signalled using, well, exceptions (python is - fortunately - not golang..)

github/compliance/issue.py
```diff
@@ -316,6 +316,27 @@ def labels_to_preserve():
...
+
+        if not issue.assignees:
+            return False
```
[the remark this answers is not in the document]
zkdev: We do use the return value, see https://github.com/gardener/cc-utils/blob/0a0f8092736a53ada5019a00a2984fa876f25d03/github/compliance/issue.py#L354.
Agree on raise instead of `False`.

ccwienk (approved):

github/compliance/report.py
```diff
@@ -887,7 +887,7 @@ def close_issues(
                 "scanned element matching its digest."
             )
             issue.create_comment('closing, because scanned element no longer present in BoM')
-            issue.close()
+            github.compliance.issue.close_issue(issue)
```
ccwienk, line 890: one more thing: I think the `close_issue` function would fit better into github.util

ccwienk: otherwise. lgtm

ccwienk:
@zkdev : see last inline-comment - otherwise lgtm. thx

github/compliance/report.py
zkdev: will mv 👍

PR #812, opened 2023-02-15
Enhance service binding client

dimityrmirchev:
This PR adds:
 - additional function called `get_service_binding` to the SBClient
 - additional fields `created_at` and `updated_at` to be returned when getting a service binding

Fixes #

/cc @MartinWeindel
 
**Release note**:

```other operator
NONE
```

### DIFF
--- a/cfg_mgmt/btp_service_binding.py
+++ b/cfg_mgmt/btp_service_binding.py
@@ -22,6 +22,12 @@ class ServiceBindingInfo:
     id: str
     name: str
     service_instance_id: str
+    created_at: str
+    updated_at: str
+
+@dataclass
+class ServiceBindingDetailedInfo(ServiceBindingInfo):
+    credentials: dict
 
 
 def _find_next_service_binding_serial_number(
@@ -62,7 +68,7 @@ class SBClient:
             raise requests.HTTPError(msg)
         logger.info(f'Deleted service binding {name} ({id})')
 
-    def create_service_binding(self, instance_id: str, binding_name: str) -> tuple[str, dict]:
+    def create_service_binding(self, instance_id: str, binding_name: str) -> ServiceBindingDetailedInfo:
         url = f'{self.sm_url}/v1/service_bindings'
         data = {
             'name': binding_name,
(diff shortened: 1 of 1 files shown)


cfg_mgmt/btp_service_binding.py
```diff
@@ -98,9 +118,21 @@ def get_service_bindings(self) -> list[ServiceBindingInfo]:
...
             name = item['name']
             instance_id = item['service_instance_id']
-            bindings.append(ServiceBindingInfo(id=id, name=name, service_instance_id=instance_id))
+            created_at = item['created_at']
+            updated_at = item['updated_at']
+            bindings.append(ServiceBindingInfo(
+                id=id, name=name, service_instance_id=instance_id, created_at=created_at, updated_at=updated_at))
         return bindings
 
+    def __response_json_to_sb_detailed_info(self, resp_json) -> ServiceBindingDetailedInfo:
```
ccwienk, line 127: codestyle: "private" methods/functions should be marked w/ a single underscore. double-underscores are typically "reserved" for python-internals or -contracts 

consider using `dacite.from_dict` for deserialising `dict` -> `dataclass`

btw, using `dacite.config.Config`'s typehooks, you might even automatically deserialise timestamp-strings into `datetime.datetime` objects (if this is helpful/desirable), like so:


```
dacite.from_dict(
  data=resp_json,
  data_class=ServiceBindingDetailedInfo,
  config=dacite.config.Config(
    type_hooks={
     datetime.datetime: dateutil.parser.isoparse, # prefer dateutil.parser over python-default as the latter is buggy
   }
)
```

(for this to work, you would have to change the typehints for updated_at, created_at)

cfg_mgmt/btp_service_binding.py
```diff
@@ -98,9 +118,21 @@ def get_service_bindings(self) -> list[ServiceBindingInfo]:
...
+        id = resp_json['id']
+        name = resp_json['name']
+        instance_id = resp_json['service_instance_id']
+        credentials = resp_json["credentials"]
+        created_at = resp_json['created_at']
+        updated_at = resp_json['updated_at']
+        return ServiceBindingDetailedInfo(id=id, name=name, service_instance_id=instance_id,
```
ccwienk, line 134: codestyle: we prefer to have one argument per line

ccwienk (changes requested):
some codestyle-nits, otherwise looks fine

dimityrmirchev:
Thanks for the feedback, @ccwienk! I made the suggested changes. PTAL

ccwienk (approved):

PR #813, opened 2023-02-16
Factor out vulnerability rescoring

zkdev:
Prepare reusability for automated rescoring.

Fixes #

**Release note**:

```other operator

```


ccwienk (changes requested):
actually, I would prefer to not add such a special-case function. this is so much boilerplate + yet another dependency just for re-using a for-loop.

zkdev:
close for now, reconsider once additional functionalities are required

PR #818, opened 2023-02-24
protecode-cfg lookup

zkdev:

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/model/protecode.py
+++ b/model/protecode.py
@@ -58,6 +58,9 @@ class ProtecodeConfig(NamedModelElement):
     def tls_verify(self) -> bool:
         return self.raw.get('tls_verify', True)
 
+    def group_ids(self) -> list[int]:
+        return self.raw.get('group_ids', [])
+
     def _required_attributes(self):
         yield from super()._required_attributes()
         yield from (
--- a/model/protecode.py
+++ b/model/protecode.py
@@ -15,6 +15,7 @@
 import enum
 import typing
 
+import ci.util
 from model.base import (
     BasicCredentials,
     NamedModelElement,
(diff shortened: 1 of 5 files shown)


ccwienk (changes requested):
as discussed, please implement this similar to model/github.py | model/container_registry.py

place actual lookup-logic into model/protecode.py; add or extend lookup-function in ccc.protecode.py

please follow the approach we took, e.g. for ccc.github:github_api, where:

- there is a single lookup-function with some convenience for different parameter-sets
- e.g. it should be possible to pass-in cfg by name, by cfg-element, or not at all
- instead of just hostname, accept a URL; parse hostname from it
- do not require group-id
- depending on parameters, lookup all matching configs, sort them by how well they match given parameters
- return best-match

as an example, of requesting a bdba-cfg for a given base-url + group-id, there might be three bdba-cfgs:
1. has different api-url, but matching group-id
2. has correct api-url, but no group-ids at all
3. has corect api-url, and matching group-id

in this example, the first cfg would be removed from candidates (because it has a specific cfg that conflicts w/ passed-in parameters), whereas the second and third cfgs would be considered candidates.

the second one has less specific cfg (w.r.t. group-ids) than the third one, so it will get a smaller priority, so consequently, the third cfg will "win"

model/protecode.py
```diff
@@ -68,3 +89,45 @@ def _required_attributes(self):
...
+        for cfg in matching_cfgs
+    ]:
+        # eliminate less specific cfgs
+        matching_cfgs = [
+            cfg
+            for cfg in matching_cfgs
+            if cfg.hostname()
+        ]
+
+    matching_cfgs = sorted(matching_cfgs, key=lambda cfg: len(cfg.group_ids()))
```
ccwienk, line 130: why would you sort by amount of configured group-IDs? if at all, less group-ids would be more specific (however, it does not make a lot of sense to sort by smaller number of group-ids as well).

a metric I would consider for measuring how (relatively) specific a given bdba-cfg is would be:
- does it, or does not not define a base-url (if not, consider it less specific than a bdba-cfg that does)
- does it, or does it not define group-ids (if not, consider it less specific than a bdba-cfg that does)

a lookup-semantics working on this could look like this:
- discard cfg-elements that do have an explicit cfg, but have cfg-values that do not match the desired values (examples: 1. the base-url does not match the passed value for base-url, 2. the group-ids do not match the passed-in value(s) for group-id)
- out of the remaining cfg-elements (i.e. those that did not formally disqualify), order cfg-elements by how specific they are
  - since each cfg will have a base-url, this attribute might actually be ignored (if base-url would not match, the cfg-element would have been discarded at this point; if the cfg-element does not have a base-url, it is invalid (note: please referain from actually adding such a check here)
  - which leaves the group-id. check whether there is a cfg-element that does declare group-ids. take e.g. the first one that has. fallback: take a bdba-cfg that does not declare group-ids

ccwienk (changes requested):

model/protecode.py
```diff
@@ -71,3 +89,43 @@ def _required_attributes(self):
...
+) -> ProtecodeConfig | None:
+    if not cfg_factory:
+        cfg_factory = ci.util.ctx().cfg_factory()
+
+    matching_cfgs = []
+
+    for cfg in cfg_factory._cfg_elements(cfg_type_name='protecode'):
+        cfg: ProtecodeConfig
+
+        if cfg.base_url() and not cfg.matches_base_url(base_url):
```
ccwienk, line 107: mv this into `matches_base_url`. or rather: add a single method `matches`, like so (pseudo-code):

```
def _matches(self, base_url, group_id):
 score = 0
 if base_url:
   # todo: normalise urls (in particular `https://foo:443/bar` should match `https://foo/bar`)
   if base_url.startswith(self.base_url()):
     score += 1
   else:
     return -1 # scores < 0 indicate explicit mis-match
   if group_id:
     if group_id in self.group_ids():
       score += 1
     else:
       if not self.group_ids():
          pass # no score change if config does not specify a group-id
       return -1 # group_id did not match
   return score
```

ccwienk (changes requested):

ccwienk (approved):

PR #821, opened 2023-02-27
Allow skipping of malware scans

AndreasBurger:
Adds support to skip some malware scans. Currently, skipping behaviour is controlled by two parameters:
- age threshold: The scan will be skipped unless the previous scans findings were based on a virus definition database older than the current by a given amount of days
- virus definition version: The scan will be skipped if there has been no change to the virus definition database at all (as determined by the signature version)
    
By default only scans with unchanged virus definition database are avoided. This should result in daily-ish scans

### DIFF
--- a/clamav/cnudie.py
+++ b/clamav/cnudie.py
@@ -12,7 +12,11 @@ import ci.util
 import clamav.client
 import clamav.model
 import clamav.scan
+import clamav.util
+import github.compliance.model as gcm
 import cnudie.iter
+import delivery.client
+import delivery.model
 import dso.model
 import github.compliance.model
 import oci.client
@@ -21,21 +25,86 @@ logger = logging.getLogger(__name__)
 ci.log.configure_default_logging()
 
 
+def _may_reuse_existing_scan_result(
+        resource_node: cnudie.iter.ResourceNode,
+        virus_db_max_age_days: int,
+        delivery_client: delivery.client.DeliveryServiceClient,
+        current_version_info: clamav.model.ClamAVVersionInfo = None,
+) -> bool:
+
+    previous_findings = delivery_client.artefact_metadata_for_resource_node(
+        resource_node=resource_node,
+        types=[dso.model.Datatype.MALWARE],
+    )
+
+    if not (latest_finding := max(
+        previous_findings,
+        default=None,
+        key=lambda f: f.meta.creation_date.astimezone(),
+    )):
+        return False
+
+    if isinstance(latest_finding.data, dict):
+        # In this case the previous scan we have available predates the current dataclasses.
+        # Returning False here results in a rescan which will put a new scan result into
+        # the database if there are any findings.
+        # Also, as an additional safeguard against future errors, we check that no such
+        # supposedly structurally outdated finding was put into the database after this
+        # change was added.
+        cutoff_date = datetime.datetime(year=2023, month=4, day=17)
+        if latest_finding.meta.creation_date > cutoff_date:
+            raise RuntimeError(
+                'Encountered a malware scan result with outdated schema that was created '
+                'after introduction of the new schema.'
+            )
+        return False
+
+    finding_clamav_metadata = latest_finding.data.metadata
+
+    if finding_clamav_metadata.signature_version == current_version_info.signature_version:
+        # No changes to virus signature database version. Safe to skip, as scan would not
+        # discover anything new.
+        return True
+
+    if (
+        current_version_info.signature_date
+        - finding_clamav_metadata.virus_definition_timestamp
+    ).days < virus_db_max_age_days:
+        return True
+
+    return False
+
+
 def scan_resources(
     resource_nodes: typing.Iterable[cnudie.iter.ResourceNode],
+    clamav_version_info: clamav.model.ClamAVVersionInfo,
+    virus_db_max_age_days: int,
     oci_client: oci.client.Client,
     clamav_client: clamav.client.ClamAVClient,
-    clamav_version_info: clamav.model.ClamAVVersionInfo,
+    delivery_client: delivery.client.DeliveryServiceClient = None,
     s3_client=None,
     max_workers:int = 16,
 ) -> typing.Generator[clamav.model.ClamAVResourceScanResult, None, None]:
-    executor = concurrent.futures.ThreadPoolExecutor(max_workers=max_workers)
 
     def scan_resource(
         resource_node: cnudie.iter.ResourceNode,
-    ):
+    ) -> clamav.model.ClamAVResourceScanResult:
         component = resource_node.component
         resource = resource_node.resource
+
+        if delivery_client and _may_reuse_existing_scan_result(
+            resource_node=resource_node,
+            delivery_client=delivery_client,
+            current_version_info=clamav_version_info,
+            virus_db_max_age_days=virus_db_max_age_days,
+        ):
+            resource_name = f'{component.name}/{resource.name}'
+            logger.info(f'Skipping scan of {resource_name}.')
+            return clamav.model.ClamAVResourceScanResult(
+                scan_result=None,
+                state=gcm.ScanState.SKIPPED,
+                scanned_element=resource_node,
+            )
 
         if isinstance(resource.access, cm.OciAccess):
             access: cm.OciAccess = resource.access
@@ -45,6 +114,7 @@ def scan_resources(
                 oci_client=oci_client,
                 clamav_client=clamav_client,
             )
+
         elif isinstance(resource.access, cm.S3Access):
             access: cm.S3Access = resource.access
             if isinstance(resource.type, enum.Enum):
@@ -65,6 +135,7 @@ def scan_resources(
                 clamav_client=clamav_client,
                 tf=tf,
             )
+
         else:
             raise NotImplementedError(type(resource.access))
 
@@ -78,18 +149,22 @@ def scan_resources(
         # pylint: disable=E1123
         return clamav.model.ClamAVResourceScanResult(
             scan_result=scan_result,
-            scanned_element=cnudie.iter.ResourceNode(
-                path=(component,),
-                resource=resource,
-            ),
+            scanned_element=resource_node,
         )
 
+    executor = concurrent.futures.ThreadPoolExecutor(max_workers=max_workers)
     tasks = [
         executor.submit(scan_resource, resource_node)
        for resource_node in resource_nodes
     ]
 
     logger.info(f'will scan {len(tasks)=} with {max_workers=}')
+
+    if not delivery_client:
+        logger.info(
+            'Unable to compare with previous findings in delivery-service, no client available. '
+            'All files will be scanned.'
+        )
 
     while True:
         done, not_done = concurrent.futures.wait(
@@ -101,6 +176,25 @@ def scan_resources(
         if not not_done:
             break
         tasks = not_done
+
+
+def _scan_result_to_malware_finding(
+        scan_result: clamav.model.ScanResult,
+) -> dso.model.MalwareFinding:
+    if (meta := scan_result.meta):
+        meta = dso.model.MalwareFindingMeta(
+            receive_duration_seconds=meta.receive_duration_seconds,
+            scanned_octets=meta.scanned_octets,
+            scanned_content_digest=meta.scanned_content_digest,
+            scan_duration_seconds=meta.scan_duration_seconds,
+        )
+    return dso.model.MalwareFinding(
+        status=scan_result.status.value,
+        details=scan_result.details,
+        meta=meta,
+        name=scan_result.name,
+        malware_status=scan_result.malware_status.name,
+    )
 
 
 def resource_scan_result_to_artefact_metadata(
@@ -121,24 +215,6 @@ def resource_scan_result_to_artefact_metadata(
         type=datatype,
         creation_date=creation_date,
     )
-
-    def _scan_result_to_malware_finding(
-            scan_result: clamav.model.ScanResult,
-    ) -> dso.model.MalwareFinding:
-        if (meta := scan_result.meta):
-            meta = dso.model.MalwareFindingMeta(
-                receive_duration_seconds=meta.receive_duration_seconds,
-                scanned_octets=meta.scanned_octets,
-                scanned_content_digest=meta.scanned_content_digest,
-                scan_duration_seconds=meta.scan_duration_seconds,
-            )
-        return dso.model.MalwareFinding(
-            status=scan_result.status.value,
-            details=scan_result.details,
-            meta=meta,
-            name=scan_result.name,
-            malware_status=scan_result.malware_status.name,
-        )
 
     aggregated_scan_result = resource_scan_result.scan_result
     clamav_version_info = aggregated_scan_result.clamav_version_info
--- a/concourse/model/traits/image_scan.py
+++ b/concourse/model/traits/image_scan.py
@@ -188,6 +188,15 @@ CLAMAV_ATTRS = (
         default=8,
     ),
     AttributeSpec.optional(
+        name='virus_db_max_age_days',
+        doc=(
+            'The maxmimum age difference (in days) the virus definition database of ClamAV may '
+            'have when being compared to the most current database before triggering rescans.'
+        ),
+        type=int,
+        default=0,
+    ),
+    AttributeSpec.optional(
         name='saf_config_name',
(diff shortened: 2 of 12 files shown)


delivery/model.py
```diff
@@ -68,3 +70,43 @@ def from_dict(raw: dict):
                 type_hooks={datetime.date | None: _parse_date_if_present},
             ),
         )
+
+
+@dataclasses.dataclass(frozen=True)
+class ComponentArtefactId:
```
ccwienk, line 76: ideally, there should also be a util for creating this from cnudie.iter.ResourceNode | SourceName, wdyt?

clamav/cnudie.py
```diff
@@ -14,29 +14,122 @@
...
+import delivery.model
+import delivery.client
 import github.compliance.model
 import oci.client
 
 logger = logging.getLogger(__name__)
 ci.log.configure_default_logging()
 
 
+def _finding_for_component_artifact(
```
ccwienk, line 26: I have multiple concerns here:

- this function mixes two aspects: 
  - filter elements
  - return first element iff filtering result is not empty, or None
- this function is actually more generic (filters/finds artefact-metadata) than its location suggests (clamav/cnudie is specific to clamav)

Also, I think we should not hardcode the special-case of only considering the (randomly determined) first element (there could be multiple metadata-entries (in case of clamav scans, this could be very relevant, because newer results might lead to decision to skip, whereas older entries might not)).

--

Also, I would prefer to make use of `cnudie.iter`'s ResourceNode, ArtefactNode, ComponentNode for new code. As I commented below, it might be a good idea to add some convenience functions, e.g. to convert to and from cnudie.iter, dso.model, dm.model, name-and-version-strings (component_name, artefact_name, ....)

I think what you should factor out is a function that matches a cnudie.iter.XX-Node against dm.ArtefactMetadata

the specialcase-function in this module should instrument such matchers, to do special-case-filtering (if necessary), specific to clamav-metadata (e.g. based on timestamp, ..)

clamav/cnudie.py
```diff
@@ -14,29 +14,122 @@
...
+    previous_findings: list[delivery.model.ArtefactMetadata],
```
ccwienk, line 27: that being said, I do not like the pattern of "return first element or return None" like that.

consider any of:
- itertools (did not check whether they have sth.)
- for-loop + return
- next + try-except (StopIteration)

reasons:
- KISS
- readability
- your pattern does not actually save a lot of lines of code (not that this actually would be a big concern)

ccwienk (changes requested):

delivery/model.py
AndreasBurger: Sure, it _could_ be done, but currently `ComponentArtefactId`  is only instantiated when retrieving data from the delivery-service. Could be useful for comparison between local and retrieved objects, but otherwise I don't see it

delivery/client.py
```diff
@@ -117,6 +124,37 @@ def upload_metadata(
...
+    def component_metadata(
+        self,
+        name: str,
+        type: str = None,
+        select: str = None,
+        version: str = None,
+    ) -> list[dso.model.ArtefactMetadata]:
+        url = self._routes.component_metadata()
+
+        # No validation done here. It is left to the caller to ensure that all parameters are
```
ccwienk, line 136: it would be nice to add a docstr explaining the parameters, or at least referencing where such a documentation can be found (esp. `select` is probably not obvious)

clamav/cnudie.py
```diff
@@ -23,20 +28,105 @@
...
     s3_client=None,
     max_workers:int = 16,
 ) -> typing.Generator[clamav.model.ClamAVResourceScanResult, None, None]:
-    executor = concurrent.futures.ThreadPoolExecutor(max_workers=max_workers)
+
+    def _skip_scan(
+            previous_finding: delivery.model.ArtefactMetadata,
+            current_version_info: clamav.model.ClamAVVersionInfo,
+    ) -> bool:
+        # helper function to determine whether a scan may be skipped
```
ccwienk, line 45: if you rename the function, e.g. to `_should_skip_scan`, its semantics would be obvious w/o the comment

clamav/cnudie.py
```diff
@@ -23,20 +28,105 @@
...
+
+        if isinstance(previous_finding.data, dict):
```
ccwienk, line 47: preferably migrate existing data, please

also, I think to remember that all of our artefact-metadata should have a timestamp?

clamav/cnudie.py
```diff
@@ -23,20 +28,105 @@
 
 def scan_resources(
     resource_nodes: typing.Iterable[cnudie.iter.ResourceNode],
+    clamav_version_info: clamav.model.ClamAVVersionInfo,
+    virus_definition_age_threshold: int,
```
ccwienk, line 32: if passing-in maximum allowed age in days as an integer (which is okay), please suffix it w/ `_days` to "document" this (as an alternative, we might use datetime.timedelta, which would be somewhat more flexible)

instead of `age_threshold`, which not simply call this `virus_db_max_age_days`?

clamav/cnudie.py
```diff
@@ -23,20 +28,105 @@
...
+
+        if isinstance(previous_finding.data, dict):
+            # In this case the previous scan we have available predates the current dataclasses.
+            # It is safe to assume that this scan result is very old.
+            return False
+
+        clamav_scan_metadata = previous_finding.data.metadata
+        if (
+            skip_if_virus_definition_unchanged
+            and clamav_scan_metadata.signature_version == current_version_info.signature_version
```
ccwienk, line 55: why check both version and timestamp? checking for timestamp should be enough, should it not?

clamav/cnudie.py
```diff
@@ -23,20 +28,105 @@
...
+        sorted_findings = sorted(
+            previous_findings,
+            key=lambda f: f.meta.creation_date.astimezone(),
+            reverse=True,
+        )
+
+        if sorted_findings:
+            # only look at the most recent data
+            finding = sorted_findings[0]
+            if _skip_scan(
```
ccwienk, line 88: this diff mixes parts of the `should-skip`-semantics w/ the regular scanning. please refactor like so:

```
def scan_resource(...):
  if _should_skip(...):
    # early-exit
  # scan
```

clamav/cnudie.py
```diff
@@ -23,20 +28,105 @@
...
+                previous_finding=finding,
+                current_version_info=clamav_version_info,
+            ):
+                resource_name = f'{component.name}/{resource.name}'
```
ccwienk, line 92: consider adding either `__str__` to cnudie.iter.Node, or a util function for that (honouring the component-path); not necessarily for this pull-request, though.

clamav/cnudie.py
```diff
@@ -23,20 +28,105 @@
...
+                logger.info(f'Skipping scan of {resource_name}.')
+                summary: dso.model.MalwareSummary = finding.data
+                if isinstance(resource.access, cm.OciAccess):
+                    resource_url = resource.access.imageReference
```
ccwienk, line 96: access-handling should be factored out
at the very least, raise for unknown resources, instead of simply ignoring/warning

clamav/cnudie.py
```diff
@@ -23,20 +28,105 @@
...
+                elif isinstance(resource.access, cm.S3Access):
+                    a = resource.access
+                    resource_url = f's3://{a.bucketName}/{a.objectKey}'
```
ccwienk, line 99: should rather be added to component-spec / S3Access

clamav/cnudie.py
```diff
@@ -23,20 +28,105 @@
...
+                else:
+                    resource_url = '<unknown>'
+
+                findings = [_malware_finding_to_scan_result(f) for f in summary.findings]
+
+                # Return a ScanResult that indicates that no new scan was performed while preserving
+                # the old data for the issue-generation.
+                return clamav.model.ClamAVResourceScanResult(
+                    scan_result=clamav.model.AggregatedScanResult(
```
ccwienk, line 108: either omit scan-result, or replicate actual data, instead of adding dummy-data.

concourse/model/traits/image_scan.py
```diff
@@ -161,6 +161,24 @@ class ClamAVRescoringEntry:
...
+        name='virus_definition_age_threshold',
+        doc=(
+            'The maxmimum age difference (in days) the virus definition database of ClamAV may '
+            'have when being compared to the most current database before triggering rescans.'
+        ),
+        type=int,
+        default=0,
+    ),
+    AttributeSpec.optional(
+        name='skip_if_virus_definition_unchanged',
```
ccwienk, line 174: I think there is not need to make this configurable.

if the data to scan did not change (e.g. determined through hash digest), and neither did the virus-db, it is always pointless to rescan (thus I think we should simply hard-code this case).

You however do have a valid point in having identified this case. Unchanged db-version should overrule exceeded max. allowed age

delivery/util.py
```diff
@@ -74,3 +78,34 @@ def update_available(
...
+    For possible values for `type` see `dso.model.Datatype`.
+    '''
+    if not delivery_client:
+        if not (delivery_client := ccc.delivery.default_client_if_available()):
+            raise ValueError('No default delivery client available.')
+
+    component = resource_node.component
+    resource = resource_node.resource
+
+    yield from filter(
```
ccwienk, line 100: use generator-comprehension, or - maybe even better - just a plain for-loop
map/filter is considered to be less readable than comprehensions

KISS / readability counts

github/compliance/model.py
```diff
@@ -67,10 +67,12 @@ class ScanState(enum.Enum):
     '''
     indicates the scan outcome of a scan (regardless of yielded contents).
 
-    SUCCEEDED: scan succeeded without errors (but potentially with findings)
-    FAILED:    scan failed (which typically implies there are not scan results)
+    SUCCEEDED:  scan succeeded without errors (but potentially with findings)
+    SKIPPED:    No scan took place (usually due to unchanged scanning environment)
```
ccwienk, line 71: the hint (in parentheses) is misleading. we will skip scans not because of "unchanged scanning env", but because we found an earlier scan that has not yet been invalidated

github/compliance/model.py
```diff
@@ -79,13 +81,17 @@ class ScanState(enum.Enum):
 
 @dataclasses.dataclass(kw_only=True)
 class ScanResult:
-    scanned_element: Target
+    scanned_element: (
+        cnudie.iter.ResourceNode
```
ccwienk, line 85: why? we defined the type-alias `Target` precisely to abstract from the concrete types; which allows us to also extend the allowed types later. by resolving it here, this adds one piece of code that will not be found e.g. by searching for this type-alias.

ccwienk (changes requested):

clamav/cnudie.py
```diff
@@ -85,28 +86,20 @@ def scan_resource(
...
-                    resource_url = f's3://{a.bucketName}/{a.objectKey}'
-                else:
-                    resource_url = '<unknown>'
-
                 findings = [_malware_finding_to_scan_result(f) for f in summary.findings]
 
                 # Return a ScanResult that indicates that no new scan was performed while preserving
                 # the old data for the issue-generation.
                 return clamav.model.ClamAVResourceScanResult(
                     scan_result=clamav.model.AggregatedScanResult(
```
AndreasBurger, line 101: omit AggregatedScanResult

delivery/util.py
```diff
@@ -97,15 +97,16 @@ def artefact_metadata_for_resource_node(
     component = resource_node.component
     resource = resource_node.resource
 
-    yield from filter(
```
AndreasBurger: consider moving to delivery-client

clamav/cnudie.py
```diff
@@ -21,22 +25,103 @@
...
 
+def _should_skip_scan(
+        resource_node: cnudie.iter.ResourceNode,
+        current_version_info: clamav.model.ClamAVVersionInfo,
+        virus_db_max_age_days: int,
+        skip_if_virus_definition_unchanged: bool,
+        delivery_client: delivery.client.DeliveryServiceClient = None,
+) -> bool:
+
+    if not delivery_client:
```
ccwienk, line 36: rather mv this check to caller, do not allow passing-in non for delivery-service-client

clamav/cnudie.py
```diff
@@ -21,22 +25,103 @@
...
+        return False
+
+    previous_findings = delivery_client.artefact_metadata_for_resource_node(
+        resource_node=resource_node,
+        types=[dso.model.Datatype.MALWARE],
+    )
+
+    sorted_findings = sorted(
```
ccwienk, line 44: nit: rather use `max` or `min`?

clamav/cnudie.py
```diff
@@ -21,22 +25,103 @@
...
+        previous_findings,
+        key=lambda f: f.meta.creation_date.astimezone(),
+        reverse=True,
+    )
+
+    if sorted_findings:
```
ccwienk, line 50: codestyle: rather place early-exit first, add non-early-exit path outside of if-block

clamav/cnudie.py
```diff
@@ -21,22 +25,103 @@
...
+        return False
+
+    if isinstance(latest_finding.data, dict):
+        # In this case the previous scan we have available predates the current dataclasses.
+        # Returning False here results in a rescan which will put a new scan result into
+        # the database if there are any findings.
+        # Also, as an additional safeguard against future errors, we check that no such
+        # supposedly structurally outdated finding was put into the database after this
+        # change was added.
+        cutoff_date = datetime.datetime(year=2023, month=4, day=1)
```
ccwienk, line 62: bump cutoff-date

clamav/cnudie.py
```diff
@@ -21,22 +25,103 @@
...
+            raise RuntimeError(
+                'Encountered a malware scan result with outdated schema that was created '
+                'after introduction of the new schema.'
+            )
+        return False
+
+    finding_clamav_metadata = latest_finding.data.metadata
+
+    if (
+        skip_if_virus_definition_unchanged
```
ccwienk, line 73: maybe we already discussed this: _why_ would we want to rescan upon unchanged virus definitions?

clamav/cnudie.py
```diff
@@ -21,22 +25,103 @@
...
-    executor = concurrent.futures.ThreadPoolExecutor(max_workers=max_workers)
 
     def scan_resource(
         resource_node: cnudie.iter.ResourceNode,
-    ):
+    ) -> clamav.model.ClamAVResourceScanResult:
         component = resource_node.component
         resource = resource_node.resource
 
+        if _should_skip_scan(
```
ccwienk, line 110: I think this should be split + renamed, still, like so:

`_may_reuse_existing_scan_result`: check whether we are "allowed" to reuse existing scan-result (we might have been called w/ a "force-scan" flag. however: why do we even want to not try to reuse existing scans?

```scan_result = _find_scan_result(
 max_age, # hardcode to ignore if version matches
 version, # make version optional
 <ocm-coordinates>, 
 ..
)
```

clamav/cnudie.py
```diff
@@ -78,19 +165,23 @@ def scan_resource(
...
     tasks = [
         executor.submit(scan_resource, resource_node)
        for resource_node in resource_nodes
     ]
 
     logger.info(f'will scan {len(tasks)=} with {max_workers=}')
 
+    if not delivery_client:
+        logger.warning(
+            'Unable to compare with previous findings in delivery-service, no client available. '
```
ccwienk, line 181: nit: I think this should be at most an info-log-entry

clamav/cnudie.py
```diff
@@ -103,6 +194,44 @@ def scan_resource(
         tasks = not_done
 
 
+def _malware_finding_to_scan_result(
+        finding: dso.model.MalwareFinding,
+) -> clamav.model.ScanResult:
+    if (meta := finding.meta):
```
ccwienk, line 200: did you miss a `not` here?

clamav/cnudie.py
```diff
@@ -103,6 +194,44 @@ def scan_resource(
...
+        meta=meta,
+        name=finding.name,
+        malware_status=clamav.model.MalwareStatus(finding.malware_status),
+    )
+
+
+def _scan_result_to_malware_finding(
+        scan_result: clamav.model.ScanResult,
+) -> dso.model.MalwareFinding:
+    if (meta := scan_result.meta):
```
ccwienk, line 219: ditto

PR #827, opened 2023-03-14
feat: new release note fetcher

darmiel:

This pull request adds the core of a new release note generator supporting release tags which are not on the default branch, thus allowing for code releases without any freezing period (the old release note generator only allowed releases on the same branch). With the new release notes generator generator, new (minor-) releases should take place on separate branches.

Example usage:

```python
import yaml, gci.componentmodel
with open("component-descriptor.yaml", "r") as fd:
    c = gci.componentmodel.ComponentDescriptor.from_dict(yaml.safe_load(fd))

notes: set[models.ReleaseNote] = fetch_release_notes(c.component, "gardener")
print('\n\n'.join(z.to_block_str() for z in notes))
```

**Release note**:

```other operator
NONE
```


gardener-robot-ci-3:
Thank you @darmiel for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

ccwienk:
@darmiel : please fix linter-issues

release_notes/fetch.py
```diff
@@ -0,0 +1,217 @@
...
+        self.github_repo: github3.repos.Repository = self.github_helper.github.repository(
+            self.github_helper.owner,
+            self.github_helper.repository_name
+        )
+
+    def _lists_for_new_initial_release(self) -> tuple[tuple[git.Commit], tuple[git.Commit]]:
+        """ returns commits for new initial-release (all commits in repository starting from the tag)
+        """
+        # just return all commits starting from the current_version_tag
+        return tuple(self.git_helper.repo.iter_commits(self.current_version_tag)), tuple()
```
ccwienk, line 90: that's awful - why?

the method's name is too generic, and misleading:

firstly, the method wil return a two-tuple containing two tuples (which contradicts the method's name)
secondly, what does that mean "lists for new initial release" / why do we need that?
thirdly, this is just a small wrapper around git_helper, w/ the special-case of passing-in "current_version_tag". why? this makes the code convoluted, and hard to understand

release_notes/fetch.py
```diff
@@ -0,0 +1,217 @@
...
+
+    def _list_for_new_patch_release(self) -> tuple[tuple[git.Commit], tuple[git.Commit]]:
+        """ return commits for new patch-release
+        """
+        logger.info(f"creating new patch release from {self.current_version_tag} to {self.previous_version_tag}")
+        if self.previous_version_tag is None:
+            raise RuntimeError("cannot create patch-release notes because previous version cannot be found")
+        return _iter_tags(self.git_helper.repo, self.current_version_tag, self.previous_version_tag), tuple()
```
ccwienk, line 98: ditto (see comments for other "list-method"

release_notes/fetch.py
```diff
@@ -0,0 +1,217 @@
...
+
+    def _list_for_new_minor_release(self) -> tuple[tuple[git.Commit], tuple[git.Commit]]:
```
ccwienk, line 100: ditto

release_notes/fetch.py
```diff
@@ -0,0 +1,217 @@
...
+
+    # otherwise the tags should be linear
+    return tuple(repo.iter_commits(f"{newest_tag.commit.hexsha}...{oldest_tag.commit.hexsha}"))
+
+
+class ReleaseNotesComponentAccess:
+    def __init__(self,
+                 component: gci.componentmodel.Component,
+                 repo_path: str,
+                 current_version: typing.Optional[semver.VersionInfo] = None):
```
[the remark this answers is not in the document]
ccwienk: also (I write this after having read through this class's methods) the name is completely misleading. I get it that it may be useful to have a struct to pass-around all those pieces of data (totally fine). however, I think those methods should be  turned into functions. use a module to group them, maybe. preferably pass-in the actually-used objects. this may look like some tedious overhead considering the amount of parameters, but I think this is just a couple of line of extra code, which will make the code a lot easier to understand considering its side-effects and semantics.

consider factoring out some of the special-cases of those "list"-methods and rather handle those in fewer functions (maybe just one)

release_notes/fetch.py
```diff
@@ -0,0 +1,217 @@
...
+
+        # new minor release
+        if self.current_version.minor != self.previous_version.minor:
+            return self._list_for_new_minor_release()
+
+        # new patch release
+        return self._list_for_new_patch_release()
+
+
+def create_note_blocks(release_notes: set[models.ReleaseNote]) -> str:
```
ccwienk, line 156: nit: release_note(s)_blocks - please use abbreviations sparingly

release_notes/fetch.py
```diff
@@ -0,0 +1,217 @@
...
+    return '\n\n'.join(z.to_block_str() for z in release_notes)
+
+
+def fetch_release_notes(
+        component: gci.componentmodel.Component,
+        repo_path: str,
+        for_version: typing.Optional[semver.VersionInfo] = None,
+):
```
ccwienk, line 164: please add a brief docstr. explain expecially what `for_version` means

release_notes/models.py
```diff
@@ -0,0 +1,210 @@
...
+import gci.componentmodel
+import git
+import github3.pulls
+from git import Commit
+
+import cnudie.util
+
+
+@dataclasses.dataclass
+class Author:
```
ccwienk, line 14: do we actually need to in addition to `git`'s Author/Commiter?

release_notes/models.py
```diff
@@ -0,0 +1,210 @@
...
+@dataclasses.dataclass
+class ReferenceType:
+    identifier: str  # identifier for release note block
+    prefix: str  # prefix for generated release notes
+
+
+REF_TYPE_PULL = ReferenceType(identifier="#", prefix="#")
+REF_TYPE_COMMIT = ReferenceType(identifier="$", prefix="@")
+
+REF_TYPES = [REF_TYPE_PULL, REF_TYPE_COMMIT]
```
ccwienk, line 49: nit: rather use tuples

I did not check: if not used outside this module, please use lowercase + prefix w/ underscore

release_notes/models.py
```diff
@@ -0,0 +1,210 @@
+import dataclasses
```
ccwienk, line 1: nit: in cc-utils, we thus far use singular (model.py)

release_notes/models.py
```diff
@@ -0,0 +1,210 @@
...
+        if isinstance(other, SourceBlock):
+            return hash(other) == hash(self)
+        return False
+
+
+pattern = re.compile(r"\x60{3}(?P<category>\w+)\s+(?P<target_group>\w+)\n(?P<note>.+?)\n\x60{3}",
+                     flags=re.DOTALL | re.IGNORECASE | re.MULTILINE)
+
+
+def list_source_blocks(content: str) -> typing.Generator[SourceBlock, None, None]:
```
ccwienk, line 119: codestyle: preferably prefix w/ `iter`

release_notes/utils.py
```diff
@@ -0,0 +1,52 @@
...
+    :param owner: Owner of the repository (on GitHub)
+    :param repo: Name of the repository (on GitHub)
+    :param sha: SHA of the commit
+    :return: a tuple with pull requests related to the specific commit
+    """
+    try:
+        url = gh._build_url("repos", owner, repo, "commits", sha, "pulls")
+        return tuple(gh._iter(-1, url, pulls.ShortPullRequest))
+    except exceptions.UnprocessableEntity as e:
+        if e.code != 422:  # pull request not found
```
ccwienk, line 24: when will this happen?

ccwienk (changes requested):

release_notes/utils.py
darmiel: when the requested commit was not found in the repository. I've updated the comment.

ccwienk: +1 for the updated comment. however: can this actually happen? You retrieve pull-requests immediately above 🤔

darmiel: Good question. I don't think this can actually happen in production. It's mostly a leftover from debugging at this point, where I got that error often when using dummy data 😅

release_notes/model.py
```diff
@@ -0,0 +1,228 @@
...
+        return self.source_block.__eq__(other)
+
+    def get_author_str(self) -> str:
+        return str(self.author)
+
+    def get_reference_str(self) -> str:
+        return f'{self.reference.type.identifier}{self.reference.get_identifier()}'
+
+    def to_block_str(self) -> str:
+        return '```{category} {group} {src_repo} {ref} {author}\n{message}\n```'.format(
```
ccwienk, line 165: preferably use format-string-literals: https://docs.python.org/3/reference/lexical_analysis.html#f-strings

release_notes/model.py
```diff
@@ -0,0 +1,228 @@
...
+            src_repo=self.source_component.name,
+            ref=self.get_reference_str(),
+            author=self.author.username or self.author.display_name.replace(' ', '-'),
+            message=self.source_block.note_message
+        )
+
+
+def create_release_note_obj(
+        block: SourceBlock,
+
```
ccwienk, line 177: codestyle: why leave blank lines? also, it might be nicer to call param `block` -> `source_block`, wdyt?

release_notes/model.py
```diff
@@ -0,0 +1,228 @@
...
+        source_commit: Commit,
+        raw_body: str,
+        author: typing.Optional[Author],
+
+        targets: typing.Union[git.Commit, github3.pulls.ShortPullRequest],
```
ccwienk, line 182: plural form typically implies that a sequence/iterable is expected. rather name this `target`?

release_notes/model.py
```diff
@@ -0,0 +1,228 @@
...
+            message=self.source_block.note_message
+        )
+
+
+def create_release_note_obj(
+        block: SourceBlock,
+
+        source_commit: Commit,
+        raw_body: str,
+        author: typing.Optional[Author],
```
ccwienk, line 180: if it is actually optional, you should define a default-value (probably =None)

ccwienk (changes requested):

release_notes/model.py
```diff
@@ -0,0 +1,228 @@
...
+@dataclasses.dataclass
+class ReleaseNote:
+    source_commit: Commit  # the commit where the release notes were initially gathered from
+    source_block: SourceBlock
+
+    raw_body: str  # the raw body of the commit / pull request
+    author: typing.Optional[Author]  # the author of the commit / pull request
+    reference: Reference
+
+    source_component: gci.componentmodel.Component = dataclasses.field(compare=False)
```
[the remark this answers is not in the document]
darmiel: Removed `dataclasses.field(compare=False)`. I re-used this object from the previous implementation, but since `__eq__` is overwritten anyways this won't make any difference :)

release_notes/model.py
darmiel: I think using format instead of f-strings helps with readability, wdyt?

```py
    def to_block_str(self) -> str:
        return '```{category} {group} {src_repo} {ref} {author}\n{message}\n```'.format(
            category=self.source_block.category,
            group=self.source_block.target_group,
            src_repo=self.source_component.name,
            ref=self.get_reference_str(),
            author=self.author.username or self.author.display_name.replace(' ', '-'),
            message=self.source_block.note_message
        )
```

vs.

```py
    def to_block_str(self) -> str:
        return f'```{self.source_block.category} {self.source_block.target_group} {self.source_component.name} ' \
               f'{self.get_reference_str()} {self.author.username or self.author.display_name.replace(" ", "-")}\n' \
               f'{self.source_block.note_message}\n```'
```

or maybe a mix of both?

```py
    def to_block_str(self) -> str:
        category = self.source_block.category
        group = self.source_block.target_group
        src_repo = self.source_component.name
        ref = self.get_reference_str()
        author = self.author.username or self.author.display_name.replace(' ', '-')
        message = self.source_block.note_message

        return f'```{category} {group} {src_repo} {ref} {author}\n{message}\n```'
```

release_notes/fetch.py
```diff
@@ -0,0 +1,227 @@
...
+
+    # find all available versions
+    component_versions: dict[semver.VersionInfo, str] = {}
+    for ver in cnudie.retrieve.component_versions(component.name, component.current_repository_ctx()):
+        parsed_version = version.parse_to_semver(ver)
+        if parsed_version.prerelease:  # ignore pre-releases
+            continue
+        component_versions[parsed_version] = ver
+
+    if current_version is None:
```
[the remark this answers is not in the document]
zkdev: noticed multiple times that you explicitly check for `== None`, just wanted to raise awareness :-)

zkdev (changes requested):
nits and codestyles :-)

release_notes/utils.py
```diff
@@ -0,0 +1,203 @@
...
+
+
+def _find_git_notes_for_commit(repo: git.Repo, commit: git.Commit) -> Optional[str]:
+    try:
+        return repo.git.notes('show', commit.hexsha)
+    except git.exc.GitCommandError:
+        return None
+
+
+def _find_payload_from_git_notes(repo: git.Repo, commit: git.Commit) -> Optional[dict]:
```
zkdev, line 86: is this function actually used/referenced?

release_notes/fetch.py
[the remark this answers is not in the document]
darmiel: I wasn't sure if `semver.VersionInfo` overwrites `__bool__`, but it looks like it doesn't, so I've changed them :)

release_notes/utils.py
darmiel: nope, it was a leftover - I've removed it.

zkdev (approved):
/lgtm

PR #838, opened 2023-03-31
Add support for OciImageIndex

AndreasBurger:
Adds support for [OCI Image Index](https://github.com/opencontainers/image-spec/blob/main/image-index.md) manifests to our oci package

We will soon have one image as part of our BoM that uses this kind of manifest, so we'll need to add support.
This new kind of manifest is very similar to image list manifests (see https://github.com/opencontainers/image-spec/blob/main/media-types.md#compatibility-matrix for a discription of the differences), so for this initial PR the new manifest is just used with the existing code.

### DIFF
--- a/oci/model.py
+++ b/oci/model.py
@@ -192,7 +192,7 @@ class OciBlobRef:
     digest: str
     mediaType: str
     size: int
-    annotations: typing.Optional[typing.Dict] = None
+    annotations: dict | None = None
 
     def as_dict(self) -> dict:
         raw = dataclasses.asdict(self)
@@ -287,7 +287,10 @@ class OciPlatform:
 
 @dataclasses.dataclass(frozen=True)
 class OciImageManifestListEntry(OciBlobRef):
-    platform: typing.Optional[OciPlatform] = None
+    artifactType: str | None = None
+    data: str | None = None
+    platform: OciPlatform | None = None
+    urls: list[str] | None = None
 
     def as_dict(self) -> dict:
         raw = OciBlobRef.as_dict(self)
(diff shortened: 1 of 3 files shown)


oci/model.py
```diff
@@ -310,3 +310,66 @@ def as_dict(self):
             'mediaType': self.mediaType,
             'schemaVersion': self.schemaVersion,
         }
+
+
+@dataclasses.dataclass
+class OciContentDescriptor:
```
ccwienk, line 316: I think this should be merged w/ `OciImageManifestListEntry`. Considering there are many optional attributes in both cases, likely, one single class should do.

oci/model.py
```diff
@@ -310,3 +310,66 @@ def as_dict(self):
...
+
+        return raw
+
+
+@dataclasses.dataclass
+class OciImageIndexDescriptor(OciContentDescriptor):
+    '''An OciContentDescriptor with an additional optional `platform` attribute as specified in
+    https://github.com/opencontainers/image-spec/blob/main/image-index.md#image-index-property-descriptions
+    '''
+    platform: OciPlatform | None
```
ccwienk, line 345: either merge w/ OciContentDescriptor, or even w/ OciImageManifestListEntry

oci/model.py
```diff
@@ -310,3 +310,66 @@ def as_dict(self):
...
+        raw = super().as_dict()
+        if self.platform is not None:
+            raw['platform'] = self.platform.as_dict()
+        else:
+            raw.pop('platform')
+
+        return raw
+
+@dataclasses.dataclass
+class OciImageIndex:
```
ccwienk, line 357: can we merge this w/ OciImageManifestList?

PR #850, opened 2023-04-24
Do not rescore further than `LOW`

zkdev:
As discussed, do not rescore findings further than `LOW`.

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/dso/cvss.py
+++ b/dso/cvss.py
@@ -18,8 +18,12 @@ class CVESeverity(enum.IntEnum):
     HIGH = 3
     CRITICAL = 4
 
-    def reduce(self, severity_classes=1) -> 'CVESeverity':
-        return CVESeverity(max(0, self.value - severity_classes))
+    def reduce(
+        self,
+        severity_classes=1,
+        minimum_severity: int=LOW,
+    ) -> 'CVESeverity':
+        return CVESeverity(max(minimum_severity, self.value - severity_classes))
 
     @staticmethod
     def from_cve_score(score: float):


dso/cvss.py
```diff
@@ -19,7 +19,8 @@ class CVESeverity(enum.IntEnum):
     CRITICAL = 4
 
     def reduce(self, severity_classes=1) -> 'CVESeverity':
-        return CVESeverity(max(0, self.value - severity_classes))
+        # do not reduce further than 'LOW'
+        return CVESeverity(max(1, self.value - severity_classes))
```
AndreasBurger, line 23:
```python3
return max(CVESeverity.LOW, CVESeverity(self.value - severity_classes))
```
no need for the comment then in my opinion. (Cannot create a suggestion 🤔)

AndreasBurger (approved):

dso/cvss.py
AndreasBurger: Well, turns out this will cause an exception if values larger than the current severity are subtracted. Probably best to leave the way you initially suggested.

ccwienk: I agree w.r.t. the comment. either remove it, OR (better IMO) explain _why_ we do not want to reduce to lower than LOW.

this might also be expressed (w/o comment) similarly as you did it in delivery-svc by defining an additional attribute `minimum_severity`, and assigning a default value to it (LOW..)

I think this approach would be nicer, because this way, the rule would not be hardcoded, but instead a caller might customise the behaviour

ccwienk (changes requested):

ccwienk (approved):

PR #851, opened 2023-04-25
Make using new release-code possible via pipeline-def (opt-in)

AndreasBurger:


concourse/model/traits/release.py
```diff
@@ -86,6 +86,21 @@ class ReleaseCommitPublishingPolicy(EnumWithDocumentation):
...
+class ReleaseNotesCreation(EnumWithDocumentation):
+    DEFAULT = EnumValueWithDocumentation(
+        value='default',
+        doc='Use default version of release note creation code.',
+    )
+
+    PREVIEW = EnumValueWithDocumentation(
+        value='preview',
+        doc=(
+            'Use preview version of release note creation code. Suppots release note creation even '
```
zkdev, line 98: `Suppots` 🙂

concourse/steps/release.py
```diff
@@ -911,6 +915,7 @@ def release_and_prepare_next_dev_cycle(
     release_commit_publishing_policy = ReleaseCommitPublishingPolicy(
         release_commit_publishing_policy
     )
+    release_notes_creation = ReleaseNotesCreation(release_notes_creation)
```
zkdev, line 918: should be index-syntax rather than call-syntax, no?

concourse/steps/release.py
```diff
@@ -883,6 +886,7 @@ def release_and_prepare_next_dev_cycle(
     release_commit_message_prefix: str=None,
     slack_channel_configs: list=[],
     version_operation: str='bump_minor',
+    release_notes_creation: str='default',
```
zkdev, line 889: musn't this match enum field-name (thus uppercase)?

release_notes/markdown.py
```diff
@@ -0,0 +1,149 @@
...
+            and (other == self or hash(other) == hash(self))
+        )
+
+
+def _simple_title(display: str) -> Title:
+    return Title(display, [display.lower()])
+
+
+categories: list[Title] = [
+    Title("⚠️ Breaking Changes", ["action", "breaking"]),
```
zkdev, line 29: use single-quotes

zkdev (changes requested):

concourse/steps/release.py
AndreasBurger: No, as with vanilla-Enums these `EnumWithDocumentation` can also be instantiated with the value.

concourse/steps/release.py
AndreasBurger: We currently render the string-values for the enums in the corresponding mako file and instantiate the actual enum-values via the passed string.
In the interest of minimal changes I've kept it this way (c.f. l 915-917 - the used `release_commit_publishing_policy` is of type `str`)

release_notes/markdown.py
AndreasBurger: That'll teach me to import other people's code I suppose 😄. Will do (also for the other cases in this file)

concourse/steps/release.py
zkdev: just check again, I mixed up `EnumValueWithDocumentation` and `EnumWithDocumentation`
you're right, lowercase 'default' is fine

release_notes/markdown.py
```diff
@@ -0,0 +1,152 @@
...
+            and isinstance(other, Title)
+            and (other == self or hash(other) == hash(self))
+        )
+
+
+def _simple_title(display: str) -> Title:
+    return Title(display, [display.lower()])
+
+
+categories: list[Title] = [
```
ccwienk, line 31: preferably use `tuple` + does typehint actually add much value here? I am not strong opposed to keeping typehint; however, considering this is a "readonly" "define", the type should be obvious from the value..

release_notes/markdown.py
```diff
@@ -0,0 +1,152 @@
...
+]
+
+target_groups: list[Title] = [
+    _simple_title('USER'),
+    _simple_title('OPERATOR'),
+    _simple_title('DEVELOPER'),
+    _simple_title('DEPENDENCY'),
+]
+
+# identifier -> Title
```
ccwienk, line 47: codestyle: how about using a type-alias for str, like so:
```
identifier = str
```

and then using `dict[identifier, Title]` as typehint.

release_notes/markdown.py
```diff
@@ -0,0 +1,152 @@
...
+    objs = []
+
+    level = start_level
+    last_index: typing.Optional[int] = None
+    for idx, line in enumerate(lines):
+        if line.startswith(' ') or line.startswith('\t'):
+            first_index = line.index(line.lstrip(' \t')[0])
+            if last_index is None or last_index < first_index:
+                last_index = first_index
+                if not idx:
```
ccwienk, line 80: codestyle: I think this should rather read `idx > 0`

ccwienk (changes requested):

concourse/steps/release.py
```diff
@@ -1043,7 +1070,7 @@ def release_and_prepare_next_dev_cycle(
                 slack_cfg_name=slack_cfg_name,
                 slack_channel=slack_channel,
                 release_version=release_version,
-                release_notes=release_notes,
+                release_notes=rls_notes,
```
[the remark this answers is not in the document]
AndreasBurger: >one might consider to use ...model.EnumName.DEFAULT(.value) to have an additional typehint, which may be helpful for future refactorings (consider grepping for all occurrences..); that being said, it might be preferable to expect the enum-value, and not a str, rather than converting within this function

Personally I'd agree with typing this more strongly. But that would break the existing convention here.
While I am intentionally refraining from using our transaction-coding, I'm still trying to adhere to the existing style for the other parts. At least for this PR.

> naming: I would prefer to not abreviate release w/ rls
That was clashing with the import of the release-notes coding. I'll think about it some more.

gitutil.py
```diff
@@ -242,16 +248,22 @@ def push(self, from_ref, to_ref):
...
             with remote.repo.git.custom_environment(**cmd_env):
                 fetch_result = remote.fetch(ref)[0]
                 return fetch_result.commit
 
 
-def url_with_credentials(github_cfg, github_repo_path):
+def url_with_credentials(github_cfg, github_repo_path, credentials=None):
     base_url = urllib.parse.urlparse(github_cfg.http_url())
-    credentials = github_cfg.credentials()
+    if not credentials:
```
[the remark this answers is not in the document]
AndreasBurger: No, you're right. I was trying to make it possible to ensure that the user set in the new git env-vars and used in the https-url is the same.
I've solved this in a slightly better way.

release_notes/markdown.py
AndreasBurger: I think there might still be a bug here. To me this reads as a function that takes an indented text-block and returns the un-indented elements together with the level they were indented at. But I'm not sure that it is successfully doing that. Will look some more into it tomorrow 🤔

PR #863, opened 2023-05-30
Support multiple ctx-repos in the update component dependencies trait

AndreasBurger:
With this PR our `update component dependencies` trait can be given an additional mapping of ctx-repo config names to component names. This will prompt the trait to look in the given ctx-repo when checking for/proposing component upgrade PRs. If no config is provided for a component, the same source as before is used (the ctx-repository as given by the `component descriptor` or, if this is not configured, the ctx-repository of the base-component itself)

Currently this is configured as a mapping of ctx-repo-config-name to a list of component-names, e.g.:
```yaml
  additional_ctx_repositories:
    onmetal:
    - github.com/onmetal/gardener-extension-provider-onmetal
```
We might also consider allowing base-urls in addition (or instead of) ctx-repository config names.

### DIFF
--- a/concourse/model/traits/update_component_deps.py
+++ b/concourse/model/traits/update_component_deps.py
@@ -160,6 +160,29 @@ ATTRIBUTES = (
         type=MergePolicy,
     ),
     AttributeSpec.optional(
+        name='additional_ctx_repositories',
+        default=dict(),
+        doc='''
+            Additional context repositories to look up components in. The entries are expected
+            to be mappings of context repo config names to a list of component names.
+
+            Example:
+
+            .. code-block:: yaml
+
+                additional_ctx_repositories:
+                    ctx_repo_config_name:
+                    - example/component/name
+                    - another/example/component
+                    ctx_repo_config_name_two:
+                    - yet/another/component
+
+            If a component name is configured for two context repository config names the last
+            mention will take precedence.
+        ''',
+        type=dict,
+    ),
+    AttributeSpec.optional(
         name='merge_policies',
         default=(),
         doc=(
@@ -218,6 +241,9 @@ class UpdateComponentDependenciesTrait(Trait):
 
     def upstream_update_policy(self):
         return UpstreamUpdatePolicy(self.raw.get('upstream_update_policy'))
+
+    def additional_ctx_repositories(self):
+        return self.raw['additional_ctx_repositories']
 
     def merge_policies(self):
         # handle default here
--- a/concourse/steps/update_component_deps.py
+++ b/concourse/steps/update_component_deps.py
@@ -205,18 +205,44 @@ def determine_reference_versions(
         raise NotImplementedError
 
 
+def create_ctx_repo_mapping(
+    map: dict[str, str],
+) -> dict[str, gci.componentmodel.OciRepositoryContext]:
+    # invert the input map and replace the ctx repo names with actual instances, i.e. turn it from
+    # a mapping of ctx repo name -> component name(s) to a mapping of component name -> ctx repo.
+    cfg_factory = ci.util.ctx().cfg_factory()
+
+    return {
+        component_name: gci.componentmodel.OciRepositoryContext(
+            baseUrl=cfg_factory.ctx_repository(ctx_repo_name).base_url()
+        )
+        for ctx_repo_name, component_name_list in map.items()
+            for component_name in component_name_list
+    }
+
+
 def determine_upgrade_prs(
     upstream_component_name: str,
     upstream_update_policy: UpstreamUpdatePolicy,
     upgrade_pull_requests: typing.Iterable[github.util.UpgradePullRequest],
-    ctx_repo: gci.componentmodel.OciRepositoryContext,
+    default_ctx_repo: gci.componentmodel.OciRepositoryContext,
+    ctx_repo_mapping: dict[str, gci.componentmodel.OciRepositoryContext],
     ignore_prerelease_versions=False,
-) -> typing.Iterable[typing.Tuple[
-    gci.componentmodel.ComponentReference, gci.componentmodel.ComponentReference, str
-]]:
+) -> typing.Iterable[
+    typing.Tuple[
+        gci.componentmodel.ComponentReference,
+        str,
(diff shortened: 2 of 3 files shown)


ccwienk:
@AndreasBurger as discussed:

consider generalising the repo-cfg like so:
```
ocm_repositories:
- ocm_repo_url: eu.gcr.io/gardener-project/xxx/.../
  component_names: []
  use_for: read # | writeonly                        
...

+ add this to component-descriptor-trait

also: extend cnudie.retrieve.Lookup such that it understands aforementioned cfg + implement a corresponding upload-function

File: concourse/model/traits/update_component_deps.py [hunk 1]
```diff
@@ -159,6 +159,29 @@ def merge_method(self):
         doc='whether or not created PRs should be automatically merged. **deprecated**',
         type=MergePolicy,
     ),
+    AttributeSpec.optional(
+        name='additional_ctx_repositories',
```
ccwienk, line 163: naming: I think we should refer to `ocm_repositories` for new code ([ocm-spec](https://github.com/open-component-model/ocm-spec/blob/main/doc/introduction/component_repository.md)).

semantics: let us rather define the contract like so:

- if no ocm-repositories are configured at all, use fallback to current behaviour (as you described in pull-request-description); if feasible, preferably return a a generated cfg (we use the same pattern for other traits)
- if ocm-repositories are configured, they are _the_ cfg to consider (no fallback)

Considering structure, consider OCM-CLI's spec:

https://github.com/open-component-model/ocm/blob/main/docs/reference/ocm_configfile.md
-> ocm.config.ocm.software (resolvers, ..)

Also, consider factoring the model-class out of the `concourse` package. ideally, this should go to package `ocm` (which does not exist yet), next-best package: `cnudie`?

concourse/model/traits/update_component_deps.py
```diff
@@ -159,6 +159,29 @@ def merge_method(self):
...
+        default=dict(),
```
ccwienk, line 164: preferably define a dataclass

concourse/model/traits/update_component_deps.py
```diff
@@ -159,6 +159,29 @@ def merge_method(self):
...
+            .. code-block:: yaml
+
+                additional_ctx_repositories:
+                    ctx_repo_config_name:
+                    - example/component/name
+                    - another/example/component
+                    ctx_repo_config_name_two:
+                    - yet/another/component
+
+            If a component name is configured for two context repository config names the last
```
ccwienk, line 180: see above. we should rather specify component-name-prefixes, rather than full component-names.

concourse/steps/update_component_deps.py
```diff
@@ -205,18 +205,44 @@ def determine_reference_versions(
         raise NotImplementedError
 
 
+def create_ctx_repo_mapping(
```
ccwienk, line 208: might inline to modelclass; alternativ: make modelclass callable so it can directly be passed as value, like so:

```
@dataclass.dataclass
class OcmRepositoryLookup:
  lookups: list[OcmRepositoryPrefixMapping]

 def __call__(self, component_id):
   for lookup in self.lookups:
     if lookup.matches(component_id):
       return lookup.ocm_repo
```

(Almost pseudo-code)

That way, instances of this type work as any other callable (thus no conversion required)

concourse/steps/update_component_deps.py
```diff
@@ -205,18 +205,44 @@ def determine_reference_versions(
...
+    default_ctx_repo: gci.componentmodel.OciRepositoryContext,
+    ctx_repo_mapping: dict[str, gci.componentmodel.OciRepositoryContext],
     ignore_prerelease_versions=False,
-) -> typing.Iterable[typing.Tuple[
-    gci.componentmodel.ComponentReference, gci.componentmodel.ComponentReference, str
-]]:
+) -> typing.Iterable[
+    typing.Tuple[
+        gci.componentmodel.ComponentReference,
+        str,
```
ccwienk, line 234: why did you change from `CREF, CREF, str` to  `CREF, str, CTX`?

concourse/steps/update_component_deps.py
```diff
@@ -205,18 +205,44 @@ def determine_reference_versions(
...
+            for component_name in component_name_list
+    }
+
+
 def determine_upgrade_prs(
     upstream_component_name: str,
     upstream_update_policy: UpstreamUpdatePolicy,
     upgrade_pull_requests: typing.Iterable[github.util.UpgradePullRequest],
-    ctx_repo: gci.componentmodel.OciRepositoryContext,
+    default_ctx_repo: gci.componentmodel.OciRepositoryContext,
```
ccwienk, line 228: I would prefer to remove `default_ctx_repo` and always pass-in a resolver-callable; ocm-repo calculation should rather be done by caller (dependency-injection)

concourse/steps/update_component_deps.py
```diff
@@ -205,18 +205,44 @@ def determine_reference_versions(
...
+    ctx_repo_mapping: dict[str, gci.componentmodel.OciRepositoryContext],
     ignore_prerelease_versions=False,
-) -> typing.Iterable[typing.Tuple[
-    gci.componentmodel.ComponentReference, gci.componentmodel.ComponentReference, str
-]]:
+) -> typing.Iterable[
+    typing.Tuple[
+        gci.componentmodel.ComponentReference,
```
ccwienk, line 233: btw, how about extending `ComponentReference` to also contain ocm-repo? just a thought - passing-along repos side-by-side will obviously also work fine.

concourse/steps/update_component_deps.py
```diff
@@ -264,7 +290,7 @@ def determine_upgrade_prs(
                 )
                 continue
             else:
-                yield(greatest_component_reference, greatest_version)
+                yield (greatest_component_reference, greatest_version, ctx_repo)
```
ccwienk, line 293: will this not break the calling code?

ccwienk (changes requested):

ccwienk:
on an additional note, please resolve conflicts :-/

AndreasBurger:
closing in Favour of #871

PR #867, opened 2023-06-15
Add limited support for downgrade pull requests

AndreasBurger:
With this, downgrade pull requests will be proposed if
- the component is configured to _strictly follow_ version changes and
- the component being tracked is itself a dependency of an upstream component

The intention is to allow components to closely follow the trajectory of their counterpart in the original component-reference.

Suppose we have two components, `A` and `B` that both depend on `C` with `B` following `A`s version of `C`. With this change, `B` can receive automatically created downgrade pull requests for `C` when `A` _downgrades_ its version of `C` for whatever reason. Downgrades of immediately tracked components are not proposed.

Downgrade pull request creation operates with the same rules as upgrade pull request creation, notably
- only one downgrade pull request will ever be proposed for a given version, unless the pull request is made unrecognizable to our coding at a later point
- merge policy is respected, so automatic merges may happen.

### DIFF
--- a/concourse/steps/update_component_deps.py
+++ b/concourse/steps/update_component_deps.py
@@ -84,12 +84,13 @@ def upgrade_pr_exists(
     component_reference: gci.componentmodel.ComponentReference,
     component_version: str,
     upgrade_requests: typing.Iterable[github.util.UpgradePullRequest],
+    request_filter: typing.Callable[[github.util.UpgradePullRequest], bool] = lambda rq: True,
 ) -> github.util.UpgradePullRequest | None:
     if any(
         (matching_rq := upgrade_rq).target_matches(
             reference=component_reference,
             reference_version=component_version,
-        )
+        ) and request_filter(upgrade_rq)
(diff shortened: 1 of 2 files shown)


concourse/steps/update_component_deps.py
```diff
@@ -189,54 +190,76 @@ def determine_upgrade_prs(
...
             upstream_update_policy=upstream_update_policy,
             ctx_repo=ctx_repo,
             ignore_prerelease_versions=ignore_prerelease_versions,
         )
-        if greatest_versions_to_consider:
+        if versions_to_consider:
             logger.info(
-                f'Found possible version{"s" if len(greatest_versions_to_consider)>1 else ""} to '
-                f'upgrade to: {greatest_versions_to_consider} for '
+                f"Found possible version(s) to upgrade to: '{versions_to_consider}' for "
```
ccwienk, line 203: nit: should read upgrade _or_ downgrade

ccwienk (commented):
apart from logging message nit, lgtm

PR #871, opened 2023-07-10
Support multiple ctx-repos in the update component dependencies trait

AndreasBurger:
With this PR our `update component dependencies` trait can be given an additional mapping of ctx-repo config names to component names. This will prompt the trait to look in the given ctx-repo when checking for/proposing component upgrade PRs. If no config is provided for a component, the same source as before is used (the ctx-repository as given by the `component descriptor` or, if this is not configured, the ctx-repository of the base-component itself)

Currently this is configured as a mapping of ctx-repo-config-name to a list of component-names, e.g.:
```yaml
  ocm_repository_mappings:
    - ocm_repo_url: <url>
      component_names: [ … ]
      use_for: 'readonly' | 'readwrite'
```

### DIFF
--- a/cnudie/upload.py
+++ b/cnudie/upload.py
@@ -1,0 +1,118 @@
+import dataclasses
+import enum
+import hashlib
+import json
+
+import ccc.oci
+import cnudie.util
+import gci.componentmodel as cm
+import gci.oci
+import oci.model as om
+
+
+class UploadMode(enum.StrEnum):
+    SKIP = 'skip'
+    FAIL = 'fail'
+    OVERWRITE = 'overwrite'
+
+
+def upload_component_descriptor(
+    component_descriptor: cm.ComponentDescriptor | cm.Component,
+    on_exist=UploadMode.SKIP,
+    ocm_repository: cm.OciRepositoryContext | str = None,
+):
+    if isinstance(component_descriptor, cm.Component):
+        component_descriptor = cm.ComponentDescriptor(
+            component=component_descriptor,
+            meta=cm.Metadata(),
+            signatures=[],
+        )
+
+    schema_version = component_descriptor.meta.schemaVersion
+    if not schema_version is cm.SchemaVersion.V2:
+        raise RuntimeError(f'unsupported component-descriptor-version: {schema_version=}')
+
+    client = ccc.oci.oci_client()
+
+    if ocm_repository:
+        if isinstance(ocm_repository, str):
+            ocm_repository = cm.OciRepositoryContext(baseUrl=ocm_repository)
+        elif isinstance(ocm_repository, cm.OciRepositoryContext):
+            pass
+        else:
+            raise TypeError(type(ocm_repository))
+
+        if  not component_descriptor.component.current_repository_ctx() == ocm_repository:
+            component_descriptor.component.repositoryContexts.append(ocm_repository)
+
+    target_ref = cnudie.util.oci_artefact_reference(component_descriptor.component)
+
+    if on_exist in (UploadMode.SKIP, UploadMode.FAIL):
+        # check whether manifest exists (head_manifest does not return None)
+        if client.head_manifest(image_reference=target_ref, absent_ok=True):
+            if on_exist is UploadMode.SKIP:
+                return
+            if on_exist is UploadMode.FAIL:
+                # XXX: we might still ignore it, if the to-be-uploaded CD is equal to the existing
+                # one
+                raise ValueError(f'{target_ref=} already existed')
+    elif on_exist is UploadMode.OVERWRITE:
+        pass
+    else:
+        raise NotImplementedError(on_exist)
+
+    raw_fobj = gci.oci.component_descriptor_to_tarfileobj(component_descriptor)
+    cd_digest = hashlib.sha256()
+    while (chunk := raw_fobj.read(4096)):
+        cd_digest.update(chunk)
+
+    cd_octets = raw_fobj.tell()
+    cd_digest = cd_digest.hexdigest()
+    cd_digest_with_alg = f'sha256:{cd_digest}'
+    raw_fobj.seek(0)
+
+    client.put_blob(
+        image_reference=target_ref,
+        digest=cd_digest_with_alg,
+        octets_count=cd_octets,
+        data=raw_fobj,
+    )
+
+    cfg = gci.oci.ComponentDescriptorOciCfg(
+        componentDescriptorLayer=gci.oci.ComponentDescriptorOciBlobRef(
+            digest=cd_digest_with_alg,
+            size=cd_octets,
+        )
+    )
+    cfg_raw = json.dumps(dataclasses.asdict(cfg)).encode('utf-8')
+    cfg_octets = len(cfg_raw)
+    cfg_digest = hashlib.sha256(cfg_raw).hexdigest()
+    cfg_digest_with_alg = f'sha256:{cfg_digest}'
+
+    client.put_blob(
+        image_reference=target_ref,
+        digest=cfg_digest_with_alg,
+        octets_count=cfg_octets,
+        data=cfg_raw,
+    )
+
+    manifest = om.OciImageManifest(
+        config=gci.oci.ComponentDescriptorOciCfgBlobRef(
+            digest=cfg_digest_with_alg,
+            size=cfg_octets,
+        ),
+        layers=[
+            gci.oci.ComponentDescriptorOciBlobRef(
+                digest=cd_digest_with_alg,
+                size=cd_octets,
+            ),
+        ],
+    )
+
+    manifest_dict = manifest.as_dict()
+    manifest_bytes = json.dumps(manifest_dict).encode('utf-8')
+
+    client.put_manifest(
+        image_reference=target_ref,
+        manifest=manifest_bytes,
+    )
--- a/cnudie/util.py
+++ b/cnudie/util.py
@@ -144,6 +144,83 @@ def determine_component_name(
     return component_name.lower() # OCI demands lowercase
 
 
+def normalise_component_name(component_name: str) -> str:
+    return component_name.lower() # oci-spec demands lowercase
+
+
+def oci_artefact_reference(
+        component: (
+            cm.Component
+            | cm.ComponentIdentity
+            | cm.ComponentReference
+            | str # 'name:version'
+            | tuple[str, str] # (name, version)
+        ),
+        ocm_repository: str | cm.OciRepositoryContext = None
+) -> str:
+    if isinstance(component, cm.Component):
+        if not ocm_repository:
+            ocm_repository = component.current_repository_ctx()
+        component_name = component.name
+        component_version = component.version
+
+    elif isinstance(component, cm.ComponentIdentity):
+        component_name = component.name
+        component_version = component.version
+
+    elif isinstance(component, cm.ComponentReference):
+        component_name = component.componentName
+        component_version = component.version
+
+    elif isinstance(component, str):
+        component_name, component_version = component.split[':']
+
+    elif isinstance(component, tuple):
+        if not len(component) == 2 or not (
+            isinstance(component[0], str) and isinstance(component[1], str)
+        ):
+            raise TypeError("If a tuple is given as component, it must contain two strings.")
+        component_name, component_version = component
+    else:
+        raise TypeError(type(component))
+
+    if not ocm_repository:
+        raise ValueError('ocm_repository must be given unless a Component is passed.')
+    elif isinstance(ocm_repository, str):
+        repo_ctx = cm.OciRepositoryContext(baseUrl=ocm_repository)
+    elif isinstance(ocm_repository, cm.OciRepositoryContext):
+        repo_ctx = ocm_repository
+    else:
+        raise TypeError(type(ocm_repository))
+
+    return repo_ctx.component_version_oci_ref(
+        name=component_name,
+        version=component_version,
+    )
+
+
+def target_oci_ref(
+    component: cm.Component,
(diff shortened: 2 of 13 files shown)


cnudie/upload.py
```diff
@@ -0,0 +1,135 @@
...
+
+import ccc.oci
+import gci.componentmodel as cm
+import gci.oci
+import json
+import oci.model as om
+import dataclasses
+
+
+class UploadMode(enum.Enum):
```
ccwienk, line 12: nit: How about using StrEnum (we can make use python3.11-features by now..); however, it is fine for me not to use it.

cnudie/upload.py
```diff
@@ -0,0 +1,135 @@
...
+    SKIP = 'skip'
+    FAIL = 'fail'
+    OVERWRITE = 'overwrite'
+
+
+def _normalise_component_name(component_name: str) -> str:
```
ccwienk, line 18: nit: I think this might better be suited for cnudie.util, wdyt?
generally speaking, it is of course good to move / (copy for now) from product.v2, as discussed earlier

cnudie/upload.py
```diff
@@ -0,0 +1,135 @@
...
+    return component_name.lower()  # oci-spec allows only lowercase
+
+
+def _target_oci_ref(
```
ccwienk, line 22: ditto - considering this is also used in replicate.py / not specific to uploading, but also required for downloading

cnudie/upload.py
```diff
@@ -0,0 +1,135 @@
...
+    component: cm.Component,
+    component_ref: cm.ComponentReference=None,
+    component_version: str=None,
+):
```
ccwienk, line 26: I think this signature should be overhauled (I know this function was copied; still I think we should take this opportunity and improve it).

suggestion (pseudo-code-ish):

```
def _oci_artefact_reference(
  component: Component | ComponentIdentity | ComponentReference | str # if str, expect `name:version` | tuple[str, str] # name, version),
  ocm_repository: str | RepoDataClass # did not lookup exact name, see gci.model = None,
) -> str: # is it actually str?
  # if component, ocm-repo can be omitted (but must be honoured if given)
  # otherwise, ocm-repo must be passed
```

cnudie/upload.py
```diff
@@ -0,0 +1,135 @@
...
+        component_name = component_ref.name
+    else:
+        component_name = component_ref.componentName
+
+    component_name = _normalise_component_name(component_name)
+    component_version = component_ref.version
+
+    last_ctx_repo = component.current_repository_ctx()
+
+    return last_ctx_repo.component_version_oci_ref(
```
ccwienk, line 38: considering the actual calculation is done in ctx-repo, I think the function should be refactored even further.

how about just (keeping the signature I discussed above) having a fucntion that will return the ocm-repo-instance (which can then be used by a caller to calculate oci-ref - said function might also be changed to accept different arguments (like this fucntion) for convenience)

cnudie/upload.py
```diff
@@ -0,0 +1,135 @@
...
+        name=component_name,
+        version=component_version,
+    )
+
+
+def _sha_digest(fileobj) -> str:
```
ccwienk, line 44: this function is new, is it not? I understand the idea of deduplicating the existing code, however, alas, I do have some concerns.

firstly, considering the name, this function calculates and returns a sha256--digest from passed-in object in hexadecimal form w/ `sha256:`-prefix. those things should be reflected in the name, e.g. `_sha256_hexdigest`

secondly, default chunksize should ideally be a parameter (it is okay to define a default-value, of course)

thirdly, not all filelike-objects are in fact seekable, and any modification to fpointer should ideally be mentioned in a docstr so callers do not have to read through the code (albeit it is short)

lastly, this function is at one place called w/ a byte-array (<class 'bytes'> as python will print its type); in this case, this will even break w/o special-case-handling depending on input-type.

from my gut-feeling, this is  little code and there is few deduplication (and too much complexity if this function would be "done right") to justify this factoring-out as a good idea.

we might factor-out the `sha256:`-prefixing. but then again: this is a contract brought upon us from Docker/OCI, and I strongly doubt it to ever change, so I do not see redundant hardcoding as an issue in this case.

cnudie/upload.py
```diff
@@ -0,0 +1,135 @@
...
+    digest = hashlib.sha256()
```
ccwienk, line 45: naming-nit: the object returned by hashlib.sha256 is not a digest, but a hash (function)

cnudie/upload.py
```diff
@@ -0,0 +1,135 @@
...
+    position = fileobj.tell()
+    while (chunk := fileobj.read(4096)):
+        digest.update(chunk)
+    fileobj.seek(position)
+    return f'sha256:{digest.hexdigest()}'
+
+
+def upload_component_descriptor(
+    component_descriptor: cm.ComponentDescriptor,
```
ccwienk, line 54: for concenience, we might also allow just `component` (and add the Component-Descriptor-boilerplate implicitly) - optional

cnudie/upload.py
```diff
@@ -0,0 +1,135 @@
...
+    on_exist=UploadMode.SKIP,
+    ctx_repo_override: cm.OciRepositoryContext = None,
```
ccwienk, line 56: I suggest to rename this to just `ocm_repository`

I also suggest to define a Union-Type `OciRepositoryContext` | str and accept also a plain-str.

cnudie/upload.py
```diff
@@ -0,0 +1,135 @@
...
+    raw_fobj = gci.oci.component_descriptor_to_tarfileobj(component_descriptor)
+    cd_sha_digest = _sha_digest(raw_fobj)
+    cd_octets = len(raw_fobj) # TODO: test
+
+    client.put_blob(
+        image_reference=target_ref,
+        digest=cd_sha_digest,
+        octets_count=cd_octets,
+        data=raw_fobj,
+        # mimetype=gci.oci.component_descriptor_mimetype, # not used in called func
```
ccwienk, line 95: if it is not used, then remove it - or let us discuss changing the called function, if this seems appropriate. remembering the oci-transport-spec, I think blob-puts are always to use application/data, so it might really not make a lot of sense to specify a mimetype.

cnudie/upload.py
```diff
@@ -0,0 +1,135 @@
...
+                # one
+                raise ValueError(f'{target_ref=} already existed')
+    elif on_exist is UploadMode.OVERWRITE:
+        pass
+    else:
+        raise NotImplementedError(on_exist)
+
+    raw_fobj = gci.oci.component_descriptor_to_tarfileobj(component_descriptor)
+    cd_sha_digest = _sha_digest(raw_fobj)
+    cd_octets = len(raw_fobj) # TODO: test
```
ccwienk, line 88: is this TODO new, or did you copy-and-paste it?

either way: preferably do not abbreviate component-descriptor to just `cd` (I know, I also did this at some places); i think this abbrev. is really misleading (e.g. might be confused with `change-dir`). how about comp_descr, or sth.? I do see not abbreviating at all means some more typing (then again, there is autocomplete..)

cnudie/upload.py
```diff
@@ -0,0 +1,135 @@
...
+    cfg_raw = json.dumps(dataclasses.asdict(cfg)).encode('utf-8')
+    cfg_sha_digest = _sha_digest(cfg_raw)
+    cfg_octets = len(cfg_raw)
+
+    client.put_blob(
+        image_reference=target_ref,
+        digest=cfg_sha_digest,
+        octets_count=cfg_octets,
+        data=cfg_raw,
+        # mimetype='application/vnd.docker.container.image.v1+json',
```
ccwienk, line 113: ditto

cnudie/upload.py
```diff
@@ -0,0 +1,135 @@
...
+    )
+
+    cfg = gci.oci.ComponentDescriptorOciCfg(
+        componentDescriptorLayer=gci.oci.ComponentDescriptorOciBlobRef(
+            digest=cd_sha_digest,
+            size=cd_octets,
+        )
+    )
+    cfg_raw = json.dumps(dataclasses.asdict(cfg)).encode('utf-8')
+    cfg_sha_digest = _sha_digest(cfg_raw)
```
ccwienk, line 105: as mentioned above, this will not work

cnudie/util.py
```diff
@@ -470,3 +472,89 @@ def enumerate_group_pairs(
...
+                self._rw_lookups.append(filter_lookup)
+            self._ro_lookups.append(filter_lookup)
+
+    def apply_to(
+        self,
+        component_name,
+        use_for: str | oa.Privileges = 'readonly',
+    ) -> typing.Iterable[cm.OciRepositoryContext]:
+
+        if isinstance(use_for, str):
```
ccwienk, line 527: no need - just always pass it to enum's ctor

cnudie/util.py
```diff
@@ -470,3 +472,89 @@ def enumerate_group_pairs(
...
+            filter_lookup = (filter_func, ctx_repo)
+
+            if mapping.use_for is oa.Privileges.READWRITE:
+                self._rw_lookups.append(filter_lookup)
+            self._ro_lookups.append(filter_lookup)
+
+    def apply_to(
+        self,
+        component_name,
+        use_for: str | oa.Privileges = 'readonly',
```
ccwienk, line 524: nit: why not init w/ enum-value?

also: rather call those (required_)privileges?

cnudie/util.py
```diff
@@ -470,3 +472,89 @@ def enumerate_group_pairs(
...
+        for mapping in mappings:
+            filter_func = reutil.re_filter(include_regexes=mapping.component_names)
+            ctx_repo = cm.OciRepositoryContext(baseUrl=mapping.ocm_repo_url)
+            filter_lookup = (filter_func, ctx_repo)
+
+            if mapping.use_for is oa.Privileges.READWRITE:
+                self._rw_lookups.append(filter_lookup)
+            self._ro_lookups.append(filter_lookup)
+
+    def apply_to(
```
ccwienk, line 521: naming: rather call this sth. like `find_ocm_repository`

cnudie/util.py
ccwienk: do we actually need to differentiate privileges / "use-for" at this point? I think the mapping might just yield _a_ oci-repository, regardless of intent

concourse/model/traits/component_descriptor.py
```diff
@@ -178,7 +178,28 @@ class StepInput:
...
         type=typing.List[StepInput],
         doc='inputs to expose to component-descriptor step',
-    )
+    ),
+    AttributeSpec.optional(
+        name='ocm_repository_mappings',
+        default=[], # cannot define a proper default here because this depends on another (optional)
+                    # config-value. At least not in a way that would be represented in our
+                    # rendered documentation.
+        type=list,
```
ccwienk, line 187: nit: specify type of list-items?

concourse/model/traits/update_component_deps.py
```diff
@@ -176,6 +176,25 @@ def merge_method(self):
         ''',
         type=ReleaseNotesHandling,
     ),
+    AttributeSpec.optional(
```
ccwienk, line 179: okay, deduplicating at this level should be fine

concourse/steps/component_descriptor.mako
```diff
@@ -80,6 +83,34 @@ component_labels = ${descriptor_trait.component_labels()}
...
+        dacite.from_dict(
+            data_class=cnudie.util.OcmLookupMapping,
+            config=dacite.Config(
+                cast=[
+                    oa.Privileges,
+                ],
+            ),
+            data={
+                'ocm_repo_url': ctx_repository_base_url,
+                'component_names': ['.*',],
```
ccwienk, line 106: does OCM-Cli also specify regex? also: did they not call this (component-)prefix? I would prefer to keep as close as reasonably possible to their [spec](https://github.com/open-component-model/ocm/blob/main/docs/reference/ocm_configfile.md)

ccwienk (changes requested):

cnudie/upload.py
AndreasBurger: Agreed - removed

cnudie/upload.py
AndreasBurger: Done

concourse/steps/component_descriptor.mako
ccwienk: @AndreasBurger : yes, I think we should keep close to OCM-cli to
a) make migration easier (should we ever need to)
b) offer a somewhat consistent interface at least

cnudie/util.py
```diff
@@ -584,50 +621,108 @@ def __init__(
...
+        cfg_factory = None,
+    ) -> str:
+
+        if not cfg_factory:
+            cfg_factory = ctx.cfg_factory()
+
+        consumers = []
+        resolvers = []
+        for index, m in enumerate(self.mappings):
+            container_registry_config = model.container_registry.find_config(
```
AndreasBurger, line 660: consider removing this and writing all necessary credentials to dockercfg

cnudie/util.py
```diff
@@ -584,50 +621,108 @@ def __init__(
...
         required_privileges: str | oa.Privileges = oa.Privileges.READONLY,
     ) -> typing.Iterable[cm.OciRepositoryContext]:
 
         required_privileges = oa.Privileges(required_privileges)
-        if required_privileges is oa.Privileges.READONLY:
-            lookups = self._ro_lookups
-
-        elif required_privileges is oa.Privileges.READWRITE:
-            lookups = self._rw_lookups
+        for mapping in self.mappings:
```
AndreasBurger, line 632: handle priority explicitly

PR #874, opened 2023-07-14
Optionally merge release-commit from tag to default-branch

zkdev:
Release commits can be pushed to tag only (publish policy `tag_only`). However, if release-commit has side-effects, e.g. version bumps via release-callback, these are missing on default branch.

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/concourse/model/traits/release.py
+++ b/concourse/model/traits/release.py
@@ -85,6 +85,12 @@ class ReleaseCommitPublishingPolicy(EnumWithDocumentation):
         doc='publish release tag to dead-end',
     )
 
+    TAG_AND_MERGE_BACK = EnumValueWithDocumentation(
+        value='tag_and_merge_back',
+        doc='publish release tag to dead-end and merge back release commit to default branch',
+    )
+
+
 
 ATTRIBUTES = (
     AttributeSpec.optional(
--- a/concourse/steps/release.py
+++ b/concourse/steps/release.py
@@ -16,6 +16,7 @@ from github3.exceptions import (
 from git.exc import (
     GitCommandError
 )
+import git.types
 
 import gci.componentmodel as cm
 
@@ -265,7 +266,10 @@ class ReleaseCommitStep(TransactionalStep):
                 from_ref=release_commit.hexsha,
                 to_ref=self.repository_branch
             )
-        elif self.publishing_policy is ReleaseCommitPublishingPolicy.TAG_ONLY:
+        elif self.publishing_policy in (
+            ReleaseCommitPublishingPolicy.TAG_ONLY,
+            ReleaseCommitPublishingPolicy.TAG_AND_MERGE_BACK,
+        ):
             # handled when creating all release tags
             pass
         else:
@@ -283,6 +287,7 @@ class CreateTagsStep(TransactionalStep):
         github_helper: GitHubRepositoryHelper,
         git_helper: GitHelper,
         publishing_policy: ReleaseCommitPublishingPolicy,
+        repository_branch: str,
     ):
         self.github_helper = github_helper
         self.git_helper = git_helper
@@ -290,6 +295,9 @@ class CreateTagsStep(TransactionalStep):
         self.publishing_policy = publishing_policy
 
         self.tags_to_set = tags_to_set
+
+        self.repository_branch = repository_branch
+        self.release_tag = tags_to_set[0]
 
     def name(self):
         return 'Create Tags'
@@ -311,6 +319,7 @@ class CreateTagsStep(TransactionalStep):
         if self.publishing_policy in [
             ReleaseCommitPublishingPolicy.TAG_ONLY,
             ReleaseCommitPublishingPolicy.TAG_AND_PUSH_TO_BRANCH,
+            ReleaseCommitPublishingPolicy.TAG_AND_MERGE_BACK,
         ]:
             def _push_tag(tag):
                 self.git_helper.push(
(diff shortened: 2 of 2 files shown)


concourse/model/traits/release.py
```diff
@@ -185,6 +185,16 @@ class ReleaseNotesHandling(EnumWithDocumentation):
         ''',
         type=str,
     ),
+    AttributeSpec.optional(
+        name='merge_release_to_default_branch',
```
ccwienk, line 189: naming: rather call this merge_release_commit_to_default_branch

concourse/model/traits/release.py
```diff
@@ -185,6 +185,16 @@ class ReleaseNotesHandling(EnumWithDocumentation):
...
+        default=False,
+        doc='''
+        merge the release-commit from tag back to default branch.
+        Useful if release-commit has side-effects, e.g. version-bumps via release-callback.
+        Only executed if `release_commit_publishing_policy` is set to `tag_only`.
```
ccwienk, line 194: setting this if publishing-policy is not tag_only should rather be considered an error

OR: how about adding a new publishing-policy:
- tag_and_merge_back
- tag_only
- <the other value, I am too lazy to lookup>

concourse/steps/release.py
```diff
@@ -740,6 +740,69 @@ def apply(self):
         logger.info('successfully posted contents to slack')
 
 
+class MergeReleaseToDefaultStep(TransactionalStep):
```
ccwienk, line 743: considering those heavyweight transactional-steps did not really work out well, please rather merge this into eh existing step

concourse/steps/release.py
```diff
@@ -740,6 +740,69 @@ def apply(self):
...
+
+        try:
+            merge_commit = create_merge_commit()
+
+            self.git_helper.push(
+                from_ref=merge_commit,
+                to_ref=self.repository_branch
+            )
+
+        except:
```
ccwienk, line 795: ideally we should not use bare-excepts

concourse/steps/release.py
```diff
@@ -740,6 +740,69 @@ def apply(self):
...
+            logger.warning(
+                f'Merging release-commit from tag {self.release_tag} failed.'
+                f'Manual actions might be required.'
```
ccwienk, line 798: nit: this warning "manual actions might be required" is not helpful. either be more specific (especially w.r.t. what this means), or omit it and only leave the first line

concourse/steps/release.py
zkdev: We discussed that in case merging-back fails, the release must not be blocked.
Therefore, I would assume the bare-exept to be acceptable.

concourse/steps/release.py
```diff
@@ -338,6 +347,41 @@ def _push_tag(tag):
         else:
             raise NotImplementedError
 
+        if self.publishing_policy == ReleaseCommitPublishingPolicy.TAG_AND_MERGE_BACK:
```
ccwienk, line 350: codestyle: enum-values should be compared by identity (`is`) rather than equality (`==`)

ccwienk: overall lgtm

concourse/steps/release.py
```diff
@@ -283,6 +287,7 @@ def __init__(
         github_helper: GitHubRepositoryHelper,
         git_helper: GitHelper,
         publishing_policy: ReleaseCommitPublishingPolicy,
+        repository_branch,
```
AndreasBurger, line 290:
```suggestion
        repository_branch: str,
```

concourse/steps/release.py
```diff
@@ -334,8 +343,51 @@ def _push_tag(tag):
...
+
+                merge_commit = create_merge_commit(head)
+                self.context().merge_release_back_to_default_branch_commit = merge_commit # pass to other steps
+
+                self.git_helper.push(
+                    from_ref=merge_commit.hexsha,
+                    to_ref=self.repository_branch
+                )
+
+                self.git_helper.repo.head.reset(
```
AndreasBurger, line 379: might be worth a comment why we hard-reset to the merge-commit here (to make sure the next dev-cycle commit does not undo the merge-commit).
At least I'm not sure I'll remember this in a few months time.

AndreasBurger (approved):
lgtm (if tested 🙈)

concourse/steps/release.py
```diff
@@ -334,8 +343,51 @@ def _push_tag(tag):
...
+                    parent_commits=(
+                        head,
+                        self.context().release_commit,
+                    ),
+                )
+
+            try:
+                head = self.git_helper.fetch_head(
+                    f'refs/heads/{self.repository_branch}'
+                ) # this is safe as release-tag is created already
```
ccwienk, line 369: I think this comment is slightly misleading. what we want/try to do is to reduce the risk for conflicts / create a merge-commit against current tgt-branch's head. it might improve readability if this code would be factored into a (local) function w/ an expressive name (alternatively, add a comment explaining what is tried to be done - and why it is okay (or so we think) to not fail release in case of errors)

concourse/steps/release.py
```diff
@@ -334,8 +343,51 @@ def _push_tag(tag):
...
+
+                merge_commit = create_merge_commit(head)
+                self.context().merge_release_back_to_default_branch_commit = merge_commit # pass to other steps
```
ccwienk, line 372: nit: `context()` is always used to pass data between other steps - no need to add this as a comment

concourse/steps/release.py
```diff
@@ -334,8 +343,51 @@ def _push_tag(tag):
...
+                    to_ref=self.repository_branch
+                )
+
+                self.git_helper.repo.head.reset(
+                    commit=merge_commit.hexsha,
+                    index=True,
+                    working_tree=True,
+                )
+
+            except:
```
ccwienk, line 385: preferably do not add bare except
also, rather print stacktrace, even if just warning (allows for easier debugging in case unexpected errors occur)

ccwienk (changes requested):

ccwienk (approved):

PR #878, opened 2023-08-08
Add support for gh oAuth token rotation

AndreasBurger:
Note: I still want to do some manual tests, do not merge


AndreasBurger:
updated to workaround the issues we foresaw:
- github-credentials _may_ now include a secondary oAuth token.
- if a secondary oAuth token is present, the secondary will be refreshed and set as the new primary
- the old primary will be set to the new secondary and thus remain valid until the next refresh happens

PR #883, opened 2023-08-21
Add option to include all refs in comp dep iterator

8R0WNI3:

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/cnudie/iter.py
+++ b/cnudie/iter.py
@@ -20,7 +20,7 @@ class Node:
 
 @dataclasses.dataclass
 class ComponentNode(Node):
-    pass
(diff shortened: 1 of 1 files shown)


cnudie/iter.py
```diff
@@ -20,7 +20,7 @@ def component_id(self):
 
 @dataclasses.dataclass
 class ComponentNode(Node):
-    pass
+    comp_refs: tuple[tuple[cm.Component]] = dataclasses.field(default_factory=lambda: tuple(tuple()))
```
ccwienk, line 23: naming: rather call this `component_references`

ccwienk: also: is it not sufficient to use `path`

ccwienk (changes requested):

PR #885, opened 2023-08-22
fix: guard against absence of release-notes for current version

ccwienk:


release_notes/fetch.py
```diff
@@ -241,7 +241,7 @@ def fetch_release_notes(
             continue
         component_versions[parsed_version] = ver
 
-    if not current_version:
+    if not current_version or current_version not in component_versions:
         current_version_tag = None
```
AndreasBurger, line 245:
```suggestion
        current_version_tag = None
        current_version = None
```
With this the code will fall back to creating the release notes from `HEAD` to previous version (if given, retrieved from oci-repository otherwise).
This might produce unexpected notes, though I suppose that's better than to break immediately. This should only happen infrequently and if this function is called with invalid parameters due to a misconfiguration beforehand.

AndreasBurger (approved):
lgtm

PR #886, opened 2023-09-01
Overhaul lookups

ccwienk:


cnudie/retrieve.py
```diff
@@ -37,6 +41,82 @@
...
+        for cfg in repository_cfg:
+            yield _iter_ocm_repositories(component, cfg)
+        return
+
+
+def iter_ocm_repositories(
+    component: str | cm.ComponentIdentity | cm.Component,
+    /,
+    *repository_cfgs: OcmRepositoryCfg,
+) -> OcmRepositoryLookup:
```
zkdev, line 95: Wrong type hint, it is a Generator of `OcmRepositoryLookup`.

cnudie/retrieve.py
```diff
@@ -37,6 +41,82 @@
...
+        return
+
+    if isinstance(repository_cfg, cnudie.util.OcmLookupMapping):
+        repository_cfg: cnudie.util.OcmLookupMapping
+        if repository_cfg.matches(component):
+            yield repository_cfg.ocm_repo_url
+        return
+
+    if isinstance(repository_cfg, cnudie.util.OcmLookupMappingConfig):
+        yield repository_cfg.iter_ocm_repositories(component)
```
zkdev, line 82: We should rather `yield from` as `iter_ocm_repositories` returns a Generator.

cnudie/retrieve.py
```diff
@@ -37,6 +41,82 @@
...
+
+    # recurse into elements in case repository_cfg is iterable
+    if hasattr(repository_cfg, '__iter__'):
+        for cfg in repository_cfg:
+            yield _iter_ocm_repositories(component, cfg)
```
zkdev, line 87: Should also be `yield from`.

zkdev (changes requested):

cnudie/retrieve.py
ccwienk: this will not work, as `_iter_ocm_repositories` is expected to return an iterable in `iter_ocm_reposositories`

cnudie/retrieve.py
ccwienk: why? if repository_cfg is iterable, the for-loop will to a `yield from` already
flattenning is already done via recursion

zkdev (approved):
/lgtm

PR #888, opened 2023-09-06
Update sprint model

zkdev:
Structure single sprint as list of sprint-events, rather than single object. Also add "display-name" attribute.

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/delivery/model.py
+++ b/delivery/model.py
@@ -26,13 +26,31 @@ class GithubUser:
     github_hostname: str
 
 
+@dataclasses.dataclass(frozen=True)
+class SprintDate:
+    name: str
+    display_name: str
+    value: datetime.datetime
+
+
 @dataclasses.dataclass(frozen=True) # TODO: deduplicate w/ modelclass in delivery-service/yp.py
 class Sprint:
     name: str
-    end_date: datetime.datetime
-    release_decision: datetime.datetime
-    rtc: datetime.datetime
-    canary_freeze: datetime.datetime
+    dates: list[SprintDate]
+
+    def find_sprint_date(
+        self,
+        name: str,
+        absent_ok: bool = False,
+    ) -> SprintDate | None:
+        for sprint_date in self.dates:
+            if sprint_date.name == name:
+                return sprint_date
+
+        if absent_ok:
+            return None
+
+        raise RuntimeError(f'did not find {name=} in {self=}')
 
     @staticmethod
     def from_dict(raw: dict):
--- a/github/compliance/milestone.py
+++ b/github/compliance/milestone.py
@@ -25,11 +25,16 @@ def find_or_create_sprint_milestone(
         return ms
 
     title = _milestone_title(sprint=sprint)
+
+    sprint_release_decision = sprint.find_sprint_date(
+        name='release_decision',
+    )
+
     ms = repo.create_milestone(
         title=title,
         state='open',
         description=f'used to track issues for upcoming sprint {title}',
-        due_on=sprint.release_decision.isoformat(),
+        due_on=sprint_release_decision.value.isoformat(),
     )
 
     return ms
--- a/github/compliance/report.py
+++ b/github/compliance/report.py
@@ -595,7 +595,7 @@ def _target_milestone(
 def _target_sprint(
     delivery_svc_client: delivery.client.DeliveryServiceClient,
     latest_processing_date: datetime.date,
-):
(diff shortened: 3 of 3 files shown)


zkdev:
requires adjustments in delivery-(service|dashboard)

-> do-not-merge yet!

delivery/model.py
```diff
@@ -26,25 +30,37 @@ class GithubUser:
...
+class SprintEvent:
+    value: datetime.datetime | str # e.g. <iso-timestamp> or <sprint-name>
+    attribute: str # e.g. "rtc"
+    attribute_display_name: str # e.g. "Release To Customer"
 
     @staticmethod
     def from_dict(raw: dict):
         return dacite.from_dict(
-            data_class=Sprint,
+            data_class=SprintEvent,
```
ccwienk, line 42: naming: not every data describing a Sprint is actually an event. So let us not choose a name implying that.

we have:
- name (obviously not an event, is it?)
- dates (start, end, rtc. ..)

Considering that there seem to be two different attributes that make up a sprint (name and dates), it might be worthwhile reflecting that in the model, maybe like so:

```
# an example sprint in yaml form
- name: 2208a # or: id?
  dates: # or: data / attributes?
  - name: end_date # name is a technical id which we will keep stable
    display_name: End Date
    value: 2023-09-09
  # ...
- name: 2208b
  # ...
```

Rationale: Sprints will for sure always have a name, so it might be more adequate to express this explicitly in the model, rather than making the model so generic that even that could be subject to customisation.

github/compliance/milestone.py
```diff
@@ -25,11 +30,17 @@ def find_or_create_sprint_milestone(
         return ms
 
     title = _milestone_title(sprint=sprint)
+
+    sprint_release_decision = delivery.model.find_sprint_event_by_attribute(
+        sprint=sprint,
+        attribute='release_decision',
+    ).isoformat()
```
ccwienk, line 37: missing error-handling. find might not yield a result

ccwienk (changes requested):

github/compliance/milestone.py
```diff
@@ -25,11 +25,20 @@ def find_or_create_sprint_milestone(
...
+
+    if not sprint_release_decision:
+        raise RuntimeError(f'sprint missing release-decision, {sprint=}')
+
     ms = repo.create_milestone(
         title=title,
         state='open',
         description=f'used to track issues for upcoming sprint {title}',
-        due_on=sprint.release_decision.isoformat(),
+        due_on=sprint_release_decision.value.isoformat(),
```
ccwienk, line 41: ohhh.... this gives me some nice ideas :-)

considering all our sprint-entries are currently "named dates", we might make `SprintDate` actually an instance of `datetime.datetime` (could be done by using inheritance - however, might be a tad too much work to be worth it)

It saddens me that `github3.py's` repository.create_milestone (this is what `repo` is, right?) does not accept datetime.datetime-instances as an alternative to passing-in a str (or does it?) :-/

ccwienk (changes requested):

github/compliance/milestone.py
zkdev: `repo` is of type `github3.repos.Repository`, yes.

Unfortunately no support for `datetime.datetime` ([source](https://github.com/sigmavirus24/github3.py/blob/43aa2d617a7d91743d7ab93c87bc9a52e300cb72/src/github3/repos/repo.py#L1112)):
```
        :param str due_on:
            (optional), ISO 8601 formatted due date
```

---

As `create_milestone` does not support `datetime.datetime` as type for its `due_on` parameter I do not see much benefit of inheriting `datetime.datetime` into `SprintDate`.

ccwienk: ahh, yes, you are correct :-/

ccwienk (approved):

delivery/model.py
```diff
@@ -26,13 +26,31 @@ class GithubUser:
...
+        absent_ok: bool = False,
+    ) -> SprintDate | None:
+        for sprint_date in self.dates:
+            if sprint_date.name == name:
+                return sprint_date
+
+        if absent_ok:
+            return None
+
+        raise RuntimeError(f'did not find {name=} in sprint {self.name=}')
```
ccwienk, line 53: it might be helpful to dump the full object (as you did before) for debugging purposes

zkdev:
consumers are ready

PR #889, opened 2023-09-14
Make secrets-replication more configurable

AndreasBurger:

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/concourse/steps/replicate_secrets.py
+++ b/concourse/steps/replicate_secrets.py
@@ -1,8 +1,5 @@
 import base64
 import logging
-import typing
-import pprint
-import re
 
 import kubernetes.client
 import kubernetes.config
@@ -17,9 +14,13 @@ import ci.log
 import ci.util
 import model
 import model.concourse
+import model.config_repo
+import model.secret
+import model.secrets_server
 
 ci.log.configure_default_logging()
 logger = logging.getLogger(__name__)
+
 
 
 def process_config_queue(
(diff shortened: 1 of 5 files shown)


concourse/steps/replicate_secrets.py
```diff
@@ -143,55 +144,79 @@ def _put_secret(
...
-        for cfg_mapping in [cfg_set._cfg_mappings()]:
-            for cfg_type_name, _ in cfg_mapping:
-                pprint.pprint(
-                    {cfg_type_name: cfg_set._cfg_element_names(cfg_type_name=cfg_type_name)}
-                )
+    for mapping in replication_target_config.replication_mappings:
+        cfg_set = cfg_factory.cfg_set(mapping.cfg_set)
+        # force cfg_set serialiser to include referenced cfg_sets
+        cfg_sets = list(cfg_set._cfg_elements('cfg_set')) + [cfg_set]
+        if isinstance(target := mapping.target, model.config_repo.KubernetesSecretTarget):
```
ccwienk, line 156: considering there will be multiple cases, rather pull this codeblock into a function.

also, would it not make sense to invert the iteration, like so:

```
for cfg_set in cfg_sets:
  serialised_cfg = <code to serialise cfg-set>
  for target in targets:
    if <is-k8s-tgt>:
     put_secret(serialised_cfg)
   elif <is-other-tgt>:
     put_secret_otherwise(serialised_cfg)
```

wdyt?

model/secret.py
```diff
@@ -57,7 +57,7 @@ def key_from_gen(self) -> bytes:
     def cipher_algorithm(self):
         return Cipher(self.raw.get('cipher_algorithm'))
 
-    def generation(self):
+    def generation(self) -> int:
         return self.raw.get('generation')
```
ccwienk, line 61: nit: while you are at it, you might add `int(...)` around the return to validate it really _is_ an int

PR #894, opened 2023-09-25
Comment problems found during issue update

AndreasBurger:


github/compliance/issue.py
```diff
@@ -226,6 +229,44 @@ def _update_issue(
...
+    if message:
+        job_url = concourse.util.own_running_build_url()
+        comment_body = f'{issue_problems_comment_header}\n\n{message}\nJob url: {job_url}'
+        if not problems_comment:
+            # Create new comment containing the message
+            issue.create_comment(comment_body)
+        else:
+            # Update comment
+            problems_comment.edit(comment_body)
+    else:
```
ccwienk, line 261: so this same function will remove an existing comment, if it is called w/o a message to report? hmmm..

this (cleanup) will have the downside of us having to issue at least one additional gh-api-request for each issue, will it not? maybe we should rather omit the cleanup functionality, and accept outdated comments to be persistent.

we might add (upon successful issue-updates) a timestamp indicating the date of last update. which would be enough for users to understand an error-indicating comment to be outdated.

ccwienk (changes requested):

ccwienk:
+ fix linter-error(s), please

PR #904, opened 2023-10-10
Delivery client enhancements

8R0WNI3:

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/delivery/client.py
+++ b/delivery/client.py
@@ -97,16 +97,19 @@ class DeliveryServiceClient:
         self,
         name: str,
         version: str,
-        ctx_repo_url: str,
+        ctx_repo_url: str=None,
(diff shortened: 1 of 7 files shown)


delivery/client.py
```diff
@@ -111,28 +121,32 @@ def component_descriptor(
 
     def upload_metadata(
         self,
-        data: dso.model.ArtefactMetadata,
+        data: typing.Iterable[dso.model.ArtefactMetadata],
```
ccwienk, line 124: stylistic nit: you might have declared data as an "arbitrary positional argument" (`*data`). that way, the API might be slightly more convenient to use

ccwienk (approved):

PR #905, opened 2023-10-17
Remove OcmLookupMapping

AndreasBurger:
Removes OcmLookupMapping and replaces its usage with its OCM-equivalent, in a mostly backwards-compatible fashion.

### DIFF
--- a/cnudie/util.py
+++ b/cnudie/util.py
@@ -11,7 +11,7 @@ import gci.componentmodel as cm
 import model.container_registry
 import oci.model as om
 
-ComponentId = cm.Component, cm.ComponentDescriptor | cm.ComponentIdentity | str | tuple[str, str]
+ComponentId = cm.Component | cm.ComponentDescriptor | cm.ComponentIdentity | str | tuple[str, str]
 
 
 def to_component_id(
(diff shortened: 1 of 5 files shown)


cnudie/util.py
```diff
@@ -536,18 +536,23 @@ def enumerate_group_pairs(
...
 @dataclasses.dataclass
 class OcmSoftwareResolverConfig:
-    repository: OcmRepositoryConfig
+    repository: cm.OciRepositoryContext | str
     prefix: str
-    priority: int
+    priority: int = 10
+
+    def matches(self, component):
+        if isinstance(component, cm.Component):
```
ccwienk, line 546: ideally use `to_component` (which also accepts e.g. ComponentDescriptor, or just a single str, or a two-tuple (name, version)

cnudie/util.py
```diff
@@ -587,24 +592,9 @@ class OcmGenericConfig:
...
-        elif isinstance(component, cm.ComponentIdentity):
-            component = component.name
-
-        return component.startswith(self.prefix)
-
-
 @dataclasses.dataclass
 class OcmLookupMappingConfig:
-    mappings: list[OcmLookupMapping]
+    mappings: list[OcmSoftwareResolverConfig]
```
ccwienk, line 597: can we rename this to just OcmResolverConfig?

ccwienk (changes requested):
what do you mean by "almost backwards-compatible"?

cnudie/util.py
```diff
@@ -537,22 +537,24 @@ def enumerate_group_pairs(
...
-    subpath: str | None = None
-    type: str = 'OCIRegistry'
+class OcmResolverConfig:
+    repository: cm.OciRepositoryContext | str
+    prefix: str
+    priority: int = 10
 
+    def matches(self, component: cm.Component | cm.ComponentDescriptor | cm.ComponentIdentity):
+        if not isinstance(component, cm.Component):
+            component = to_component(component)
```
AndreasBurger, line 547: use to_component_id()

Issue #226, opened 2019-02-28
Add built docker images to release notes


rfranzke:
The pipeline definition describes which Docker images shall be build, e.g. https://github.com/gardener/machine-controller-manager/blob/master/.ci/pipeline_definitions#L11-L19.

For each of these Docker images, can we please have the following section at the end of the release notes:

## Docker Images
machine-controller-manager: `eu.gcr.io/gardener-project/gardener/machine-controller-manager:0.3.0`

ccwienk:
good idea. I shall add this to our backlog discussion

rfranzke:
ping

rfranzke:
ping2

rfranzke:
@petersutter can you help with this?

petersutter:
@rfranzke I understood there is a belief that the release notes handling is broken and that there is no point to add any feature to it. 
As there seems to be a concrete idea on how to improve the release notes handling in general I assume this feature will be picked up then.

vlerenc:
/remove lifecycle/rotten

@ccwienk Can this be done?

ccwienk:
related pull-request: https://github.com/gardener/cc-utils/issues/226

rfranzke:
I can't believe it... 😳😲🫨

AndreasBurger:
Todays gardener-releases seem to include the newly generated resource-release-notes and look good to me.
@rfranzke any further wishes/remarks? If not I'll close this issue

rfranzke:
Love it. Minor ask: Maybe sort this alphabetically?

AndreasBurger:
Will do.

AndreasBurger:
Release note categories (of which there currently are 1, `Docker Image`) and their entries will each be sorted with our next release, keeping both the categories and the entries in a stable order across releases while also making the lookup of entries for larger lists a bit more convenient.

Closing this issue.

PR #906, opened 2023-10-17
Add (some of) component's resources to release-note

AndreasBurger:
With this PR, additional information about a components resources will be added to the release-notes upon release. The format is identical to the way the gardener colleagues set it by hand.

- Currently only set up for OciImages with OciAccess and will return/add nothing for other resource- and access-types.
- Categories are simply sorted alphabetically.
- Will not be added to draft-releases.

### DIFF
--- a/release_notes/markdown.py
+++ b/release_notes/markdown.py
@@ -4,6 +4,8 @@ import logging
 import typing
 from collections import defaultdict
 
+import cnudie.iter as cni
+import gci.componentmodel as cm
 import release_notes.model as rnm
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


ccwienk:
related to https://github.com/gardener/cc-utils/issues/226

release_notes/markdown.py
```diff
@@ -178,3 +180,56 @@ def render(notes: set[rnm.ReleaseNote]):
...
+        ):
+            title = resource_type_titles[resource_type]
+            release_note_line = f'{resource.name}: `{url}`'
+            return (
+                title,
+                release_note_line
+            )
+
+    lines_by_title = defaultdict(list)
+    for resource_node in cni.iter_resources(component=component, recursion_depth=0):
```
ccwienk, line 223: if there is a recursion-depth of zero, you might just as well:

```
for resource in component.resources:
 ...
```

release_notes/markdown.py
```diff
@@ -178,3 +180,56 @@ def render(notes: set[rnm.ReleaseNote]):
                     message = note.source_block.note_message
                     objs.append(list_item_from_note(message, note, group))
     return objs
+
+
+def component_resources(
```
ccwienk, line 185: factor-out conversion of resource (actually it would be sufficient to accept _artefact_, considering we only care about type + access-type), like so:

```
def release_notes_for_ocm_component(component):
   "collect release note string for data gathered from ocm-component - including artefacts"


def release_notes_for_ocm_artefact(artefact):
  "check if type + access-type are supported, return None if not"
  "create release-notes-markdown-fragment for artefact and return it"
```

ccwienk (changes requested):

PR #907, opened 2023-10-18
Improve release-note fetching behaviour

AndreasBurger:
This PR reverts the previous change that introduced hard limits on the amount of commits we process when creating release-notes and introduces throttling behaviour instead. The release-notes-coding will now process commits in batches of 200 when the github-api is involved, with pauses in between, which should not impact most release-jobs.
Together with these changes, the timeout for `draft-release` and `release` steps is also increased to two hours.

### DIFF
--- a/release_notes/utils.py
+++ b/release_notes/utils.py
@@ -1,5 +1,6 @@
 import collections
 import dataclasses
+import itertools
 import logging
 import time
 import typing
@@ -15,6 +16,7 @@ import yaml
 import yaml.scanner
 
 import ccc
+import ccc.github
 import gci.componentmodel
 import github.util
 import gitutil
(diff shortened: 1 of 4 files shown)


release_notes/utils.py
```diff
@@ -189,35 +205,55 @@ def request_pull_requests_from_api(
...
+                    continue
+                data = dataclasses.asdict(
+                    rnm.ReleaseNotesMetadata(round(time.time() * 1000), [z.number for z in prs])
+                )
+                meta = rnm.get_meta_obj(_meta_key, data)
+                _upsert_document(yaml_documents, _meta_key, meta)
+                git_helper.add_note(body=yaml.safe_dump_all(yaml_documents), commit=commit)
+
+        end_time = time.time()
+        time_elapsed = end_time - start_time # in seconds
```
AndreasBurger, line 249: fetch new github-user to equally spread the load

PR #912, opened 2023-10-27
Fix bdba scans

8R0WNI3:

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/protecode/scanning.py
+++ b/protecode/scanning.py
@@ -1,3 +1,4 @@
+import collections
 import concurrent.futures
 import logging
 import typing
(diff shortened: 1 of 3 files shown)


protecode/scanning.py
```diff
@@ -111,80 +111,101 @@ def iter_vulnerabilities_with_assessments(
...
+                yield pm.ScanRequest(
+                    component=c,
+                    artefact=r,
+                    scan_content=iter_content(),
+                    display_name=display_name,
+                    target_product_id=target_product_id,
+                    custom_metadata=component_artifact_metadata,
+                )
         elif (
             isinstance(r.type, str)
```
ccwienk, line 171: nit (not actually part of this pull-request): why check for type - we might immediately check for equality w/ str-literal

protecode/scanning.py
```diff
@@ -464,53 +490,108 @@ def result_matches(resource: cm.Resource, result: pm.AnalysisResult):
...
+        # groups are assumed to belong to same component + resource version name, and type so it
+        # is okay to choose first one as representative
+        c = resource_group[0].component
+        r = resource_group[0].resource
+
+        # special-handling for vm-image-rootfs - we keep different component-versions in parallel
+        # for those
+        if (
+            isinstance(r.type, str)
+            and r.type.startswith('application/tar+vm-image-rootfs')
```
ccwienk, line 512: ditto (also: in previous occurrence, str-equality is checked; in this place, we check for startswith; -> should  be consistent)

protecode/scanning.py
```diff
@@ -464,53 +490,108 @@ def result_matches(resource: cm.Resource, result: pm.AnalysisResult):
...
+        group_id = _resource_group_id((resource_node,))
+        resource_groups[group_id].append(resource_node)
+
+    logger.info(f'{len(resource_groups)=}')
+    yield from (tuple(g) for g in resource_groups.values())
+
+
+# TODO: Hacky cache. See _retrieve_existing_scan_results
+@functools.lru_cache
+def _proxy_list_apps(
```
ccwienk, line 582: why not use an explicit cache (or, as you suggested: integrate caching in bdba-client)?

protecode/scanning.py
```diff
@@ -320,103 +341,108 @@ def process(
...
+                        status=pm.UploadStatus.DONE,
+                        result=scan_result,
+                        state=state,
+                        affected_package=affected_package,
+                        vulnerability=vulnerability,
+                    )
+                for license in affected_package.licenses:
+                    if license.name in seen_license_names:
+                        continue
+                    seen_license_names.add(license.name)
```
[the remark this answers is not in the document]
8R0WNI3: Yes, however the check is in place to only yield a `LicenseScanResult` in case it is a license which has not been seen before.

PR #918, opened 2023-11-03
Move sprint allocation to ticket creation and properly handle edge cases for sprint milestone assignment

8R0WNI3:
During the assignment of the relevant milestone to the issue, there are edge cases, e.g. when the relevant milestone is already closed or the relevant sprint ended in the past.

Fixes #

**Release note**:

```other operator

```


github/compliance/milestone.py
```diff
@@ -3,26 +3,51 @@
...
-    sprint: delivery.model.Sprint,
-):
-    for ms in repo.milestones():
-        if ms.title == _milestone_title(sprint=sprint):
-            return ms
-    return None
+    sprints: tuple[delivery.model.Sprint],
+) -> tuple[delivery.model.Sprint, github3.repos.repo.milestone.Milestone]:
+    sprint_milestone = None
+    all_milestones = repo.milestones(state='all')
```
zkdev, line 15: does `repo.milestones()` perform an HTTP request?
if so, we could consider some caching (or passing in as for sprints) to reduce HTTP load

github/compliance/model.py
```diff
@@ -119,15 +119,17 @@ def calculate_latest_processing_date(
...
+                    sprints_count=2,
                 )
                 target_milestone = gcr._target_milestone(
                     repo=repository,
-                    sprint=target_sprint,
+                    sprints=target_sprints,
                 )
-                date = target_milestone.due_on.date()
+                if target_milestone:
+                    date = target_milestone.due_on.date()
```
zkdev, line 132: Please add return type-hint to function, as date can also be `None`

github/compliance/report.py
```diff
@@ -627,39 +627,77 @@ def _scanned_element_ctx_label(
...
+
+    if latest_processing_date:
+        date = latest_processing_date
+        offset = 1 # find the sprint that ends before the specified date
+    else:
+        date = sprint_end_date
+        offset = 0 # find the sprint that includes the specified date
+
+    def get_sprints_starting_from(
+        starting_from: int,
```
zkdev, line 679: I prefer to rather accept a date (than an int offset) to specify the starting point.

github/compliance/issue.py
```diff
@@ -219,6 +219,12 @@ def _create_issue(
 
             issue.create_comment(comment_body)
 
+        if not milestone:
+            issue.create_comment(
+                'Unable to automatically assign ticket to milestone. ' +
+                'Relevant milestone(s) were probably closed before their due date.'
```
zkdev, line 225: Would be helpful to (also) print which milestone (candidate) was determined.

zkdev (changes requested):

github/compliance/milestone.py
ccwienk: from how github3.py behaves elsewhere, I think this should return an iterator that will issue an http-request upon consuming results. rather than caching (which would be problematic in this case, I think), we might consider limiting the amount of milestones + order by date. for the time being, we have some few tens of milestones (rate of growth: ~30-ish / year). -> nothing to worry about short-term..

github/compliance/model.py
```diff
@@ -119,15 +119,17 @@ def calculate_latest_processing_date(
         if delivery_svc_client and repository and max_days > 0:
             try:
                 import github.compliance.report as gcr
```
zkdev, line 121: Even though not introduced with this PR, why late import?
If it is just to avoid circular imports, I would rather consider to factor out this function to another module.

zkdev: A good pattern to avoid circular imports is to have `model` classes not dependent on other modules. Same goes for `util` modules by the way.
These should be importable in any location.

github/compliance/report.py
```diff
@@ -627,39 +627,85 @@ def _scanned_element_ctx_label(
...
+    for idx, sprint in enumerate(sprints):
+        if len(targets_sprints) == sprints_count:
+            # found enough sprints -> early exiting
+            break
+
+        end_date = sprint.find_sprint_date(name='end_date').value.date()
+        if end_date >= date:
+            if idx + offset == -1: # compare to "-1" instead of "<0" to print warning _once_
+                logger.warning(
+                    f'did not found not ended sprints starting from {date} with an offset '
```
zkdev, line 693: typo: did not find

github/compliance/report.py
```diff
@@ -627,39 +627,85 @@ def _scanned_element_ctx_label(
...
+                )
+            elif idx + offset >= len(sprints):
+                # index + offset keep being out of bounds -> early exiting
+                break
+            else:
+                targets_sprints.append(sprints[idx + offset])
+
+    if len(targets_sprints) < sprints_count:
+        logger.warning(
+            f'did not found {sprints_count} sprints starting from {date} with ' +
```
zkdev, line 704: same here

github/compliance/report.py
```diff
@@ -627,39 +627,85 @@ def _scanned_element_ctx_label(
...
     return github.compliance.milestone.find_or_create_sprint_milestone(
         repo=repo,
-        sprint=sprint,
+        sprints=sprints,
     )
 
 
 @functools.cache
-def _target_sprint(
+def _upcoming_sprints(
```
zkdev, line 642: good idea!
that way we must not deal with filtering later

github/compliance/report.py
```diff
@@ -627,39 +627,85 @@ def _scanned_element_ctx_label(
...
-            )
-        elif sprint_end_date:
-            target_sprint = delivery_svc_client.sprint_current(
-                before=sprint_end_date,
-            )
-    except requests.HTTPError as http_error:
-        logger.warning(f'error determining tgt-sprint {http_error=} - falling back to current')
-        target_sprint = delivery_svc_client.sprint_current()
+    today = datetime.date.today()
+    sprints = _upcoming_sprints(
```
zkdev, line 669: As we rely on sprints being ordered by date, I would prefer to either explicitly order here (again), or at least add a comment / choose a variable name expressing that.

zkdev (changes requested):

github/compliance/model.py
8R0WNI3: yep, was to avoid circular imports. Will add a commit which moves affected functions to a different module

github/compliance/report.py
8R0WNI3: The ordering of sprints is done a few lines below. The function `_upcoming_sprints` does not do any ordering.

zkdev: got it!

zkdev (approved):
/lgtm

PR #924, opened 2023-11-23
Remove bdba scan from pipeline template

8R0WNI3:
As a restructuring, the compliance scanning tools shall be moved to standalone kubernetes deployments instead of running as a pipeline job. These kubernetes deployments will be part of the so-called delivery-gear.
As a first step, the BDBA compliance scan for vulnerabilities and licenses as well as the replication of the compliance-tracking-issues were included in the delivery-gear recently. So, this pipeline template is not required anymore.

Fixes #

**Release note**:

```other operator

```


ccwienk:
@8R0WNI3 which usages are missing?

8R0WNI3:
@ccwienk the landscaper-live repository still has an open PR to remove bdba from the scan job

ccwienk (changes requested):
two remarks:
- should also include concourse/steps/scan_container_images.py
- this will also remove malware-scanning, will it not? -> we need to migrate malware-scanning into a standalone extension, too (prior to removing from template)

PR #926, opened 2023-12-05
Automatically set version_interface to 'callback' when callbacks are set

voelzmo:
Currently the `version_interface` still uses its default of `file`, even if you specify callbacks in your pipeline. You'd need to set it specifically to `version_interface: callback`, otherwise the callbacks would be ignored and the `VERSION` file is taken.
With this PR, `version_interface` will be automatically changed if callbacks are specified. Additionally, we introduce a validation that both callbacks are set – only being able to read but not write a version is probably not what we want?

**Release note**:

```other operator
Set `version_interface: callback` automatically when callback scripts are defined. No need to do this manually in your pipelines anymore.
```

### DIFF
--- a/concourse/steps/version.mako
+++ b/concourse/steps/version.mako
@@ -4,6 +4,7 @@ import os
 from makoutil import indent_func
 from concourse.steps import step_lib
 from concourse.model.traits.release import TagConflictAction
+from concourse.model.traits.version import VersionInterface
 
 main_repo = job_variant.main_repository()
 head_sha_file = main_repo.head_sha_path()
@@ -21,12 +22,21 @@ path_to_repo_version_file = os.path.join(
 output_version_file = os.path.join(job_step.output('version_path'), 'version')
 legacy_version_file = os.path.join(job_step.output('version_path'), 'number')
 
-if (read_callback := version_trait.read_callback()):
+# Assign empty string to calbacks if None, as we'd template 'None' (the string) later otherwise
+if (read_callback := version_trait.read_callback() or ''):
   read_callback = os.path.join(main_repo.resource_name(), read_callback)
 
-# Assign empty string if None, as we'd template 'None' (the string) later otherwise
+
 if (write_callback := version_trait.write_callback() or ''):
   write_callback = os.path.join(main_repo.resource_name(), write_callback)
+
+# Validate that either both callbacks are set or none
+if (write_callback == '' and read_callback != '') or (write_callback != '' and read_callback == ''):
+  raise ValueError(f"write_callback is '{write_callback}' and read_callback is '{read_callback}'. Either set both callbacks or none!")
+
+# Set version_interface to 'callback' if write_callback and read_callback are set
+if (write_callback != '' and read_callback != ''):
+  version_trait.raw['version_interface']='callback'
 
 version_operation = version_trait.preprocess
 branch_name = main_repo.branch()
--- a/concourse/steps/version.mako
+++ b/concourse/steps/version.mako
@@ -4,7 +4,6 @@ import os
 from makoutil import indent_func
 from concourse.steps import step_lib
 from concourse.model.traits.release import TagConflictAction
-from concourse.model.traits.version import VersionInterface
 
 main_repo = job_variant.main_repository()
 head_sha_file = main_repo.head_sha_path()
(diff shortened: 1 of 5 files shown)


concourse/steps/version.mako
```diff
@@ -21,13 +22,22 @@ path_to_repo_version_file = os.path.join(
 output_version_file = os.path.join(job_step.output('version_path'), 'version')
 legacy_version_file = os.path.join(job_step.output('version_path'), 'number')
 
-if (read_callback := version_trait.read_callback()):
+# Assign empty string to calbacks if None, as we'd template 'None' (the string) later otherwise
+if (read_callback := version_trait.read_callback() or ''):
```
ccwienk, line 26: you might initialise `read_callback` w/ empty str as initial value; otoh: why actually default to empty str? None should be okay as a marker for "not set"

concourse/steps/version.mako
```diff
@@ -21,13 +22,22 @@ path_to_repo_version_file = os.path.join(
...
   read_callback = os.path.join(main_repo.resource_name(), read_callback)
 
-# Assign empty string if None, as we'd template 'None' (the string) later otherwise
+
 if (write_callback := version_trait.write_callback() or ''):
   write_callback = os.path.join(main_repo.resource_name(), write_callback)
 
+# Validate that either both callbacks are set or none
+if (write_callback == '' and read_callback != '') or (write_callback != '' and read_callback == ''):
```
ccwienk, line 34: I think it would be more "pythonic" to check this leveraging coercion to bool, like so:
`if write_callback and (not read_callback) or (not write_callback) and read_callback`

You might also use XOR (not specifically pythonic), like so (I personally prefer this style as it is shorter/more expressive):
`if bool(write_callback) ^ bool(read_callback)`

On an additional note: how about moving this check to `concourse.model.traits.version.py` (define a new method `validate` - this is automatically called if it exists. [example](https://github.com/gardener/cc-utils/blob/master/concourse/model/traits/component_descriptor.py#L435-L447)?

concourse/steps/version.mako
```diff
@@ -21,13 +22,22 @@ path_to_repo_version_file = os.path.join(
...
+  raise ValueError(f"write_callback is '{write_callback}' and read_callback is '{read_callback}'. Either set both callbacks or none!")
+
+# Set version_interface to 'callback' if write_callback and read_callback are set
+if (write_callback != '' and read_callback != ''):
+  version_trait.raw['version_interface']='callback'
```
ccwienk, line 39: similar to validation, I think it would be nicer to move this code into `concourse.model.traits.version`. More specifically into the "TraitProcessor" (an awkwardly-named hook that is called after parsing pipeline-definitions, but before executing the pipeline-rendering:

https://github.com/gardener/cc-utils/blob/master/concourse/model/traits/version.py#L127

I think processing might be done in an initialiser (would need to be defined). [example](https://github.com/gardener/cc-utils/blob/master/concourse/model/traits/component_descriptor.py#L457-L459). It might also be done earlier (e.g. in the trait's initialiser)

ccwienk (changes requested):

concourse/steps/version.mako
ccwienk: also (nit): pep8 suggests there should be whitespace (blank) around assignment operator in this case (`] = 'call..`)

concourse/steps/version.mako
voelzmo: I only introduced the same check that was already existing for the `write_callback` also for the `read_callback`. Do you want me to change how this is done for both? My understanding was that we're specifically trying to avoid having the literal string `None` in the rendered pipeline.

concourse/steps/version.mako
voelzmo: done. I decided against the XOR (unless you really insist on it) as, at least for me, it is less readable. I rarely encounter this in any code, so I'm trying to avoid it, if possible.

concourse/steps/version.mako
voelzmo: done.

concourse/steps/version.mako
ccwienk: Of course, it would be nicer to not render the str-literal `None` into the template. However, I think this might be done even better by not rendering that part at all by guarding the line(s) in question with a
```
% if <check for callback>
# do stuff with callback
% endif
```

That way, there would also not be a str-literal `None` rendered into the code

voelzmo: I think I understand what you want this to look like, but couldn't do it easily without re-structuring `version.mako` quite a bit. 
* There's quite a few assumptions that e.g. `write_callback` always exists as a variable
* It feels weird to me to access the callbacks based on `version_interface: callback` in `version.mako`, but you cannot see why it could never happen that this is set, but the callback variables are undefined

So for this PR, I wouldn't include that change, if that's ok with you?

PR #927, opened 2023-12-08
Add new uploader for images that were not filtered

AndreasBurger:
We've recently changed the way images are referenced in our Component Descriptors after processing them with CTT. This causes issues further downstream, as there is currently a codepath that modifies image-references during transport to `latest` for images that are sha-digest-tagged.
This change adds another uploader that will not change the tags it encounters, but will otherwise behave the same. 

Not yet tested.


AndreasBurger:
Closing as we're solving it differently

PR #928, opened 2023-12-10
Add optional cross-rotation for AWS and GCP

zkdev:
Adds optional "cross-rotation" (rotate User A with credentials of user B) for GCP and AWS ServiceAccounts.
Can be opt-in by adding attribute `rotation_cfg_name`.
AWS also requires to specify ServiceAccount name via attribute `iam_user_name`.

If `rotation_cfg_name` is not specified the regular "self-rotation" is performed.

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/cfg_mgmt/aws.py
+++ b/cfg_mgmt/aws.py
@@ -8,9 +8,11 @@ import boto3
 import dacite
 
 import cfg_mgmt
+import cfg_mgmt.util as cmu
 import ci.log
 import model
 import model.aws
+import model.base
 
 from cfg_mgmt.model import (
     CfgQueueEntry,
@@ -51,21 +53,47 @@ class ListAccessKeysResponse:
     Marker: typing.Optional[str] # only present when IsTruncated is True
 
 
+def rotation_cfg_or_none(
+    aws_cfg: model.aws.AwsProfile,
+    cfg_factory: model.ConfigFactory,
+):
+    if (rotation_cfg_reference := aws_cfg.rotation_cfg()):
+        return cfg_factory.resolve_cfg_element_reference(
+            cfg_element_reference=rotation_cfg_reference,
+        )
+
+    return None
+
+
 def rotate_cfg_element(
     cfg_element: model.aws.AwsProfile,
     cfg_factory: model.ConfigFactory,
 ) ->  typing.Tuple[cfg_mgmt.revert_function, dict, model.NamedModelElement]:
+    rotation_cfg = rotation_cfg_or_none(
+        aws_cfg=cfg_element,
+        cfg_factory=cfg_factory,
+    ) or cfg_element
+
+    logger.info(f'using {rotation_cfg.name()=} for rotation')
+
+    iam_user_name = cfg_element.iam_user_name()
 
     iam_client = boto3.client(
         'iam',
-        aws_access_key_id=cfg_element.access_key_id(),
-        aws_secret_access_key=cfg_element.secret_access_key(),
-        region_name=cfg_element.region(),
+        aws_access_key_id=rotation_cfg.access_key_id(),
+        aws_secret_access_key=rotation_cfg.secret_access_key(),
+        region_name=rotation_cfg.region(),
     )
+
+    if iam_user_name:
+        data = iam_client.create_access_key(UserName=iam_user_name)
+
+    else:
+        data = iam_client.create_access_key()
 
     response: CreateAccessKeyResponse = dacite.from_dict(
         data_class=CreateAccessKeyResponse,
-        data=iam_client.create_access_key(),
+        data=data,
     )
     access_key = response.AccessKey
 
@@ -77,7 +105,13 @@ def rotate_cfg_element(
     new_element.raw['secret_access_key'] = access_key.SecretAccessKey
 
     def revert_function():
-        iam_client.delete_access_key(AccessKeyId=access_key.AccessKeyId)
+        if iam_user_name:
+            iam_client.delete_access_key(
+                AccessKeyId=access_key.AccessKeyId,
+                UserName=iam_user_name,
+            )
+        else:
+            iam_client.delete_access_key(AccessKeyId=access_key.AccessKeyId)
 
     secret_id = {'accessKeyId': cfg_element.access_key_id()}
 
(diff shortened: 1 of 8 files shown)


zkdev:
not yet tested E2E -> do not merge (yet)

zkdev:
@ccwienk @8R0WNI3 adjusted as discussed, ready for review

cfg_mgmt/aws.py
```diff
@@ -89,10 +95,16 @@ def delete_config_secret(
     cfg_factory: model.ConfigFactory,
     cfg_queue_entry: CfgQueueEntry,
 ) -> model.aws.AwsProfile | None:
+    rotation_cfg = cmu.rotation_cfg_or_none(
+        cfg_element=cfg_element,
+        cfg_factory=cfg_factory,
+    ) or cfg_element
+    logger.info(f'using {rotation_cfg.name()=} for rotation')
```
8R0WNI3, line 102: maybe rather `... for deletion`?

cfg_mgmt/gcp.py
```diff
@@ -115,9 +122,15 @@ def delete_config_secret(
     cfg_queue_entry: cmm.CfgQueueEntry,
     cfg_factory: model.ConfigFactory,
 ) -> GcpServiceAccount | None:
+    rotation_cfg = cmu.rotation_cfg_or_none(
+        cfg_element=cfg_element,
+        cfg_factory=cfg_factory,
+    ) or cfg_element
+    logger.info(f'using {rotation_cfg.name()=} for rotation')
```
8R0WNI3, line 129: same here

8R0WNI3 (approved):
/lgtm

zkdev:
@8R0WNI3 [flagged you as maintainer](https://github.com/gardener/cc-utils/commit/55016dd262e4e31be28445f283435d18b1cee9f2)

cfg_mgmt/aws.py
zkdev: copy-pasta ... , will adjust, thanks!

model/aws.py
```diff
@@ -28,5 +28,11 @@ def access_key_id(self):
     def secret_access_key(self):
         return self.raw['secret_access_key']
 
+    def rotation_cfg_name(self):
+        return self.raw.get('rotation_cfg_name')
+
+    def iam_user_name(self):
```
ccwienk, line 34: considering presense of `rotation_cfg_name` implies `iam_user_name` must also be set, how about adding this into validation. there are two options:

1. calculate required attributes in `_required_attributes` method
2. validate in `validate` method (see e.g. model/concourse.py for an example)

model/container_registry.py
```diff
@@ -81,6 +82,9 @@ def registry_type(self):
     def api_base_url(self):
         return self.raw.get('api_base_url')
 
+    def rotation_cfg_name(self):
+        return self.raw.get('rotation_cfg_name')
```
ccwienk, line 86: check for registry-type. if it is not GCR/AR, setting this attribute should at least emit a warning (or rather fail). Also, considering the cfg-name is of a different type (-> GCP I assume?), the type should also be configured, rather than being implied (implying type seems okay-ish if "rotation-cfg-element" is of same type as to-be-rotated element).

otoh: how about defining (-> model/base.py) a new type for rotation-cfg. it could be as simple as:
```
@dataclasses.dataclass
class CfgElementReference:
  type_name: str
  element_name: str
  purpose: str | None = None # might be useful for documenting in-place what the cfg-element-reference is used for
```

ccwienk (changes requested):

model/aws.py
zkdev: Will add to required attributes, however we must ensure to explicitly validate in `aws.py`, as secret-replication handles validation-errors with warnings.

model/container_registry.py
zkdev: This is a bit fuzzy as we have the generic cfg-type "container_registry" with its subclassification via "registry_type", but also the dedicated cfg-type "gcp".
However, both cfg-types use the very same rotation coding.

Therefore, if a cfg-element of type "container_registry" specifies a "rotation_cfg_name", it is implied that the referenced cfg is also of type "container_registry"
-> no need for explicit cfg-type specification

zkdev: Thus, the only remaining value in introducing a `CfgElementReference` class would be the purpose attribute, however one could also simply inline that as Yaml comment.

zkdev: > check for registry-type. if it is not GCR/AR, setting this attribute should at least emit a warning (or rather fail). 
good point, will add to validation

cfg_mgmt/util.py
```diff
@@ -424,3 +425,15 @@ def determine_status(
...
         responsible=responsible,
     )
+
+
+def rotation_cfg_or_none(
+    cfg_element,
+    cfg_factory: model.ConfigFactory,
+    rotation_cfg_reference_attribute_name: str='rotation_cfg_reference',
+):
+    if (rotation_cfg_reference := getattr(cfg_element, rotation_cfg_reference_attribute_name)()):
```
ccwienk, line 435: rather make this type-specific and explicit, please.

model/aws.py
```diff
@@ -28,5 +28,25 @@ def access_key_id(self):
     def secret_access_key(self):
         return self.raw['secret_access_key']
 
+    def rotation_cfg_reference(self) -> model.base.CfgElementReference:
```
ccwienk, line 31: naming: while the typename is perfectly fine (CfgElementReference), the method's name is not ideal. it is about the cfg-element for rotation. the chosen name, however, drops the `element` part of the name (which reduces emphasis), and puts an emphesis on the aspect of the returned object being a a reference.

please rather name this:
- rotation_cfg_element
- rotation_cfg_element_reference
- rotation_cfg # this would be in sync with other namings of cfg-element-references we already have elsewhere

the same holds true for the attribute-name

model/aws.py
```diff
@@ -28,5 +28,25 @@ def access_key_id(self):
...
+        return self.raw.get('iam_user_name')
+
     def _required_attributes(self):
-        return ['region','access_key_id','secret_access_key']
+        static_attributes = [
+            'region',
+            'access_key_id',
+            'secret_access_key',
+        ]
+        dynamic_attributes = ['iam_user_name'] if self.rotation_cfg_reference() else []
```
ccwienk, line 50: it would be nice to document (preferably at the getters for rotation-cfg-element + iam_user_name) what those are intended for.

also, is it actually true that we always need a IAM user for credentials-rotation?

model/base.py
```diff
@@ -175,3 +177,19 @@ def token(self):
...
         return ['token']
+
+
+@dataclasses.dataclass(frozen=True)
+class CfgElementReference:
+    type_name: str
+    element_name: str
+    purpose: str | None = None
+
+    def cfg_element(
```
ccwienk, line 188: rather add this to cfg-factory as a convenience-method than here.

model/container_registry.py
```diff
@@ -81,6 +95,16 @@ def registry_type(self):
     def api_base_url(self):
         return self.raw.get('api_base_url')
 
+    def rotation_cfg_reference(self) -> model.base.CfgElementReference:
```
ccwienk, line 98: see comment to aws-cfg.

btw, if you _do_ want do deduplicate this, you might define a "CrossRotationMixin". If going for that approach, it would be fine to standardise on the name.

However, it is perfectly fine (and probably better, as we will have specific code for each element-type anyhow) to keep this type-specific. In which case, however, there should be no hardcoded assumptions as to the attribute-names.

model/gcp.py
```diff
@@ -25,5 +25,15 @@ def client_email(self) -> str:
     def private_key_id(self) -> str:
         return self.service_account_key()['private_key_id']
 
+    def rotation_cfg_reference(self) -> model.base.CfgElementReference:
```
ccwienk, line 28: ditto

ccwienk (changes requested):

model/aws.py
zkdev: AWS cross-rotation **always** requires `iam_user_name`, yes.

As for the reasoning:
boto `iam_client` either implicitly uses its own iam-user (== self-rotation), or requires to explicitly give an user-name (== cross-rotation).
Btw, this holds also true for GCP rotation, however for GCP we [construct the user-name with ServiceAccount E-Mail and project-id](https://github.com/gardener/cc-utils/blob/3311905313346232547b52c8abebd0e24037b706/ccc/gcp.py#L72C5-L72C5).

ccwienk (approved):

PR #934, opened 2024-01-03
Remove `snapshot_ctx_repository` from component descriptor trait

8R0WNI3:

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/concourse/enumerator.py
+++ b/concourse/enumerator.py
@@ -95,6 +95,9 @@ class DefinitionEnumerator:
             target_team = self.job_mapping.team_name() # noqa
 
         for name, definition in raw_definitions.items():
+            if not definition.get('render_pipeline', True):
+                continue
+
             pipeline_definition = deepcopy(definition)
             yield DefinitionDescriptor(
                 pipeline_name=name,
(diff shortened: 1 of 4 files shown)


ccwienk (commented):
considering I recently adjusted almost all pipelines to explicitly configure ocm-repositories, how about we check all pipelines for whether there still are pipelines that actually set a "snapshot-ctx-repository-url" (my assumption: this should only be true for some few outdated hotfix-release-pipelines). if my assumption is correct, we might entirely remove this snapshot-special-case (and thus further simplify component-descriptor-step's code

concourse/enumerator.py
```diff
@@ -354,6 +354,10 @@ def _scan_repository_for_definitions(
             # handle inheritance
             definitions = merge_dicts(definitions, override_definitions)
 
+            for pipeline_name in tuple(definitions.keys()):
```
ccwienk, line 357: ~~why create a tuple from `definitions.keys()`?~~

instead of working around issues from manipulating iterated-over dict, why not create a new one, like so:

```
definitions = {
  name: definition for name, definition in definitions.items()
  if definition.get('render_pipeline', True)
}
```

wdyt? not 100% sure this variant is easier to read

ccwienk: as an alternative, why not add a simple `continue` into  `_wrap_into_descriptors`?

8R0WNI3: yep the latter sounds good to me 👍🏼

ccwienk (approved):

PR #935, opened 2024-01-05
Add option to delete old BDBA products

8R0WNI3:
Using this option, an interval X can be set after which "old" products (i.e. products which were not (re-)scanned for the last X period) are going to be deleted. Hence, this interval must not be set smaller than the specified rescan interval since then the products are going to be deleted continuously. Also, products belonging to resources which are not scanned anymore (in no version) are not going to be deleted.

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/protecode/scanning.py
+++ b/protecode/scanning.py
@@ -1,8 +1,10 @@
 import collections
 import concurrent.futures
+import datetime
+import dateutil.parser
 import functools
 import logging
-import random
+import pytz
 import typing
 
 import botocore.exceptions
@@ -117,11 +119,21 @@ class ResourceGroupProcessor:
             ))
             yield from products
 
-    def iter_components_with_vulnerabilities_and_assessments(
+    def iter_products(
         self,
         products_to_import_from: list[pm.Product],
         use_product_cache: bool=True,
+        delete_inactive_products_after_seconds: int=None,
     ) -> typing.Generator[tuple[pm.Component, pm.Vulnerability, tuple[pm.Triage]], None, None]:
+        '''
+        Used to retrieve the triages of the supplied products grouped by components and
+        their vulnerabilities. Also, if `delete_inactive_products_after` is set, old
+        bdba products will be deleted according to it.
+        Note: `delete_inactive_products_after` must be greater than the interval in which
+        the resources are set or otherwise the products are going to be deleted immediately.
+        Also, old products of resources which are not scanned anymore at all (meaning in no
+        version) are _not_ going to be deleted.
+        '''
         def _iter_vulnerabilities(
             result: pm.AnalysisResult,
         ) -> typing.Generator[tuple[pm.Component, pm.Vulnerability], None, None]:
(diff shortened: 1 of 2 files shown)


protecode/scanning.py
```diff
@@ -121,6 +121,7 @@ def iter_components_with_vulnerabilities_and_assessments(
         self,
         products_to_import_from: list[pm.Product],
         use_product_cache: bool=True,
+        delete_inactive_products_after: int=0,
```
ccwienk, line 124: codestyle: add unit as suffix (`_seconds`) . It would probably be nice to add a short docstr to the function where the parameters are briefly explained (specifically the cleanup-logic).

as for the default-value, rather use either `-1` or `None` (`0` might be a value a caller might want to set (generally speaking - _we_ do not plan to pass-in `0`, of course; if passing-in `0`, this would mean "delete immediately". which is kind of reasonable)

protecode/scanning.py
```diff
@@ -137,20 +138,24 @@ def iter_vulnerabilities_with_assessments(
...
-        )
-        products_to_import_from = products_to_import_from[:50]
-
-        # do a shuffle to increase cache hits as otherwise same product ids are
-        # more likely to be handled simultaneously among the different threads
-        random.shuffle(products_to_import_from)
-
         for product in products_to_import_from:
+            if delete_inactive_products_after:
+                if deletion_after := product.custom_data().get('DELETION_AFTER'):
```
ccwienk, line 143: naming-nit: how about calling this `delete_after` | `DELETE_AFTER`? to me, this sounds a bit more obvious (considering this is what the code will do (delete)) :-)

protecode/scanning.py
```diff
@@ -137,20 +138,24 @@ def iter_vulnerabilities_with_assessments(
...
-            reverse=True,
-        )
-        products_to_import_from = products_to_import_from[:50]
-
-        # do a shuffle to increase cache hits as otherwise same product ids are
-        # more likely to be handled simultaneously among the different threads
-        random.shuffle(products_to_import_from)
-
         for product in products_to_import_from:
+            if delete_inactive_products_after:
```
ccwienk, line 142: I would structure this slightly different (also to maybe reduce nesting):

- calculate delete_after-timestamp outside of (before, obviously..) for-loop + fixate "now" outside of for-loop (rationale: if processing will take long, this way, there will be consistent (and even identical) timestamps; too my experience, this often has benefits)
- first add deletion-date (if not already present), then act upon deletion-date (if present); rationale: this way, passing-in of "delete after 0s" will work without special-case handling

protecode/scanning.py
```diff
@@ -137,20 +138,24 @@ def iter_vulnerabilities_with_assessments(
...
+                        logger.info(f'deleted old bdba product {product.product_id()}')
+                        continue
+                else:
+                    deletion_after = datetime.datetime.now() + datetime.timedelta(
+                        seconds=delete_inactive_products_after,
+                    )
+                    self.protecode_client.set_metadata(
+                        product_id=product.product_id(),
+                        custom_attribs={
+                            'DELETION_AFTER': deletion_after.isoformat(),
```
ccwienk, line 155: should we use tz-aware timestamps?

protecode/scanning.py
```diff
@@ -137,20 +138,24 @@ def iter_vulnerabilities_with_assessments(
...
-        products_to_import_from = products_to_import_from[:50]
-
-        # do a shuffle to increase cache hits as otherwise same product ids are
-        # more likely to be handled simultaneously among the different threads
-        random.shuffle(products_to_import_from)
-
         for product in products_to_import_from:
+            if delete_inactive_products_after:
+                if deletion_after := product.custom_data().get('DELETION_AFTER'):
+                    if datetime.datetime.now() > datetime.datetime.fromisoformat(deletion_after):
```
ccwienk, line 144: preferably use dateutil.parser - datetime.datetime's fromisoformat is (I think) buggy and does not properly parse all valid iso8601-timestamps (dateutil.parser is more robust and a drop-in replacement, so I recommend we use it instead of datetime.datetime.isoparse always)

protecode/scanning.py
```diff
@@ -478,6 +486,15 @@ def process(
                     product_id=scan_result.product_id(),
                 )
 
+            if delete_inactive_products_after and scan_request.target_product_id:
+                # remove deletion flag for current product as it is still in use
+                self.protecode_client.set_metadata(
+                    product_id=scan_request.target_product_id,
+                    custom_attribs={
+                        'DELETION_AFTER': '', # must be empty str and not `None`
```
ccwienk, line 494: +1 for adding a comment indicating empty str should remain.

however, two remarks:
- we might adjust `set_metadata` in such a way that the function will translate `None` -> `''` . that way, our API might maybe feel a bit more "natural" (not sure whether you agree)
- it would be nice to explain _why_ we need empty str (I assume to prevent `None` (or `null`?) (str literal) to be stored, which would cause parsing-issues (and also not semantically be what we want - which is to remove the attribute)

ccwienk (changes requested):

ccwienk (approved):

PR #936, opened 2024-01-05
Preliminaries for adding `bandit` (security-focussed linter)

ccwienk:


ccc/secrets_server.py
```diff
@@ -131,7 +131,7 @@ def retrieve_secrets(self):
                     return json.load(f)
 
         request_url = urljoin(self.url, self.concourse_secret_name)
-        response = requests.get(request_url)
+        response = requests.get(request_url, timeout=(3, 30))
```
8R0WNI3, line 134: Since [here](https://requests.readthedocs.io/en/latest/user/advanced/#timeouts:~:text=slightly%20larger%20than%20a%20multiple%20of%203) it is stated that it would be beneficial to set the timeouts "slightly larger than a multiple of 3" because of TCP transmission insides, wouldn't it make sense to set the timeouts to e.g. (4, 31)?

8R0WNI3 (changes requested):

ccc/secrets_server.py
ccwienk: sure, why not. will change it to 4/31 (although I have to admit, I also am not fully aware of the technical background of the recommendation you quoted (I saw this before creating this pull-request, too).

I _assume_, this relates to the tcp handshake (IIRC this consists of three steps). However, I would be quite surprised if each of those steps typically is assumed to take 1s. so I think `<typical max allowed time per tcp handshake step> * 3 + "a bit more"` kind of makes sense, provided there is a commonly accepted value for "typical max allowed time per tcp handshake step"

8R0WNI3 (approved):
/lgtm

PR #937, opened 2024-01-15
Oci layer blobs from local blob accesses

ccwienk:
**What this PR does / why we need it**

If publishing a component-descriptor containing local blob accesses, it is necessary to also reference those blobs as OCI Image Layers in the (synthesised) OCI Image Manifest. Do this by inspecting all accesses of a given Component Descriptor (note: this code assumes that referenced blobs were uploaded prior to invoking `upload_component_descriptor` function).

Add non-standard `size` attribute to local blob access, use it if set (size is required for OCI Image Layer entries) to avoid a http-head request in case of absent globalAccess attribute. As fallback, determine blob's size from passed OCI Image Reference.

**Special notes for your reviewer**

This change is backwards-incompatible, as it removes the `extra_layers` parameter (no longer required after this change).

### DIFF
--- a/cnudie/upload.py
+++ b/cnudie/upload.py
@@ -91,7 +91,6 @@ def upload_component_descriptor(
     on_exist:UploadMode|str=UploadMode.SKIP,
     ocm_repository: cm.OciOcmRepository | str = None,
     oci_client: oci.client.Client=None,
-    extra_layers: list[gci.oci.OciBlobRef]=[],
 ):
     if not oci_client:
         oci_client = ccc.oci.oci_client()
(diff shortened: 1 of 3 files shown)


oci/model.py
```diff
@@ -200,6 +200,30 @@ def as_dict(self) -> dict:
         raw = {k:v for k,v in raw.items() if v is not None}
         return raw
 
+    def __hash__(self):
+        annotations = tuple(self.annotations.items()) if self.annotations else ()
```
zkdev, line 204: Did you consider normalising (especially sorting) annotations?
If so, why did you decide against it?

cnudie/upload.py
```diff
@@ -48,10 +118,10 @@ def upload_component_descriptor(
         else:
             raise TypeError(type(ocm_repository))
 
-        if  not component_descriptor.component.current_repository_ctx() == ocm_repository:
-            component_descriptor.component.repositoryContexts.append(ocm_repository)
+        if  not component.current_repository_ctx() == ocm_repository:
```
zkdev, line 121: nit: two blanks between `if` and `not`

oci/model.py
ccwienk: good point. w/o sorting, I would also have to also enforce strict order-comparison in __eq__

zkdev (approved):
/lgtm

PR #941, opened 2024-01-19
Github token replication for Solinas service

MartinWeindel:
This is a follow-up of #939.
The Solinas service (SUGAR) only needs the new github token, but no rotation of a user.
Therefore instead of using the rotation functionality, it is integrated with the secrets replication to simplify it.  

Fixes #

**Release note**:

```other operator
Github token replication for Solinas service
```

### DIFF
--- a/model/config_repo.py
+++ b/model/config_repo.py
@@ -26,13 +26,14 @@ class KubernetesSecretTarget:
 @dataclasses.dataclass
 class SugarSolinasServiceTarget:
     type: str
+    sugar_config_name: str
     team_id: str
     github: str
 
 
 @dataclasses.dataclass
 class ReplicationMapping:
-    target: SecretsServerTarget | KubernetesSecretTarget
+    target: SecretsServerTarget | KubernetesSecretTarget | SugarSolinasServiceTarget
     cfg_set: str
 
 
--- a/model/config_repo.py
+++ b/model/config_repo.py
@@ -24,6 +24,13 @@ class KubernetesSecretTarget:
 
 
 @dataclasses.dataclass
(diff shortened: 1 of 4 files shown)


ccwienk (approved):

model/config_repo.py
```diff
@@ -23,6 +23,13 @@ class KubernetesSecretTarget:
...
+@dataclasses.dataclass
+class SugarSolinasServiceTarget:
+    type: str
+    team_id: str
+    github: str
+
+
 @dataclasses.dataclass
 class ReplicationMapping:
     target: SecretsServerTarget | KubernetesSecretTarget
```
zkdev, line 35: Also reference `SugarSolinasServiceTarget` for sake of deserialisation.

model/config_repo.py
```diff
@@ -23,6 +23,13 @@ class KubernetesSecretTarget:
     secret_key: str
 
 
+@dataclasses.dataclass
+class SugarSolinasServiceTarget:
+    type: str
+    team_id: str
```
zkdev, line 29: As `team_id` is used as sugar config name I suggest a more expressive name, e.g. `sugar_cfg_name`.
wdyt?

zkdev (changes requested):

model/config_repo.py
MartinWeindel: fixed

model/config_repo.py
MartinWeindel: added `sugar_config_name` field

zkdev (approved):
/lgtm

PR #943, opened 2024-01-25
[btp_application_certificate] Store intermediate certificate and send it on client authentication

MartinWeindel:
The certificate service introduced a new intermediate certificate which causes problems with client authentication like
```
requests.exceptions.SSLError: HTTPSConnectionPool(host='agw0s98rn.accounts.ondemand.com', port=443): Max retries exceeded with url: /service/sps/74e3f610-893d-4e75-86e6-f480a120a156/apiCertificate (Caused by SSLError(SSLError(1, '[SSL: TLSV1_ALERT_UNKNOWN_CA] tlsv1 alert unknown ca (_ssl.c:1006)')))
```

The issue is fixed by sending both certificate and the intermediate certificate on client authentication.

Additionally, small improvements have been included:
- use new v3 for certificate generation
- change key size to 3072

Fixes #

**Release note**:

```other operator
[btp_application_certificate] Store intermediate certificate and send it on client authentication
```

### DIFF
--- a/cfg_mgmt/btp_application_certificate.py
+++ b/cfg_mgmt/btp_application_certificate.py
@@ -55,10 +55,6 @@ class CertServiceClient:
         self.access_token = resp.json()['access_token']
 
     def create_client_certificate_chain(self, csr_pem: str, validity_in_days: int) -> dict:
-        '''
-        see https://pages.github.tools.sap/sap-pki-certificate-service/consumer-guide/getting-started/get-client-certificate/
-        see https://pages.github.tools.sap/sap-pki-certificate-service/consumer-guide/specifications/api/
-        '''
         headers = {
             'Accept': 'application/json',
             'Authorization': f'Bearer {self.access_token}',
@@ -214,10 +210,10 @@ def _extract_client_certificate(cert_response: dict) -> str:
     certs = pkcs7.load_pem_pkcs7_certificates(pkcs7_pem.encode('utf-8'))
     if not certs:
         raise ValueError('no certificates found in response')
-    allCerts = ""
+    all_certs = ""
     for c in certs:
-        allCerts += c.public_bytes(serialization.Encoding.PEM).decode('utf-8')
-    return allCerts
+        all_certs += c.public_bytes(serialization.Encoding.PEM).decode('utf-8')
+    return all_certs
 
 
 def rotate_cfg_element(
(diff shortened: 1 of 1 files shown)


cfg_mgmt/btp_application_certificate.py
```diff
@@ -55,16 +55,22 @@ def _setup_oauth_token(self):
         self.access_token = resp.json()['access_token']
 
     def create_client_certificate_chain(self, csr_pem: str, validity_in_days: int) -> dict:
+        '''
+        see https://pages.github.tools.sap/sap-pki-certificate-service/consumer-guide/getting-started/get-client-certificate/
```
zkdev, line 59: Please refrain from linking non-public resources.
Rather link this as comment within configuration.

cfg_mgmt/btp_application_certificate.py
```diff
@@ -208,7 +214,10 @@ def _extract_client_certificate(cert_response: dict) -> str:
     certs = pkcs7.load_pem_pkcs7_certificates(pkcs7_pem.encode('utf-8'))
     if not certs:
         raise ValueError('no certificates found in response')
-    return certs[0].public_bytes(serialization.Encoding.PEM).decode('utf-8')
+    allCerts = ""
```
zkdev, line 217: nit: [PEP8](https://peps.python.org/pep-0008/#function-and-variable-names) suggests snake_case for variable names

zkdev: I would prefer to avoid mutation, e.g. with `.join()`:

```python
all_certs = ''.join([cert.public_bytes(...) for cert in certs])
```

cfg_mgmt/btp_application_certificate.py
```diff
@@ -271,3 +280,10 @@ def delete_config_secret(
             gbaas_client.delete_certificate(info.cn, info.id)
 
     return None
+
+
+def test(cfg_factory: model.ConfigFactory):
```
zkdev, line 285: Is it intended to include this hunk?
If so, I would rather prefer to move this to a dedicated (unit)test module and mock external services (cfg-mgmt, gbaas-service).

zkdev (changes requested):

cfg_mgmt/btp_application_certificate.py
MartinWeindel: Why is this project on a public repo?

zkdev (approved):
/lgtm

PR #944, opened 2024-01-26
Assure re-usage of existing bdba scans if possible

8R0WNI3:

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/protecode/util.py
+++ b/protecode/util.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 
 import datetime
+import hashlib
 import logging
 import time
 import typing
@@ -23,6 +24,8 @@ import delivery.client
 import dso.model
 import gci.componentmodel as cm
 import github.compliance.model
+import oci.client
+import oci.model
 import protecode.model as pm
 
 
@@ -250,6 +253,7 @@ def component_artifact_metadata(
     artefact: cm.Artifact,
     omit_component_version: bool,
     omit_resource_version: bool,
+    oci_client: oci.client.Client,
 ):
     ''' returns a dict for querying bdba scan results (use for custom-data query)
     '''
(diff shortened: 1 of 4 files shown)


protecode/util.py
```diff
@@ -310,3 +315,18 @@ def filter_func(other_dict: typing.Dict[str, str]):
         return result.product_id()
     else:
         return None
+
+
+def image_ref_with_digest_if_possible(
+    image_reference: str | oci.model.OciImageReference,
```
ccwienk, line 321: for OCI Images, it is always possible to calculate digest (the digest is calculated as the SHA-256-digest from the image's manifest (`f'sha256:{hashlib.sha256(oci_client.manifest_raw().content).hexdigest()}'`). How about doing so? We can still omit doing so in cases where we e.g. do not have an OCI Client (however, if we do not have such a client, we will not be able to upload the OCI Image, so I assume this case is actually not relevant in practise).

protecode/util.py
```diff
@@ -310,3 +315,18 @@ def filter_func(other_dict: typing.Dict[str, str]):
...
+) -> str:
+    if not (digest and digest.value):
+        return image_reference
+
+    image_reference = oci.model.OciImageReference.to_image_ref(image_reference=image_reference)
+
+    if image_reference.has_digest_tag:
+        return image_reference.original_image_reference
+
+    return image_reference.ref_without_tag + '@sha256:' + digest.value
```
ccwienk, line 332: stylistic nit: preferably use f-str?

as an alternative idea - considering replacing the tag is a somewhat common problem: how about we add a method like `replace_tag(tag: str)` (or `with_tag(tag: str)`) to OciImageReferences as a convenience-shortcut?

as another idea for convenience, we might add `as_oci_tag` to cm.DigestSpec

In both cases, it is perfectly fine to not add those convenience methods for me.

ccwienk (changes requested):

ccwienk (approved):

PR #952, opened 2024-02-22
Rotation of Service Account kubeconfigs from Garden

guewa:

Automatic rotation for kubeconfigs of Service Account tokens in the Garden. 
This introduces a new attribute for "rotation_strategy" for the kubeconfig configs, for example
```
cluster-management:
  service_account:
    name: cluster-manager
    namespace: garden-mycluster

  rotation_strategy: tokenRequest

  kubeconfig:
  [...]
  bound_secret_name: cluster-manager-token
```
If set to the value "tokenRequest", a new token for the Service Account with max. validity of 90d is created and bound to a new Secret (opaque) created. The secret name is stored in kubeconfig with attribute "bound_secret_name",  which will be used in the next rotation cycle to added to the deletion queue. 

If no attribute "rotation_strategy", function works like before as default.

Fixes #

**Release note**:

```feature operator
NONE
```

### DIFF
--- a/cfg_mgmt/kubernetes.py
+++ b/cfg_mgmt/kubernetes.py
@@ -47,72 +47,127 @@ def rotate_cfg_element(
 
     # find service-account
     service_account_config = cfg_to_rotate.service_account()
+
     sa: V1ServiceAccount = core_api.read_namespaced_service_account(
         name=service_account_config.name(),
         namespace=service_account_config.namespace(),
     )
-    old_tokens = [{"name": token.name} for token in sa.secrets] if sa.secrets else []
-    secret_id = {
-        'old_tokens': old_tokens,
-    }
 
-    # create new token
-    new_access_token_name = _create_token_for_sa(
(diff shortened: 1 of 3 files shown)


cfg_mgmt/kubernetes.py
```diff
@@ -47,73 +47,125 @@ def rotate_cfg_element(
...
+            service_account=sa,
         )
-        raise RuntimeError(
-            'Service account token-secret was not populated with the required data in time.'
+
+        new_sa_access_token = core_api.create_namespaced_service_account_token(
+            name=service_account_config.name(),
+            namespace=service_account_config.namespace(),
+            body={
+                "spec": {"expirationSeconds": 7776000}, # 90d is max
```
ccwienk, line 73: codestyle: preferably use this form: `60 * 60 * 24 * 90`

zkdev (changes requested):

zkdev (approved):
/lgtm

PR #954, opened 2024-03-04
retain attributes from unknown `access`-types

ccwienk:
OCM is designed as a typed, extensible model. For operations not requiring the processing of certain aspects of model elements (e.g. opaque replication of OCM Component Descriptors), it is the desired behaviour for tooling to accept unknonwn model elements and retain them unmodified.

Therefore, add `dict` as fallback deserialisation type for `dacite`. For the sake of keeping code api-compatible, use custom `AccessDict` to still expose attribute `type`, which is expected to be accessible as attribute from existing users.

### DIFF
--- a/gci/componentmodel.py
+++ b/gci/componentmodel.py
@@ -66,6 +66,21 @@ class Access:
     type: typing.Optional[AccessTypeOrStr] = AccessType.NONE
 
 
+class AccessDict(dict):
+    '''
+    fallback for unknown access-types; it is api-compatible to `Access` in that it exposes its type
+    via the `type` attribute mimicking behaviour of `dataclasses` from this module, but otherwise
+    behaves as a `dict` (thus allowing de/reserialisation using dacite/dataclasses.asdict w/o losing
+    attributes).
+    '''
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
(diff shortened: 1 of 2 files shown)


zkdev (approved):
/lgtm

gci/componentmodel.py
```diff
@@ -437,6 +452,15 @@ class Resource(Artifact, LabelMethodsMixin):
     labels: typing.List[Label] = dataclasses.field(default_factory=tuple)
     srcRefs: list[SourceReference] = dataclasses.field(default_factory=tuple)
 
+    def __post_init__(self):
+        if dataclasses.is_dataclass((access := self.access)):
```
8R0WNI3, line 456: nit: only use single parentheses

gci/componentmodel.py
```diff
@@ -437,6 +452,15 @@ class Resource(Artifact, LabelMethodsMixin):
...
+            return
+
+        if isinstance(access, dict):
+            if not 'type' in access:
+                raise ValueError('attribute `type` must be present')
+            self.access = AccessDict(access)
```
8R0WNI3, line 462: Maybe keep dataclass frozen and instead set the value here using `object.__setattr__(self, 'access', AccessDict(access))`? That way, unwanted mutation would still be prevented

PR #957, opened 2024-03-12
Redesign compliance scan model classes

8R0WNI3:

Fixes #

**Release note**:

```other operator

```


dso/model.py
```diff
@@ -54,7 +53,7 @@ def as_frozenset(self) -> frozenset[str]:
 
 def component_artefact_id_from_ocm(
     component: cm.Component,
-    artefact: cm.Resource | cm.ComponentSource
+    artefact: cm.Resource | cm.Source,
```
ccwienk, line 56: good catch!

dso/model.py
```diff
@@ -114,64 +110,25 @@ class Metadata:
...
-    affected_package_name: str | None
-    affected_package_version: str | None
-    reportUrl: str
-    product_id: int
-    group_id: int
-    base_url: str
-    bdba_cfg_name: str
+            object.__setattr__(self, 'creation_date', now)
+        if not self.last_update:
+            object.__setattr__(self, 'last_update', now)
```
ccwienk, line 124: some thoughts w.r.t. creation-date + mutability.

generally speaking, if we need our model objects to be mutable for good reasons, and there are no strong reasons for making them immutable, we might just as well consider making our model-classes mutable, instead of making them mutable and then patch them anyway :-)

In addition, I am not exactly sure whether implicitly setting creation-date + last-update to "now" is such a good idea.

If we want to enforce _all_ instances must _always_ have those attributes set, let is instead add a validation (raise ValueError or the like), so programming errors will be detected early. If we want to ensure instances must have them set in certain cases (e.g. prior to uploading to delivery-service), we might add validation logic there.

protecode/util.py
```diff
@@ -25,19 +28,15 @@
 
 def sync_results_with_delivery_db(
     delivery_client: delivery.client.DeliveryServiceClient,
-    results: typing.Iterable[pm.BDBAScanResult],
-    bdba_cfg_name: str,
+    results: collections.abc.Iterable[dso.model.ArtefactMetadata],
```
ccwienk, line 31: do we actually still need this function?

dso/model.py
8R0WNI3: In fact, we do not require this model object to be mutable since the `creation_date` and `last_update` can be set during initialisation already. However, this hack, to set them in the `__post_init__` function, was just for convenience.

I suggest I remove it from here again and update usages to explicitly set those attributes.

protecode/util.py
8R0WNI3: No you're right, actually we can drop this wrapper by now :-)

zkdev (approved):
/lgtm

PR #960, opened 2024-04-05
Refactoring, fixes and support for extra-component-references-label

8R0WNI3:

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/cnudie/iter.py
+++ b/cnudie/iter.py
@@ -1,8 +1,8 @@
+import collections.abc
 import dataclasses
-import typing
 
-import gci.componentmodel as cm
 import cnudie.retrieve
+import gci.componentmodel as cm
 
 
 @dataclasses.dataclass
@@ -31,14 +31,16 @@ class ArtefactNode:
     - iterable (useful for pattern-matching, e.g. c,a = node)
     '''
     @property
-    def artefact(self) -> cm.Resource | cm.ComponentSource:
+    def artefact(self) -> cm.Resource | cm.Source:
         if isinstance(self, ResourceNode):
             return self.resource
         if isinstance(self, SourceNode):
             return self.source
         raise TypeError('must be of type ResourceNode or SourceNode')
 
-    def __iter__(self) -> typing.Generator[cm.Component|cm.Resource|cm.ComponentSource, None, None]:
+    def __iter__(
+        self,
+    ) -> collections.abc.Generator[cm.Component | cm.Resource | cm.Source, None, None]:
         # pylint: disable=E1101
         yield self.component
         yield self.artefact
@@ -51,7 +53,7 @@ class ResourceNode(Node, ArtefactNode):
 
 @dataclasses.dataclass
 class SourceNode(Node, ArtefactNode):
-    source: cm.ComponentSource
+    source: cm.Source
 
 
 class Filter:
@@ -73,10 +75,10 @@ def iter(
     lookup: cnudie.retrieve.ComponentDescriptorLookupById=None,
     recursion_depth: int=-1,
     prune_unique: bool=True,
-    node_filter: typing.Callable[[Node], bool]=None,
+    node_filter: collections.abc.Callable[[Node], bool]=None,
     ocm_repo: cm.OcmRepository | str=None,
     ctx_repo: cm.OcmRepository | str=None, # deprecated, use `ocm_repo` instead
-) -> typing.Generator[Node, None, None]:
+) -> collections.abc.Generator[Node, None, None]:
     '''
     returns a generator yielding the transitive closure of nodes accessible from the given component.
 
@@ -140,14 +142,14 @@ def iter(
                 name=cref.componentName,
                 version=cref.version,
             )
+
             if ocm_repo:
                 referenced_component_descriptor = lookup(cref_id, ocm_repo)
             else:
                 referenced_component_descriptor = lookup(cref_id)
-            referenced_component = referenced_component_descriptor.component
 
             yield from inner_iter(
-                component=referenced_component,
+                component=referenced_component_descriptor.component,
                 lookup=lookup,
                 recursion_depth=recursion_depth,
                 path=path,
@@ -176,7 +178,7 @@ def iter_resources(
     lookup: cnudie.retrieve.ComponentDescriptorLookupById=None,
     recursion_depth: int=-1,
     prune_unique: bool=True,
-) -> typing.Generator[ResourceNode, None, None]:
+) -> collections.abc.Generator[ResourceNode, None, None]:
     '''
     curried version of `iter` w/ node-filter preset to yield only resource-nodes
     '''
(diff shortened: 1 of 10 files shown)


cnudie/iter.py
```diff
@@ -1,22 +1,41 @@
...
 
-import gci.componentmodel as cm
 import cnudie.retrieve
+import dso.labels
+import gci.componentmodel as cm
+
+
+class NodeReferenceType(enum.StrEnum):
+    COMPONENT_REFERENCE = 'componentReference'
+    EXTRA_COMPONENT_REFS_LABEL = dso.labels.ExtraComponentReferencesLabel.name
```
ccwienk, line 12: should we prefix the label (e.g. with `label:`)?

cnudie/iter.py
```diff
@@ -1,22 +1,41 @@
...
     def component(self):
-        return self.path[-1]
+        return self.path[-1].component
 
     @property
     def component_id(self):
         return self.component.identity()
 
+    def __post_init__(self):
+        self.path = tuple(
```
ccwienk, line 34: shouldn't we rather expect callers to pass correct values (we might check + raise ValueError if values are incorrect)?

cnudie/iter.py
```diff
@@ -31,14 +50,16 @@ class ArtefactNode:
     - iterable (useful for pattern-matching, e.g. c,a = node)
     '''
     @property
-    def artefact(self) -> cm.Resource | cm.ComponentSource:
+    def artefact(self) -> cm.Resource | cm.Source:
```
ccwienk, line 53: good catch

cnudie/iter.py
```diff
@@ -140,17 +162,43 @@ def inner_iter(
...
+                referenced_component_descriptor = lookup(extra_cref_id, ocm_repo)
+            else:
+                referenced_component_descriptor = lookup(extra_cref_id)
+
+            yield from inner_iter(
+                component=referenced_component_descriptor.component,
                 lookup=lookup,
                 recursion_depth=recursion_depth,
                 path=path,
+                reftype=NodeReferenceType.EXTRA_COMPONENT_REFS_LABEL,
```
ccwienk, line 201: IMHO, we should only mark the exact component that declared reference via label (sub-components from such a component should again be marked as regular crefs (if they are regular crefs, of course)

ccwienk (changes requested):

cnudie/iter.py
8R0WNI3: I was not sure about this either, so I'm with you to define it like that :-)

ctt/uploaders.py
```diff
@@ -149,7 +149,7 @@ def process(self, processing_job, target_as_source=False):
 
         src_prefix, src_name, src_tag = oci.client._split_image_reference(src_ref)
 
-        if ':' in src_tag:
+        if '@' in src_tag:
```
ccwienk, line 152: preferably use oci.model.OciImageReference + `has_tag`

cnudie/iter.py
```diff
@@ -105,9 +107,12 @@ def iter(
                             dependencies; -1 will resolve w/o recursion limit, 0 will not resolve
                             component dependencies
     @param prune_unique: if true, redundant component-versions will only be traversed once
-    @node_filter:        use to filter emitted nodes (see Filter for predefined filters)
+    @param node_filter:  use to filter emitted nodes (see Filter for predefined filters)
     @param ocm_repo:     optional OCM Repository to be used to override in the lookup
     @param ctx_repo:     deprecated, use `ocm_repo` instead
+    @param component_filter: use to exclude components (and their references) from the iterator
```
ccwienk, line 113: it would be nice to state that returning `True` means component should be filtered out

cnudie/iter.py
```diff
@@ -105,9 +107,12 @@ def iter(
...
+    @param reftype_filter: use to exclude components (and their references) from the iterator if
```
ccwienk, line 114: ditto

ctt/process_dependencies.py
```diff
@@ -420,12 +424,20 @@ def process_images(
 
     executor = concurrent.futures.ThreadPoolExecutor(max_workers=16)
 
+    reftype_filter = None
+    if skip_labels and skip_labels(dso.labels.ExtraComponentReferencesLabel.name):
+        def filter_extra_component_refs(reftype: cnudie.iter.NodeReferenceType) -> bool:
+            return reftype == cnudie.iter.NodeReferenceType.EXTRA_COMPONENT_REFS_LABEL
```
ccwienk, line 430: nit: use `is` for comparison w/ enum-values?

did you use `==` to gracefully handle cases where we encounter a str-value we do not know as of our enum?

8R0WNI3: No good point, I did not do this on purpose. The `reftype` here stems from `cnudie.iter`, so we can assume that it's a value of our enum or at least fail if it's not 👍🏼

ctt/process_dependencies.py
```diff
@@ -688,6 +690,8 @@ def reupload_component(component: cm.Component):
 
         if processing_mode is ProcessingMode.REGULAR:
             if component.name == source_comp.name and component.version == source_comp.version:
+                if skip_root_cd_upload:
```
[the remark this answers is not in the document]
8R0WNI3: I also thought about it again. I came out with the solution to keep the `skip_component_upload` parameter with its current semantics (preventing just the cd-replication, not the replication of its resources and component references). Instead, I added a newly introduced parameter `component_filter` which has the semantics of excluding the resources and transitive component references from the replication as well. I also added some comments explaining this behaviour 😅

dso/labels.py
```diff
@@ -91,6 +91,18 @@ class CveCategorisationLabel(Label):
...
 
+@dataclasses.dataclass(frozen=True)
+class ExtraComponentReference:
+    component_reference: cm.ComponentIdentity
+    purpose: tuple[str] | None
+
+
+@dataclasses.dataclass(frozen=True)
+class ExtraComponentReferencesLabel(Label):
+    name = 'ocm.software/ocm-gear/extra-component-references'
```
zkdev, line 102: Shouldn't this rather be a general "defined / well-known" ocm label, and not an `ocm-gear` / `delivery-gear` specific?

ctt/process_dependencies.py
```diff
@@ -648,6 +661,15 @@ def append_ctx_repo(ctx_base_url: str | cm.OciOcmRepository, component: cm.Compo
         ocm_repository = component.current_repository_ctx()
         oci_ref = ocm_repository.component_version_oci_ref(component)
 
+        if skip_labels:
+            labels: list[cm.Label] = []
+
+            for label in component.labels:
+                if not skip_labels(label.name):
+                    labels.append(label)
```
zkdev, line 669: rather use list comprehension 🙂

zkdev (changes requested):

zkdev (approved):
/lgtm

PR #962, opened 2024-04-25
Make version lookup more robust against (common) errors

8R0WNI3:

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/cnudie/retrieve.py
+++ b/cnudie/retrieve.py
@@ -485,11 +485,29 @@ def oci_component_descriptor_lookup(
     return lookup
 
 
+def error_code_indicating_not_found(image_reference: str | om.OciImageReference) -> int:
+    '''
+    Since some oci registries don't comply with the open containers spec
+    (https://github.com/opencontainers/distribution-spec/blob/main/spec.md#endpoints)
+    regarding the error codes for failure upon listing tags, this function tries to
+    guess the oci registry type based on the supplied `image_reference` and returns
+    the expected error code.
+    '''
+    oci_registry_type = om.OciRegistryType.from_image_ref(image_reference=image_reference)
+
+    if oci_registry_type == om.OciRegistryType.DOCKERHUB:
+        return 401
+    if oci_registry_type == om.OciRegistryType.GHCR:
+        return 403
+
+    return 404
+
+
 def version_lookup(
     default_ctx_repo: cm.OcmRepository=None,
     ocm_repository_lookup: OcmRepositoryLookup=None,
     oci_client: oc.Client=None,
-    default_absent_ok=True,
+    default_absent_ok: bool=True,
 ) -> VersionLookupByComponent:
     if not oci_client:
         import ccc.oci
@@ -530,8 +548,15 @@ def version_lookup(
                     oci_client=oci_client,
                 ):
                     versions.add(version_tag)
-            except requests.exceptions.HTTPError:
-                continue
+            except requests.exceptions.HTTPError as e:
+                if (error_code := e.response.status_code) == 404:
+                    continue
+
+                image_reference = ocm_repo.component_oci_ref(component_name)
+                if error_code == error_code_indicating_not_found(image_reference=image_reference):
+                    continue
+
+                raise
 
         if not versions and not absent_ok:
             raise om.OciImageNotFoundException()
--- a/oci/client.py
+++ b/oci/client.py
@@ -7,6 +7,7 @@ import json
 import logging
 import tempfile
 import threading
+import time
 import typing
 
 import dacite
@@ -268,6 +269,7 @@ class Client:
         self,
         image_reference: typing.Union[str, om.OciImageReference],
         scope: str,
+        remaining_retries: int=3,
     ):
         if isinstance(image_reference, om.OciImageReference):
             image_reference = str(image_reference)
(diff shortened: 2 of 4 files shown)


oci/client.py
```diff
@@ -268,6 +269,7 @@ def _authenticate(
         self,
         image_reference: typing.Union[str, om.OciImageReference],
         scope: str,
+        retry_on_missing_quota: bool=True,
```
ccwienk, line 272: how about making this an `int` (amount of remaining retries) + keep it abstract (just: retries).

we can still only do retries for http-429-case for now. however we might add other cases later

cnudie/retrieve.py
```diff
@@ -489,7 +489,8 @@ def version_lookup(
     default_ctx_repo: cm.OcmRepository=None,
     ocm_repository_lookup: OcmRepositoryLookup=None,
     oci_client: oc.Client=None,
-    default_absent_ok=True,
+    default_absent_ok: bool=True,
+    default_raise_on_error: bool=False,
```
ccwienk, line 493: how about making this more generic by passing-in status-codes to be considered as error?

cnudie/retrieve.py
```diff
@@ -530,7 +532,14 @@ def lookup(
                     oci_client=oci_client,
                 ):
                     versions.add(version_tag)
-            except requests.exceptions.HTTPError:
+            except requests.exceptions.HTTPError as e:
+                # status codes/respones for container registries in case no tags are found:
```
ccwienk, line 536: nit: typo (responSes)

cnudie/retrieve.py
```diff
@@ -530,7 +532,14 @@ def lookup(
...
+                # acr (Alibaba) -> 404
+                # artifactory -> 404
+                # gar/gcr (Google) -> []
+                # ghcr (GitHub) -> 403
+                if raise_on_error and not (absent_ok and e.response.status_code in (403, 404)):
```
ccwienk, line 541: so for ghcr, this will raise if there is no tag? that's not nice :-(

oci/model.py
```diff
@@ -43,8 +43,24 @@ class OciRegistryType(enum.Enum):
     GCR = 'gcr'
     DOCKERHUB = 'dockerhub'
     ARTIFACTORY = 'artifactory'
+    GHCR = 'ghcr'
+    ALIBABA = 'alibaba'
     UNKNOWN = 'unknown'
 
+    @staticmethod
+    def from_image_ref(image_reference: str) -> 'OciRegistryType':
+        if 'pkg.dev' in image_reference:
```
ccwienk, line 52: nit: how about using OciImageReference + its `netloc` property to avoid "false-positives"?

oci/model.py
```diff
@@ -43,8 +43,24 @@ class OciRegistryType(enum.Enum):
     GCR = 'gcr'
     DOCKERHUB = 'dockerhub'
     ARTIFACTORY = 'artifactory'
+    GHCR = 'ghcr'
+    ALIBABA = 'alibaba'
```
zkdev, line 47: nit: keep abbreviations consistent and use `ACR` or `ACCR` for `Alibaba Cloud Container Registry`

oci/model.py
```diff
@@ -43,8 +43,24 @@ class OciRegistryType(enum.Enum):
...
+    def from_image_ref(image_reference: str) -> 'OciRegistryType':
+        if 'pkg.dev' in image_reference:
+            return OciRegistryType.GAR
+        if 'gcr.io' in image_reference:
+            return OciRegistryType.GCR
+        if 'docker.io' in image_reference:
+            return OciRegistryType.DOCKERHUB
+        if 'ghcr.io' in image_reference:
+            return OciRegistryType.GHCR
+        if 'cr.aliyuncs.com' in image_reference:
```
zkdev, line 60: A simple `in` check is not sufficient because we copy over images during release and keep old img-ref as part of the new img-ref.
Rather check with `startswith` (+ some trimming?)

cnudie/retrieve.py
```diff
@@ -485,11 +485,30 @@ def lookup(
     return lookup
 
 
+def status_code_if_not_found(image_reference: str) -> int:
+    oci_registry_type = om.OciRegistryType.from_image_ref(
+        image_reference=image_reference,
+    )
+
+    if oci_registry_type in (om.OciRegistryType.GAR, om.OciRegistryType.GCR):
+        return 200
```
zkdev, line 494: really sad this mapping is required considering there is an open specification ...
-> https://github.com/opencontainers/distribution-spec/blob/main/spec.md#endpoints

zkdev (changes requested):

oci/model.py
ccwienk: isn't it rather AliYun / AliCloud?

ccwienk: also, I would prefer not to introduce abbreviations, such as ACR and rather go through the lengths of spending a few more characters for the sake of being more descriptive ("well-known" abbreviations are okay)

zkdev: was refering to this document:
https://www.alibabacloud.com/help/en/acr/product-overview/what-is-container-registry

oci/model.py
8R0WNI3: By using the netloc as @ccwienk suggested, this should also be handled.

ccwienk (commented):
.

zkdev (approved):
/lgtm

ccwienk (approved):

PR #963, opened 2024-04-26
Update metadata by type instead of in one request

8R0WNI3:
Rationale: Reduce the request size of metadata updates in order to lower risks of running into http 413 errors.

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/delivery/client.py
+++ b/delivery/client.py
@@ -1,5 +1,6 @@
 import dataclasses
 import datetime
+import itertools
 import logging
 import requests
 import time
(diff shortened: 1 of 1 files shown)


delivery/client.py
```diff
@@ -200,18 +201,31 @@ def update_metadata(
...
+                ]},
+                timeout=(4, 121),
+            )
 
-        res.raise_for_status()
+            if not res.ok:
+                logger.warning(
+                    f'error occured during metadata update of {type=}; {processed_types=}'
+                )
+                res.raise_for_status()
```
zkdev, line 226: We should not immediately raise upon failed request but rather process the remaining data-types.

delivery/client.py
```diff
@@ -200,18 +201,41 @@ def update_metadata(
...
-                ) for artefact_metadata in data
-            ]},
-            timeout=(4, 121),
+        # update metadata by type to reduce the per-request-size to lower risks of running into
+        # http 413 errors. However, don't split data of the same type into different requests
+        # because the delivery services requires all data of the same type to be in the same
+        # request (the request is being used as a kind of context, so that not containing elements
+        # are being removed from the delivery database)
+        data_by_types = itertools.groupby(
+            sorted(data, key=lambda datum: datum.meta.type),
```
zkdev, line 210: why do we need to sort previously to grouping?

8R0WNI3: Because the `itertools.groupby` function only groups consecutive items together, so you have to sort first to establish this order and then you can group by each value.

zkdev (approved):
/lgtm

PR #967, opened 2024-05-02
Import bdba triages

8R0WNI3:
- query existing bdba scan results for artefact prior to uploading to determine those findings, which are not included in the current result anymore and thus need to be removed
- also store triaged findings together with their triage(s) as rescoring in the delivery db
- remove stale findings from delivery db

Fixes #

**Release note**:

```other operator

```


protecode/util.py
```diff
@@ -24,10 +26,45 @@
...
+        components=[resource_node.component_id],
+        type=finding_type,
+    )
+
+    return [
+        delivery.model.ArtefactMetadata.from_dict(raw).to_dso_model_artefact_metadata()
+        for raw in findings_raw
+        if (
+            raw.get('meta').get('datasource') == datasource
+            and raw.get('artefactId').get('artefactName') == resource_node.artefact.name
```
ccwienk, line 47: we really need to factor-out this matching-gear eventually

protecode/util.py
```diff
@@ -106,23 +154,69 @@ def iter_artefact_metadata(
...
+                        relation_kind=dso.model.RelationKind.RESCORE,
+                    ),
+                    creation_date=triage.modified.astimezone(datetime.UTC),
+                )
+
+                vulnerability_rescoring = dso.model.CustomRescoring(
+                    finding=dso.model.RescoringVulnerabilityFinding(
+                        id=dso.model.BDBAPackageId(
+                            package_name=package_id.package_name,
+                            # don't specify version to store only one rescoring per package name
```
ccwienk, line 192: should we not rather honour triages' scopes (app vs group, ..) and translate those into our scoping?

protecode/util.py
```diff
@@ -141,59 +235,22 @@ def iter_artefact_metadata(
...
-            summary=None,
-        )
-
-        yield dso.model.ArtefactMetadata(
-            artefact=artefact_ref,
-            meta=meta,
-            data=dummy_finding,
-            discovery_date=discovery_date,
-        )
+            # remove vuln finding because it is still relevant and must not be deleted afterwards
```
ccwienk, line 239: might be made slightly more self-describing by moving code from condition into a function and call it in list-comprehension

protecode/util.py
8R0WNI3: I thought about this too. However, I came to the conclusion _not_ to do it because
- we can't actually translate them 100% correctly since the BDBA scopes are based on a different model (groups, apps, ...). We _could_ of course translate them in the way it fits Gardener, however, if someone utilises the BDBA groups and so on differently, this mapping wouldn't fit anymore.
- BDBA includes the triages in every scan result which matches its scope. So, if we just convert it to "our" rescoring for every scan and set it to our `single` scope (which means it just applies to the exact tuple of component-id and artefact-id), we basically copy the BDBA scope because we mirror where the rescoring is applied to. The only downside I see to this is that we store one rescoring artefact per version this way, but I think that is okay-ish.
---
So, I would prefer to always use the `single` scope for BDBA triages and import them for every scan result separately.

protecode/util.py
```diff
@@ -24,10 +26,45 @@
 ci.log.configure_default_logging(print_thread_id=True)
 
 
+def iter_existing_findings(
```
zkdev, line 29: nit: name implies generator as return-type

protecode/util.py
```diff
@@ -106,23 +154,69 @@ def iter_artefact_metadata(
...
-            has_license_findings = True
 
-        meta = dso.model.Metadata(
-            datasource=dso.model.Datasource.BDBA,
-            type=dso.model.Datatype.VULNERABILITY,
-            creation_date=now,
-            last_update=now,
-        )
+            # remove license finding because it is still relevant and must not be deleted afterwards
+            existing_findings = [
```
zkdev, line 159: After filtering this name (`existing_findings`) is not really accurate anymore.
How about some `stale_findings` ?

zkdev (changes requested):
Overall /lgtm

protecode/util.py
ccwienk: good point - this way, we need less logic on our end and the acceptable price of redundant data.

maybe you could add a comment explaining why we decided against interpreting the scope from bdba

ccwienk (commented):
see my last reply (add comment) - otherwise: lgtm

ccwienk (approved):

zkdev (approved):
/lgtm

PR #968, opened 2024-05-02
Add gzip compression to delivery client metadata routes

8R0WNI3:

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/delivery/client.py
+++ b/delivery/client.py
@@ -15,6 +15,7 @@ import cnudie.retrieve
 import cnudie.util
 import delivery.model as dm
 import dso.model
+import http_requests
 
 
 logger = logging.getLogger(__name__)
@@ -179,14 +180,24 @@ class DeliveryServiceClient:
         self,
         data: typing.Iterable[dso.model.ArtefactMetadata],
     ):
-        res = self.session.post(
-            url=self._routes.upload_metadata(),
+        headers = {
+            'Content-Type': 'application/json',
+        }
+
+        data, headers = http_requests.encode_request(
             json={'entries': [
                 dataclasses.asdict(
                     artefact_metadata,
                     dict_factory=ci.util.dict_to_json_factory,
                 ) for artefact_metadata in data
             ]},
+            headers=headers,
+        )
+
+        res = self.session.post(
+            url=self._routes.upload_metadata(),
+            data=data,
+            headers=headers,
             timeout=(4, 121),
         )
 
@@ -196,14 +207,24 @@ class DeliveryServiceClient:
         self,
         data: typing.Iterable[dso.model.ArtefactMetadata],
     ):
-        res = self.session.put(
-            url=self._routes.update_metadata(),
+        headers = {
+            'Content-Type': 'application/json',
+        }
+
+        data, headers = http_requests.encode_request(
             json={'entries': [
                 dataclasses.asdict(
                     artefact_metadata,
                     dict_factory=ci.util.dict_to_json_factory,
                 ) for artefact_metadata in data
             ]},
+            headers=headers,
+        )
+
+        res = self.session.put(
+            url=self._routes.update_metadata(),
+            data=data,
+            headers=headers,
             timeout=(4, 121),
         )
 
@@ -213,14 +234,24 @@ class DeliveryServiceClient:
         self,
         data: typing.Iterable[dso.model.ArtefactMetadata],
     ):
-        res = self.session.delete(
-            url=self._routes.delete_metadata(),
+        headers = {
+            'Content-Type': 'application/json',
+        }
+
+        data, headers = http_requests.encode_request(
             json={'entries': [
                 dataclasses.asdict(
                     artefact_metadata,
                     dict_factory=ci.util.dict_to_json_factory,
                 ) for artefact_metadata in data
             ]},
+            headers=headers,
+        )
+
+        res = self.session.delete(
+            url=self._routes.delete_metadata(),
+            data=data,
+            headers=headers,
             timeout=(4, 121),
         )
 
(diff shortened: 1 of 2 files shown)


http_requests.py
```diff
@@ -221,3 +226,66 @@ def delete(self, url: str, return_type=None, **kwargs):
...
+            raise NotImplementedError(f'Encoding of iterable {type(data)} is not (yet) supported')
+
+        else:
+            data = _encode(data)
+
+            if encoder == Encoder.GZIP:
+                compressor = zlib.compressobj(wbits=31)
+                yield compressor.compress(data) + compressor.flush()
+
+    if headers is not None:
```
zkdev, line 287: can't we just check whether headers is truthy?

http_requests.py
```diff
@@ -221,3 +226,66 @@ def delete(self, url: str, return_type=None, **kwargs):
...
+        if isinstance(obj, bytes):
+            return obj
+        elif isinstance(obj, str):
+            return obj.encode(encoding=encoding)
+        elif isinstance(obj, dict):
+            return js.dumps(obj).encode(encoding=encoding)
+        else:
+            raise NotImplementedError(f'Encoding of type {type(obj)} is not (yet) supported')
+
+    def _compress(data, encoder) -> collections.abc.Generator[bytes, None, None]:
```
zkdev, line 267: nit: `encoding_method` rather than `encoder` as this is not used directly but as a compressor mapping

zkdev (changes requested):

http_requests.py
8R0WNI3: No, because we also want to add the `Content-Encoding` header in case an empty header/dictionary is passed-in.

http_requests.py
```diff
@@ -221,3 +226,66 @@ def delete(self, url: str, return_type=None, **kwargs):
...
+        elif isinstance(obj, str):
+            return obj.encode(encoding=encoding)
+        elif isinstance(obj, dict):
+            return js.dumps(obj).encode(encoding=encoding)
+        else:
+            raise NotImplementedError(f'Encoding of type {type(obj)} is not (yet) supported')
+
+    def _compress(data, encoding_method) -> collections.abc.Generator[bytes, None, None]:
+        if isinstance(data, io.BufferedIOBase):
+            if encoding_method == EncodingMethod.GZIP:
```
ccwienk, line 269: nit: enum-values ought to be compared using identity (`is` instead of `==`)

http_requests.py
```diff
@@ -221,3 +226,66 @@ def delete(self, url: str, return_type=None, **kwargs):
...
+
+    def _encode(obj, encoding: str='utf-8') -> bytes:
+        if isinstance(obj, bytes):
+            return obj
+        elif isinstance(obj, str):
+            return obj.encode(encoding=encoding)
+        elif isinstance(obj, dict):
+            return js.dumps(obj).encode(encoding=encoding)
+        else:
+            raise NotImplementedError(f'Encoding of type {type(obj)} is not (yet) supported')
```
ccwienk, line 265: nit: as we found out some months ago, NotImplementedError is actually intended for marking a missing method; `ValueError` would likely be more appropriate. OTOH, we are using NIE in a lot of places in our codebase. So I would leave it up to you to either be more consistent with stdlib's intention, or our codebase's conventions.

http_requests.py
```diff
@@ -221,3 +226,66 @@ def delete(self, url: str, return_type=None, **kwargs):
...
+            if encoding_method == EncodingMethod.GZIP:
+                compressor = zlib.compressobj(wbits=31)
+                data.seek(0)
+
+                while (chunk := data.read(4096)):
+                    yield compressor.compress(chunk)
+                yield compressor.flush()
+
+        elif hasattr(data, '__iter__') and not isinstance(data, (str, bytes, dict)):
+            raise NotImplementedError(f'Encoding of iterable {type(data)} is not (yet) supported')
```
ccwienk, line 278: ditto (NIE vs VE)

http_requests.py
```diff
@@ -221,3 +226,66 @@ def delete(self, url: str, return_type=None, **kwargs):
...
+
+        else:
+            data = _encode(data)
+
+            if encoding_method == EncodingMethod.GZIP:
```
ccwienk, line 283: ditto: is vs ==

http_requests.py
```diff
@@ -221,3 +226,66 @@ def delete(self, url: str, return_type=None, **kwargs):
...
+    json: dict=None,
+    headers: dict[str, str]=None,
+    encoding_method: EncodingMethod=EncodingMethod.GZIP,
+) -> (
+    tuple[collections.abc.Generator[bytes, None, None], dict[str, str]]
+    | collections.abc.Generator[bytes, None, None]
+):
+    '''
+    Encodes the given `data` or `json` property based on the selected `encoding_method`. If both
+    `data` and `json` are provided, the `data` property is used (same behaviour as implemented by
```
ccwienk, line 246: really? this is how requests does it? that's horrifying.
I personally would prefer to raise a VE in that case to signal programming errors, tbh

http_requests.py
```diff
@@ -221,3 +226,66 @@ def delete(self, url: str, return_type=None, **kwargs):
...
                 **kwargs
         )
+
+
+class EncodingMethod(enum.StrEnum):
+    GZIP = 'gzip'
+
+
+def encode_request(
+    data: str | bytes | dict | typing.IO=None,
```
ccwienk, line 236: should we actually allow  `dict` here (in conjunction w/ gzip)? I know `requests` allows it. However, I think, requests will translate a dict into a html-form-request IIRC (which is something we likely do not want to do here).

-> propsal: if gzip + dict: raise VE?

delivery/client.py
```diff
@@ -179,14 +180,24 @@ def upload_metadata(
         self,
         data: typing.Iterable[dso.model.ArtefactMetadata],
     ):
-        res = self.session.post(
-            url=self._routes.upload_metadata(),
+        headers = {
+            'Content-Type': 'application/json',
+        }
```
ccwienk, line 185: nit: you might offload setting content-type to encode_request. otoh: explicit is better than implicit.

ccwienk (changes requested):

http_requests.py
8R0WNI3: You're right, you can, however, compare string enums directly to strings using `==`. If you're using `is` for comparison, you have to compare against `enum.value`.

8R0WNI3: Of course I can change it, if you'd prefer the latter?

PR #971, opened 2024-05-13
Refactor `ArtefactMetadata` model classes

8R0WNI3:

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/delivery/client.py
+++ b/delivery/client.py
@@ -385,6 +385,14 @@ class DeliveryServiceClient:
         components: collections.abc.Iterable[cm.Component]=(),
         type: dso.model.Datatype | tuple[dso.model.Datatype]=None,
     ):
+        '''
+        Query artefact metadata from the delivery-db and return the raw json response.
+
+        @param components: component identities used for filtering; if no identities are specified,
+                           no component filtering is done
+        @param type:       datatype(s) used for filtering; if no datatype(s) is (are) specified, no
+                           datatype filtering is done
+        '''
         if type:
             params = {
                 'type': type,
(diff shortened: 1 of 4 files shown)


dso/model.py
```diff
@@ -381,3 +384,21 @@ class ArtefactMetadata:
...
+    @staticmethod
+    def from_dict(raw: dict):
+        return dacite.from_dict(
+            data_class=ArtefactMetadata,
+            data=raw,
+            config=dacite.Config(
+                type_hooks={
+                    datetime.datetime: dateutil.parser.isoparse,
+                    datetime.date: lambda date: datetime.datetime.fromisoformat(date).date()
+                        if date else None,
```
zkdev, line 397: how can `date` be falsy here?

8R0WNI3: Good catch, it should only be falsy if it's `None` but in that case the type hook should not be used 👍🏼

zkdev (approved):
/lgtm

delivery/client.py
```diff
@@ -416,6 +416,21 @@ def query_metadata_raw(
 
         return res.json()
 
+    def query_metadata(
+        self,
+        components: collections.abc.Iterable[cm.Component]=(),
+        type: dso.model.Datatype | tuple[dso.model.Datatype]=None,
+    ) -> tuple[dso.model.ArtefactMetadata]:
```
ccwienk, line 423: it would be nice to add a short docstr, esp. considering parameters (what does it mean if multiple / no components are passed..)

delivery/client.py
```diff
@@ -416,6 +416,21 @@ def query_metadata_raw(
...
+        artefact_metadata_raw = self.query_metadata_raw(
```
ccwienk, line 424: nit: if `raw` flavour is not intended to be used, mark it as private (leading underscore)?

dso/model.py
```diff
@@ -24,13 +27,41 @@ class Datasource:
     CC_UTILS = 'cc-utils'
 
 
+def normalise_artefact_extra_id(
+    artefact_extra_id: dict[str, str],
+    artefact_version: str=None,
+) -> str:
+    '''
+    generate stable representation of `artefact_extra_id` and remove `version` key if
+    the specified version is identical to the given artefact version
```
ccwienk, line 36: why special-handling for version? how about adding this to gci.componentmodel instead, like so:
-> id_tuple: returns (alphabetically) ordered tuple of (<id>, <value>)-pairs, which will always contain as many id-attrs as are available (always `name`, `version` if available, every attr from extra-id)?

this function's output might be used to also normalise into a str (we might use tuple's __str__ for that or invent a custom format, as you suggested). however, as separation character, I would suggest to use one that is forbidden within keys/values - did you check ocm-spec whether this is true for underscore?

8R0WNI3: We added the special-handling for the version already a while ago ([ref](https://github.com/gardener/cc-utils/commit/ae7c03d76cfe7c3242829eb08400da033493639c)). The reason behind this was that some artefacts contained their version also in the extra identity field, however, not always stable across different OCI repositories. Because of that, the comparison of the extra identities failed sometimes, depending of which OCI repository was used. To allow a stable representation of the extra identity independent of the OCI repository, we decided to allow the removal of the version attribute from the extra identity if desired.

Also, I just moved the `normalise_artefact_extra_id` function from `delivery.model` to `dso.model`, so I did not look into it in more detail, e.g. the separating characters. Nevertheless, I figure the underscore character is very basic in Python variable names and thus not forbidden within keys/values.

delivery/client.py
8R0WNI3: The `query_metadata` route is just added as a convenience parser, however, parsing the raw data to `dso.model.ArtefactMetadata` comes with some data loss (i.e. patched-in data by the delivery-service). For most cases, this is perfectly fine as those extra information is not required, but sometimes it is and so `query_metadata_raw` has to be instrumented.

PR #978, opened 2024-05-27
Authenticate against delivery-service in delivery-service client

8R0WNI3:

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/delivery/client.py
+++ b/delivery/client.py
@@ -9,13 +9,18 @@ import dacite
 
 import gci.componentmodel as cm
 
+import ccc.github
 import ci.util
 import cnudie.iter
 import cnudie.retrieve
 import cnudie.util
+import delivery.jwt
 import delivery.model as dm
 import dso.model
 import http_requests
+import model
+import model.base
+import model.github
 
 
 logger = logging.getLogger(__name__)
@@ -24,6 +29,30 @@ logger = logging.getLogger(__name__)
 class DeliveryServiceRoutes:
     def __init__(self, base_url: str):
         self._base_url = base_url
+
+    def auth(self):
+        return ci.util.urljoin(
+            self._base_url,
+            'auth',
+        )
+
+    def auth_configs(self):
+        return ci.util.urljoin(
+            self._base_url,
+            'auth',
+            'configs',
+        )
+
+    def openid_configuration(self):
+        '''
+        endpoint according to OpenID provider configuration request
+        https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfigurationRequest
+        '''
+        return ci.util.urljoin(
+            self._base_url,
+            '.well-known',
+            'openid-configuration',
+        )
 
     def component_descriptor(self):
         return ci.util.urljoin(
@@ -109,9 +138,151 @@ class DeliveryServiceClient:
     def __init__(
         self,
         routes: DeliveryServiceRoutes,
+        github_cfgs: tuple[model.github.GithubConfig]=(),
+        cfg_factory: model.ConfigFactory | None=None,
     ):
+        '''
+        Initialises a client which can be used to interact with the delivery-service.
+
+        :param DeliveryServiceRoutes routes
+            object which contains information of the base url of the desired instance of the
+            delivery-service as well as the available routes
+        :param tuple[GithubConfig] github_cfgs (optional):
+            tuple of the available GitHub configurations which are used to authenticate against the
+            delivery-service
+        :param ConfigFactory cfg_factory (optional):
+            the config factory is used to retrieve available GitHub configurations in case they are
+            not provided anyways (i.e. can be safely omitted in case `github_cfgs` is specified)
+        '''
         self._routes = routes
-        self.session = requests.sessions.Session()
+        self.github_cfgs = github_cfgs
+        self.cfg_factory = cfg_factory
+
+        self._bearer_token = None
+        self._json_web_keys = None
+        self._session = requests.sessions.Session()
+
+    def _openid_configuration(self):
+        '''
+        response according to OpenID provider configuration response
+        https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfigurationResponse
+        '''
+        res = self._session.get(
+            url=self._routes.openid_configuration(),
+            timeout=(4, 31),
+        )
+
+        res.raise_for_status()
+
+        return res.json()
+
+    def _openid_jwks(self):
+        openid_configuration = self._openid_configuration()
+
+        res = self._session.get(
+            url=openid_configuration.get('jwks_uri'),
+            timeout=(4, 31),
+        )
+
+        res.raise_for_status()
+
+        return res.json()
+
+    def _authenticate(self):
+        if self._bearer_token and not delivery.jwt.is_jwt_token_expired(
+            token=self._bearer_token,
+            json_web_keys=self._json_web_keys,
+        ):
+            return
+
+        res = self._session.get(
+            url=self._routes.auth_configs(),
+            timeout=(4, 31),
+        )
+
+        res.raise_for_status()
+
+        auth_configs = res.json()
+
+        for auth_config in auth_configs:
+            api_url = auth_config.get('api_url')
+
+            try:
+                github_cfg = ccc.github.github_cfg_for_repo_url(
+                    api_url=api_url,
+                    cfg_factory=self.cfg_factory,
+                    require_labels=(),
+                    github_cfgs=self.github_cfgs,
+                )
+                break
+            except model.base.ConfigElementNotFoundError:
+                continue
+        else:
+            logger.info('no valid credentials found - attempting anonymous-auth')
+            return
+
+        params = {
+            'access_token': github_cfg.credentials().auth_token(),
+            'api_url': github_cfg.api_url(),
+        }
+
+        res = self._session.get(
+            url=self._routes.auth(),
+            params=params,
+            timeout=(4, 31),
+        )
+
+        if not res.ok:
+            logger.warning(
+                'authentication against delivery-service failed: '
+                f'{res.status_code=} {res.reason=} {res.content=}'
+            )
+
+        res.raise_for_status()
+
+        self._bearer_token = res.cookies.get(delivery.jwt.JWT_KEY)
+
+        if not self._bearer_token:
+            raise ValueError('delivery-service returned no bearer token upon authentication')
+
+        # current json web keys are required to verify the signature of the bearer token
+        openid_jwks = self._openid_jwks()
+        self._json_web_keys = [
+            delivery.jwt.JSONWebKey.from_dict(raw_key)
+            for raw_key in openid_jwks.get('keys')
+        ]
+
+    def request(
+        self,
+        url: str,
+        method: str='GET',
+        headers: dict=None,
+        **kwargs,
+    ):
+        self._authenticate()
+
+        headers = headers or {}
+
+        if self._bearer_token:
+            headers = {
+                'Authorization': f'Bearer {self._bearer_token}',
+                **headers,
+            }
+
+        try:
+            timeout = kwargs.pop('timeout')
+        except KeyError:
+            timeout = (4, 31)
+
+        res = self._session.request(
+            method=method,
+            url=url,
+            headers=headers,
+            timeout=timeout,
+            **kwargs,
+        )
+
+        return res
 
     def component_descriptor(
         self,
@@ -132,10 +303,9 @@ class DeliveryServiceClient:
         if version_filter is not None:
             params['version_filter'] = version_filter
 
-        res = self.session.get(
+        res = self.request(
             url=self._routes.component_descriptor(),
             params=params,
-            timeout=(4, 31),
         )
 
         res.raise_for_status()
@@ -166,10 +336,9 @@ class DeliveryServiceClient:
         if version_filter is not None:
             params['version_filter'] = version_filter
 
-        res = self.session.get(
+        res = self.request(
             url=self._routes.greatest_component_versions(),
             params=params,
-            timeout=(4, 31),
         )
 
         res.raise_for_status()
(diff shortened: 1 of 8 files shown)


delivery/client.py
```diff
@@ -109,9 +123,93 @@ class DeliveryServiceClient:
     def __init__(
         self,
         routes: DeliveryServiceRoutes,
+        cfg_factory=None,
     ):
         self._routes = routes
+        self.cfg_factory = cfg_factory or ci.util.ctx().cfg_factory()
```
ccwienk, line 129: I would actually prefer to require the Client's initialiser to accept GitHub-Cfg-Elements (or even bearer-tokens) + add convenience functionality to util functions (e.g. in ccc.delivery)

delivery/client.py
```diff
@@ -109,9 +123,93 @@ class DeliveryServiceClient:
...
         self.session = requests.sessions.Session()
+        self.bearer_token = None
+
+    def _authenticate(self):
+        if self.bearer_token:
```
ccwienk, line 134: nit: should not we also check whether jwt-token is still valid (not expired) and re-authenticate if this is the case?

delivery/client.py
```diff
@@ -109,9 +123,93 @@ class DeliveryServiceClient:
...
+            logger.warning(
+                'authentication against delivery-service failed: '
+                f'{res.status_code=} {res.reason=} {res.content=}'
+            )
+
+        res.raise_for_status()
+
+        self.bearer_token = res.cookies.get(jwt_key)
+
+    def _request(
```
ccwienk, line 183: nit: considering with this change, `_session` should no longer be used directly, I think it might be more intuitive to make this method "public" (remove leading underscore) and the session-object "private", or make both public (+ add docstr explaining when self.session vs self.request ought to be used).

delivery/client.py
```diff
@@ -109,9 +123,93 @@ class DeliveryServiceClient:
...
+        headers: dict=None,
+        **kwargs,
+    ):
+        self._authenticate()
+
+        headers = headers or {}
+
+        headers = {
+            'Authorization': f'Bearer {self.bearer_token}',
+            **headers,
```
ccwienk, line 196: nit: you might inline initialisation (headers = headers or {}) here

ccwienk (changes requested):

delivery/client.py
```diff
@@ -19,12 +19,26 @@
 
 
 logger = logging.getLogger(__name__)
+jwt_key = 'bearer_token'
```
zkdev, line 22: nit: use upper case for constants as described in [PEP Style Guide](https://peps.python.org/pep-0008/#constants)

zkdev (changes requested):

delivery/client.py
```diff
@@ -14,17 +14,41 @@
...
+            self._base_url,
+            'auth',
+            'configs',
+        )
+
+    def auth_public_signature_key(self):
+        return ci.util.urljoin(
+            self._base_url,
+            'auth',
+            'public-signature-key',
```
ccwienk, line 49: Considering there might be different public-keys (1. upon rotation, 2. for different principals), I suggest to prepare for that by adding another path component. We might hardcode a default value (maybe just: `default`) for now, for example (`auth/public-key/default`). The route (on server-side) should interpret the last part as key-id.
We _might_ write (and read) key-id into JW-tokens (although I think this might be considered an insecure practise).

delivery/client.py
```diff
@@ -109,9 +133,123 @@ class DeliveryServiceClient:
     def __init__(
         self,
         routes: DeliveryServiceRoutes,
+        github_configs: collections.abc.Iterable[model.github.GithubConfig]=[],
+        cfg_factory: model.ConfigFactory | None=None,
```
ccwienk, line 137: do we actually need to have the option to pass-in cfg-factory?

The reason I ask (+ would prefer not to) is the following. We will over time (hopefully) reduce the tight integration of delivery-svc-client to eventually mv it from cc-utils into open-component-model-org (and consume it from cc-utils as external dependency). Such that we can offer an easy-to-consume delivery-service-clients as part of ocm-gear.

cfg-factory is highly specific for our cicd environment. which is why we should avoid using it in code we eventually want to factor-out.

requirements.txt
```diff
@@ -35,6 +35,7 @@ oss2==2.18.5
 passlib<2.0.0
 pycryptodome
 pydash<8.0.0
+pyjwt
```
ccwienk, line 38: interesting. did we miss to add this dependency earlier? if so: how come this was not an issue before? 🤔

ccwienk (changes requested):

delivery/client.py
8R0WNI3: As discussed, superseded by OpenID standard configuration.

delivery/client.py
8R0WNI3: At this point in time, I think the configuration factory is too tightly coupled with the ocm-gear, so that it cannot be removed easily. I suggest to postpone factoring-out of it until we have a suitable replacement, to prevent a list of workarounds until that point in time. Wdyt?

ccwienk: agreed

delivery/client.py
```diff
@@ -25,6 +28,26 @@ class DeliveryServiceRoutes:
...
+        )
+
+    def auth_configs(self):
+        return ci.util.urljoin(
+            self._base_url,
+            'auth',
+            'configs',
+        )
+
+    def openid_configuration(self):
```
ccwienk, line 44: add ref to https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfigurationRequest in docstr

delivery/client.py
```diff
@@ -109,9 +132,142 @@ class DeliveryServiceClient:
...
-        self.session = requests.sessions.Session()
+        self.github_configs = github_configs
+        self.cfg_factory = cfg_factory
+
+        self._bearer_token = None
+        self._json_web_keys = None
+        self._session = requests.sessions.Session()
+
+    def _openid_configuration(self):
+        res = self._session.get(
```
ccwienk, line 160: ditto: https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfigurationRequest

delivery/client.py
```diff
@@ -109,9 +132,142 @@ class DeliveryServiceClient:
...
+
+        if not res.ok:
+            logger.warning(
+                'authentication against delivery-service failed: '
+                f'{res.status_code=} {res.reason=} {res.content=}'
+            )
+
+        res.raise_for_status()
+
+        self._bearer_token = res.cookies.get(delivery.jwt.JWT_KEY)
```
ccwienk, line 231: should we check such a cookie was actually returned (and emit an error if this was not the case)?

delivery/jwt.py
```diff
@@ -0,0 +1,192 @@
...
+@dataclasses.dataclass(frozen=True)
+class JSONWebKey:
+    kty: KeyType
+    use: Use
+    alg: Algorithm
+    kid: str
+
+    @property
+    def key(self) -> bytes:
+        # has to be defined by derived classes
```
ccwienk, line 44: raise NotImplemented?

delivery/jwt.py
```diff
@@ -0,0 +1,192 @@
...
+
+    return Crypto.Util.number.bytes_to_long(b)
+
+
+@functools.cache
+def decode_jwt(
+    token: str,
+    verify_signature: bool=True,
+    json_web_key: JSONWebKey | None=None,
+    **kwargs,
```
ccwienk, line 142: add docstr stating this function wraps jwt.decode (esp. so callers know about the allowed kwargs)

ccwienk (commented):
generally lgtm. considering there is quite some new code for creating / serialising. how about adding a unittest for the e2e-scenario (generate key + token, serialise it, then jwt.decode + verify)?

delivery/jwt.py
```diff
@@ -0,0 +1,192 @@
...
+
+import model.delivery
+
+logger = logging.getLogger(__name__)
+JWT_KEY = 'bearer_token'
+
+
+class KeyType(enum.StrEnum):
+    RSA = 'RSA'
+    OCTET_SEQUENCE = 'oct' # used to represent symmetric keys
```
ccwienk, line 22: Considering this (+ subsequent) enums/dataclasses are implemented according to an existing spec, it would make sense to add a module-level docstr stating this

delivery/jwt.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+def encodeBase64url(b: bytes) -> str:
+    url_encoding = base64.urlsafe_b64encode(b)
+
+    return url_encoding.rstrip(b'=').decode('utf-8')
+
+
+def decodeBase64url(s: str) -> bytes:
+    return base64.urlsafe_b64decode(s + '==')
```
zkdev, line 127: why adding padding chars without checking str-length?
IIRC. those are used to ensure the str-length is a multiple of "4".

zkdev (changes requested):

model/delivery.py
```diff
@@ -111,18 +111,22 @@ def deployment_name(self):
...
-                if purpose_label in labels:
-                    return signing_cfg
+        if not purpose_label:
+            yield from [
+                SigningCfg(raw) for raw in cfgs
+            ]
+            return
 
-        return [SigningCfg(raw) for raw in cfgs]
+        for raw_singing_Cfg in cfgs:
```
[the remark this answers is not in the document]
8R0WNI3: Had to read it at least 10 times to find the mistake (apart from the capital `C`) 🎶😅

delivery/jwt.py
8R0WNI3: According to the [documentation](https://docs.python.org/3/library/base64.html#base64.b64decode), an extra padding is being truncated prior to padding check:
> If validate is False (the default), characters that are neither in the normal base-64 alphabet nor the alternative alphabet are discarded prior to the padding check.

Hence, adding the maximum possible padding and letting `b64decode` check on the correct padding will just work fine.

ccwienk: I only I had known that earlier some years back [here](https://github.com/gardener/cc-utils/blob/master/oci/client.py#L34-L45) 🙈

delivery/jwt.py
```diff
@@ -0,0 +1,201 @@
...
+    alg: Algorithm
+    kid: str
+
+    @property
+    def key(self) -> bytes:
+        # has to be defined by derived classes
+        raise NotImplementedError
+
+    @staticmethod
+    def from_dict(data: dict) -> 'JSONWebKey':
```
ccwienk, line 53: nit: you might use (as of python3.12 I think) typing.Self :-)

delivery/jwt.py
```diff
@@ -0,0 +1,201 @@
...
+        return dacite.from_dict(
+            data_class=class_by_algorithm.get(algorithm),
+            data=data,
+            config=dacite.Config(
+                cast=[enum.Enum],
+            ),
+        )
+
+    @staticmethod
+    def from_signing_cfg(signing_cfg: model.delivery.SigningCfg) -> 'JSONWebKey':
```
ccwienk, line 70: ditto

model/github.py
```diff
@@ -201,3 +205,26 @@ def _optional_attributes(self):
             'secondary_authToken',
             'recovery-codes',
         )
+
+
+def find_github_cfg_for_api(
+    api_url: str | collections.abc.Iterable[str],
```
ccwienk, line 211: hmm... I am not fully happy with this helper function. Firstly, it has some overlaps w/ ccc.github's lookup by repo-url, secondly it mixes two aspects (lookup for _one_ vs lookup first matching cfg).

Do you think we might merge the required functionality into ccc.github (or mv from ccc.github into model/github)?

test/delivery/jwt_test.py
```diff
@@ -0,0 +1,105 @@
...
+        signing_cfg=signing_cfg,
+    )
+
+    # token was just created and thus is not expired yet
+    assert delivery.jwt.is_jwt_token_expired(
+        token=token,
+        json_web_keys=[
+            json_web_key,
+        ],
+    ) == False
```
ccwienk, line 63: nit: compare using identity-check (is intead of ==), or just use `assert not <expr>`

test/delivery/jwt_test.py
```diff
@@ -0,0 +1,105 @@
...
+
+    # this is a correct token validation
+    try:
+        validation_failed = delivery.jwt.decode_jwt(
```
ccwienk, line 67: use pytest.raises if you expect an exception. Otherwise, I think there is no need to do explicit exception-handling. unittest execution will abort upon first (unexpected) exception and will thus fail the test.

test/delivery/jwt_test.py
```diff
@@ -0,0 +1,105 @@
...
+            token=token,
+            verify_signature=True,
+            json_web_key=json_web_key,
+            issuer=ISSUER,
+        )
+        validation_failed = False
+    except:
+        validation_failed = True
+
+    assert validation_failed == False
```
ccwienk, line 77: no need to check non-occurrence of an exception explicitly. but still: use identity-comparison or simply `asset not validation_failed` instead of equality-check if comparing w/ global singleton 
values (True, False, None, Enum-Values, ..)

ccwienk (changes requested):

test/delivery/jwt_test.py
```diff
@@ -0,0 +1,105 @@
...
+
+    assert validation_failed == True
+
+    # token validation but with wrong issuer
+    try:
+        validation_failed = delivery.jwt.decode_jwt(
+            token=token,
+            verify_signature=True,
+            json_web_key=json_web_key,
+            issuer=f'{ISSUER}s',
```
ccwienk, line 99: nit: while it is technically correct to change issue by appending a single letter, how about making the code more self-descriptive, and choose an issuer name like `wrong-issuer`?

PR #980, opened 2024-06-05
Refactor signing

ccwienk:


cosign.py
```diff
@@ -16,48 +15,39 @@
 ci.log.configure_default_logging()
 logger = logging.getLogger(__name__)
 
-COSIGN_SIGNATURE_TYPE = "gardener.vnd/oci/cosign-signature"
```
8R0WNI3: nit: maybe keep this constant and also reference it in `cosign_test.py`?

8R0WNI3 (approved):
/lgtm

cosign.py
ccwienk: We might. OTOH, duplicating the mimetype in unittest could actually considered to be better, as this will fail the test if the mimetype were to be changed unintentionally (typically, such values very rarely change)

ccwienk:
@8R0WNI3 : thx for your review. I added some more refactorings (actually a re-implementation of signing-server-client + signature-creation/publishing)

PR #983, opened 2024-06-17
support helmcharts in publish-trait

ccwienk:
Extend publish-trait to also allow creation and publishing of helmcharts from pipelines. Let users define mappings between built images (via OCM Component-Descriptor) and helmcharts to ensure consistency (between published images and helmcharts' default values), and also allow for later changes of image-references.

The helmchart's version will be set to pipelinerun's effective version (as defined via `version` trait). It will be included as a resource in component-descriptor.

Example usage (inspired by
https://github.com/gardener/gardener/tree/master/charts/gardener/operator):

In this example, a helmchart is expected in relative path `charts/my-chart`. Due to specified mappings, the attributes `image.repository` and `image.tag` in values.yaml will be patched to match corresponding values (repository, tag) from built image "my-image".

```
traits:
  publish:
    dockerimages:
      my-image:
        image: europe-docker.pkg.dev/gardener-project/snapshots/my-image

    helmcharts:
      - name: my-chart
        dir: charts/my-chart # expected to be relative path in repo 
                                            # containing pipeline-definition containing
                                            # chart-dir
        registry: europe-docker.pkg.dev/gardener-project/snapshots/charts
        mappings:
        - ref: ocm-resource:my-image.repository
          attribute: image.repository
        - ref: ocm-resource:my-image.tag
          attribute: image.tag
```

Fixes #

WIP: need to include mapping as additional release-artefact

### DIFF
--- a/concourse/model/traits/publish.py
+++ b/concourse/model/traits/publish.py
@@ -9,9 +9,11 @@ import textwrap
 import typing
 
 import dacite
+import jsonpath_ng
 
 from ci.util import not_none
 from model import NamedModelElement
+import concourse.model.traits.component_descriptor
 import gci.componentmodel as cm
 import oci.model as om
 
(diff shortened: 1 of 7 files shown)


concourse/model/traits/publish.py
```diff
@@ -457,6 +525,16 @@ def oci_builder(self) -> OciBuilder:
     def use_buildkit(self) -> bool:
         return not self.raw['no-buildkit']
 
+    @property
+    def helmcharts(self) -> tuple[HelmchartCfg]:
+        return tuple([
```
8R0WNI3, line 530: Why parsing to `list` prior to parsing to `tuple`?

concourse/steps/component_descriptor.mako
```diff
@@ -165,6 +170,26 @@ component_v2.resources.append(
...
+% for helmchart in helmcharts:
+<%
+name = helmchart.name
+target_ref_prefix = f'{helmchart.registry}/{name}'
+%>
+component_v2.resources.append(
+  cm.Resource(
+    name='${helmchart.name}',
+    version=effective_version, # always inherited from component
+    type='helmchart',
```
8R0WNI3, line 183: Why not re-using `cm.ResourceType.HELM_CHART` (`helmChart/v1`)?

concourse/steps/helmcharts.mako
```diff
@@ -0,0 +1,191 @@
...
+with open(values_yaml_path, 'w') as f:
+  yaml.safe_dump(values, f)
+
+charts_yaml = os.path.join(helmchart_dir, 'Chart.yaml')
+with open(charts_yaml) as f:
+  charts_yaml_values = yaml.safe_load(f)
+
+if charts_yaml_values.get('name') != helmchart_name:
+  logger.warning(f'{charts_yaml}.name differs from {helmchart_name} - patching')
+  charts_yaml_values['name'] = helmchart_name
```
8R0WNI3, line 105: Shouldn't we also patch the chart's version?

concourse/steps/helmcharts.mako
```diff
@@ -0,0 +1,191 @@
...
+    mediaType='application/vnd.cncf.helm.config.v1+json',
+    size=cfg_blob_leng,
+  ),
+  layers=[oci.model.OciBlobRef(
+    digest=digest,
+    mediaType='application/vnd.cncf.helm.chart.content.v1.tar+gzip',
+    size=leng,
+  )],
+  mediaType='application/vnd.oci.image.manifest.v1+json',
+  )
```
8R0WNI3, line 184: nit: remove indentation

concourse/model/traits/publish.py
```diff
@@ -344,6 +346,65 @@ class OciBuilder(enum.Enum):
...
+@dataclasses.dataclass
+class HelmchartCfg:
+    '''
+    A configuration for a helmchart that should be published. Data is also injected into
+    component-descriptor (if trait is present).
+
+    name: Chart's name (also used as resource-name for component-descriptor); must be unique
+    registry: OCI Registry prefix to publish helmchart to (<name>:<version> will be appended)
+    dir: relative path to directory containing helmchart
+    values: attributes to patch in values.yaml
```
8R0WNI3, line 400: typo: parameter name `values` -> `mappings`

concourse/steps/helmcharts.mako
```diff
@@ -187,5 +189,29 @@ oci_client.put_manifest(
...
+
+## dump mapping as suggested by OCM-CLI:
+## https://github.com/open-component-model/ocm/blob/2b9ed814dee16e351636cb0d4ea0203f72224c0d/components/helmdemo/README.md
+mapping_outpath = 'helmcharts/${helmchart_cfg.name}.mapping.json'
+print(f'writing  mapping to {mapping_outpath}')
+<%
+# <resource-name>: {tag: attr, repository: attr, image: attr}
+attrs_by_resource = collections.defaultdict(dict)
+for mapping in helmchart_cfg.mappings:
+  resource_name, ref_name = mapping.ref.removeprefix('ocm-resource:').split('.')
```
8R0WNI3, line 201: seen this multiple times now, maybe add a util function to parse `ref`?

8R0WNI3 (changes requested):

concourse/model/traits/publish.py
ccwienk: good point. changed to generator-expression

concourse/steps/component_descriptor.mako
ccwienk: valid ask. I used the same type-str as used by ocm-cli if asking it to package a helmchart (ocm resource add). Although, according to [source](https://github.com/open-component-model/ocm/blob/main/pkg/contexts/ocm/resourcetypes/const.go#L16), I would actually expect OCM-CLI to acually use `helmChart` as typename.

@mandelsoft | @fabianburth : Could you shed some light into which resource-typename we should use for helmcharts so we are consistent w/ ocm-cli?

@8R0WNI3 : I suggest to change to helmChart for now, and change later, depending on Uwe/Fabian's responses.

concourse/steps/helmcharts.mako
ccwienk: that attribute is patched by `helm package` via the `--version`-argument, if I am not mistaken.

8R0WNI3: ah that makes sense, ty

8R0WNI3 (approved):

PR #985, opened 2024-06-18
Use common model-class for malware findings

zkdev:

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/dso/model.py
+++ b/dso/model.py
@@ -283,38 +283,16 @@ class RescoringLicenseFinding:
 
 
 @dataclasses.dataclass(frozen=True)
-class MalwareFindingBase():
+class ClamAVMalwareFinding(Finding):
     filename: str
     content_digest: str
     virus_name: str
-
-
-@dataclasses.dataclass(frozen=True)
-class OCIMalwareFinding(MalwareFindingBase):
-    layer_digest: str
-
-    @property
-    def key(self) -> str:
-        return f'{self.content_digest}|{self.filename}|{self.layer_digest}|{self.virus_name}'
-
-
-@dataclasses.dataclass(frozen=True)
-class S3MalwareFinding(MalwareFindingBase):
-
-    @property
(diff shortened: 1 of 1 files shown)


8R0WNI3 (approved):

dso/model.py
```diff
@@ -283,38 +283,17 @@ def key(self) -> str:
 
 
 @dataclasses.dataclass(frozen=True)
-class MalwareFindingBase():
+class ClamAVMalwareFinding(Finding):
+    layer_digest: str | None # oci-image specific
```
ccwienk, line 287: wo do not always have a _layer_-digest, do we? how about we either omit it (might be included in filename), or make it more generic (e.g. artefact_digest, or similar)

ccwienk (changes requested):
see inline comment (published too early)

dso/model.py
zkdev: the layer-digest is specific to oci-artefacts
e.g. gardenlinux stores (some) of its artefacts in s3 buckets without oci manifests

zkdev: > we do not always have a layer-digest, do we
We do not always have a layer-digest.

zkdev: We could make it more generic to allow re-use, I don't have a strong opinion here.

ccwienk: in that case, I suggest to rename it (ideally also fill value for s3-artefacts). do you happen to know for which purpose we actually keep layer-digest? as mentioned before, I think we could include layer-digest in path (filename), which should be "good enough"; I do not recall storing layer-digest to yield an additional benefit 🤔

zkdev: > do you happen to know for which purpose we actually keep layer-digest?

I think we never really discussed this in detail, but I think this is required due to overwrite semantics of multiple oci layers for the very same filepath

PR #986, opened 2024-06-19
Cosign append

ccwienk:
append to and keep track of cosign signatures
    
Add new `append` signing mode for cosign-OCI Image signing, and make it
the new default. Keep track of used signing algorithm and public key in
preparation for usage of signing algorithms that will yield different
signatures for repeated signing of unchanged payload and private keys
(e.g. rssa-pss; not currently supported by cosign, though).
    
With this change, changes of private key or signing algorithm will lead
to new signatures to be appended to existing signature artefacts, thus
allowing for key rotation. Signing code will avoid appending redundant
signatures so it is safe for callers to omit such checks.


cli/gardener_ci/_cosign.py
```diff
@@ -44,8 +44,9 @@ def sign(
     signing_server_url: str,
     signing_server_client_cert: str,
     signing_server_client_cert_key: str,
+    on_exist: str=cosign.OnExist.APPEND,
```
8R0WNI3, line 47: nit: rather use type hint `cosign.OnExist | str` or just `cosign.OnExist`

cosign.py
```diff
@@ -102,13 +128,18 @@ def sign_image(
         ).hexdigest()
         image_reference = f'{image_reference.ref_without_tag}@sha256{digest}'
 
-    if on_exist is OnExist.SKIP:
+    if on_exist in (OnExist.SKIP, OnExist.APPEND):
         if oci_client.head_manifest(
```
8R0WNI3, line 132: Maybe use `exists = bool(oci_client.head_manifest(...))`. In general, I think patterns which set a variable to `True` if a condition is true and to `False` if the condition is false should be prevented if possible.

cli/gardener_ci/_cosign.py
```diff
@@ -89,7 +90,9 @@ def sign(
     cosign.sign_image(
         image_reference=image_reference,
         signature_image_reference=signature_image_ref,
-        signature=signature.signature,
-        on_exist=cosign.OnExist.OVERWRITE,
+        signature=signature.signature.replace('\n', ''),
+        signing_algorithm=signing_algorithm,
+        public_key=signature.public_key.replace('\n', ''),
+        on_exist=cosign.OnExist(on_exist),
```
8R0WNI3, line 96: Parsing to enum could be omitted here since `sign_image` also accepts string values and checks if it is part of the enum.

cosign.py
```diff
@@ -118,57 +149,116 @@ def sign_image(
...
-        octets_count=payload_size,
-        data=payload,
-    )
+    if on_exist is OnExist.APPEND and exists:
+        manifest = oci_client.manifest(
+            image_reference=signature_image_reference,
+        )
+        for layer_idx, layer in enumerate(manifest.layers):
+            a = layer.annotations
+            existing_signature = a.get('dev.cosignproject.cosign/signature', None)
```
8R0WNI3, line 158: I'd prefer to factor-out this annotation as well (just like the public key and signature annotation). Or you could also introduce a new enum like `SignatureAnnotations` or something like that.

cosign.py
```diff
@@ -118,57 +149,116 @@ def sign_image(
...
+            octets_count=cfg_blob_size,
+            data=cfg_blob,
+        )
+
+    signature_layer = om.OciBlobRef(
+        digest=payload_digest,
+        size=payload_size,
+        mediaType='application/vnd.dev.cosign.simplesigning.v1+json',
+        annotations={
+            'dev.cosignproject.cosign/signature': signature,
```
8R0WNI3, line 239: You could re-use the constant/enum here then

cosign.py
ccwienk: okay, changed accordingly. this saves two lines of code, while making it very slightly less readable (imho). which is okay in such simple cases.

why do you think 
```
if <condition>:
  marker = True
else:
  marker = False
```
should be avoided? I think it is very easy to read, albeit not very elegant (-> KISS)

8R0WNI3: I think this is just duplicated code, since the condition is equivalent to the value of `marker` and I think duplicated code should be avoided unless it has any extra value. If you think it is more readable to choose your variant, I'm fine to keep it since the extra value is a better readability then.

8R0WNI3 (approved):
/lgtm

zkdev (approved):
/lgtm

PR #987, opened 2024-07-01
Add support for build metadata in version string

8R0WNI3:
To be compatible with SemVer, we should be able to handle versions which include optional build metadata
([see](https://semver.org/#spec-item-10)). In general, this is already the case. However, (some) OCI registries don't allow `+` in their tags, which is used as separator for these build metadata. That's why, the version has to be sanitised prior to uploading to a registry or after retrieving it from there.

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/ccc/oci.py
+++ b/ccc/oci.py
@@ -1,3 +1,4 @@
+import collections.abc
 import functools
 import logging
 import traceback
@@ -67,6 +68,8 @@ def oci_client(
     install_logging_handler: bool=True,
     cfg_factory=None,
     http_connection_pool_size:int=16,
+    tag_preprocessing_callback: collections.abc.Callable[[str], str]=None,
+    tag_postprocessing_callback: collections.abc.Callable[[str], str]=None,
 ) -> oc.Client:
     def base_api_lookup(image_reference):
         registry_cfg = model.container_registry.find_config(
@@ -107,6 +110,8 @@ def oci_client(
         credentials_lookup=credentials_lookup,
         routes=routes,
         session=session,
+        tag_preprocessing_callback=tag_preprocessing_callback,
+        tag_postprocessing_callback=tag_postprocessing_callback,
     )
 
 
--- a/oci/client.py
+++ b/oci/client.py
@@ -1,4 +1,5 @@
 import base64
+import collections.abc
 import dataclasses
 import datetime
 import enum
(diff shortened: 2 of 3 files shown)


oci/client.py
```diff
@@ -701,7 +707,10 @@ def tags(self, image_reference: str):
             method='GET'
         )
 
-        return res.json()['tags']
+        return [
```
ccwienk, line 710: it might be worthwhile to define this conversion in oci/util.py (as a function). while code duplication might not be that important for a trivial str-replacement, it would slightly make code more self-describing, and would allow us to document why it is required at said function.

in addition, I am not quite sure whether OCI-Client is the best place for this, as it is actually slightly too low-level. semver/version-handling is actually done on OCM-level (the oci-client thus far handles content stored in OCI-registries as opaque).

ccc/oci.py
```diff
@@ -78,7 +79,10 @@ def base_api_lookup(image_reference):
             return base_url
         return oc.base_api_url(image_reference)
 
-    routes = oc.OciRoutes(base_api_lookup)
+    routes = oc.OciRoutes(
+        base_api_url_lookup=base_api_lookup,
+        sanitise_tags=sanitise_tags,
```
ccwienk, line 84: hmmm... do you think it is a good match to store this as part of the routes? For manifest-route, it _is_ part of URL. otoh, for `tags` method, it is not. -> I think it would be cleaner to either make this a parameter of oci-client, or wrap it into a second "OciOptions" cfg object (although that might be a bit too heavyweight)

it would also be nice to document semantics of this flag (also) at params for oci.client's initialiser (__init__)

oci/util.py
```diff
@@ -42,6 +44,27 @@ def urljoin(*parts):
     return '/'.join([first] + middle + [last])
 
 
+def sanitise_tag(tag: str) -> str:
+    '''
+    Additional build metadata as defined in SemVer can be added via `+` to the version. However,
```
ccwienk, line 49: nit: considering this package is about OCI, it may a bit off to start talking about semver (from OCI's perspective, tags are just strings (with some limitations to allowed characters).

which makes it slightly problematic (imo) to add such logic to oci package :-/

maybe it would make more sense to pass-in a tag-processing callback (similar to git's clean/smudge callbacks)

8R0WNI3: Changed it to callbacks and moved the respective sanitisation functions to the `cnudie.util` module instead.

ccwienk (approved):

PR #989, opened 2024-07-01
fix: properly validate uniqueness of resource-names

ccwienk:
Remove old codepath, originally intended for kaniko (+ remove check for extra-push-target, which was a limitation for (removed) kaniko-builder). Instead, validate uniqueness of resource-names.

### DIFF
--- a/concourse/model/traits/publish.py
+++ b/concourse/model/traits/publish.py
@@ -566,21 +566,22 @@ class PublishTrait(Trait):
             raise ModelValidationError(
                 'must not specify empty list of platforms (omit attr instead)'
             )
-        resource_names = set()
-        if not self.oci_builder() in (OciBuilder.DOCKER, OciBuilder.DOCKER_BUILDX):
-            for image in self.dockerimages():
-                resource_names.add(image.name)
-                if image.extra_push_targets:
-                    raise ModelValidationError(
-                        f'must not specify extra_push_targets if using {self.oci_builder()}'
-                    )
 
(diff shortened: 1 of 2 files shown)


concourse/model/traits/publish.py
```diff
@@ -566,19 +570,15 @@ def validate(self):
...
+            resource_names.add(image._base_name)
 
         for helmchart in self.helmcharts:
-            if helmchart.name in resource_names:
+            if helmchart.resource_name in resource_names:
                 raise ModelValidationError(
-                    f'duplicate resource: {helmchart.name=}',
+                    f'duplicate resource: {helmchart.resource_name=}',
                 )
             resource_names.add(helmchart.name)
```
8R0WNI3, line 583: Also change to `helmchart.resource_name`

concourse/steps/component_descriptor.mako
```diff
@@ -178,7 +178,7 @@ target_ref_prefix = f'{helmchart.registry}/{name}'
 %>
```
8R0WNI3, line 178: Ditto -> `name = helmchart.resource_name`

8R0WNI3 (changes requested):

concourse/model/traits/publish.py
ccwienk: unless I am not seeing something - is this not exactly already the case? 🤔

8R0WNI3 (approved):

PR #992, opened 2024-07-05
Don't store component version for BDBA scan results

8R0WNI3:
We do already re-use existing BDBA scans for an artefact version in case only the component version has changed. However, we have still stored the scan results for each component version separately, resulting in duplicated findings in case the artefact version did not change. To deduplicate those findings, don't store the component version anymore. However, this also requires downstream adjustments since with this change it is not possible anymore to retrieve all findings for a component version directly, but instead the included artefact versions must be specified as well.

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/protecode/util.py
+++ b/protecode/util.py
@@ -53,14 +53,13 @@ def iter_artefact_metadata(
     license_cfg: image_scan.LicenseCfg=None,
     delivery_client: delivery.client.DeliveryServiceClient=None,
 ) -> collections.abc.Generator[dso.model.ArtefactMetadata, None, None]:
-    now = datetime.datetime.now()
+    now = datetime.datetime.now(tz=datetime.timezone.utc)
     discovery_date = datetime.date.today()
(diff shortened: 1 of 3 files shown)


protecode/util.py
```diff
@@ -53,14 +54,23 @@ def iter_artefact_metadata(
...
     discovery_date = datetime.date.today()
     datasource = dso.model.Datasource.BDBA
 
-    artefact = gcm.artifact_from_node(node=scanned_element)
+    # don't include component version here since it is also not considered in the BDBA scan
+    # -> this will deduplicate findings of the same artefact version across different
+    # component versions
+    component = dataclasses.replace(scanned_element.component, version=None)
+    resource = scanned_element.resource
+    node = cnudie.iter.ResourceNode(
```
zkdev, line 66: can't we just patch scanned_element path instead?

zkdev (changes requested):

ccwienk (approved):

PR #995, opened 2024-07-15
Rm clamav related coding

zkdev:
Malware scan capabilities (based on ClamAV) were migrated to OCM-Gear extension.

Fixes #

**Release note**:

```other operator

```


ccwienk (approved):
<3

8R0WNI3 (approved):

concourse/model/traits/filter.py
```diff
@@ -1,155 +0,0 @@
-# SPDX-FileCopyrightText: 2024 SAP SE or an SAP affiliate company and Gardener contributors
```
8R0WNI3, line 1: This module is still used in delivery-service

ccwienk: @8R0WNI3 : it _is_? in that case, I suggest we remove usage asap. Code from concourse package is typically (exception: concourse/client.py) not intended for reuse outside of pipeline-template

ccwienk (changes requested):
.

PR #997, opened 2024-07-22
Add support for merged diki report

AleksandarSavchev:
Merged diki reports contain compliance data from more than 1 evaluated instance. The targets of the merged checks is a map with keys the id of a specific instance and value the found targets of that instance.

Fixes #

**Release note**:

```other operator
NONE
```


8R0WNI3 (approved):
/lgtm

PR #998, opened 2024-07-23
Do not risk inconsistencies during component updates

zkdev:
Rebasing git repository is not sufficient as component updates also consider component version range for release notes. Rather fail current component update build and trigger resource check, so next iteration will succeed.

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/concourse/steps/update_component_deps.mako
+++ b/concourse/steps/update_component_deps.mako
@@ -12,6 +12,7 @@ import gci.componentmodel as cm
 
 main_repo = job_variant.main_repository()
 repo_name = main_repo.repo_name()
+repo_hostname = main_repo.repo_hostname()
 repo_relpath = main_repo.resource_name()
 repo_owner = main_repo.repo_owner()
 repo_branch = main_repo.branch()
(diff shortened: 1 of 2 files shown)


concourse/steps/update_component_deps.mako
```diff
@@ -101,16 +101,6 @@ pull_request_util = github.util.PullRequestUtil(
...
-## hack / workaround: rebase to workaround concourse sometimes not refreshing git-resource
-git_helper = gitutil.GitHelper(
-    repo=REPO_ROOT,
-    github_cfg=github_cfg,
-    github_repo_path=f'{REPO_OWNER}/{REPO_NAME}',
-)
-git_helper.rebase(
-    commit_ish=REPO_BRANCH,
-)
-
```
ccwienk, line 113: as discussed: I am fine w/ simply removing rebase-code. however, we might trigger resource-check (to increase chance for next run to succeed)

zkdev: Yes, triggering resource-check seems like a good approach.
However, I still have to check auto-merge behaviour upon merge conflict, especially error handling. That's why PR is still on draft :-)

zkdev:
We could also consider to check whether there will be merge conflicts prior to opening PR so there will be less noise, however I am not sure as this seems more error prone to me.

concourse/steps/update_component_deps.py
```diff
@@ -563,6 +593,25 @@ def  _merge_pr(
                 f'Unable to merge upgrade pull request #{pull_request.number} '
                 f'({pull_request.html_url}).'
             )
+            pull_request.create_comment(
+                'Unable to auto-merge PR. '
```
ccwienk, line 597: nit: I still prefer multiline-strings..

ccwienk (approved):
nit wr.t. codestyle, otherwise lgtm

PR #1000, opened 2024-07-24
Refactor ctt's `process_images`

8R0WNI3:

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/gci/oci.py
+++ b/gci/oci.py
@@ -1,4 +1,5 @@
 import dataclasses
+import hashlib
 import io
 import logging
 import os
@@ -6,7 +7,10 @@ import tarfile
 import typing
 import yaml
 
+import ccc.oci
 import gci.componentmodel
+import oci.client
+import oci.model
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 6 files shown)


gci/oci.py
```diff
@@ -103,3 +106,26 @@ def component_descriptor_from_tarfileobj(
           raise ValueError('Component Descriptor appears to be empty')
 
         return gci.componentmodel.ComponentDescriptor.from_dict(raw_dict)
+
+
+def image_ref_with_digest(
```
ccwienk, line 111: I think this function might better fit into oci.util (w/o DigestSpec, of course) - wdyt?

protecode/scanning.py
```diff
@@ -226,7 +227,7 @@ def scan_requests(
                     logger.info(f'{group_name=}: did not find old scan')
 
                 def iter_content():
-                    image_reference = protecode.util.image_ref_with_digest(
```
ccwienk, line 229: good catch..

ccwienk (commented):
not too sure about gci/oci (but I have not too strong opinion) - otherwise lvgtm

gci/oci.py
8R0WNI3: If you move it to `oci.util`, there would be some obstacles because of circular imports. I'd suggest to keep it in `gci/oci` since it contains both OCM and OCI related aspects 🤔

ccwienk (approved):

PR #1001, opened 2024-07-25
allow adding plain files as release-assets

ccwienk:
Make it configurable whether release-assets should be wrapped into a tape-archive (aka tar). Also, improve documentation.

### DIFF
--- a/concourse/model/traits/release.py
+++ b/concourse/model/traits/release.py
@@ -96,6 +96,35 @@ class TagConflictAction(enum.StrEnum):
 class Asset:
     '''
     base class for release assets. Not intended to be instantiated
+
+    Assets are references to data from build (build log, files, or directory trees) that are
+    to be included in Component-Descriptor as resources. Each asset is added as one resource.
+
+    Parameters for OCM Resource:
+
+    - ocm_labels: OCM-Labels to be added to resource
+    - name: resource name
+    - artefact_type: resource type
+    - purposes: value for label `gardener.cloud/purposes`
+
+    Other Parameters:
+
+    - step_name: pipelinestep-name of data-source
+    - comment: unused
+    - type: asset-type - overwritten by subclasses
+
+    Purpose-labels will be added as OCM-Label `gardener.cloud/purposes`.
+    They are used to identify (in a machine-readable manner) the purpose / semantics
+    of the included assets. Values are not (yet) standardised. It is recommended to use the
+    following labels (multiple may be specified):
+
+    build
+    codecoverage
+    integrationtest
+    linter
+    sast          static code analysis results
+    test
+    unittest
     '''
     ocm_labels: list[dict[str, object]] = dataclasses.field(default_factory=list)
     type: str = None # must overwrite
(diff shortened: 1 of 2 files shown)


concourse/model/traits/release.py
```diff
@@ -124,23 +173,6 @@ class BuildstepLogAsset(Asset):
     - type
     - target
```
8R0WNI3, line 174: I think `target` is obsolete by now

concourse/model/traits/release.py
```diff
@@ -124,23 +173,6 @@ class BuildstepLogAsset(Asset):
...
-
-    build
-    codecoverage
-    integrationtest
-    linter
-    sast          static code analysis results
-    test
-    unittest
 
     Purpose-labels will be added as OCM-Label:
```
8R0WNI3, line 177: Maybe this should also be moved to base class docstring?

concourse/model/traits/release.py
```diff
@@ -328,6 +360,9 @@ def deserialise(raw_asset: dict):
             return dacite.from_dict(
                 data_class=data_class,
                 data=raw_asset,
+                config=dacite.Config(
+                    cast=(enum.Enum,),
```
8R0WNI3, line 364: Does this work actually? I think you have to cast to `FileAssetMode` 🤔

concourse/steps/release.mako
```diff
@@ -107,6 +109,13 @@ import github.util
 import gitutil
 
 import git
+try:
+  import magic
+except ImportError:
+  # hack: install libmagic as workaround until libmagic will be installed into job-image
+  # -> remove this again
```
8R0WNI3, line 116: `libmagic` is installed into job-image by now -> https://github.com/gardener/cc-utils/commit/108702b5a0f7c9cedf09524bde863cd49a1aad0a

8R0WNI3 (changes requested):

concourse/model/traits/release.py
ccwienk: well, I tested it (and also pass enum.Enum at several other invocations of dacite.from_dict:

```
ccc/grafeas_model.py:129:                cast=[Enum],
concourse/model/traits/component_descriptor.py:285:                cast=(enum.Enum,),
concourse/model/traits/release.py:364:                    cast=(enum.Enum,),
concourse/steps/component_descriptor.mako:332:  config=dacite.Config(cast=(enum.Enum,)),
# ... (there are more, but I think those examples should do) :-)
```

concourse/steps/release.mako
ccwienk: good catch - removed

concourse/model/traits/release.py
ccwienk: done

concourse/model/traits/release.py
8R0WNI3: hmm, and I wonder how dacite is able to find the correct enum class 🤔

ccwienk: well that's easy, is it not? The concrete enum is added as typehint (as for all other types as well). To me, it is strange why dacite will not by default cast enums.

Consider the following code:
```
class E(enum.Enum):
 X = 'x'


@dataclasses.dataclass
class D:
 a: E
```

Then the following expression will create an instance of E:

```
D.__dataclass_fields__['a'].type('x')
```

-> as with all typehints for dataclass-attrs, dacite can simply use them (provided the values _are_ types) to "cast" to enum by invoking them.

So to my understanding, by passing `cast=[Enum]` to dacite, we merely tell it that it _should_ try such a cast. We do not tell it _how_ to cast (that we could to using type_hooks parameter). However, I do not quite understand _why_ dacite will not by default also try enum-casting.

8R0WNI3 (approved):

PR #1004, opened 2024-08-01
fix: guard against adding step-output multiple times

ccwienk:
If multiple release-assets are added, we must not add input multiple times (as, at least currently, doing so will result in a validation error).

It might also be considered to simply remove said check; which was originally intended to detect programming or configuration errors.

### DIFF
--- a/concourse/model/step.py
+++ b/concourse/model/step.py
@@ -422,8 +422,8 @@ class PipelineStep(ModelBase):
         ci.util.not_none(name)
         ci.util.not_none(variable_name)
 
-        if variable_name in self._inputs_dict:
-            raise ValueError(f'input already exists: {variable_name}')
+        if variable_name in self._inputs_dict and name != self._inputs_dict[variable_name]:
+            raise ValueError(f'conflict: {variable_name} was tried to be reset to different value')
         self._inputs_dict[variable_name] = name
 
     def remove_input(self, name):


concourse/model/traits/release.py
```diff
@@ -489,6 +489,7 @@ def process_pipeline_args(self, pipeline_args: JobVariant):
             if not pipeline_args.has_step(asset.step_name):
                 raise ValueError(f'{asset=}\'s step_name refers to an absent build-step')
 
+            seen_inputs = set()
```
8R0WNI3, line 492: Hmm, I do not really like the idea that it would cause an error if you manually define an output directory as input + use the output directory for additional release assets. I'd maybe either fill the initial set of `seen_inputs` with the existing inputs of the release step or vote, as you already suggest, to simply remove said check. Wdyt?

zkdev: Why would this cause an error?
With this change the output dir would be added as input, once.

It is simply mounted to the next task-container, containing everything an user added to said dir 🤔 
When using additional release assets the user must anyway ensure that desired files are moved to output dir.

8R0WNI3: Yeah okay I was wrong here, I figured it is possible to define `inputs` also for the release trait, which might have caused an error here if I'm not mistaken. However, as manually defined `inputs` don't seem to be allowed for the release trait, this should not be an issue.

8R0WNI3 (approved):

PR #1005, opened 2024-08-01
Add unit test for cosign `sign_image` function

8R0WNI3:

Fixes #

**Release note**:

```other operator

```


cosign.py
```diff
@@ -98,7 +98,9 @@ def sign_image(
     on_exist: OnExist|str=OnExist.APPEND,
     signature_image_reference: str=None,
     oci_client: oc.Client=None,
-):
+    manifest: om.OciImageManifest=None,
+    dry_run: bool=False,
```
ccwienk, line 102: hmmm :-/
How about passing-in a mocked oci_client, instead? As an alternative, we might factor-out the signature-creation-part (however, I think mocking might actually not be a bad approach here)

ccwienk (changes requested):

test/cosign_test.py
```diff
@@ -3,48 +3,233 @@
...
+import dacite
 import pytest
+import requests.exceptions
 
 import cosign
+import model.signing_server
+import oci.model
+
+
+class OciClientMock:
```
ccwienk, line 20: how about using `unittest.mock` intead of a custom mock? this would have the advantage that it has builtin method-calling-capturing-capabilities (see https://docs.python.org/3/library/unittest.mock.html#quick-guide), which can be quite useful in unittests

test/cosign_test.py
```diff
@@ -3,48 +3,233 @@
...
+    3. Re-signing using `rsassa-pkcs1-v1-5` algorithm and `public_key_1`
+    4. Re-signing using `rsassa-pss` algorithm again but with `public_key_2` and changed signature
+    '''
+    oci_client = OciClientMock()
+    signature_image_reference = cosign.default_signature_image_reference(image_reference)
+
+    signature = 'rsassa-pss-signature-dummy'
+    signing_algorithm = model.signing_server.SigningAlgorithm.RSASSA_PSS
+
+    cosign.sign_image(
```
ccwienk, line 169: seeing the unittest code, I wonder whether we should really split `sign_image` into `sign_manifest` (which expects a manifest-object as input, and returns the signed manifest. we can keep sign_image as a "convenience-function" that will include the retrieval and upload of image-manifest. wdyt?

test/cosign_test.py
8R0WNI3: By factoring-out a `sign_manifest` function, there is no need for an `oci_client`-mock for the unit test anymore.

ccwienk (approved):

test/cosign_test.py
```diff
@@ -3,48 +3,235 @@
...
+        payload_size=payload_size,
+        payload_digest=payload_digest,
+        signature=signature,
+        signing_algorithm=signing_algorithm,
+        public_key=public_key_1,
+        on_exist=cosign.OnExist.APPEND,
+    )
+
+    assert len(manifest_different_algorithm.layers) == 2
+    signature_layer = manifest_different_algorithm.layers[1]
```
ccwienk, line 213: we might also re-check existing signatures are not change. but I think this should be good-enough like this

8R0WNI3: As part of the integration test, we also check if the public key of the first signature is still able to verify the signature after appending a new one.

PR #1010, opened 2024-08-15
OCI nits

ccwienk:
Set default-mimetypes to match recommendations from OCI-Distribution-Spec, and actually pass-through mimetype-arguments. Also, add docstr + fix typehints.

### DIFF
--- a/oci/client.py
+++ b/oci/client.py
@@ -942,7 +942,7 @@ class Client:
         image_reference: typing.Union[str, om.OciImageReference],
         digest: str,
         octets_count: int,
-        data: requests.models.Response,
+        data: requests.models.Response | collections.abc.Generator | bytes | typing.IO,
         max_chunk=1024 * 1024 * 1, # 1 MiB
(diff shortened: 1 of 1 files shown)


oci/client.py
```diff
@@ -942,7 +942,7 @@ def put_blob(
         image_reference: typing.Union[str, om.OciImageReference],
         digest: str,
         octets_count: int,
-        data: requests.models.Response,
+        data: requests.models.Response | typing.Generator | bytes | typing.IO,
```
8R0WNI3, line 945: Please use `collections.abc.Generator` instead ([PEP 585](https://peps.python.org/pep-0585/)) 🙃

8R0WNI3 (approved):
Just a nit, otherwise /lgtm

oci/client.py
ccwienk: done. I really hate this lengthy package-name, but have to acknowledge pep-585 😤

PR #1011, opened 2024-08-15
Add option to set timestamp as prerelease in version trait

8R0WNI3:

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/concourse/steps/component_descriptor.py
+++ b/concourse/steps/component_descriptor.py
@@ -45,6 +45,9 @@ def base_component_descriptor_v2(
     if parsed_version.finalize_version() == parsed_version:
         # "final" version --> there will be a tag, later (XXX hardcoded hack)
         src_ref = f'refs/tags/{effective_version}'
+    elif parsed_version.prerelease.startswith('timestamp'):
+        # prerelease is the build timestamp, not a "real" ref
+        src_ref = None
     else:
         # let's hope the version contains something committish
         if parsed_version.build:
--- a/concourse/steps/version.mako
+++ b/concourse/steps/version.mako
@@ -1,5 +1,6 @@
 <%def name="version_step(job_step, job_variant, indent)", filter="indent_func(indent),trim">
 <%
+import datetime
 import os
 from makoutil import indent_func
 from concourse.steps import step_lib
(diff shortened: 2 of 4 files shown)


concourse/steps/component_descriptor.py
```diff
@@ -45,6 +45,9 @@ def base_component_descriptor_v2(
     if parsed_version.finalize_version() == parsed_version:
         # "final" version --> there will be a tag, later (XXX hardcoded hack)
         src_ref = f'refs/tags/{effective_version}'
+    elif parsed_version.prerelease.startswith('timestamp'):
+        # prerelease is the build timestamp, not a "real" ref
+        src_ref = None
```
ccwienk, line 50: good catch! + thanks for adding a (very helpful) comment

concourse/steps/version.mako
```diff
@@ -43,6 +44,9 @@ elif version_operation in ('finalize-skip-patchlevel-zero', 'finalise-skip-patch
   version_operation_kwargs['skip_patchlevel_zero'] = True
 elif version_operation == 'noop':
   version_operation_kwargs['operation'] = 'noop'
+elif version_operation == 'inject-timestamp':
+  version_operation_kwargs['operation'] = 'set_prerelease'
+  prerelease = f'timestamp-{int(datetime.datetime.now(tz=datetime.timezone.utc).timestamp())}'
```
ccwienk, line 49: I am fine w/ epoch instead of  RFC3339-timestamp. we should check w/ @rfranzke whether he is also fine w/ that, as I think he explicitly mentioned an rfc-3339-ts

8R0WNI3: I've decided to use an epoch timestamp instead of the RFC3339 one to stay compliant with [SemVer](https://semver.org/). Also, the epoch timestamp is just as good for sorting purposes as the RFC3339 timestamp.

@rfranzke is there any other requirement which enforces RFC3339?

rfranzke: I think epoch should also be fine. I was following https://fluxcd.io/flux/guides/image-update/#imagepolicy-examples (see very last example at the bottom of the page), and here they talk about `RFC3339`. However, epoch is also alphabetically-sortable, and that's all that counts, I guess.

ccwienk (approved):

PR #1013, opened 2024-08-15
Also consider "soft" component references in update deps step

8R0WNI3:

Fixes #

The `cnudie.iter.iter` function, which already has support for considering "soft" extra component references, was not used here
   (1) to omit not required lookups of referenced components and
   (2) to keep `componentReference.name` property next to `componentReference.componentName` property.

**Release note**:

```other operator

```

### DIFF
--- a/concourse/steps/update_component_deps.py
+++ b/concourse/steps/update_component_deps.py
@@ -16,6 +16,7 @@ import cnudie.util
 import concourse.model.traits.update_component_deps as ucd
 import concourse.steps.component_descriptor_util as cdu
 import dockerutil
+import dso.labels
 import github.util as gu
 import gitutil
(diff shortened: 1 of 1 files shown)


concourse/steps/update_component_deps.py
```diff
@@ -239,18 +229,45 @@ def greatest_references(
             yield matching_refs[-1]
 
 
+def deserialise_extra_component_references(
```
ccwienk, line 232: generally-speaking, I would prefer to split into:
a) a function that filter for cref-labels
b) a function that does the deserialisation

generally-speaking, I do not like the pattern of a function that claims to do X and silently skips over elements for which it cannot do X

ccwienk (approved):

PR #1020, opened 2024-08-26
cleanup: rm unnecessary checks

ccwienk:
fopen will already fail of parent-directory does not exist - there is no need to put an additional check upfront.


zkdev (approved):
/lgtm

PR #1023, opened 2024-08-28
fix: ignore case for github-url-matching

ccwienk:
GitHub effectively does case-folding (i.e. URLs with cases deviating from org/repo-names are redirected). Therefore, ignore cases to avoid unnecessary lookup-failures for irrelevant misconfiguration.


zkdev (approved):
/lgtm

PR #1024, opened 2024-08-28
Cleanup build

ccwienk:


.ci/pipeline_definitions
```diff
@@ -47,19 +48,6 @@ cc-utils:
...
-        component_descriptor:
-          retention_policy:
-            name: 'cleanup-snapshots-and-releases'
-            dry_run: False
-            rules:
-              - name: 'snapshots'
-                match: 'snapshots'
-                restrict: 'none'
-                keep: 15
-              - name: 'releases'
```
8R0WNI3, line 59: Since there is only an automated cleanup for snapshots repository configured, shouldn't we keep the retention policy for releases?

ccwienk: hmmm... yeah, I suppose so

ccwienk: I moved cleanup to release-job, though (there is no need to run it more often, is there..)

8R0WNI3 (approved):

PR #1025, opened 2024-08-28
Don't copy signatures of root component to referenced ones

8R0WNI3:

Fixes #

**Release note**:

```other operator

```


ccwienk (approved):
good catch!

PR #1027, opened 2024-08-29
add dev-version-injection compliant w/ python-setuptools

ccwienk:
Python's setuptools do not allow for multiple dashes to occur. Therefore, add a new option `inject-commit-hash-nodash`. Keep existing default to not risk breaking existing scripts which might parse commits-digests from such versions.

This change (once published) will allow us to reconfigure our pipelines for _this_ repository such that resulting versions will be accepted by setuptools.


8R0WNI3 (approved):
/lgtm

PR #1028, opened 2024-08-29
Add flag to disable delivery service ocm mapping in lookup

8R0WNI3:

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/cnudie/retrieve.py
+++ b/cnudie/retrieve.py
@@ -149,10 +149,10 @@ def in_memory_cache_component_descriptor_lookup(
         component_id: cm.ComponentIdentity,
         component_descriptor: cm.ComponentDescriptor,
     ):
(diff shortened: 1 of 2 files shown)


cnudie/retrieve.py
```diff
@@ -661,7 +663,8 @@ def create_default_component_descriptor_lookup(
     cache_dir: str | None=None,
     oci_client: oc.Client=None,
     delivery_client=None,
-    default_absent_ok=False,
+    default_absent_ok: bool=False,
+    fallback_to_service_mapping: bool=True,
```
ccwienk, line 667: I think we should document the semantics of fallback_to_service_mapping in docstr

ccwienk (changes requested):

ccwienk (approved):

PR #1029, opened 2024-08-30
add `prefix` option for release-assets

ccwienk:
In cases where generated files have names one may not safely predict (or may not want to spend the effort of explicitly describing), it is likely a common practise to output them into a directory, from where they can be picked up using GLOBBING, as already supported by release-assets.

Allow specifying an optional `prefix` to conveniently strip such directories from resulting TAR-files (maybe there will also be other usages).


concourse/model/traits/release.py
```diff
@@ -152,6 +152,10 @@ class BuildstepFileAsset(Asset):
     `path` is interpreted relative to `step_output_dir`. Using Unix globbing, any matching files
     will be included (use `*` to include all, except files starting with period (`.`) character).
 
+    `prefix` will be removed from fnames in tar (similar to tar's -C option). Only effective in
```
8R0WNI3, line 155: Wouldn't it be more convenient to allow overriding of `arcname`? 🤔

ccwienk: then, we would have to provide a mapping for each single file. which - considering there are variations in filenames (e.g. in case of our package-build there is a version-number) - I think would be quite clunky / would mean one would have to use e.g. regex (sed-ish). imo, prefix is a lot easier and should be "good enough".

the arcnames can already be altered indirectly by simply renaming contained files 🤷🏻‍♂️

8R0WNI3 (approved):
/lgtm

PR #1030, opened 2024-08-30
fix: also strip trailing `/`

ccwienk:
If retaining trailing slash, arcnames will have a leading `/` (which - depending on parameterisation might cause them to be extracted into root directory). Also, it will require `prefix` to include a leading `/`, which is quite unintuitive.


zkdev (approved):
/lgtm

PR #1034, opened 2024-09-05
Resolve artefact access via callback

zkdev:
Resolve artefact access (thus its contents) from callback to prepare support of various artefact types.

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/cnudie/access.py
+++ b/cnudie/access.py
@@ -1,6 +1,14 @@
+import collections.abc
+
 import gci.componentmodel as cm
 
 import ioutil
+
+
+ResolveArtefactAccess = collections.abc.Callable[
+    [cm.Access],
+    collections.abc.Generator[bytes, None, None],
+]
 
 
 def s3_access_as_blob_descriptor(
--- a/protecode/scanning.py
+++ b/protecode/scanning.py
@@ -20,14 +20,12 @@ import dso.cvss
 import dso.labels
 import dso.model
 import gci.componentmodel as cm
-import gci.oci
 import oci.client
 import protecode.assessments
 import protecode.client
 import protecode.model as pm
 import protecode.rescore
 import protecode.util
-import tarutil
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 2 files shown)


ccwienk:
please fix linter issues @zkdev

cnudie/util.py
```diff
@@ -589,3 +595,87 @@ def desanitise_version(version: str) -> str:
     desanitised_version = version.replace(META_SEPARATOR, '+')
 
     return desanitised_version
+
+
+def resource_content(
+    resource_node: 'cnudie.iter.ResourceNode',
+    oci_client: oci.client.Client,
```
8R0WNI3, line 602: Rather make `oci_client` and `s3_client` optional attributes and only check their existence in case it is actually required for the resource's access type (or set reasonable defaults).

cnudie/util.py
```diff
@@ -589,3 +595,87 @@ def desanitise_version(version: str) -> str:
...
+    raises if resource or access type is not supported
+    '''
+    component: cm.Component = resource_node.component
+    resource: cm.Resource = resource_node.resource
+
+    if resource.access.type is cm.AccessType.OCI_REGISTRY:
+        if resource.type is not cm.ResourceType.OCI_IMAGE:
+            return
+
+        def iter_content():
```
8R0WNI3, line 616: Why using an extra function here?

cnudie/util.py
```diff
@@ -589,3 +595,87 @@ def desanitise_version(version: str) -> str:
...
+            )
+
+        return iter_content()
+
+
+    elif resource.access.type is cm.AccessType.S3:
+        if resource.type != 'application/tar+vm-image-rootfs':
+            return
+
+        def as_blob_descriptors():
```
8R0WNI3, line 636: Ditto

cnudie/util.py
```diff
@@ -589,3 +595,87 @@ def desanitise_version(version: str) -> str:
...
+
+    elif resource.access.type is cm.AccessType.LOCAL_BLOB:
+        img_ref = oci_artefact_reference(
+            component=f'{component.name}:{component.version}',
+            ocm_repository=component.current_repository_ctx(),
+        )
+        digest = resource.access.localReference
+        size = resource.access.size
+
+        if resource.access.globalAccess:
```
8R0WNI3, line 654: Since you're overwriting all variables from before, why not putting both into the conditional block?

cnudie/util.py
```diff
@@ -589,3 +595,87 @@ def desanitise_version(version: str) -> str:
...
+            img_ref = resource.access.globalAccess.ref
+            digest = resource.access.globalAccess.digest
+            size = resource.access.globalAccess.size
+
+        print(f'{img_ref=}')
```
8R0WNI3, line 659: Debugging leftover?

cnudie/util.py
```diff
@@ -589,3 +595,87 @@ def desanitise_version(version: str) -> str:
...
+        print(f'{digest=}')
+        print(f'{size=}')
+
+        blob = oci_client.blob(
+            image_reference=img_ref,
+            digest=digest,
+            stream=True,
+        )
+
+        def as_blob_descriptors():
```
8R0WNI3, line 669: Again: Why using an extra function here?

protecode/util.py
```diff
@@ -303,7 +304,19 @@ def component_artifact_metadata(
...
+    elif resource.access.type is cm.AccessType.LOCAL_BLOB:
+        metadata['IMAGE_REFERENCE_NAME'] = resource.name
+        metadata['RESOURCE_TYPE'] = cm.AccessType.LOCAL_BLOB
+        # do not use global access as it is optional
+        img_ref = cnudie.util.oci_artefact_reference(
+            component=f'{component.name}:{component.version}',
+            ocm_repository=component.current_repository_ctx(),
+        )
+        digest = resource.access.localReference
+        metadata['IMAGE_REFERENCE'] = f'{img_ref}:{digest}'
```
8R0WNI3, line 318: Use digest reference syntax, i.e. `<image-ref>@sha256:<digest>`

8R0WNI3 (changes requested):

protecode/util.py
ccwienk: also, I removed `current_repository_ctx`. replace w/ current_ocm_repo (caveat: this is a property)

cnudie/util.py
ccwienk: in addition, I would prefer to pass-in an `access`, rather than a resource, or even a resource-node

even more specifically, I do not think this should be implemented in cc-utils. instead, have the bdba-function accept a acess-resolver-callback, and pass those in in using code -> delivery-service

ccwienk (changes requested):
see inline comment.

This pullrequest should focus on enabling bdba-code to accept callback, not handle it.

zkdev:
linter fails as this PR depends on:
[-](https://github.com/gardener/cc-utils/pull/1045)

ccwienk:
> linter fails as this PR depends on: #1045

please rework such that:

- bdba allows for an optional blob-resolving-callback to be passed-in (and use it to retrieve contents from)
- actual resolving to accesses is done in delivery-service-repository

rationale: we should keep dependencies between cc-utils and delivery-service-repository to a minimum / work towards reducing dependencies.

PR #1036, opened 2024-09-09
replace daily alpine-image-build w/ github-action

ccwienk:
also: rm obsolete github-action


.github/workflows/alpine-sap.yaml
```diff
@@ -0,0 +1,31 @@
+name: Build Alpine-SAP-Image
+run-name: Building Alpine-SAP-Image
+on:
+  schedule:
+    - cron: "0 0 * * *" # daily
```
8R0WNI3, line 5: Hmm, does GHCR offer support for some kind of clean-up? Since this image was published to a GAR snapshots repository before, the images were cleaned-up every 30 days.

ccwienk: @8R0WNI3 : there seems to be an action for that: https://github.com/marketplace/actions/ghcr-io-cleanup-action

ccwienk: I added cleanup-action..

8R0WNI3 (approved):
/lgtm

PR #1040, opened 2024-09-12
Prepare standalone

ccwienk:
will hopefully help w/ GithubActions.


oci/auth.py
```diff
@@ -133,3 +136,90 @@ def lookup_credentials(
...
+        image_netloc = image_reference.split('/')[0]
+        image_host = image_netloc.split(':')[0]
+
+        for netloc, auth_dict in auths.items():
+            host = netloc.split(':')[0]
+            if host == image_host:
+                break
+        else:
+            if not absent_ok:
+                raise ValueError(f'no auth-cfg found in {docker_cfg=} for {image_reference=}')
```
8R0WNI3, line 208: Maybe change the error messages here and the two above a little bit to indicate more precisely what lead to the error, e.g. `no _matching_ auth-cfg found...`. Rationale: This will allow to distinguish what was the root cause in the case of an error.

oci/auth.py
```diff
@@ -133,3 +136,90 @@ def lookup_credentials(
         return valid_cfgs[0].credentials
 
     return lookup_credentials
+
+
+def docker_credentials_lookup(
+    docker_cfg: str | None=None,
+    absent_ok: bool=False,
+) -> typing.Callable[[image_reference, Privileges, bool], OciConfig]:
```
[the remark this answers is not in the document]
ccwienk: @8R0WNI3 :  alas, you are correct. However, this is copy-paste from above, so I will also change cp-source into a separate commit

oci/auth.py
```diff
@@ -133,3 +136,90 @@ def lookup_credentials(
...
+        else:
+            if not absent_ok:
+                raise ValueError(f'no auth-cfg found in {docker_cfg=} for {image_reference=}')
+            return None # no matching cfg was found
+
+        # we found a cfg
+        # docker's auth-cfgs only have a single value `auth` (or so we hope / assume)
+        auth = auth_dict.get('auth', None)
+        if not auth:
+            raise ValueError('did not find expected attr `auth` in {docker_cfg=} for {image_host=}')
```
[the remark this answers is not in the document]
ccwienk: nit: it's not just a nit, but rather a bug / error (on my end) :-)
-> done / thanks

8R0WNI3 (approved):
/lgtm

setup.ocm.py
```diff
@@ -0,0 +1,40 @@
...
+        return f.read().strip()
+
+
+setuptools.setup(
+    name='ocm',
+    version=version(),
+    description='Open-Component-Model (OCM) language bindings',
+    python_requires='>=3.11',
+    py_modules=modules(),
+    packages=['ocm', 'gci'],
```
zkdev, line 33: if we include the whole `gci` module, we will also have to install `oci` requirements
see https://github.com/gardener/cc-utils/blob/84731c906ceab2026c3ff9b95f1969b10989a857/gci/oci.py#L11

zkdev (changes requested):

setup.ocm.py
ccwienk: that is indeed correct. I think I will, however, remove reference to `gci`, rather than adding those dependencies, for now.

PR #1045, opened 2024-09-16
Create flexible artefact access resolver

zkdev:
Useful to support various artefact types.
Can be used as a callback for code which should not bother with resolving ocm artefacts.

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/cnudie/access.py
+++ b/cnudie/access.py
@@ -1,6 +1,10 @@
-import gci.componentmodel as cm
+import collections.abc
 
+import cnudie.access
+import gci.componentmodel as cm
 import ioutil
+import oci.client
+import tarutil
 
 
 def s3_access_as_blob_descriptor(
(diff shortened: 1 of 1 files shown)


cnudie/access.py
```diff
@@ -22,3 +26,94 @@ def s3_access_as_blob_descriptor(
         size=size,
         name=name or f's3://{s3_access.bucketName}/{s3_access.objectKey}',
     )
+
+
+def create_resolve_access(
```
ccwienk, line 31: please mv to using code (-> delivery-service)

ccwienk (changes requested):

cnudie/access.py
zkdev: Is required in cc-utils for BDBA cli

zkdev:
see https://github.com/open-component-model/delivery-service/pull/179

PR #1046, opened 2024-09-17
Overhaul CTT processing behaviour

8R0WNI3:

This change will adjust the CTT behaviour in such that:
- "jobs" are only being created for components which don't have an existing component descriptor in the target OCM repo -> previously, _all_ resources were processed (i.e. accesses were patched) and replicated to the new target url -> however, component descriptors were not overwritten resulting in replicated resources which were not referenced _anywere_ -> now, only those resources are being processed and replicated, which will be also reflected in the component descriptor
- only the components which don't exist in the target OCM repo are being tried to be uploaded there (less OCI requests as OCM lookup is used instead)
- it is worked on component descriptor level instead of component level now, which allows to keep signature information of referenced components as well
- removal of special handling for root component during upload process -> there is no case anymore, where the root component does not exist in the source OCM repo, hence the special handling is not required anymore
- the yielded component and resource nodes are now directly retrieved from the target OCM repo, so they reflect the actual state -> before, they only reflected the "optimistic" state that no component descriptor existed in the target before and hence all patches were applied successfully

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/ctt/process_dependencies.py
+++ b/ctt/process_dependencies.py
@@ -227,16 +227,72 @@ def enum_processing_cfgs(
     )
 
 
-def create_jobs(
-    processing_cfg_path: str,
-    component_descriptor_v2: cm.ComponentDescriptor,
-    inject_ocm_coordinates_into_oci_manifests: bool,
+def determine_changed_components(
+    component_descriptor: cm.ComponentDescriptor,
+    tgt_ocm_repo_url: str,
     component_descriptor_lookup: cnudie.retrieve.ComponentDescriptorLookupById,
+    tgt_component_descriptor_lookup: cnudie.retrieve.ComponentDescriptorLookupById,
     component_filter: collections.abc.Callable[[cm.Component], bool]=None,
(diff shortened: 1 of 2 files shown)


ctt/process_dependencies.py
```diff
@@ -521,135 +587,66 @@ def wrap_process_job(processing_job: processing_model.ProcessingJob):
...
-        if component.name == root.name and component.version == root.version:
-            src_component = root
-        else:
-            src_component = component_descriptor_lookup(component).component
-        src_ocm_repo = src_component.current_ocm_repo
-        append_ctx_repo(src_ocm_repo, component)
-        append_ctx_repo(tgt_ctx_base_url, component)
+        append_ocm_repo(
+            component=component,
+            ocm_repo=tgt_ctx_base_url,
```
ccwienk, line 631: nit: also rename to tgt_ocm_repo_url?

ccwienk (approved):

zkdev (approved):
overall /lgtm

PR #1048, opened 2024-09-19
Rename to ocm

ccwienk:


gci/componentmodel.py
```diff
@@ -65,7 +85,7 @@ class AccessType(enum.StrEnum):
 AccessTypeOrStr = typing.Union[AccessType, str]
 
 
-@dc(frozen=True, kw_only=True)
```
zkdev, line 68: I prefer to keep them frozen.
Most of the time mutating objects is a bad idea, and for the remaining cases where this might be valid, one can still use workarounds, e.g. `dataclasses.replace`

gci/componentmodel.py
```diff
@@ -539,10 +554,6 @@ def __post_init__(self):
             self.access = AccessDict(access)
 
 
-# backwards-compatibility
-ComponentSource = Source
```
zkdev, line 543: There are still references in delivery-service (did **not** check other repos)

ccwienk: good catch - let us rename those, too (this should be a trivial search-and-replace task. I did check our "internal" repositories, though.

gci/componentmodel.py
ccwienk: I have to disagree. The way we use those model-classes, e.g. in CTT, we _do_ mutate. I do not actually see we gain benefits from keeping (parts of the model, anyways..) immutable.

I wanted to do some cleanups (like removal of as many as possible dependencies towards non-standard-libs), but also make the model-elements mutable, to thus improve ease-of use.

ccwienk:
https://github.com/open-component-model/delivery-service/pull/186

ccwienk:
@jensh007 : you recently showed interest for a stand-alone version of `gci.componentmodel` (`ocm` as of this change). In a subsequent step, I think we can (again) expose those sources as a standalone package. still need to figure out how to proceed w/ the remains of `gci` (e.g. oci.py), though

gci/componentmodel.py
```diff
@@ -654,7 +651,7 @@ def dateparse(v):
                 },
             )
         )
-        if not validation_mode is ValidationMode.NONE:
+        if validation_mode is not None:
```
8R0WNI3, line 673: It is enough to check for a boolean value here, isn't it? We don't have to _exactly_ require `None` (ditto for equivalent change in `ocm` module).

bin/component-cli
```diff
@@ -8,7 +8,7 @@ import sys
 
 import yaml
 
-import gci.componentmodel as cm
+import ocm as cm
```
8R0WNI3, line 11: Let us also get rid of the alias cm here, I think ocm is already short enough :-)

github/util.py
```diff
@@ -24,8 +24,8 @@
 from github3.pulls import PullRequest
 from github3.repos.release import Release
 
-import gci.componentmodel
-import gci.componentmodel as cm
+import ocm
+import ocm as cm
```
8R0WNI3, line 28: There is no need to import this module twice (I know this doesn't stem from this PR), but I think we shouldn't keep it that way.

bin/component-cli
ccwienk: yes, I agree. Let's see how well my `sed`-skills are

gci/componentmodel.py
ccwienk: technically, you are correct. However, I wanted to be slightly stricter here to catch programming-errors.

by explictly checking for `None` + otherwise "casting" to ValidationMode, passing-in values, such as `False`, an empty string, or 0 (or actually any value that will evaluate to `False`) would silently shortcut validation. This made me go for this more specific check. WDYT?

8R0WNI3 (approved):
/lgtm

PR #1051, opened 2024-09-25
Gha headupdates pipeline

ccwienk:


ocm/__main__.py
```diff
@@ -0,0 +1,229 @@
...
+    import oci.auth
+    import oci.client
+    _have_oci = True
+except ImportError:
+    _have_oci = False
+
+
+def _parse_yaml_or_json(path) -> dict:
+    with open(path) as f:
+        if _have_yaml:
```
8R0WNI3, line 28: Shouldn't we rather decide whether to you `yaml` or `json` based on file contents (of course only if `yaml` module is available)?

ocm/__main__.py
```diff
@@ -0,0 +1,229 @@
...
+
+def dump(component_descriptor: ocm.ComponentDescriptor, parsed):
+    if parsed.out == '-':
+        outfh = sys.stdout
+    else:
+        outfh = open(parsed.out, 'w')
+
+    raw_dict = dataclasses.asdict(component_descriptor)
+
+    if _have_yaml:
```
8R0WNI3, line 44: Again, maybe we should add a `--json` and `--yaml` flag and determine the default based on the presence of the `yaml` module?

zkdev (approved):
/lgtm

ocm/__main__.py
ccwienk: well, it is always safe to use `yaml`, as each json-document can always also be parsed using a yamlparser. I wanted to at least _allow_ usage of ocm-package w/o installed pyyaml package (not sure whether this will really ever be a reasonable use-case).
we might do a pre-check and prefer json if document is json-compliant; however, the only disadvantage I see from always using yaml if it is present is a slightly worse performance (which is not actually a concern for us at this point, I think)

ocm/__main__.py
ccwienk: yes, we could. do you see an actual use-case where someone would want to force usage of json, even if yaml module is present?

8R0WNI3 (approved):

PR #1052, opened 2024-09-27
Omit `features` field from oci-platform if no contents

zkdev:
Some oci-registries do not support features field yet. As we do not use this field (at least for now), omit entirely on serialisation if it has no entries.

Fixes #

**Release note**:

```other operator

```


8R0WNI3 (approved):
/lgtm

PR #1053, opened 2024-10-02
Add async OCI and OCM packages

8R0WNI3:

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/ccc/oci.py
+++ b/ccc/oci.py
@@ -61,6 +61,52 @@ def oci_request_handler_requirements_fulfilled() -> bool:
 
 
 @functools.cache
+def oci_client_async(
+    credentials_lookup: collections.abc.Callable=None,
+    cfg_factory=None,
+    http_connection_pool_size: int | None=None,
+    tag_preprocessing_callback: collections.abc.Callable[[str], str]=None,
+    tag_postprocessing_callback: collections.abc.Callable[[str], str]=None,
+) -> 'oci.client_async.Client':
+    import aiohttp
+    import oci.client_async
+
+    def base_api_lookup(image_reference):
+        registry_cfg = model.container_registry.find_config(
+            image_reference=image_reference,
+            privileges=None,
+            cfg_factory=cfg_factory,
+        )
+        if registry_cfg and (base_url := registry_cfg.api_base_url()):
+            return base_url
+        return oc.base_api_url(image_reference)
+
+    routes = oc.OciRoutes(base_api_lookup)
+
+    if not credentials_lookup:
+        credentials_lookup = oci_cfg_lookup(cfg_factory=cfg_factory)
+
+    if http_connection_pool_size is None: # 0 is a valid value here (meaning no limitation)
+        connector = aiohttp.TCPConnector()
+    else:
+        connector = aiohttp.TCPConnector(
+            limit=http_connection_pool_size,
+        )
+
+    session = aiohttp.ClientSession(
+        connector=connector,
+    )
+
+    return oci.client_async.Client(
+        credentials_lookup=credentials_lookup,
+        routes=routes,
+        session=session,
+        tag_preprocessing_callback=tag_preprocessing_callback,
+        tag_postprocessing_callback=tag_postprocessing_callback,
+    )
+
+
+@functools.cache
 def oci_client(
     credentials_lookup: collections.abc.Callable=None,
     install_logging_handler: bool=True,
--- a/cnudie/iter_async.py
+++ b/cnudie/iter_async.py
@@ -1,0 +1,172 @@
+import collections.abc
+
+import cnudie.iter
+import cnudie.retrieve_async
+import dso.labels
+import ocm
+
+
+async def iter(
+    component: ocm.Component,
+    lookup: cnudie.retrieve_async.ComponentDescriptorLookupById=None,
+    recursion_depth: int=-1,
+    prune_unique: bool=True,
+    node_filter: collections.abc.Callable[[cnudie.iter.Node], bool]=None,
+    ocm_repo: ocm.OcmRepository | str=None,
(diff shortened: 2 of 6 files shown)


ccwienk:
@8R0WNI3 : how about picking the first two commits:
- [Replace some typing occurrences by collections.abc](https://github.com/gardener/cc-utils/pull/1053/commits/5023bf4c1b725664f2055ee488c850855ea81e55)
- [Remove unused parameter](https://github.com/gardener/cc-utils/pull/1053/commits/e053197c81ec052981f9b73195278c50500c2904)

directly to `master` (and thus reduce the size of this pullrequest)?

ccwienk:
as for the "actual" changes, I have some concerns w.r.t. new package-names. considering other patterns I have seen, it seem to be more common to have (few) different entrypoints (e.g. `client.Client` vs `client.AsyncClient`) or use `async`-suffixes (e.g. `client` + `client_async`).

Specifically for `oci`-package (or more specifically `oci/__init__.py`, I think we either might drop some parts (e.g. kaniko-related code should not be actually in-use any longer, so we might discard an async flavour of it). From the diffs I thus-far inspected, I can see there is quite some duplication w.r.t. dataclasses - I think we should deduplicate those (there should not be any differences between them (sync vs async), should there?.

How about a layout like the following (limited to `oci` package for now):

```
oci/__init__.py # functions are replicated w/ _async-suffix: replicate_artifact_async, replicate_blobs_async, drop kaniko (at least: do not re-implement/duplicate as async)
oci/client,py # Client + AsyncClient (same module); as an alternative: client + client_base + client_async (factor-out re-usable code (login/token-handling, routes); this has the slight disadvantage of not being fully backwards-compatible (although we might fix that by exporting imports from client_base in client-module)
oci/platform.py # check which functions we actually use; create async versions of needed ones (in same module)
oci/workarounds.py # ditto
```

oci/__init__.py
```diff
@@ -490,6 +826,77 @@ def replicate_blob(blob: om.OciBlobRef) -> om.OciBlobRef:
...
+
+    Note that the uploaded artifact must be finalised after the upload by a "manifest-put".
+    '''
+    blob_overwrites = {k.digest:v for k,v in blob_overwrites.items()}
+
+    async def replicate_blob_async(blob: om.OciBlobRef) -> om.OciBlobRef:
+        if blob_overwrite_bytes := blob_overwrites.get(blob.digest):
+            logger.info(f'Replicate with overwriting {blob=}')
+
+            if hasattr(blob_overwrite_bytes, 'read'):
```
ccwienk, line 849: hmmm...... does it actually make sense to accept a (synchronous) reader/filelike-object in async-flavour of this function?

cnudie/iter.py
```diff
@@ -249,3 +249,169 @@ def iter_resources(
         component_filter=component_filter,
         reftype_filter=reftype_filter,
     )
+
+
+async def iter_async(
+    component: ocm.Component,
+    lookup: cnudie.retrieve.ComponentDescriptorLookupById=None,
```
ccwienk, line 256: do we need an async-lookup (or should we at least encourage one)? -> I think there should be dedicated typedefs for async lookups (+ wrong lookups should be rejected)

cnudie/retrieve.py
```diff
@@ -193,6 +195,69 @@ def lookup(
     return lookup
 
 
+def in_memory_cache_component_descriptor_lookup_async(
```
ccwienk, line 198: will it make sense to create separate modules (cnudie/retrieve + cnudie/retrieve_async) to make this more of a drop-in replacement? we should deduplicate Dataclasses / Typedefs, though

oci/__init__.py
```diff
@@ -8,6 +8,7 @@
 import tarfile
 import zlib
 
+import aiohttp
```
ccwienk, line 11: as aiohttp is a new dependency, which I would like to avoid, esp. in oci-package, can we also make it optional?
(ignore import-errors + raise runtime-error in async functions); imports for typehints should be possible to be worked around by converting them to str-literatls

oci/client.py
```diff
@@ -11,6 +12,8 @@
 import threading
 import time
 
+import aiohttp
```
ccwienk, line 15: ditto. preferably either make import optional (and raise if aiohttp is absent)? maybe an `client_async` module might also be an option

users might then use API like so:

```
import oci
import oci.client_async as ocm_client # OR: import ocm.client as oci_client
```

admittedly, it does not seem nice to have asymetry w.r.t. functions in `oci/__init__` :-/
-> maybe we should consider moving them into a submodule (e.g. `oci.replicate`) and deprecate the old ones (it should be feasible to make our userbase migrate so we can eventually cleanup) ?

such an approach would have the benefit that:
- we only claim one rootlevel-package
- modules w/o sync/async-relevance stay as they are
- others have companions, allowing users not needing / desiring async to avoid dependency by simply not importing async-modules
- we still have means to deduplicate / share common dataclasses / helper functions (e.g. async-companion-modules might import their non-async friends)

requirements.oci.txt
```diff
@@ -1,3 +1,4 @@
+aiohttp
```
ccwienk, line 1: should we make this dependency optional?

requirements.txt
```diff
@@ -2,6 +2,7 @@
 GitPython
 Mako<2.0.0
 Sphinx
+aiohttp
```
zkdev, line 5: I think this can be dropped from `requirements.txt`, as `aiohttp` is explicitly listed as dependency for optional `async` feature.
For developer convenience we might introduce an additional `requirements-async.txt`(?) file containing `async` dependencies + reference to base `requirements.txt` (`-r requirements.utils.txt`)

PR #1056, opened 2024-10-08
Add Release-Steps to GHA-Workflow

ccwienk:


.github/workflows/build-and-test.yaml
```diff
@@ -22,9 +31,19 @@ jobs:
           snapshots_repo="${repo_base}/snapshots"
           releases_repo="${repo_base}/releases"
 
+          if ${{ inputs.release || false }}; then
```
8R0WNI3, line 34: What is the purpose of `|| false`? Won't it anyways always depend on `inputs.release` evaluating to `true`? Also, is this testing artefact intended to be still part of the PR?

.github/workflows/build-and-test.yaml
```diff
@@ -22,9 +31,19 @@ jobs:
...
+            # "erroneously" still publish to snapshots for testing purporses
+            # XXX this is a debugging / testing artefact
+            ocm_repository=${releases_repo}
+            oci_repository=${releases_repo}
+          else
+            ocm_repository=${snapshots_repo}
+            oci_repository=${snapshots_repo}
+          fi
+
           # hardcode to "snapshots" for now
```
8R0WNI3, line 44: > hardcode to "snapshots" for now

This comment also seems to be outdated.

.github/workflows/build-and-test.yaml
```diff
@@ -50,10 +71,22 @@ jobs:
         minor="$(echo ${src_version} | cut -d. -f2)"
         patch="$(echo ${src_version} | cut -d. -f3 | cut -d- -f1)"
 
-        effective_version=${major}.${minor}.${patch}-${commit}
+        if ${{ inputs.release || false }}; then
```
8R0WNI3, line 74: ditto (`|| false`)?

.github/workflows/build-and-test.yaml
```diff
@@ -184,8 +252,25 @@ jobs:
...
+                  {
+                    "type": "githubTeam",
+                    "teamname": "gardener/ci-maintainers",
+                    "github_hostname": "github.com"
+                  }
+                ]
+            }' \
             > component-descriptor.yaml
 
+          if ${{ inputs.release || false }}; then
```
8R0WNI3, line 267: ditto (`|| false`)?

.github/workflows/build-and-test.yaml
ccwienk: if `release` is not defined (which is the case if triggering was done by head-update or pullrequest), this expression will evaluate to the empty string otherwise, leaving the shell-script with a syntax-error:
```
if ;then
...
```

.github/workflows/build-and-test.yaml
ccwienk: see above. I have yet to find a more elegant approach. one option might be to pull this logic into the `params` step, so we at least onyl have to do this once, wdyt?

8R0WNI3 (approved):

PR #1058, opened 2024-10-16
rm /v1-suffix for serialisation of helmchart-type

ccwienk:
OCM-CLI/OCM-Spec seem to not allow/know v1-suffix for helmchart-artefacts. Thus, change deserialisation output to just helmChart. Keep support for alias _with_ v1-suffix for compatibility.


Diaphteiros (approved):
Looks good.

PR #1059, opened 2024-10-16
fix: store a list of image-mappings

ccwienk:
see:
https://github.com/open-component-model/ocm/tree/main/components/helmdemo#using-transported-images-image-mapping


zkdev (approved):
/lgtm

PR #1062, opened 2024-10-21
Cleanups (rm gci.componentmodel, rm platform-logic from pipeline-template)

ccwienk:


cnudie/upload.py
```diff
@@ -110,10 +110,7 @@ def upload_component_descriptor(
     component = component_descriptor.component
 
     schema_version = component_descriptor.meta.schemaVersion
-    if isinstance(schema_version, enum.Enum):
-        # hack: convert from gci.componentmodel
-        schema_version = ocm.SchemaVersion(schema_version.value)
-    else:
+    if not isinstance(schema_version, ocm.SchemaVersion):
```
8R0WNI3, line 113: nit: You can drop this check and always try to parse to `ocm.SchemaVersion` (it doesn't make a difference if the object is already of this type).

concourse/resources/defaults.mako
```diff
@@ -1,5 +1,5 @@
 <%namespace file="/resources/image.mako" import="task_image_resource"/>
-<%def name='task_image_defaults(registry_cfg, platform=None, indent=0)'
+<%def name='task_image_defaults(registry_cfg, indent=0)'
```
8R0WNI3, line 2: nit: I think we can also drop the comment from line 5.

8R0WNI3 (approved):
Only nits, otherwise /lgtm

PR #1063, opened 2024-10-21
Add component-diff to release notes

TuanAnh17N:

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/cnudie/util.py
+++ b/cnudie/util.py
@@ -1,6 +1,7 @@
 import collections.abc
 import dataclasses
 import graphlib
+import textwrap
 
 import ci.util
 import ocm
@@ -411,6 +412,142 @@ def diff_components(
         names_only_left=left_component_names,
         names_only_right=right_component_names,
         names_version_changed=names_version_changed,
+    )
+
+
+def format_component_diff(
+    component_diff: ComponentDiff,
+    delivery_dashboard_url_view_diff: str | None=None,
+    delivery_dashboard_url: str | None=None
+):
+    if delivery_dashboard_url_view_diff:
+        bom_diff_header = f'## <a href="{delivery_dashboard_url_view_diff}">BoM Diff</a>\n'
+    else:
+        bom_diff_header = '## BoM Diff\n'
+
+    added_components = [
+        f'\U00002795 {component.name} {component.version}'
+        for component in component_diff.cidentities_only_right
+        if component.name not in component_diff.names_version_changed
+    ]
+
+    removed_components = [
+        f'\U00002796 {component.name} {component.version}'
+        for component in component_diff.cidentities_only_left
+        if component.name not in component_diff.names_version_changed
+    ]
+
+    changed_components = [
+        f'\U00002699 {new_component.name}: {old_component.version} → {new_component.version}'
+        for old_component, new_component in component_diff.cpairs_version_changed
+    ]
+
+    summary_counts = textwrap.dedent(f'''\
+        Added components: {len(added_components)}
+        Changed components: {len(changed_components)}
+        Removed components: {len(removed_components)}\n
+    ''')
+
+    summary_details = []
+    if added_components:
+        summary_details.append('### Added Components:\n' + '\n'.join(added_components) + '\n')
+
+    if removed_components:
+        summary_details.append('### Removed Components:\n' + '\n'.join(removed_components) + '\n')
+
+    if changed_components:
+        summary_details.append('### Changed Components:\n' + '\n'.join(changed_components) + '\n')
+
+    component_details = []
+    for old_component, new_component in component_diff.cpairs_version_changed:
+        if delivery_dashboard_url:
+            component_link = (
+                f"<a href='{delivery_dashboard_url}/#/component?name={new_component.name}'>"
+                f'{new_component.name}</a>'
+            )
+        else:
+            component_link = new_component.name
+
+        component_header = (
+            f'<details><summary>\U00002699 {component_link}:'
+            f'{old_component.version} → {new_component.version}</summary>\n'
+        )
+
+        added_resources = []
+        removed_resources = []
+        changed_resources = []
+
+        # Group resources by name
+        old_resources_grouped = {}
+        new_resources_grouped = {}
+
+        for res in old_component.resources:
+            old_resources_grouped.setdefault(res.name, []).append(res)
+
+        for res in new_component.resources:
+            new_resources_grouped.setdefault(res.name, []).append(res)
+
+        # Process each resource in the new component
+        for res_name, new_res_list in new_resources_grouped.items():
+            old_res_list = old_resources_grouped.get(res_name, [])
+
+            if (
+                old_res_list
+                and sorted([res.version for res in old_res_list])
+                == sorted([res.version for res in new_res_list])
+            ):
+                # Skip resource as all versions are identical in both the old and new components
+                continue
+
+            if len(new_res_list) == 1 and len(old_res_list) == 1:
+                # Single occurrence in both -> Compare versions
+                new_res = new_res_list[0]
+                old_res = old_res_list[0]
+                if new_res.version != old_res.version:
+                    changed_resources.append(
+                        [f'\U0001F504 {res_name}', f'{old_res.version} → {new_res.version}']
+                    )
+            else:
+                # Multiple occurrences -> Display all versions for each resource name
+                if old_res_list:
+                    removed_resources.extend(
+                        [[f'\U00002796 {res_name}', res.version] for res in old_res_list]
+                    )
+                if new_res_list:
+                    added_resources.extend(
+                        [[f'\U00002795 {res_name}', res.version] for res in new_res_list]
+                    )
+
+        # Process resources that only exist in the old component
+        for res_name, old_res_list in old_resources_grouped.items():
+            if res_name not in new_resources_grouped:
+                removed_resources.extend(
+                    [[f'\U00002796 {res_name}', res.version] for res in old_res_list]
+                )
+
+        # Aggregate results for resources into `resources_data`
+        if not (added_resources or removed_resources or changed_resources):
+            resources_data = [['No resources added, removed, or changed', '']]
+        else:
+            resources_data = added_resources + removed_resources + changed_resources
+
+        #Import `tabulate` only when the function is called to avoid loading it during module import
+        import tabulate
+        # Generate HTML table with tabulate
+        resources_table = tabulate.tabulate(
+            tabular_data=resources_data,
+            headers=['Resource', 'Version Change'],
+            tablefmt='html'
+        )
+
+        component_details.append(component_header + resources_table + '\n</details>')
+
+    return (
+        bom_diff_header
+        + summary_counts
+        + '\n'.join(summary_details)
+        + '\n## Component Details:\n'
+        + '\n'.join(component_details)
     )
 
 
--- a/concourse/steps/update_component_deps.py
+++ b/concourse/steps/update_component_deps.py
@@ -13,6 +13,7 @@ import github3.repos.repo
 import ccc.github
 import ci.util
 import cnudie.util
+import cnudie.retrieve
 import concourse.model.traits.update_component_deps as ucd
 import concourse.steps.component_descriptor_util as cdu
 import dockerutil
@@ -392,6 +393,7 @@ def create_upgrade_pr(
     merge_method: ucd.MergeMethod,
     version_lookup,
     component_descriptor_lookup,
+    delivery_dashboard_url: str=None,
     after_merge_callback=None,
     container_image:str=None,
     pullrequest_body_suffix: str=None,
@@ -409,6 +411,35 @@ def create_upgrade_pr(
         absent_ok=False,
     )
     from_component = from_component_descriptor.component
+
+    to_component_descriptor = component_descriptor_lookup(
+        ocm.ComponentIdentity(
+            name=to_ref.componentName,
+            version=to_version,
+        )
+    )
+
+    to_component = to_component_descriptor.component
+
+    bom_diff = cnudie.retrieve.component_diff(
+        from_component,
+        to_component
+    )
+
+    if delivery_dashboard_url:
+        delivery_dashboard_url_view_diff = (
+            f'{delivery_dashboard_url}/#/component?name={to_component.name}&view=diff'
+            f'&componentDiff={from_component.name}:{from_component.version}'
+            f':{to_component.name}:{to_component.version}'
+        )
+    else:
+        delivery_dashboard_url_view_diff = None
+
+    formatted_diff = cnudie.util.format_component_diff(
+        component_diff=bom_diff,
+        delivery_dashboard_url_view_diff=delivery_dashboard_url_view_diff,
+        delivery_dashboard_url=delivery_dashboard_url
+    )
 
     # prepare env for upgrade script and after-merge-callback
     cmd_env = os.environ.copy()
@@ -503,19 +534,32 @@ def create_upgrade_pr(
             '\n\nRelease notes were shortened since they exceeded the maximum length allowed for a '
             'pull request body. The remaining release notes will be added as comments to this PR.'
         )
-        if max_pr_body_length < len(release_notes):
-            step_size = max_pr_body_length - len(max_body_length_exceeded_remark)
+        additional_notes = []
+        # Split the formatted_diff if it exceeds max body length (line-by-line)
+        if len(formatted_diff) > max_pr_body_length:
+            truncated_diff = formatted_diff[:formatted_diff.find('## Component Details:')]
+            truncated_diff += '\n... [Component details omitted]\n'
+
+            pr_body = truncated_diff + max_body_length_exceeded_remark
+        else:
+            pr_body = formatted_diff
+
+        available_length = max_pr_body_length - len(pr_body)
+
+        if available_length < len(release_notes):
+            step_size = available_length - len(max_body_length_exceeded_remark)
             split_release_notes = [
-                release_notes[start:start+step_size]
+                release_notes[start:start + step_size]
                 for start in range(0, len(release_notes), step_size)
             ]
         else:
             split_release_notes = [release_notes]
 
-        if not (additional_notes := split_release_notes[1:]):
-            pr_body = split_release_notes[0]
+        if len(split_release_notes) > 1:
+            pr_body += max_body_length_exceeded_remark
+            additional_notes = split_release_notes
         else:
-            pr_body = split_release_notes[0] + max_body_length_exceeded_remark
+            pr_body += '\n\n' + split_release_notes[0]
     else:
         pr_body = ''
         additional_notes = []
--- a/test/cnudie/cnudie_util_test.py
+++ b/test/cnudie/cnudie_util_test.py
@@ -64,6 +64,36 @@ def iref():
     return image_ref
 
 
+def test_componentdiff_to_str(cid):
+    diff = cnudie.util.ComponentDiff(
+        cidentities_only_left=(cid(name='a', version='1.0.0'), cid(name='b', version='2.0.0')),
+        cidentities_only_right=(cid(name='c', version='3.0.0'),),  # added
+        cpairs_version_changed=[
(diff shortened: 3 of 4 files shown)


gardener-robot-ci-1:
Thank you @TuanAnh17N for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

concourse/steps/update_component_deps.py
```diff
@@ -520,6 +661,12 @@ def create_upgrade_pr(
         pr_body = ''
         additional_notes = []
 
+    dashboard_url = (
+        f'https://delivery-dashboard.ci.gardener.cloud.sap/#/'
```
zkdev, line 665: use cfg exposed by ctx_factory, 'delivery_endpoints'

zkdev (changes requested):

concourse/steps/update_component_deps.py
```diff
@@ -410,6 +530,22 @@ def create_upgrade_pr(
...
+    )
+
+    to_component = to_component_descriptor.component
+
+    bom_diff = cnudie.retrieve.component_diff(
+        from_component,
+        to_component
+    )
+
+    formatted_diff = format_bom_diff(bom_diff, True)
```
zkdev, line 547: nit: please use keyword args + one arg per line

concourse/steps/update_component_deps.py
```diff
@@ -503,7 +639,12 @@ def create_upgrade_pr(
             '\n\nRelease notes were shortened since they exceeded the maximum length allowed for a '
             'pull request body. The remaining release notes will be added as comments to this PR.'
         )
-        if max_pr_body_length < len(release_notes):
+
+        formatted_diff_length = len(f'## BoM Diff:\n{formatted_diff}\n\n')
```
zkdev, line 643: did you also consider that BoM Diff could exceed max PR body size?

concourse/steps/update_component_deps.py
zkdev: also, only link delivery-dashboard if an instance is available

concourse/steps/update_component_deps.py
```diff
@@ -377,6 +378,125 @@ def _import_release_notes(
...
 
 
+def format_bom_diff(component_diff, include_internal_links=False):
+    changed_components = []
+    added_components = []
+    removed_components = []
+    changed_versions = []
+
+    if (hasattr(component_diff, 'cidentities_only_left') and
+        hasattr(component_diff, 'cidentities_only_right')):
```
zkdev, line 388: nit: I prefer following formatting (due to readability)
```python
if (
  hasattr(...)
  and hasattr(...)
):
  pass
```

concourse/steps/update_component_deps.py
```diff
@@ -377,6 +378,125 @@ def _import_release_notes(
...
+        left_components = component_diff.cidentities_only_left
+        right_components = component_diff.cidentities_only_right
+
+        for component_identity in left_components:
+            matching_component = next((comp for comp in right_components
+                                       if comp.name == component_identity.name), None)
+            if matching_component:
+                if matching_component.version != component_identity.version:
+                    changed_versions.append(
+                        f'🔄 {component_identity.name}:'
```
zkdev, line 398: I would like to avoid emojis in source code, can we rather use codepoints? (e.g. `"\u2764"`)

zkdev: also, I suggest to group components (changed, removed, ...) as objects and only apply (markdown) formatting and the very end

concourse/steps/update_component_deps.py
ccwienk: good catch! we should determine whether max-body-size would be exceeded, and in that case resort to just adding a link in that case

concourse/steps/update_component_deps.py
ccwienk: however, IIRC, component-diff is a dataclass - what's the point of checking for attributes to exist (they should always be present)?

concourse/steps/update_component_deps.py
ccwienk: @zkdev : well... python3 does allow use of unicode, so I do not see a strong reason not to (see [documentation on unicode](https://docs.python.org/3/howto/unicode.html); we might add an encoding-hint to the top of this module - however, considering utf-8 is the default, it should be safe (and formally correct) to omit it.

concourse/steps/update_component_deps.py
```diff
@@ -377,6 +378,125 @@ def _import_release_notes(
...
+            )
+        else:
+            component_link = component_name
+
+        component_header = (
+            f'<details><summary>⚙️ {component_link}:'
+            f'{old_component.version} → {new_component.version}</summary>\n')
+
+        resources_details = []
+        resources_details.append("<table>")
```
ccwienk, line 441: consider using tabulate (which also has an option to output html)

zkdev (changes requested):

concourse/steps/update_component_deps.py
```diff
@@ -503,8 +682,22 @@ def create_upgrade_pr(
             '\n\nRelease notes were shortened since they exceeded the maximum length allowed for a '
             'pull request body. The remaining release notes will be added as comments to this PR.'
         )
-        if max_pr_body_length < len(release_notes):
-            step_size = max_pr_body_length - len(max_body_length_exceeded_remark)
+
+        if len(formatted_diff) > max_pr_body_length:
+            split_formatted_diff = [
+                formatted_diff[start:start+max_pr_body_length]
```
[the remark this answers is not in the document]
TuanAnh17N: makes sense, thanks for pointing it out :)

concourse/steps/update_component_deps.py
```diff
@@ -377,6 +382,163 @@ def _import_release_notes(
...
+
+    class ComponentChange:
+        def __init__(self, name, old_version=None, new_version=None, change_type=None):
+            self.name = name
+            self.old_version = old_version
+            self.new_version = new_version
+            self.change_type = change_type  # 'added', 'removed', 'changed'
+
+        def __str__(self):
+            if self.change_type == ChangeType.CHANGED:
```
ccwienk, line 403: use identity-comparison to check for enum-values (is instead of ==)

concourse/steps/update_component_deps.py
```diff
@@ -377,6 +382,163 @@ def _import_release_notes(
...
+    component_diff: cnudie.util.ComponentDiff,
+    to_component_name: str,
+    base_url: str | None
+):
+    class ChangeType(enum.StrEnum):
+        ADDED = 'added'
+        REMOVED = 'removed'
+        CHANGED = 'changed'
+
+    class ComponentChange:
```
ccwienk, line 395: why use a class with basically just a single function?
-> define as a plain function

concourse/steps/update_component_deps.py
```diff
@@ -377,6 +382,163 @@ def _import_release_notes(
...
+        def __init__(self, name, old_version=None, new_version=None, change_type=None):
+            self.name = name
+            self.old_version = old_version
+            self.new_version = new_version
+            self.change_type = change_type  # 'added', 'removed', 'changed'
```
ccwienk, line 400: use typehint rather than comment to indicate allowed values

concourse/steps/update_component_deps.py
```diff
@@ -377,6 +382,163 @@ def _import_release_notes(
...
+
+        def __str__(self):
+            if self.change_type == ChangeType.CHANGED:
+                return f'\U00002699 {self.name}: {self.old_version} → {self.new_version}'
+            elif self.change_type == ChangeType.REMOVED:
+                return f'\U00002796 {self.name} (v{self.old_version})'
+            elif self.change_type == ChangeType.ADDED:
+                return f'\U00002795 {self.name} (v{self.new_version})'
+            return ''
```
ccwienk, line 409: raise ValueError upon unexpected value rather than silently returning an empty string

concourse/steps/update_component_deps.py
```diff
@@ -377,6 +382,163 @@ def _import_release_notes(
...
+                        old_version=component_identity.version,
+                        new_version=matching_component.version,
+                        change_type=ChangeType.CHANGED)
+                )
+        else:
+            removed_components.append(
+                ComponentChange(
+                    name=component_identity.name,
+                    old_version=component_identity.version,
+                    change_type=ChangeType.REMOVED)
```
ccwienk, line 443: codestyle: rather place closing parenthesis in new line

concourse/steps/update_component_deps.py
```diff
@@ -377,6 +382,163 @@ def _import_release_notes(
     return release_notes
 
 
+def format_bom_diff(
+    component_diff: cnudie.util.ComponentDiff,
+    to_component_name: str,
+    base_url: str | None
```
ccwienk, line 388: `base_url` is not a good name, as it gives no indication as to which kind of URL is expected here. rather call it `delivery_dashboard_url`. It is also not that nice to hard-code assumptions about delivery-dashboard's query-parameters (later in code), but that's okay-ish as long as we do not have a better approach (e.g. via delivery-service).

concourse/steps/update_component_deps.py
```diff
@@ -410,6 +572,32 @@ def create_upgrade_pr(
...
+    to_component_descriptor = component_descriptor_lookup(
+        ocm.ComponentIdentity(
+            name=to_ref.componentName,
+            version=to_version,
+        )
+    )
+
+    to_component = to_component_descriptor.component
+
+    cfg_factory = ctx.cfg_factory()
```
ccwienk, line 584: do not create cfg-factories within nested functions. instead, pass-in a cfg-factory into the function, if needed. or better yet: pass-in the needed bom_diff-object

concourse/steps/update_component_deps.py
```diff
@@ -410,6 +572,32 @@ def create_upgrade_pr(
...
+    cfg_set = cfg_factory.cfg_set(ci.util.current_config_set_name())
+    delivery_endpoints_cfg = cfg_set.delivery_endpoints()
+
+    delivery_base_url = delivery_endpoints_cfg.dashboard_url()
```
ccwienk, line 588: as the getter-method suggests, this is the delivery-dashboard-url. so I suggest to call the variable such, to make the name more expressive, and self-explanatory.

concourse/steps/update_component_deps.py
```diff
@@ -410,6 +572,32 @@ def create_upgrade_pr(
...
+
+    bom_diff = cnudie.retrieve.component_diff(
+        from_component,
+        to_component
+    )
+
+    formatted_diff = format_bom_diff(
```
ccwienk, line 595: I think format-diff might rather be a good fit for cnudie.util

concourse/steps/update_component_deps.py
```diff
@@ -503,19 +691,47 @@ def create_upgrade_pr(
             '\n\nRelease notes were shortened since they exceeded the maximum length allowed for a '
             'pull request body. The remaining release notes will be added as comments to this PR.'
         )
-        if max_pr_body_length < len(release_notes):
-            step_size = max_pr_body_length - len(max_body_length_exceeded_remark)
+
+        # Split the formatted_diff if it exceeds max body length (line-by-line)
+        if len(formatted_diff) > max_pr_body_length:
```
ccwienk, line 696: why don't you simply calculate the allowed length, and shorten the diff accordingly, like so:

```
max_leng = allowed - len(diff)
diff = diff[:len(diff + max_leng) - 3] + '..'
```

Assuming we wanted to remove entire lines (why?), it would be preferable to calculate location of last line-break (by looking for linebreak backwards from last allowed character). That way, we less str-operations, and also do not need to duplicate the strs.

However, even better than just cutting diff might be generating a shorter version (e.g. only print components and omit resource-changes).

There should also rather be a better indication of shortening than the `..` I used above as a short example, to make it obvious to readers that a portion of the diff is missing.

ccwienk (changes requested):

cnudie/util.py
```diff
@@ -414,6 +416,121 @@ def find_changed_component(
...
 
+def format_component_diff(
+    component_diff: ComponentDiff,
+    to_component_name: str,
+    delivery_dashboard_url: str | None
+):
+    if delivery_dashboard_url:
+        dashboard_view_diff = (
+            f'{delivery_dashboard_url}/#/component?name='
+            f'{to_component_name}&view=diff'
```
8R0WNI3, line 427: Instead of only linking to the diff page, you could also already display the calculated diff by adding:
```
&componentDiff=<left-component-name>:<left-component-version>:<right-component-name>:<right-component-version>
```

cnudie/util.py
```diff
@@ -414,6 +416,121 @@ def find_changed_component(
...
+        )
+        bom_diff_header = f'## <a href="{dashboard_view_diff}">BoM Diff</a>\n'
+    else:
+        bom_diff_header = '## BoM Diff\n'
+
+    added_components = [
+        f'\U00002795 {component.name} (v{component.version})'
```
8R0WNI3, line 434: Let us not add a leading `v` here. Some components use a `v` in their versions, some don't, so it's confusing to always add the `v` here.

cnudie/util.py
```diff
@@ -414,6 +416,121 @@ def find_changed_component(
...
+
+    removed_components = [
+        f'\U00002796 {component.name} (v{component.version})'
+        for component in component_diff.cidentities_only_left
+        if component.name not in component_diff.names_version_changed
+    ]
+
+    changed_components = [
+        f'\U00002699 {name}: {old_component.version} → {new_component.version}'
+        for name in component_diff.names_version_changed
```
8R0WNI3, line 447: What's the benefit of iterating `names_version_changed` first and then searching the corresponding entry in `cpairs_version_changed` instead of directly only iterating `cpairs_version_changed`?

cnudie/util.py
```diff
@@ -414,6 +416,121 @@ def find_changed_component(
...
+        for old_component, new_component in component_diff.cpairs_version_changed
+        if old_component.name == name
+    ]
+
+    summary_counts = (
```
8R0WNI3, line 452: Rather use a multiline string, that way you can omit the explicit line breaks after each line:
```
summary_counts = textwrap.dedent(f'''\
    Added components: {len(added_components)}
    Changed components: {len(changed_components)}
    Removed components: {len(removed_components)}\n
''')
```

cnudie/util.py
```diff
@@ -414,6 +416,121 @@ def find_changed_component(
...
+            f'<details><summary>\U00002699 {component_link}:'
+            f'{old_component.version} → {new_component.version}</summary>\n')
+
+        # Sort resources into categories: added, removed and changed
+        added_resources = []
+        removed_resources = []
+        changed_resources = []
+
+        for new_resource in new_component.resources:
+            matching_old_resource = next(
```
8R0WNI3, line 489: Hmm I'm not sure about this semantics: In case we have the same resource-names multiple times referenced by the component (in different versions), it would always use the same first `matching_old_resource` to display the diff.
What do you think of the following approach instead:
- in case a resource-name is only referenced once by the component -> display the diff
- in case a resource-name is referenced multiple times -> display all "old" resources vs. all "new" resources?

/cc @ccwienk

cnudie/util.py
```diff
@@ -414,6 +416,121 @@ def find_changed_component(
...
+                 if old_resource.name == new_resource.name), None
+            )
+
+            if matching_old_resource:
+                if matching_old_resource.version != new_resource.version:
+                    changed_resources.append(
+                        [f'\U0001F504 {new_resource.name}',
+                         f'{matching_old_resource.version} → {new_resource.version}'])
+            else:
+                added_resources.append(
```
8R0WNI3, line 500: Also, this heuristics wouldn't determine if a resource was added in a new version, if the resource-name already existed in any version before, would it?

cnudie/util.py
```diff
@@ -414,6 +416,121 @@ def find_changed_component(
...
+                    [f'\U00002795 {new_resource.name}',
+                     f'{new_resource.version} (added)'])
+
+        for old_resource in old_component.resources:
+            if not any(new_resource.name == old_resource.name
```
8R0WNI3, line 505: Same here, it wouldn't determine if a resource-version was removed if any resource with the same name was still existing.

concourse/steps/update_component_deps.py
```diff
@@ -392,6 +393,7 @@ def create_upgrade_pr(
     merge_method: ucd.MergeMethod,
     version_lookup,
     component_descriptor_lookup,
+    delivery_dashboard_url: str = None,
```
8R0WNI3, line 396: nit: remove whitespace around operand `=`

8R0WNI3 (changes requested):

concourse/steps/update_component_deps.py
```diff
@@ -410,6 +412,35 @@ def create_upgrade_pr(
...
+
+    bom_diff = cnudie.retrieve.component_diff(
+        from_component,
+        to_component
+    )
+
+    if delivery_dashboard_url:
+        delivery_dashboard_url_view_diff = (
+            f'{delivery_dashboard_url}/#/component?name={to_component.name}&view=diff'
+            f'&componentDiff={from_component.name}:{from_ref.version}'
```
8R0WNI3, line 432: nit: Although it is not wrong, rather use consistently either `from_component` or `from_ref` here.

cnudie/util.py
```diff
@@ -414,6 +417,140 @@ def find_changed_component(
     )
 
 
+def format_component_diff(
+    component_diff: ComponentDiff,
+    delivery_dashboard_url_view_diff: str,
```
8R0WNI3, line 422: Adjust type hint to allow `None` as value + preferably also set it as default value (same for `delivery_dashboard_url`).

cnudie/util.py
```diff
@@ -414,6 +417,140 @@ def find_changed_component(
...
+        # Process each resource in the new component
+        for res_name, new_res_list in new_resources_grouped.items():
+            old_res_list = old_resources_grouped.get(res_name, [])
+
+            if (
+                old_res_list
+                and sorted([res.version for res in old_res_list])
+                == sorted([res.version for res in new_res_list])
+            ):
+                # skip
```
8R0WNI3, line 501: It's obvious that the iteration is skipped here because of the following `continue` keyword, rather state _why_ it is skipped here :-)

cnudie/util.py
```diff
@@ -414,6 +417,140 @@ def find_changed_component(
...
+
+    if removed_components:
+        summary_details.append('### Removed Components:\n' + '\n'.join(removed_components) + '\n')
+
+    if changed_components:
+        summary_details.append('### Changed Components:\n' + '\n'.join(changed_components) + '\n')
+
+    component_details = []
+    for old_component, new_component in component_diff.cpairs_version_changed:
+        component_name = f'{new_component.name}'
```
8R0WNI3, line 465: nit: I think this variable is rather unnecessary, you can simply use `new_component.name` directly in the following conditional block (also it would not be required to explicitly parse it to a string here again).

cnudie/util.py
```diff
@@ -414,6 +417,140 @@ def find_changed_component(
...
+            component_link = (
+                f"<a href='{delivery_dashboard_url}/#/component?name={new_component.name}'>"
+                f'{component_name}</a>'
+            )
+        else:
+            component_link = component_name
+
+        component_header = (
+            f'<details><summary>\U00002699 {component_link}:'
+            f'{old_component.version} → {new_component.version}</summary>\n')
```
8R0WNI3, line 476: stylistic-nit: Put closing bracket in new line.

8R0WNI3 (approved):
Only nits, otherwise /lgtm

test/cnudie/cnudie_util_test.py
```diff
@@ -64,6 +64,26 @@ def image_ref(name, version):
...
+        names_only_right={'c'},
+        names_version_changed={'d'},
+    )
+    result = cnudie.util.format_component_diff(
+        component_diff=diff,
+        delivery_dashboard_url_view_diff=None,
+        delivery_dashboard_url=None,
+    )
+
+    assert '### Added Components:' in result
```
ccwienk, line 84: might be nice to also test case where diff is too long to increase codecoverage

PR #1067, opened 2024-10-30
Dismantle Concourse Pipeline

ccwienk:


.github/workflows/build-and-test.yaml
```diff
@@ -373,6 +377,24 @@ jobs:
           python -m ocm upload \
             --file component-descriptor.yaml \
             --blobs-dir /tmp/dist/blobs.d
+      - name: Generate and publish Documentation
```
8R0WNI3, line 380: Why did you choose to publish the documentation as part of the `component_descriptor` step? Also, I think `Generate` is quite misleading here as name since it's actually generated as part of the `documentation` step.

ccwienk: @8R0WNI3 : you are correct about the misleading name (will fix this).

I can, however, elaborate on why I chose to do this as part of component-descriptor step. This is mostly a shortcut to save the efforts of (again) having to initialise git-cfg (set username/email-addr). Which admittedly is quite not such a strong argument (instead, it might rather be a reason to factor-out such code into re-usable action(s)). 🙈 

btw, even the order of steps is significant (as git-checkout + git-clean would otherwise cause other steps to fail). I suppose doing different things w/ steps for the same job is quite brittle (as steps have virtually no isolation, different than to concourse's steps..)

8R0WNI3 (approved):

8R0WNI3:
One remark: We should also consider GitHub actions now instead of Concourse pipelines to display recent `tests` and `build` state in the README.

ccwienk:
> One remark: We should also consider GitHub actions now instead of Concourse pipelines to display recent `tests` and `build` state in the README.

valid point! will adopt the URLs

PR #1069, opened 2024-11-04
Define github-hostname for trusted teams

zkdev:
Useful for scenarios where trusted-teams from multiple github instances have to be considered.

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/model/concourse.py
+++ b/model/concourse.py
@@ -3,6 +3,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 
+import collections.abc
 import dataclasses
 from enum import Enum
 import typing
@@ -385,6 +386,12 @@ def secret_name_from_team(team_name: str, key_generation: int) -> str:
         return f'{cfg_name_from_team(team_name)}-{key_generation}'
 
 
+@dataclasses.dataclass(frozen=True)
+class TrustedTeams:
+    github_hostname: str
+    teams: list[str]
+
+
 class JobMapping(NamedModelElement):
     def team_name(self) -> str:
         # todo: use `name` attr for that (thus enforce unique mappings)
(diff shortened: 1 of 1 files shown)


zkdev:
/hold

do not merge (yet), requires cfg changes

model/concourse.py
```diff
@@ -385,6 +386,12 @@ def secret_name_from_team(team_name: str, key_generation: int) -> str:
         return f'{cfg_name_from_team(team_name)}-{key_generation}'
 
 
+@dataclasses.dataclass(frozen=True)
+class TrustedTeams:
+    github_hostname: str
+    teams: list[str]
```
ccwienk, line 392: list is mutable. declaring the dataclass as frozen is inconsistent w/ that. suggestion: declare the class as mutable (non-frozen)

model/concourse.py
```diff
@@ -455,11 +462,31 @@ def expose_deployed_pipelines(self) -> bool:
...
         return self.raw.get('expose_pipelines', True)
 
-    def trusted_teams(self) -> typing.Iterable[str]:
+    def trusted_teams(self) -> collections.abc.Iterable[TrustedTeams]:
         '''Pull requests created/synchronized by members of this team will automatically have
         the required label set by the webhook-dispatcher
         '''
-        return self.raw.get('trusted_teams', [])
+        return [
+            dacite.from_dict(
```
ccwienk, line 470: can we keep this simpler and just keep a list of strings? it should be fairly simple to discriminate between qualified and non-qualified teams, as github does not allow `/` (or `.` for that matter) characters in team-names.

Hence, entries of the form `foo/bar` (two parts if splitting by `/`) should be safe to translate to `<org>/<teamname>`, whereas entries of the form `host.tld/foo/bar` (three parts if splitting by `/`) should be safe to interpret as qualified

ccwienk (changes requested):

zkdev:
merge with https://github.com/gardener/cc-utils/pull/1070 , let's follow-up there

PR #1071, opened 2024-11-06
Refactor gha

ccwienk:
First (small) refactoring - more to come.

### DIFF
--- a/.github/actions/gar-auth/action.yaml
+++ b/.github/actions/gar-auth/action.yaml
@@ -1,0 +1,70 @@
+name: Authenticate against GAR
+description: |
+  Authenticates against GCP using OIDC. This action defines a default Service-Account with read-
+  and write-access for Google Artifact Registries from `gardener-project` as identity.
+
+  For convenience, it will generate a docker-cfg (`$HOME/.docker/config.json`) for subsequent steps.
+  This is equivalent to running `docker login`.
+inputs:
+  gcp_service_account:
+    description: The Service-Account to impersonate
+    required: false
+    default: ci-opensource-operator@gardener-project.iam.gserviceaccount.com
+  workload_identity_provider:
+    description: |
+      GCP Workload Identity Provider. The default value exclusively allows GitHub-Actions from
+      github.com/gardener (no forks).
+    required: false
+    default: projects/694386720375/locations/global/workloadIdentityPools/github-actions/providers/github-actions-provider
+  gcp_project:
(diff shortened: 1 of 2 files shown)


.github/actions/gcp-auth/action.yaml
```diff
@@ -0,0 +1,66 @@
+name: Authenticate against GAR
+description: |
+  Authenticates against GCP using OIDC. This action defines a default Service-Account as
+  identity.
+inputs:
+  gcp_service_account:
+    description: The Service-Account to impersonate
+    required: false
+    default: ocm-ci-opensource@gardener-project.iam.gserviceaccount.com
```
zkdev, line 9: this one is only used within ocm org, rather use `ci-gcr-operator@gardener-project.iam.gserviceaccount.com` as default value

.github/actions/gcp-auth/action.yaml
```diff
@@ -0,0 +1,66 @@
+name: Authenticate against GAR
```
zkdev, line 1: there is an inconsistency, `gcp-auth` vs `Authenticate against GAR`
is this action only working for GAR, or for GCP in general?

zkdev (changes requested):

.github/actions/gcp-auth/action.yaml
ccwienk: hmm, valid point. Maybe even better yet, we might create a dedicated service-account just for OIDC / GHAs

.github/actions/gcp-auth/action.yaml
```diff
@@ -0,0 +1,66 @@
...
+      The hostname of the GAR (for generating docker-cfg). For convenience, it is also
+      okay to pass a full image-reference (in which case hostname is parsed from parsed value).
+      Note that no special-handling for docker-hub (e.g. `alpine` ->
+      registry-1.docker.io/library/alpine:latest) is done.
+    required: false
+    default: europe-docker.pkg.dev
+outputs:
+  access_token:
+    description: The obtained access-token
+    value: ${{ steps.authenticate.outputs.access_token }}
```
8R0WNI3, line 31: I'm wondering, how is this working? I'm not seeing any step with id `authenticate`, and the "Export Access-Token" step does not define any id 🤔

ccwienk: @8R0WNI3 : good catch - fixed

zkdev (approved):
/lgtm

PR #1075, opened 2024-11-11
factor-out ocm-oci-imagebuild into separate action

ccwienk:


.github/actions/ocm-oci-build/action.yaml
```diff
@@ -0,0 +1,119 @@
...
+  repository:
+    description: |
+      OCI Repository to push to (including registry, but excluding tag)
+      For example: `europe-docker.pkg.dev/gardener-project/snapshots/my-image`
+    required: true
+  version:
+    description: |
+      Used for both image-tag, and OCM Resource fragment.
+    required: true
+  extra-tags:
```
8R0WNI3, line 21: The inputs `extra-tags`, `target` and `dockerfile` seem to not being actually used. Also, I think we should allow to specify a custom extra identity for the emitted OCM resource.

.github/workflows/build-and-test.yaml
```diff
@@ -542,15 +536,16 @@ jobs:
           image_ref=${oci_repo}/cicd/job-image:${image_tag}
 
           echo "oci_image_ref=${image_ref}" >> ${GITHUB_OUTPUT}
```
8R0WNI3, line 538: I think we can drop calculation of `oci_repo`, `image_tag` and `image_ref` here as well as setting of `oci_image_ref` output. Instead, the `image-build` job already sets the `image_reference` output which should be equivalent to `oci_image_ref` (has to be also adjusted in the declared outputs of this job of course).

8R0WNI3 (changes requested):

.github/actions/ocm-oci-build/action.yaml
ccwienk: done - still need to do some testing, though

.github/workflows/build-and-test.yaml
ccwienk: good catch. I think  I already switched `image`-step's output accordingly, but missed to clean the - now obsolete - code-snippet calculating image-ref (again)

.github/workflows/build-and-test.yaml
```diff
@@ -507,20 +506,15 @@ jobs:
       - package
       - params
     outputs:
-      oci_image_ref: ${{ steps.prepare.outputs.oci_image_ref }}
-      ocm_resources: ${{ steps.prepare.outputs.ocm_resources }}
+      oci_image_ref: ${{ steps.image-build.outputs.image_reference }}
+      ocm_resources: ${{ steps.image-build.outputs.ocm_resource }}
```
8R0WNI3, line 510: nit: Let us rather consistently use singular here, i.e. `ocm_resource`.

8R0WNI3 (approved):
/lgtm

.github/workflows/build-and-test.yaml
ccwienk: @8R0WNI3 sure, we should eventually refactor this part, as well.

rationale behind using plural forms consistently within this workflow: there are multiple jobs each emitting either one ore multiple ocm-resources + the `ocm append resource` command is built to allow both.

hence, using plural form made sense to me for jobs defined in workflow (as opposed to the (new) action emitting only one resource / image)

at any rate, it will still have to be seen how we can best support handling of multiple resources, considering [limitations of matrix-build](https://github.com/orgs/community/discussions/17245), for which there are already some workarounds like [this](https://github.com/cloudposse/github-action-matrix-outputs-write).

PR #1077, opened 2024-11-12
rm almost all remainders of `ctx_repository`

ccwienk:


concourse/model/traits/component_descriptor.py
```diff
@@ -297,16 +279,16 @@ def validation_policies(self):
...
         # - a ocm-repository-url (type: str)
         # - a OCM Repository (type: dict / gcm.OcmRepository
 
+        # XXX hack for unittests
+        if not self.cfg_set:
+            return None
+
         if ocm_repo:
             try:
                 ctx_repo_cfg = self.cfg_set.ctx_repository(ocm_repo)
```
8R0WNI3, line 294: I wonder, you removed the comment "a cfg-element-name (type: CtxRepositoryCfg)" above, but kept resolving `ocm_repo` as such. Please keep this consistent.

concourse/model/traits/component_descriptor.py
```diff
@@ -383,12 +357,13 @@ def transformer(self):
...
         super().validate()
 
-        ocm_repository_url = self.ctx_repository_base_url()
-        ocm_repository_mappings: list[dict] = self.raw['ocm_repository_mappings']
-        if not ocm_repository_url and ocm_repository_mappings:
-            raise ModelValidationError(
-                'ocm_repository_url must be defined if ocm_repository_mappings are'
-            )
+        if self.ocm_repository:
+            ocm_repository_url = self.ocm_repository.oci_ref
```
8R0WNI3, line 361: If `self.ocm_repository` evaluates to `True`, we'll always have an `ocm_repository_url`, don't we?

concourse/model/traits/component_descriptor.py
ccwienk: good catch. Will rm the resolving-logic as well (it did not prove to be a good idea to have an indirection via cfg-element for ocm-repository-definition (this is one of the few cases where it is preferable IMO to have redundant configuration requiring explicit changes, if desired).

ccwienk: done

concourse/model/traits/component_descriptor.py
ccwienk: that's correct m(

cnudie/retrieve.py
```diff
@@ -163,17 +163,12 @@ def writeback(
 
     def lookup(
         component_id: ocm.ComponentIdentity,
-        ctx_repo: ocm.OcmRepository=None,
```
8R0WNI3, line 166: Please also adjust type definition of `ComponentDescriptorLookupById` (ditto for `cnudie.retrieve_async.ComponentDescriptorLookupById`).

8R0WNI3 (approved):
Only one remark, otherwise /lgtm

cnudie/retrieve.py
ccwienk: again: good catch (however, the signature was not accurate before this change :-/)

PR #1078, opened 2024-11-18
Prepare property `artefact_extra_id` handling

8R0WNI3:

Fixes #

**Release note**:

```other operator

```


zkdev (approved):
/lgtm

dso/model.py
```diff
@@ -39,37 +39,46 @@ def has_scan_info(datasource: str) -> bool:
...
+            self.artefact_name,
+            self.artefact_type,
+            self.artefact_version,
+            self.normalised_artefact_extra_id,
+        ))
+
+    def __hash__(self) -> int:
+        return hash(self.as_frozenset())
+
+    def __eq__(self, other: 'LocalArtefactId') -> bool:
```
ccwienk, line 75: nit: you might use `typing.Self`

dso/model.py
```diff
@@ -39,37 +39,46 @@ def has_scan_info(datasource: str) -> bool:
...
+        return self.as_frozenset() == other.as_frozenset()
```
ccwienk, line 76: I think you need to guard against `other` having different type (which might not have `as_frozenset`)

dso/model.py
```diff
@@ -101,21 +110,22 @@ def as_frozenset(self) -> frozenset[str]:
...
+            props += (self.artefact.as_frozenset(),)
 
         return frozenset(props)
 
-    def __hash__(self):
+    def __hash__(self) -> int:
         return hash(self.as_frozenset())
 
-    def __eq__(self, other: 'ComponentArtefactId'):
+    def __eq__(self, other: 'ComponentArtefactId') -> bool:
```
ccwienk, line 120: typing.Self?

ccwienk (changes requested):

dso/model.py
```diff
@@ -39,37 +40,48 @@ def has_scan_info(datasource: str) -> bool:
...
+            self.artefact_type,
+            self.artefact_version,
+            self.normalised_artefact_extra_id,
+        ))
+
+    def __hash__(self) -> int:
+        return hash(self.as_frozenset())
+
+    def __eq__(self, other: typing.Self) -> bool:
+        if not type(self) == type(other):
```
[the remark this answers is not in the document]
8R0WNI3: I get that `isinstance` is usually the preferable way to compare types, however, as this overwrites the `__eq__` function I suppose we should stick with comparing the exact type (as it should evaluate to "not equal" if it's not the exact same type).

PR #1079, opened 2024-11-18
also expose ocm-resource-fragments as artefacts

ccwienk:
see: https://github.com/orgs/community/discussions/17245

As workaround, expose ocm-resource-fragments as artefacts, using naming-convention to avoid collisions. Leverage OCM already mandates to have unique names; ignore extra-identity for now, and hardcode type (oci-image) instead, for now.


8R0WNI3 (approved):
/lgtm

PR #1081, opened 2024-11-19
Refactor gci and cnudie

ccwienk:


gci/oci.py
```diff
@@ -106,34 +106,3 @@ def component_descriptor_from_tarfileobj(
           raise ValueError('Component Descriptor appears to be empty')
 
         return ocm.ComponentDescriptor.from_dict(raw_dict)
-
-
-def image_ref_with_digest(
```
8R0WNI3, line 111: Actually, this function is being used [here](https://github.com/open-component-model/delivery-service/blob/cf8ea385c6b6940c0bc0a2ad95313228c8f5d515/bdba/util.py#L299) in delivery-service. That being said, I'm not against dropping this function (here), but we'll have to adjust delivery-service accordingly first.

cnudie/upload.py
```diff
@@ -6,7 +6,7 @@
 
 import cnudie.util
 import ocm
-import gci.oci
+import ocm.oci
```
8R0WNI3, line 15: nit: Please order imports alphabetically.

ctt/replicate.py
```diff
@@ -8,11 +8,10 @@
 import json
 import logging
 
-import gci.oci
-
 import ccc.oci
-import ocm
 import oci
+import ocm
+import ocm.oci
```
8R0WNI3, line 14: ditto

gci/oci.py
ccwienk: indeed. I think moving this function to delivery-service will make sense. maybe we can drop some codepaths

8R0WNI3 (approved):
/lgtm

PR #1083, opened 2024-11-21
Enhance pr body generation to omit component details when length exceeds

TuanAnh17N:
When the combined length of the release notes and formatted diff surpasses the allowed pr body length, the component details are now omitted

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/concourse/steps/update_component_deps.py
+++ b/concourse/steps/update_component_deps.py
@@ -373,7 +373,7 @@ def _import_release_notes(
     )
 
     if not release_notes:
-        release_notes = pull_request_util.retrieve_pr_template_text()
+        release_notes = ''
 
     return release_notes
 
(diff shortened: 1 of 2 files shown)


test/cnudie/cnudie_util_test.py
```diff
@@ -75,23 +75,65 @@ def test_componentdiff_to_str(cid):
...
+    )
+
+    combined_length = len(formatted_diff) + len(release_notes)
 
-    assert '### Added Components:' in body
-    assert '... [Component details omitted]' in body
-    assert '## Component Details:' not in body
+    if combined_length > max_pr_body_length:
+        # remove component details from formatted_diff
+        truncated_diff = formatted_diff[:formatted_diff.find('## Component Details:')]
```
ccwienk, line 97: how about adding a switch to the function generating the diff and simply re-generate it (with truncation set to true on the second try?)

ccwienk: also: why did you duplicate so much code into unittest?

concourse/steps/update_component_deps.py
```diff
@@ -562,10 +551,22 @@ def create_upgrade_pr(
             split_release_notes = [release_notes]
 
         if len(split_release_notes) > 1:
-            pr_body += max_body_length_exceeded_remark
-            additional_notes = split_release_notes
+            pr_body += split_release_notes[0] + max_body_length_exceeded_remark
+            additional_notes = split_release_notes[1:]
```
zkdev, line 555: what if the additional_notes block also exceeds the size limit?

zkdev (changes requested):

concourse/steps/update_component_deps.py
TuanAnh17N: does `step_size` not already ensure that the chunks in `split_release_notes` are wihtin the size limit? since `step_size` is calculated based on `max_pr_body_length`, each chunk in `additional_notes` should already respect the max allowed size. am i wrong?

zkdev (approved):
/lgtm

PR #1086, opened 2024-11-27
Introduce sastfinding as a new finding type

TuanAnh17N:

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/dso/model.py
+++ b/dso/model.py
@@ -22,6 +22,20 @@ def _as_key(
     return separator.join(absent_indicator if arg is None else arg for arg in args)
 
 
+class SastStatus(enum.StrEnum):
+    NO_LINTER = 'no-linter'
+
+
+class SastSubType(enum.StrEnum):
+    LOCAL_LINTING = 'local-linting'
+    CENTRAL_LINTING = 'central-linting'
+
+
+@dataclasses.dataclass
+class MatchCondition:
+    component_name: str
+
+
 @dataclasses.dataclass
 class ScanArtifact:
     name: str
@@ -33,6 +47,7 @@ class ScanArtifact:
 class Datasource:
     ARTEFACT_ENUMERATOR = 'artefact-enumerator'
     BDBA = 'bdba'
+    SAST_LINT_CHECK = 'sast-lint-check'
     CHECKMARX = 'checkmarx'
     CLAMAV = 'clamav'
     CC_UTILS = 'cc-utils'
@@ -51,6 +66,11 @@ class Datasource:
                 Datatype.VULNERABILITY,
                 Datatype.LICENSE,
                 Datatype.STRUCTURE_INFO,
+                Datatype.RESCORING,
+            ),
+            Datasource.SAST_LINT_CHECK: (
+                Datatype.ARTEFACT_SCAN_INFO,
+                Datatype.SAST_FINDING,
                 Datatype.RESCORING,
             ),
             Datasource.CHECKMARX: (
@@ -214,6 +234,7 @@ class Datatype:
     LICENSE = 'finding/license'
     VULNERABILITY = 'finding/vulnerability'
     MALWARE_FINDING = 'finding/malware'
+    SAST_FINDING = 'finding/sast'
     DIKI_FINDING = 'finding/diki'
     CODECHECKS_AGGREGATED = 'codechecks/aggregated'
     OS_IDS = 'os_ids'
@@ -234,6 +255,7 @@ class Datatype:
             Datatype.DIKI_FINDING: Datasource.DIKI,
             Datatype.CRYPTO_ASSET: Datasource.CRYPTO,
             Datatype.FIPS_FINDING: Datasource.CRYPTO,
+            Datatype.SAST_FINDING: Datasource.SAST_LINT_CHECK,
         }[datatype]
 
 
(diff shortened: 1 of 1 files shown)


dso/model.py
```diff
@@ -24,6 +24,7 @@ class ScanArtifact:
 class Datasource:
     ARTEFACT_ENUMERATOR = 'artefact-enumerator'
     BDBA = 'bdba'
+    CM06 = 'cm06'
```
8R0WNI3, line 27: Please also add handling for `CM06` datasource in `datasource_to_datatypes` function.

dso/model.py
```diff
@@ -193,6 +194,7 @@ class Datatype:
     LICENSE = 'finding/license'
     VULNERABILITY = 'finding/vulnerability'
     MALWARE_FINDING = 'finding/malware'
+    SAST_FINDING = 'finding/sast'
```
8R0WNI3, line 197: ditto here for `datatype_to_datasource`

dso/model.py
```diff
@@ -355,6 +357,12 @@ def key(self) -> str:
         return self.finding.key
 
 
+@dataclasses.dataclass(frozen=True)
+class SastFinding(Finding):
+    sast_status: str
+    component_context: str
```
8R0WNI3, line 363: I think we should use enums for both the `sast_status` and the `component_context`, as both only allow a set of specific values, don't they? 🤔

dso/model.py
```diff
@@ -355,6 +357,12 @@ def key(self) -> str:
         return self.finding.key
 
 
+@dataclasses.dataclass(frozen=True)
+class SastFinding(Finding):
```
8R0WNI3, line 361: We'll also have to add the `key` property here since it's required for objects of `CustomRescoring` class ([see](https://github.com/gardener/cc-utils/blob/fd2771aff40506c10094136cc3045faaf4eb9988/dso/model.py#L428)).

8R0WNI3 (changes requested):

dso/model.py
```diff
@@ -43,6 +45,10 @@ def datasource_to_datatypes(datasource: str) -> tuple[str]:
                 Datatype.STRUCTURE_INFO,
                 Datatype.RESCORING,
             ),
+            Datasource.CM06: (
+                Datatype.ARTEFACT_SCAN_INFO,
+                Datatype.SAST_FINDING,
```
8R0WNI3, line 50: CM06 will also emit rescorings -> `Datatype.RESCORING`

dso/model.py
```diff
@@ -13,6 +13,18 @@
 import unixutil.model
 
 
+class ComponentContext(enum.StrEnum):
+    SAP_INTERNAL = 'sap-internal'
```
8R0WNI3, line 17: We should keep this more general, e.g. just `internal` or something similar.

dso/model.py
```diff
@@ -13,6 +13,18 @@
...
+    PUBLIC = 'public'
+
+
+class SastStatus(enum.StrEnum):
+    LOCAL_LINTING = 'local-linting'
+    NO_LINTING = 'no-linting'
+    CENTRAL_LINTING = 'central-linting'
+    CENTRAL_AND_LOCAL_LINTING = 'central-and-local-linting'
```
8R0WNI3, line 25: I'm not sure if we should really use extra enum values for combinations of other values, this can become quite cumbersome I'd imagine if we liked to add other statuses in the future. What about allowing `SastFinding` either a `SastStatus` or a list of `SastStatus`?

dso/model.py
```diff
@@ -13,6 +13,17 @@
 import unixutil.model
 
 
+class ComponentContext(enum.StrEnum):
+    INTERNAL = 'internal'
```
ccwienk, line 17: rather call it `relation` + reuse `ocm`'s definition

dso/model.py
```diff
@@ -13,6 +13,17 @@
...
+    PUBLIC = 'public'
+
+
+class SastStatus(enum.StrEnum):
+    LOCAL_LINTING = 'local-linting'
```
ccwienk, line 22: can we inline those definitions into using code?

dso/model.py
```diff
@@ -370,6 +387,17 @@ def key(self) -> str:
         return self.finding.key
 
 
+@dataclasses.dataclass(frozen=True)
+class SastFinding(Finding):
+    sast_statuses: list[SastStatus] | SastStatus
+    component_context: str
```
8R0WNI3, line 393: As also mentioned in other [comment](https://github.com/open-component-model/delivery-service/pull/270#discussion_r1881691225), I think we can drop the `component_context` property.

dso/model.py
```diff
@@ -380,6 +397,16 @@ def key(self) -> str:
...
 
 
+@dataclasses.dataclass(frozen=True)
+class SastFinding(Finding):
+    sast_statuses: list[SastStatus] | SastStatus
+    sub_type: SastSubType
+
+    @property
+    def key(self) -> str:
+        return _as_key(self.sast_statuses, self.sub_type)
```
8R0WNI3, line 407: In case `sast_statuses` is of type `list[SastStatus]`, you'll have to take care of sorting it first (to keep the key stable) + unpack the list so that it can be processed by `_as_key`.
On another note: Do we actually still require `sast_status(es)` to allow lists? The only status currently is `no-linter`, and if we'd like to add something like `has-linting-findings` in the future, it is logically not possible that both of them occur at the same time.

dso/model.py
```diff
@@ -33,6 +47,7 @@ class ScanArtifact:
 class Datasource:
     ARTEFACT_ENUMERATOR = 'artefact-enumerator'
     BDBA = 'bdba'
+    SAST_LINT_CHECK = 'sastLintCheck'
```
ccwienk, line 50: nit: for reasons of consistency, preferably use "kebap-case" (dash-separated) rather than snake-case

ccwienk (commented):
one last naming-nit, otherwise lgtm

PR #1087, opened 2024-11-27
Upload gh assets

ccwienk:
also upload release-assets as github-release-assets

### DIFF
--- a/concourse/steps/release.py
+++ b/concourse/steps/release.py
@@ -416,30 +416,25 @@ def github_release(
         )
 
 
-def upload_github_release_asset(
-    github_helper: GitHubRepositoryHelper,
-    github_release_tag: str,
+def upload_component_descriptor_as_release_asset(
+    github_release,
     component,
 ):
-    # upload copy as release-asset
-    release_tag_name = github_release_tag.removeprefix('refs/tags/')
+    component_descriptor = ocm.ComponentDescriptor(
+        component=component,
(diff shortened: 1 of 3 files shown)


concourse/steps/release.py
```diff
@@ -416,30 +416,25 @@ def github_release(
...
+    descriptor_str = yaml.dump(
+        data=dataclasses.asdict(component_descriptor),
+        Dumper=ocm.EnumValueYamlDumper,
+    )
 
-        normalized_component_name = component.name.replace('/', '_')
-        asset_name = f'{normalized_component_name}.component_descriptor.cnudie.yaml'
-        release.upload_asset(
+    normalized_component_name = component.name.replace('/', '_')
+    asset_name = f'{normalized_component_name}.component_descriptor.cnudie.yaml'
```
8R0WNI3, line 435: Do we have any automation in place which depends on this asset name? Otherwise, I'd suggest to move away from "cnudie" by now 🙈

8R0WNI3 (approved):

concourse/steps/release.py
ccwienk: @8R0WNI3 : _we_ do not have automation still relying on those copies of ocm-component-descriptors (and I also thought about renaming this, when I touched those lines of code yesterday). There might be some external automation we do not know about. however, I think we might just rename, and see whether someone complains

PR #1090, opened 2024-12-03
Refactor GitHub-Actions-Pipeline some more

ccwienk:
- split release-jobs into separate workflow
- create release-commit upfront via capture-commit-action
- import release-commit in release-workflow via import-commit-action

### DIFF
--- a/.github/actions/capture-commit/action.yaml
+++ b/.github/actions/capture-commit/action.yaml
@@ -68,5 +68,13 @@ runs:
 
         echo "commit-digest=$(cat ${commit_out})" >> ${GITHUB_OUTPUT}
         echo "commit-objects=$(cat ${objects_out} | base64 -w0)" >> ${GITHUB_OUTPUT}
+        cat << EOF > ${GITHUB_STEP_SUMMARY}
+        ## Capture-Commit Summary
+        commit-digest: \`$(cat ${commit_out})\`
+        objects-listing (tarfile):
+        $(tar tf ${objects_out})
+        commit-message:
+        $(git show)
+        EOF
         git clean -dfx
       shell: bash
--- a/.github/actions/capture-commit/action.yaml
+++ b/.github/actions/capture-commit/action.yaml
@@ -1,0 +1,72 @@
+name: capture-commit
+description: |
+  captures the head commit in serialised form in order to be shared w/ other jobs.
+
+  This action is intended for the use-case of sharing a "release-commit" (where typically
+  some metadata-bearing files (e.g. containing version) are modified) with other jobs, without
+  persisting it, yet (as subsequent jobs might still fail).
+
+  This action will only work if git-repository is cloned using `git` (i.e. git-repository must be
+  present). It will work well if repository was cloned as `shallow`. The action needs a reference
+  timestamp, which is passed via regular file. Said file *must* be created _prior_ to creating the
+  commit to serialise (/"capture"). It leverages git's internal organisation of objects, where new
+  objects will be created in separate files; such newly created object-files will be picked up, thus
+  containing both added/changed blobs, as well as tree-objects, and commit.
+
+  As release-commits are commonly rather small (typically just changing some versions in a few
+  files), the resulting data is typically only a few kiBs in size, which allows it to be exported
+  as a (base64-encoded) output (thus making passing-around somewhat less cumbersome).
+
+  Thus-exported commits can be imported, retaining the exact same commit-digest, provided the
+  importing repository has the same head-commit, into which the commit passed to this action had.
+
+  Note: this action makes use of git-internals (mostly layout of object-db). It will only work if
+  those are not violated (e.g. no git-gc or repack must be done prior to calling this action).
+
+inputs:
+  timestamp-reference:
+    required: true
+    description: |
+      a filepath pointing to a file that was created before the head-commit
+
+outputs:
+  commit-digest:
+    description: |
+      the commit-digest of the head-commit (for convenience - previous step might already know
+      it, of course)
+    value: ${{ steps.capture.outputs.commit-digest }}
+  commit-objects:
+    description: |
+      a base64-encoded tarfile containing the objects required to restore the head-commit.
+    value: ${{ steps.capture.outputs.commit-objects }}
+
+runs:
+  using: composite
+  steps:
+    - name: install-git
+      run: |
+        if which git &>/dev/null; then exit 0; fi
+        apt-get install -y git
+      shell: bash
+    - name: capture-commit
+      id: capture
+      run: |
+        ts_ref=${{ inputs.timestamp-reference }}
+        if [ ! -f ${ts_ref} ]; then
+          echo "no timestamp-reference-file at expected path: ${ts_ref}"
+          exit 1
+        fi
+        objects_out='commit-objects.tar'
+        commit_out='commit-digest'
+        GIT_DIR=$PWD/.git \
+          ${GITHUB_ACTION_PATH}/capture-commit \
+            ${objects_out} \
+            ${ts_ref} \
+            ${commit_out}
+
+        tar tf commit-objects.tar
+
+        echo "commit-digest=$(cat ${commit_out})" >> ${GITHUB_OUTPUT}
+        echo "commit-objects=$(cat ${objects_out} | base64 -w0)" >> ${GITHUB_OUTPUT}
+        git clean -dfx
+      shell: bash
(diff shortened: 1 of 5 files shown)


.github/workflows/build-and-test.yaml
```diff
@@ -75,13 +81,33 @@ jobs:
...
+        git config --global --add safe.directory $PWD
+        git config user.name 'Gardener-CICD Bot'
+        git config user.email 'gardener.ci.user@gmail.com'
+        echo ${{ steps.version.outputs.effective_version }} | .ci/write-version
+        touch /tmp/timestamp-ref
+        git add .
+        if ${{ inputs.release || false }}; then
+          git commit -m "Release ${{ steps.version.outputs.effective_version }}"
+        else
+          git commit -m "Build ${{ steps.version.outputs.effective_version }}"
```
8R0WNI3, line 108: I'm not sure if I understood it right: This "build" commit is actually not required, is it? Or more generally, the whole capture/import-commit logic is only required if `inputs.release` is set, correct? If so, I would suggest to drop the condition here and pull it up to the "create commit with effective version" step (same for "capture-commit").

.github/workflows/release.yaml
```diff
@@ -0,0 +1,116 @@
...
+    inputs:
+      release:
+        required: false
+        type: boolean
+        default: false
+      release-to-pypi:
+        required: false
+        type: boolean
+        default: false
+    push: # todo: rm push-trigger again after testing (we want release to be triggered manually)
```
8R0WNI3, line 13: Todo: Remove push trigger :-)

.github/workflows/release.yaml
```diff
@@ -0,0 +1,116 @@
...
+            apt-get install git -y
+          fi
+      - uses: actions/checkout@v4
+      - name: Import Release-Commit
+        uses: ./.github/actions/import-commit
+        with:
+          commit-objects: ${{ needs.build.outputs.release-commit-objects }}
+          commit-digest: ${{ needs.build.outputs.release-commit-digest }}
+
+      - name: debug
```
8R0WNI3, line 44: remove?

.github/actions/capture-commit/action.yaml
```diff
@@ -0,0 +1,73 @@
...
+        objects_out='commit-objects.tar'
+        commit_out='commit-digest'
+        GIT_DIR=$PWD/.git \
+          ${GITHUB_ACTION_PATH}/capture-commit \
+            ${objects_out} \
+            ${ts_ref} \
+            ${commit_out}
+
+        echo "debug: printing contents of commit-objects"
+        tar tf commit-objects.tar
```
8R0WNI3, line 68: remove debug print?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,116 @@
...
+          next_version=${{ needs.version.outputs.next_version }}
+          echo "next version: ${next_version}"
+          git reset --hard @~
+          echo "${next_version}" | .ci/write-version
+          git add .
+          git commit -m"Prepare next Dev-Cycle"
+          git pull --rebase
+          git push origin
+      - name: Retrieve Documentation
+        uses: actions/download-artifact@v4
```
8R0WNI3, line 69: This is only required if `inputs.release` is set I suppose?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,116 @@
+name: Release
```
8R0WNI3, line 1: Hmm, I'm not quite sure about this structure/naming. Since one can trigger this "Release" workflow with `release` set to `false`, I think it is a little bit misleading. Also, if I'm not mistaken, all jobs except `build` are skipped if `release` is set to `false`. Can't we also allow manual triggering of the `build-and-test` workflow (equivalent to running this workflow without enable `release`), and remove the `release` input from here and assume it's always an actual release?

.github/workflows/build-and-test.yaml
ccwienk: @8R0WNI3 : you are correct in that it is not needed in _our_ case (asamof, I also first thought simply not creating this commit in else-case might be the better choice).

However, I figured it might be preferable to consistently create a "commit bearing effective version". Rationale: also for snapshot-builds, there are cases where build-tools need to see changed versionfiles.

OTOH, creating a commit that will not actually be published indeed has the downside of making it difficult (actually impossible w/o looking into buildlogs) to retrospectively map back to build-commit.

So I suppose we also need a second codepath for transporting version-diff (either by leveraging "capture-commit" (but only doing a checkout, rather than a rebase), or by using a different mechanism (e.g. re-create the diff, or expose a patchfile)). Either way, non-release-builds should - different than release-builds - indeed behave differently.

That's unfortunately many different codepaths downstream :-(

.github/workflows/release.yaml
ccwienk: good catch!

.github/workflows/release.yaml
ccwienk: @8R0WNI3 : you are correct (actually, I think I felt the urge of adding a comment hinting towards removal).

My idea for leaving it in, and setting default-value to `false` was two-fold:

for once, this might be an option to trigger a test-release-run (admittedly of limited value, as you pointed out)
secondly, I assumed it might be a nice "are-you-sure"-thingy avoiding accidental triggering. OTOH reconsidering, I think that might not actually be an issue. will change it as suggested.

.github/workflows/build-and-test.yaml
ccwienk: okay, considering we do not need this transient commit except for release, I agree w/ your proposal of skipping the capturing entirely. I suppose that's a task for another increment.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,106 @@
...
+          commit_digest=$(git rev-parse @)
+          tgt_ref="refs/tags/${effective_version}"
+          echo "pushing release-commit ${commit_digest} to ${tgt_ref}"
+          git push origin "@:${tgt_ref}"
+          next_version=${{ needs.build.outputs.next_version }}
+          echo "next version: ${next_version}"
+          git reset --hard @~
+          echo "${next_version}" | .ci/write-version
+          git add .
+          git commit -m"Prepare next Dev-Cycle"
```
8R0WNI3, line 56: nit: I found it useful to see the actual version from the commit message, e.g. when scrolling through the commit history. May we add the `next_version` here?

.github/workflows/build-and-test.yaml
```diff
@@ -280,7 +284,7 @@ jobs:
             > component-descriptor.yaml
 
           if ${{ inputs.release || false }}; then
-            commit_digest=${{ steps.releasecommit.outputs.release_commit_digest || '' }}
+            commit_digest=${{ needs.version.outputs.release-commit-digest || '' }}
```
8R0WNI3, line 287: In which case is the `|| ''` required?

.github/workflows/build-and-test.yaml
```diff
@@ -75,13 +94,31 @@ jobs:
...
         echo "repo_version=${src_version}" >> "${GITHUB_OUTPUT}"
         echo "setuptools_version=${setuptools_version}" >> "${GITHUB_OUTPUT}"
+    - name: create commit with effective version
+      if: ${{ inputs.release }}
+      run: |
+        git config --global --add safe.directory $PWD
+        git config user.name 'Gardener-CICD Bot'
+        git config user.email 'gardener.ci.user@gmail.com'
+        echo ${{ steps.version.outputs.effective_version }} | .ci/write-version
+        touch /tmp/timestamp-ref
```
8R0WNI3, line 104: Don't we have to guard against cases where `/tmp` is not part of the `.gitignore` file?

.github/workflows/release.yaml
ccwienk: good idea!

.github/workflows/build-and-test.yaml
ccwienk: by default, `/tmp` is outside of repository-tree, so: no :-)

However, I think we should anyhow maybe switch to a mechanism that does not need a reference-file. For example, we might consider to explicitly set old timestamps (as in epoch-0) after git-clone. that way, any file from worktree might be used as reference-file, thus making the interface less clumsy

.github/workflows/build-and-test.yaml
ccwienk: good catch! it is in our workflow not required at all, considering we always create a crelease-commit in release-case.

PR #1091, opened 2024-12-03
refactor/cleanup: mv `main_source` -> ocm.util

ccwienk:
As another step towards discarding `cnudie` package, mv cnudie.util's main_source to ocm.util. Slightly overhaul while doing so (give callers more control about lookup/raising behaviour).

Replace usages within this repository (there outside usages, hence keeping legacies in cnudie.util for smooth migration). Also, rm some unused imports.

### DIFF
--- a/ocm/util.py
+++ b/ocm/util.py
@@ -1,0 +1,46 @@
+import ocm
+
+
+def as_component(
+    component: ocm.Component | ocm.ComponentDescriptor,
+    /,
+) -> ocm.Component:
+    if isinstance(component, ocm.Component):
+        return component
+    if isinstance(component, ocm.ComponentDescriptor):
+        return component.component
+
+    raise ValueError(component)
+
+
+def main_source(
+    component: ocm.Component | ocm.ComponentDescriptor,
+    *,
+    no_source_ok: bool=True,
+    ambiguous_ok: bool=True,
+) -> ocm.Source | None:
+    '''
+    returns the "main source" of the given OCM Component. Typically, components will have exactly
+    one source, in which the applied logic is to return the sole source-artefact.
+
+    For other cases, behaviour can be controlled via kw-(only-)params:
+
+    no_source_ok: if component has _no_ sources, return None
+    ambiguous_ok: if component has more than one source, return first
+
+    In cases where no main-source can be determined, raises ValueError.
+    '''
+    component = as_component(component)
+
+    if len(component.sources) == 1:
+        return component.sources[0]
+    elif not component.sources:
+        if no_source_ok:
+            return None
+        else:
+            raise ValueError('no sources', component)
+
+    if ambiguous_ok:
+        return component.sources[0]
+
+    raise ValueError('could not umambiguously determine main-source', component)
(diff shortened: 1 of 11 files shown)


ocm/util.py
```diff
@@ -0,0 +1,58 @@
+import ocm
+
+
+def as_component(
+    component: ocm.Component | ocm.ComponentDescriptor,
+    /,
+):
```
8R0WNI3, line 7: nit: Add type-hint for return value.

ocm/util.py
```diff
@@ -0,0 +1,58 @@
...
+    if isinstance(component, ocm.ComponentDescriptor):
+        component = component.component
```
8R0WNI3, line 9: This `component` value is actually never returned (instead the `ValueError` will be raised).

ocm/util.py
```diff
@@ -0,0 +1,58 @@
...
+
+    raise ValueError(component)
+
+
+def main_source(
+    component: ocm.Component | ocm.ComponentDescriptor,
+    *,
+    no_source_ok: bool=True,
+    ambiguous_ok: bool=True,
+    honour_label: bool=False,
```
8R0WNI3, line 21: Are you aware that this is not backwards compatible? Before, the label was honoured by default/always.

ccwienk: @8R0WNI3 : yes, I am, and this change is intentional (I pondered a bit about that yesterday)

rationale: `ocm` package is (at least that's my hope/vision) intended to be "vanilla-ocm", whereas our gardener-specific labels are more of a "gardener-ocm-flavour". Hence, I actually considered to drop support entirely (esp. considering that in current gardener-bom there is _no_ case of more than one source-artefact). then again, I thought it might be nice to still keep it in, but figured it might be better to not enable it by default

asamof, I actually considered "hiding" it even more by accepting it via `**kwargs` 🤔 
wdyt, should I rather go for that? I am a bit undecided.

8R0WNI3: Hmm, I see your point here. As (yet) another approach, wdyt of passing in a callback `Callable[[ocm.Source], bool]` which allows determining the source in case multiple sources are available? 😅

ocm/util.py
ccwienk: are you sure? in line 10, there is an `if` (not an `elif`)

However, we can save an unnecessary invocation of `isinstance` by using `return`, so I will change it

ocm/util.py
ccwienk: believe it or not, but that's exactly what I was also pondering. However, I decided against the callback for pragmatic reasons (we would have had to define this callback somewhere outside `ocm` package + use it in quite a few places.

However, considering we _currently_ do not need it, so I suppose we might just as well drop it for now, and re-introduce it, as needed, later (which might be in form of a callback).

Yet again, maybe this is actually too special a function to be part of `ocm.util`. OCM as such does not know a concept of a source being special ("the main source"). Still, in current codebase (including referencing code), there are several places that could not reasonably handle the presence of multiple sources + in context of gardener-cicd (I assume probably also in other CICD-systems), it _is_ typically the case that there is a de-facto "main source".

Another idea I had (actually in two slightly similar flavours) was to:
- pass-in label-name to honour
- pass-in "mode" (with `gardener` leading to the behaviour currently implemented for `honour_label`)

I tended to like this idea, as it is quite explicit. However, I disliked it, as it suggested a degree of genericity the implementation actually does not have (and probably should not have).

PR #1094, opened 2024-12-09
Purge `DeliveryCfg`

zkdev:
Was mved to delivery-service in order to (further) decouple ocm-gear from cc-utils.

Fixes #

**Release note**:

```other operator

```


8R0WNI3 (approved):
/lgtm

ccwienk (changes requested):
won't this break ccc.delivery?

zkdev:
> won't this break ccc.delivery?

Why would it?
`ccc.delivery` only relies on delivery_endpoints cfg, which is decoupled from delivery cfg 🤔

ccwienk (approved):

PR #1097, opened 2024-12-10
allow configuring release-trait to not publish release-commits

ccwienk:
As (yet)  a(nother) workaround for an inconsistency w.r.t. ocm-component-descriptor-creation/publishing, allow configuring release-trait such that it will neither create nor publish release-commits (both release-commit itself and bump-commit).

If configured like that, release-trait can be added to head-update-pipelines to thus publish component-descriptors consistent w/ those published from regular release-jobs.

### DIFF
--- a/concourse/steps/release.mako
+++ b/concourse/steps/release.mako
@@ -70,12 +70,23 @@ release_commit_publishing_policy = release_trait.release_commit_publishing_polic
 if release_commit_publishing_policy is ReleaseCommitPublishingPolicy.TAG_ONLY:
   merge_back = False
   push_release_commit = False
+  create_release_commit = True
+  bump_commit = True
 elif release_commit_publishing_policy is ReleaseCommitPublishingPolicy.TAG_AND_PUSH_TO_BRANCH:
   merge_back = False
   push_release_commit = True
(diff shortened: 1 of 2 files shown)


concourse/steps/release.mako
```diff
@@ -70,12 +70,23 @@ release_commit_publishing_policy = release_trait.release_commit_publishing_polic
...
+  bump_commit = True
 elif release_commit_publishing_policy is ReleaseCommitPublishingPolicy.TAG_AND_MERGE_BACK:
   push_release_commit = False
   merge_back = True
+  create_release_commit = True
+  bump_commit = True
+elif release_commit_publishing_policy is ReleaseCommitPublishingPolicy.SKIP:
+  push_release_commit = False
+  merge_back = False
+  create_release_commit = false
```
8R0WNI3, line 88: nit: `false` -> `False`

8R0WNI3 (approved):
/lgtm

concourse/steps/release.mako
ccwienk: again: good catch. I wonder why this did not break my local test 🤔

PR #1104, opened 2024-12-19
Lookup job-mapping in multiple sets for WHD PR processing

zkdev:
Necessary for cases where a single WHD instance should handle resource triggers on multiple Concourse instances.

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/whd/pull_request.py
+++ b/whd/pull_request.py
@@ -335,10 +335,23 @@ def set_pr_labels(
     repo = pr_event.repository()
     repository_path = repo.repository_path()
     owner, _ = repository_path.split('/')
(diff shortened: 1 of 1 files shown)


whd/pull_request.py
```diff
@@ -335,10 +335,23 @@ def set_pr_labels(
...
+                cfg_set=cfg_set,
+            )
+        except ValueError:
+            job_mapping = None
+
+        if job_mapping:
+            break
+
+    else:
+        raise ValueError(f'no job-mapping found for {repo.repository_url()}')
```
ccwienk, line 354: this raise is new. I assume this will not be an issue for WHD, but improve logging /traceability?

ccwienk (approved):

PR #1111, opened 2025-01-07
cleanup: rm dead code

ccwienk:
also, replace last remaining usage of FluentIterable w/ Generator.

**Release note**:

```other developer
drop some unused code, mainly from ci.util as another refactoring towards making cc-utils less heavy to consume (reduce dependencies).
```


ci/util.py
```diff
@@ -212,14 +211,6 @@ def warning(msg:str):
         _print('WARNING: ' + str(msg), colour='yellow', outfh=sys.stderr)
 
 
-@deprecated.deprecated
-def verbose(msg:str):
-    if not _verbose():
```
8R0WNI3, line 217: I think `_verbose` can be dropped as well 🤔

github/util.py
```diff
@@ -311,36 +314,26 @@ def enumerate_upgrade_pull_requests(
         state: str='all',
         pattern: re.Pattern=None,
     ):
-        '''returns a sequence of `UpgradePullRequest` for all found pull-requests
+        '''returns a generator yielding `UpgradePullRequest` for all found pull-requests
```
8R0WNI3, line 317: Let us rather put this into a type hint for the return value.

ccwienk: great idea!

8R0WNI3 (approved):

PR #1112, opened 2025-01-08
cleanup: do not unncecessarily configure logger

ccwienk:
Configuring `logging` gear should be to leaf components. This does not apply to either ccc or model packages, hence drop corresponding calls.

**Release note**:

```bugfix developer
remove unnecessary invocations of ci.log.configure_default_logging in common imports (model + ccc) to avoid issues from multiple invocations, which may result in logger with broken cfg (which results in logging-outputs not to be emitted at all).

Note: this _may_ require leaf-scripts that previously relied on implicit calling of ci.log.configure_default_logging to now need to explicitly initialise logging-package.
```

### DIFF
--- a/ccc/delivery.py
+++ b/ccc/delivery.py
@@ -7,7 +7,6 @@ import ctx
 import delivery.client
 import model.base
 
-ci.log.configure_default_logging()
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


zkdev (approved):
/lgtm

PR #1117, opened 2025-01-10
Base component descriptor workflow

ccwienk:
**Release note**:

```other developer
implement a reusable workflow for conveniently generating `base ocm-component-descriptors`
```

### DIFF
--- a/.github/workflows/base-component-descriptor.yaml
+++ b/.github/workflows/base-component-descriptor.yaml
@@ -1,0 +1,161 @@
+name: Generate Base Component-Descriptor
+on:
+  workflow_call:
+    inputs:
+      version:
+        required: true
+        type: string
+      component-name:
+        required: false
+        type: string
+        description: |
+          Sets the Component-Name. If not passed, defaults to repository-URL
+      ocm-repo:
+        required: true
+        type: string
+        description: |
+          the OCM-Repository the Component-Descriptor is intended to be published to
+      commit-digest:
+        required: false
+        type: string
+        description: |
+          the commit-digest to use for declaring main source. If not passed, will default to
+          current HEAD. Useful in conjunction with `capture-commit` / `import-commit`, if
+          release-commit is created upfront.
+      provider:
+        required: false
+        type: string
+        default: SAP SE
+      labels:
+        required: false
+        type: string
+        description: |
+          Labels to set for the component in YAML-form (caveat: need to quote). May either be a
+          single object, or an array.
+          Example:
+            # single label
+            name: cloud.gardener.cnudie/responsibles
+            value:
+              - type: githubTeam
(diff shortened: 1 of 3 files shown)


ocm/__main__.py
```diff
@@ -4,6 +4,7 @@
 import json
 import os
 import sys
+import typing
```
8R0WNI3, line 7: Since this is newly introduced code, rather use `collections.abc` instead of `typing` from the beginning ([PEP-585](https://peps.python.org/pep-0585/)).

ocm/__main__.py
```diff
@@ -131,9 +131,23 @@ def append(parsed):
...
 
+    def inject_labels(artefact: dict):
+        if not (labels := parsed.labels):
+            return
+
+        labels = list(_iter_parsed_labels(labels=labels))
+        if not 'labels' in artefact:
+            artefact['labels'] = []
+
+        artefact['labels'] = [dataclasses.asdict(l) for l in labels]
```
8R0WNI3, line 142: nit: You may factor-out the conversion from `parsed.labels` to this list here since it's the same for all passed-in artefacts.

.github/workflows/base-component-descriptor.yaml
```diff
@@ -0,0 +1,161 @@
...
+        type: string
+        description: |
+          Labels to be set for main-source. Same syntax as for `labels`
+      artefact-name:
+        default: base-component-descriptor
+        type: string
+        description: |
+          Base-Component-Descriptor is exposed both via output (component-descriptor) and as
+          artefact. If needed, target-artefact-name can be configured through this input.
+    inputs:
```
8R0WNI3, line 59: I assume this is supposed to be `outputs`.

8R0WNI3 (commented):
Also, please remove your WIP commit from this PR (again) :^)

ocm/__main__.py
ccwienk: dang.... I was not aware this is also recommended for `Generator`. +1

.github/workflows/base-component-descriptor.yaml
ccwienk: yes. I already fixed that last friday (but apparently, I failed to include this patch into this pullrequest m()

ccwienk:
> Also, please remove your WIP commit from this PR (again) :^)

🙈 
while working, I used to do:
- add new changes/fixes for existing commits (except the topmost WIP-one) + git-commit using --fixup
- run git rebase --autosquash
- run git-push against my fork + this pullrequest (using @~ as ref to push)

apparently, the last rebase did not properly do autosquashing, resulting in the wrong commit to be skipped 🤦🏻‍♂️

PR #1119, opened 2025-01-15
Remove dead code

8R0WNI3:
This code was used for finding-specific SLA tracking (i.e. for BDBA findings). However, said handling of findings was moved to delivery-service and does not rely on `ScanResultGroups` anymore.

**Release note**:

```other operator

```


ccwienk (approved):
very nice - thx for tidying our codebase

PR #1124, opened 2025-01-17
fix: do not overwrite existing labels

ccwienk:
**Release note**:

```bugfix developer
fix a bug where `gardener-ocm resource|source add` would overwrite labels from passed artefact-fragments.
```


zkdev (approved):
/lgtm

PR #1127, opened 2025-01-21
migrate cc pipeline to gha

TuanAnh17N:
**Release note**:

```other operator

```


.github/workflows/cc-job-image-base.yaml
```diff
@@ -0,0 +1,97 @@
...
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+          registry: ghcr.io
+
+      - name: Build and Publish CC-Job-Image-Base
+        uses: docker/build-push-action@v6
+        with:
+          push: true
+          platforms: ${{ env.platforms }}
+          tags: ghcr.io/${{ github.repository_owner }}/cc-utils/job-image-base:latest
```
ccwienk, line 47: while it is okay to (also) set `latest`-tag, I would prefer to stick to also setting version-specific tags

.github/workflows/cc-job-image-base.yaml
```diff
@@ -0,0 +1,97 @@
...
+          file: Dockerfile.job-image-base
+
+      - name: Get latest Docker version
+        id: docker_version
+        run: |
+          greatest_version=$(curl -sL \
+            https://download.docker.com/linux/static/stable/x86_64 \
```
ccwienk, line 54: nit: if you append a `/` to the url, you will safe (curl) a redirect 🤓

.github/workflows/cc-job-image-base.yaml
```diff
@@ -0,0 +1,97 @@
...
+            | cut -d\" -f2 \
+            | grep "docker-[[:digit:]]" \
+            | cut -d- -f2 \
+            | cut -d. -f1,2,3 \
+            | sort -Vr \
+            | head -1)
+          echo "DOCKER_VERSION=$greatest_version" >> $GITHUB_ENV
```
ccwienk, line 61: is there a reason to pass values via env rather than outputs? I think it will not make much of a difference in this context; however, using outputs seemed more idiomatic to me thus-far (at the price of having to type lengthier code for dereferentiation). interested to learn about your thoughts

.github/workflows/cc-job-image-base.yaml
```diff
@@ -0,0 +1,97 @@
...
+      - name: Build and Publish CC-Job-Image-Base
+        uses: docker/build-push-action@v6
+        with:
+          push: true
+          platforms: ${{ env.platforms }}
+          tags: ghcr.io/${{ github.repository_owner }}/cc-utils/job-image-base:latest
+          file: Dockerfile.job-image-base
+
+      - name: Get latest Docker version
+        id: docker_version
```
ccwienk, line 51: nit: currently, you never use the ID (as you would to e.g. access step's outputs (`${{ steps.<id>.outputs.<output>}}`). hence, you might drop those id-attrs (it is not wrong to set them of course).

.github/workflows/cc-job-image-base.yaml
```diff
@@ -0,0 +1,97 @@
...
+        run: |
+          sed -i "s/DOCKER_VERSION=.*/DOCKER_VERSION=${{ env.DOCKER_VERSION }}/" \
+          Dockerfile.job-image-base
+          sed -i "s/DOCKER_BUILDX_VERSION=.*/DOCKER_BUILDX_VERSION=${{ env.BUILDX_VERSION }}/" \
+          Dockerfile.job-image-base
+          git add Dockerfile.job-image-base
+          git commit -m "Update Dockerfile to use Docker ${{ env.DOCKER_VERSION }} and \
+          Buildx ${{ env.BUILDX_VERSION }}"
+          git checkout -b update-dockerfile
+          git push origin update-dockerfile
```
ccwienk, line 85: nit: it is not necessary (but of course also not wrong) to do a checkout prior to push. however, hardcoding target-branch (which is a reasonable way to avoid spamming branches / reduce need to do cleanups) will result in this push to fail upon second time the job is executed. this might be solved by doing a force-push. as an alternative, we might of course generate random branch-names (at the disadvantage I pointed out before).

However, I think this update-job should be separated from the build-and-push-job, resulting in two jobs:
```
- build-and-push (and release / bump base-image-version)
- update base-image's dockerfile (note: we might leverage dependabot for that - according to @zkdev , it features custom update-hooks)
```

after job-image-base's publishing-pipeline will be switched to setting version-specific tags, it will also be nice to have the release/build-and-push-job create a second kind of pullrequests: one that will suggest to bump base-image in job-image's Dockerfile

.github/workflows/cc-job-image-base.yaml
```diff
@@ -0,0 +1,97 @@
...
+
+      - name: Create Pull Request
+        run: |
+          gh pr create \
+            -B master \
+            -H update-dockerfile \
+            --title "Update Dockerfile with latest Docker and Buildx versions" \
+            --body "This PR updates the Dockerfile to use the latest versions:
```
ccwienk, line 93: wording-nit: you might drop mentioning in the pullrequest's body that the pullrequest is a pullrequest (-> Smurf-Typing..) ;-)

.github/workflows/cc-job-image-base.yaml
```diff
@@ -0,0 +1,97 @@
...
+              - Docker: `${{ env.DOCKER_VERSION }}`
```
ccwienk, line 94: I wonder: will it not cause issues to have unescaped ``-pairs? 🤔

Dockerfile.job-image-base
```diff
@@ -0,0 +1,63 @@
...
+&& pip3 install --upgrade --no-cache-dir \
+   pip \
+&& pip3 install --root /pkgs --upgrade --no-cache-dir \
+   $(grep -v '#' requirements.txt | grep -v 'gardener' | tr '\n' ' ')
+
+FROM alpine:3
+
+
+ENV DOCKER_CHANNEL=stable
+ENV DOCKER_VERSION=27.4.1
```
ccwienk, line 20: change to ARGs?

Dockerfile.job-image-base
```diff
@@ -0,0 +1,63 @@
...
+&& echo "before second aia" \
+&& curl http://aia.pki.co.sap.com/aia/SAPNetCA_G2.crt -o \
+    /usr/local/share/ca-certificates/SAPNetCA_G2.crt \
+&& curl -Lo /usr/local/share/ca-certificates/SAPNetCA_G2_2.crt \
+    http://aia.pki.co.sap.com/aia/SAPNetCA_G2_2.crt \
+&& update-ca-certificates \
+&& dos2unix /etc/ssl/certs/ca-certificates.crt \
+&& ln -sf /etc/ssl/certs/ca-certificates.crt "$(python3 -m certifi)" \
+&&  mkdir -p ~/.ssh \
+&& ssh-keyscan -H github.wdf.sap.corp github.com >> ~/.ssh/known_hosts
```
ccwienk, line 63: nit: missing newline

apk-packages.blacklist
```diff
@@ -0,0 +1,5 @@
+gcc
+libc-dev
+libffi-dev
+openssl-dev
+python3-dev
```
ccwienk, line 5: newline

ccwienk (changes requested):

.github/workflows/cc-job-image-base.yaml
TuanAnh17N: i initially opted for `env` because it felt more concise, and using `outputs` seemed a bit too lengthy. that said, i agree that `outputs` is more explicit

ccwienk: I see. I think, for such small workflows, it should be okay to use env, specifically for the reason you mentioned (I assumed this was your motivation, just wanted to clarify)

zkdev (changes requested):

Dockerfile.job-image-base
```diff
@@ -0,0 +1,63 @@
...
+&& echo "before docker" \
+&& curl -Lf \
+ "https://download.docker.com/linux/static/${DOCKER_CHANNEL}/${DOCKER_ARCH}/docker-${DOCKER_VERSION}.tgz" \
+ | tar zx -C /opt \
+&& mkdir -p /usr/lib/docker/cli-plugins \
+&& echo "before docker-buildx" \
+&& curl -Lo /usr/lib/docker/cli-plugins/docker-buildx \
+ https://github.com/docker/buildx/releases/download/${DOCKER_BUILDX_VERSION}/buildx-${DOCKER_BUILDX_VERSION}.linux-${TARGETARCH} \
+&& chmod +x /usr/lib/docker/cli-plugins/docker-buildx \
+&& echo "before first aia" \
```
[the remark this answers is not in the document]
ccwienk: @zkdev : yes, I once added those as a debugging artefact (already some years ago..).

we _might_ consider factoring those commands into a shellscript (which will allow us to add debugging output / get a helpful error indication (line-number)). the reason for using chained commands is having less image-layers. by grouping commands in a shellscript (which we might transiently mount using RUN's `--mount`-option) will allow us to still only create one layer, but also no longer have to do command-chaining (which makes debugging very cumbersome). wdyt?

ccwienk (approved):

PR #1128, opened 2025-01-22
Heldkat patch 1

heldkat:
**Release note**:

```other operator
update of output file path
```

### DIFF
--- a/cli/gardener_ci/compliance.py
+++ b/cli/gardener_ci/compliance.py
@@ -48,13 +48,13 @@ class DiffArguments:
     right_name: str
     left_version: str
     right_version: str
+    outfile_prefix: str
     ocm_repo_urls: list[str]
     exclude_component_names: list[str] = None
     exclude_component_resource_names: list[ComponentResourceNames] = None
     resource_types: list[str] = None
     name_template: str = None
     name_template_expr: str = None
-    outfile_prefix: str = 'resource-diff'
 
 
 def diff(
@@ -91,6 +91,8 @@ def diff(
         params['name_template_expr'] = name_template
     if resource_types:
         params['resource_types'] = resource_types
+    if outfile_prefix:
+        params['outfile_prefix'] = outfile_prefix
 
     try:
         parsed = dacite.from_dict(


cli/gardener_ci/compliance.py
```diff
@@ -256,8 +256,8 @@ def resource_as_dict(component, resource, resource_id):
         resource_as_dict(c,r,i) for c,r,i in removed_resource_version_ids
     ])))
 
-    outfile_new = f'{parsed.outfile_prefix}-added.yaml'
-    outfile_removed = f'{parsed.outfile_prefix}-removed.yaml'
+    outfile_new = f'os.environ['SOURCE_PATH']+'/versionDiff/{parsed.outfile_prefix}-added.yaml'
```
ccwienk, line 259: that won't work (syntax-error). also I think it is not nice to hardcode assumptions about presence of SOURCE_PATH-env-var. as discussed out-of-bands, check whether passing `--outfile-prefix` via ARGV might solve what you are trying to achieve

ccwienk (changes requested):

cli/gardener_ci/compliance.py
heldkat: I adapted the file to set the `outfile_prefix` parameter to `None`, so that I can call it with a self-defined string. Would this work?

ccwienk: @heldkat : well... in principal: yes. however, this will de-facto turn the (thus-far optional) parameter into a required one

heldkat: How about now? This would give me the option to define it in the `defaults-file` but does not require it in my understanding

cli/gardener_ci/compliance.py
```diff
@@ -68,7 +69,7 @@ def diff(
     ocm_repo_urls: [str]=None,
     cache_dir: str=_cfg.ctx.cache_dir,
     defaults_file: str=None,
-    outfile_prefix: str='resource-diff'
+    outfile_prefix: str= 'resource-diff'
```
ccwienk, line 72: I think according to our linter-rules (and pep8), there should not be a whitespace after `=` here (also, whitespace should be used consistently at least within one codeblock)

ccwienk (commented):
please fix whitespace-nit. otherwise lgtm

cli/gardener_ci/compliance.py
heldkat: ah, perfect. I was wondering why it is not passing the linter test

cli/gardener_ci/compliance.py
```diff
@@ -48,13 +48,14 @@ class DiffArguments:
...
     right_version: str
+    outfile_prefix: str
     ocm_repo_urls: list[str]
     exclude_component_names: list[str] = None
     exclude_component_resource_names: list[ComponentResourceNames] = None
     resource_types: list[str] = None
     name_template: str = None
     name_template_expr: str = None
-    outfile_prefix: str = 'resource-diff'
+
```
ccwienk, line 58: I think this is one line too much (pep8 and our linter mandates exactly two lines - see linter output (linked to your pullrequest))

ccwienk (commented):
also, please squash commits into just one

ccwienk (approved):

PR #1129, opened 2025-01-22
factor draft-release-updater into reusable workflow

ccwienk:
**Release note**:

```other developer
add new reusable workflow `draft-release.yaml`
```

### DIFF
--- a/release_notes/fetch.py
+++ b/release_notes/fetch.py
@@ -235,17 +235,17 @@ def _determine_blocks_to_include(
 
 
 def fetch_draft_release_notes(
-    current_version: str,
     component: ocm.Component,
     component_descriptor_lookup: cnudie.retrieve.ComponentDescriptorLookupById,
     version_lookup: cnudie.retrieve.VersionLookupByComponent,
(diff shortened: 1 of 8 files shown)


release_notes/fetch.py
```diff
@@ -304,66 +303,66 @@ def fetch_release_notes(
     version_lookup: cnudie.retrieve.VersionLookupByComponent,
     git_helper: gitutil.GitHelper,
     github_api_lookup: rnu.GithubApiLookup,
-    current_version: typing.Optional[str] = None,
-    previous_version: typing.Optional[str] = None,
+    version_whither: str | None = None,
+    version_whence: str | None = None,
```
8R0WNI3, line 307: nit: Shouldn't we keep naming consistent with `fetch_draft_release_notes`?

8R0WNI3 (approved):
/lgtm

release_notes/fetch.py
ccwienk: yes, so I also figured yesterday evening. but I also figured that's for another commit (will add to this pullrequest) :-)

PR #1135, opened 2025-01-29
extract update-dockerfile step to its own workflow

TuanAnh17N:
**Release note**:

```other operator
Extracted `update-dockerfile` step into a standalone workflow, running on a daily schedule
```


.github/workflows/update-dockerfile.yaml
```diff
@@ -0,0 +1,87 @@
+name: Dockerfile Version Bump
+run-name: Update Dockerfile with latest Docker and Buildx versions
```
8R0WNI3, line 2: nit: I'd also express in the workflow name that this job will only update the job-image-base's Dockerfile

8R0WNI3 (approved):
One nit, otherwise /lgtm

PR #1136, opened 2025-01-29
Drop `cfg_name` property from compliance snapshots

8R0WNI3:
**Release note**:

```other developer
Drop cfg_name property from compliance snapshots
```

### DIFF
--- a/dso/model.py
+++ b/dso/model.py
@@ -602,14 +602,13 @@ class ComplianceSnapshotState:
 
 @dataclasses.dataclass(frozen=True)
 class ComplianceSnapshot:
-    cfg_name: str
     latest_processing_date: datetime.date
     correlation_id: str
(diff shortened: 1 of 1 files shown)


ccwienk (approved):

zkdev (approved):
What will this mean for migration path?
I assume all existing compliance snaptshots will be re-created, but do we have to take care of cleaning up the old ones manually?

8R0WNI3:
> What will this mean for migration path? I assume all existing compliance snaptshots will be re-created, but do we have to take care of cleaning up the old ones manually?

This will be part of another DB migration which will be added to the deployment script, including the removal of the `cfg_name` column as well as the deletion of all existing compliance snapshots (those will be re-created automatically the next time the artefact enumerator takes action).

PR #1138, opened 2025-01-29
fix another regression: set draft-name

ccwienk:
**Release note**:

```bugfix user
fix regression: set draft-name again to avoid name-clashes
```


zkdev (approved):
/lgtm

PR #1144, opened 2025-02-03
introduce automated credential rotation for BDBA

TuanAnh17N:
**Release note**:

```other operator
Introduce automated credential rotation for BDBA
```

### DIFF
--- a/cfg_mgmt/bdba.py
+++ b/cfg_mgmt/bdba.py
@@ -1,0 +1,50 @@
+import logging
+
+import bdba.client
+import cfg_mgmt
+import model.bdba
+
+
+logger = logging.getLogger(__name__)
+
+
+def rotate_cfg_element(
+    cfg_element: model.bdba.BDBAConfig,
+    cfg_factory: model.ConfigFactory,
+) -> tuple[cfg_mgmt.revert_function, dict, model.bdba.BDBAConfig]:
+    logger.info(f'Rotating API key for {cfg_element.name()}')
+
+    bdba_client = bdba.client.BDBAApi(
+        api_routes=bdba.client.BDBAApiRoutes(base_url=cfg_element.api_url()),
+        token=cfg_element.credentials().token(),
+        tls_verify=cfg_element.tls_verify(),
+    )
+
+    response = bdba_client.create_key(validity_seconds=60 * 60 * 24 * 178)  # 178 days
+
+    new_key_info = response.json()
+
+    new_key = new_key_info['key']['value']
+
+    raw_cfg = cfg_element.raw.copy()
+    raw_cfg['credentials']['token'] = new_key
+
+    updated_cfg_element = model.bdba.BDBAConfig(
+        name=cfg_element.name(),
+        raw_dict=raw_cfg,
+        type_name=cfg_element._type_name
+    )
+
+    secret_id = {'api_key': new_key}
+
+    def no_op():
+        logger.critical(
+            'No rollback possible for BDBA key rotation!\n'
+            'Manuel intervention required:\n'
+            '1. A new API Key was generated but may not be saved in the config\n'
+            '2. The old key is immediatly invalid after rotation\n'
+            '3. Check logs for the new key\n'
+            '4. Update the config with the new key'
+        )
+
+    return no_op, secret_id, updated_cfg_element
--- a/cfg_mgmt/util.py
+++ b/cfg_mgmt/util.py
@@ -2,6 +2,7 @@ import dataclasses
 import datetime
 import logging
 import os
+import time
 import typing
 
 import pytimeparse
@@ -17,6 +18,25 @@ import model.base
 
 
 logger = logging.getLogger(__name__)
+
+
+class CredentialWritebackError(Exception):
+    def __init__(self, message, new_secret=None, original_exception=None):
+        super().__init__(message)
+        self.new_secret = new_secret
+        self.original_exception = original_exception
+
+    def __str__(self):
+        base_msg = super().__str__()
+        if self.new_secret:
+            return f'{base_msg} [UNPERSISTED_SECRET_REDACTED]'
+        return base_msg
+
+    def log_secret(self):
+        if self.new_secret:
+            logger.critical(
+                f'UNPERSISTED_SECRET for {type(self).__name__}: {self.new_secret}'
+            )
 
 
 def iter_cfg_elements(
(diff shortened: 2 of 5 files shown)


cfg_mgmt/bdba.py
```diff
@@ -0,0 +1,109 @@
...
+logger = logging.getLogger(__name__)
+
+
+def get_api_key_expiry(
+    cfg_element: model.bdba.BDBAConfig
+) -> datetime.datetime:
+    credentials = cfg_element.credentials()
+    headers = {'Authorization': f'Bearer {credentials.token()}'}
+
+    response = requests.get(
```
ccwienk, line 18: damn... I think it would be better to add this to delivery-service's `bdba.client`. However, this would introduce a cyclic dependency :-(

Could you:
- also add this into mentioned bdba-client
- add a reference to delivery-service-repository + a todo (to deduplicate)

OTOH, from python-package's pov, the dependency would not actually be cyclic.... sooo wdyt about declaring dependency against delivery-service's bdba-package?

/cc @zkdev | @8R0WNI3

cfg_mgmt/bdba.py
```diff
@@ -0,0 +1,109 @@
...
+import requests
+
+import cfg_mgmt
+import model.bdba
+
+logger = logging.getLogger(__name__)
+
+
+def get_api_key_expiry(
+    cfg_element: model.bdba.BDBAConfig
```
ccwienk, line 13: see comment below. Also, I think we should not build such functions around our cfg-model-classes (separation of concerns). It is okay to have a util-function. however, the logic for issueing requests against bdba should be handled by our bdba-client.

cfg_mgmt/bdba.py
```diff
@@ -0,0 +1,109 @@
...
+
+import requests
+
+import cfg_mgmt
+import model.bdba
+
+logger = logging.getLogger(__name__)
+
+
+def get_api_key_expiry(
```
ccwienk, line 12: naming-nit: expiration_date?

cfg_mgmt/bdba.py
```diff
@@ -0,0 +1,109 @@
...
+        verify=cfg_element.tls_verify(),
+        timeout=30,
+    )
+
+    response.raise_for_status()
+
+    key_info = response.json()
+    expiry_timestamp = key_info['key'].get('expires')
+
+    if not expiry_timestamp:
```
ccwienk, line 30: might save a line by using assignment-expression (:=)

cfg_mgmt/bdba.py
```diff
@@ -0,0 +1,109 @@
...
+        raise ValueError('No expiration timestamp found for the API key.')
+
+    return datetime.datetime.fromisoformat(expiry_timestamp).replace(tzinfo=datetime.timezone.utc)
```
ccwienk, line 33: why change timezone?

cfg_mgmt/bdba.py
```diff
@@ -0,0 +1,109 @@
...
+) -> tuple[cfg_mgmt.revert_function, dict, model.bdba.BDBAConfig]:
+    logger.info(f'Rotating API key for {cfg_element.name()}')
+
+    credentials = cfg_element.credentials()
+    headers = {'Authorization': f'Bearer {credentials.token()}'}
+
+    # Request a new API key
+    response = requests.post(
+        f'{cfg_element.api_url()}/api/key/',
+        json={'validity': 15379200},  # 178 days
```
ccwienk, line 48: nit: you might write this as `60 * 60 * 24 * 178` to make it easier to edit (still nice to keep comment about amount of days) / the fact that those are seconds

cfg_mgmt/bdba.py
```diff
@@ -0,0 +1,109 @@
...
+    updated_cfg_element = model.bdba.BDBAConfig(
+        name=cfg_element.name(),
+        raw_dict=raw_cfg,
+        type_name=cfg_element._type_name
+    )
+
+    secret_id = {'api_key': new_key}
+
+    def no_op():
+        logger.warning('No rollback possible for BDBA key rotation.')
```
ccwienk, line 72: If this happens, the result will (or might) be fatal, right? As in: manual resetting of apikey will be needed. It might be a good idea to print this in more detail.

cfg_mgmt/bdba.py
```diff
@@ -0,0 +1,109 @@
...
+
+    return no_op, secret_id, updated_cfg_element
+
+
+def delete_config_secret(
+    cfg_element: model.bdba.BDBAConfig,
```
ccwienk, line 78: ditto -> add to bdba.client

cfg_mgmt/bdba.py
```diff
@@ -0,0 +1,109 @@
...
+    headers = {'Authorization': f'Bearer {credentials.token}'}
+
+    response = requests.delete(
+        f'{cfg_element.api_url()}/api/key/',
+        headers=headers,
+        verify=cfg_element.tls_verify(),
+        timeout=30,
+    )
+
+    if response.status_code == 400:
```
ccwienk, line 92: will bdba really return 400 (instead of 404) in this case?

cfg_mgmt/bdba.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def no_op():
+        logger.warning('No rollback possible for BDBA key rotation.')
+
+    return no_op, secret_id, updated_cfg_element
+
+
+def delete_config_secret(
+    cfg_element: model.bdba.BDBAConfig,
+) -> model.bdba.BDBAConfig | None:
```
ccwienk, line 79: this function never returns bdbaconfig (-> hence typehint is not accurate)

cfg_mgmt/util.py
```diff
@@ -18,6 +19,9 @@
 
 logger = logging.getLogger(__name__)
 
+MAX_RETRIES = 5
```
ccwienk, line 22: I think we need more retries / longer delay (or (better)  exponential backoff).

consider we sometimes see maintenance windows of a couple of hours for GHE. so we should make our code be able to bridge such a gap

also, instead of setting module-level-constants, and read from multiple places in (one) function, preferably set those as default values in function-parameters.

cfg_mgmt/util.py
```diff
@@ -297,7 +301,31 @@ def rotate_config_element_and_persist_in_cfg_repo(
         git_helper.add_and_commit(
             message=commit_message,
         )
-        git_helper.push('@', target_ref)
+        for attempt in range(1, MAX_RETRIES + 1):
```
ccwienk, line 304: I suggest to stick to using recursion (as we commonly do in other places in our codebase). like so:

```
def function_with_retry(retries=123):
  try:
   something()
  except ExpectedException as e:
   if retries > 0:
     function_with_retry(retries=retries - 1)
```

cfg_mgmt/util.py
```diff
@@ -297,7 +301,31 @@ def rotate_config_element_and_persist_in_cfg_repo(
...
+                git_helper.push('@', target_ref)
+                logger.info(f'Successfully pushed changes to GitHub on attempt {attempt}')
+                break
+            except Exception as e:
+                if attempt < MAX_RETRIES:
+                    logger.warning(f'GitHub push failed (attempt {attempt}/{MAX_RETRIES}): {e}')
+                    logger.info(f'Retrying in {RETRY_DELAY} seconds...')
+
+                    # pull the latest changes and rebase
+                    latest_commit = git_helper.fetch_head(target_ref)
```
ccwienk, line 315: chance for success will be greater if we do fetch + hard-reset + recreate changes

cfg_mgmt/util.py
```diff
@@ -297,7 +301,31 @@ def rotate_config_element_and_persist_in_cfg_repo(
...
+                    git_helper.rebase(latest_commit.hexsha)
+
+                    time.sleep(RETRY_DELAY)
+        else:
```
ccwienk, line 319: invert (early-exit) to decrease indent, like so:

```
if <error-condition>:
  raise ....
# regular code (w/o indentation) goes here
```

ccwienk (changes requested):

cfg_mgmt/bdba.py
zkdev: not too sure about introducing a dependency towards delivery-service's bdba-package ...
this would also contain recently introduced secret-mgmt.

in my opinion it is fine to keep this specific bdba implementation in parallel, as it is not to be expected that this will be expanded in future

cfg_mgmt/bdba.py
```diff
@@ -0,0 +1,109 @@
+import datetime
+import logging
+
+import requests
+
+import cfg_mgmt
+import model.bdba
+
```
zkdev, line 8: nit: two blanks (pep8)

zkdev (changes requested):

ccwienk:
I think there is a missing package-installation (bdba-package), leading to linter-errors. Please also address those (will need extension of github-action preparing linting/test steps.

ccwienk (approved):

ccwienk:
@TuanAnh17N : I think you should assess the codeql-finding for reasons of "hygiene"

cfg_mgmt/bdba.py
```diff
@@ -0,0 +1,75 @@
...
+        )
+
+    return no_op, secret_id, updated_cfg_element
+
+
+def validate_for_rotation(cfg_element: model.bdba.BDBAConfig) -> None:
+    remaining_days = (expiration_date(cfg_element) - datetime.datetime.now()).days
+
+    if remaining_days >= 10:
+        raise ValueError(f'API key for {cfg_element.name()} does not need rotation')
```
8R0WNI3, line 72: Shouldn't this raise `cfg_mgmt.model.ValidationError` instead (looking at [ref](https://github.com/gardener/cc-utils/blob/36997fc97e47ffc1e5c304497b01b9e2c3e001db/cfg_mgmt/rotate.py#L179))?

ccwienk: hmmm..... I missed that line (apparently). I am not so convinced about hardcoding this threshold - do we do this for other occurrences as well?

I think we do not need to add such a check. We have a configuration through which we can define how often rotation should be done. With this hardcoded check, we would raise in cases where a shorter rotation-period is configured.

-> I would vote for dropping this check. If keeping, it should rather be a RE, rather than a VE, wdyt @8R0WNI3 / @TuanAnh17N ?

ccwienk (approved):

PR #1143, opened 2025-02-03
Rm some hyperscaler client factories

zkdev:
Mved to sole user, glci:
https://github.com/gardenlinux/glci/pull/154

**Release note**:

```breaking developer
`gardener-cicd` python distribution packages no longer feature client-factories for AWS and AliCloud, and GCP SDKs.
```

PR #1145, opened 2025-02-03
Factor out SDKs into cfg-mgmt package

zkdev:
**Release note**:

```breaking developer
Hyperscaler SDKs are moved to dedicated python distribution package, "gardener-cicd-cfg-mgmt".
```


zkdev:
/hold
wait for https://github.com/gardener/cc-utils/pull/1143 to be merged

ccwienk:
how about splitting this into a separate workflow (so we can have independent releases of job-image and cfg-mgmt-image)? Admittedly, this will cause some more refactoring later-on (planned gha-changes), but that's okay from my perspective.

Rationale: cfg-mgmt-image is quite heavy, and we will less often want to release new versions

downside: we will have to maintain reference from cfg-mgmt -> gardener-cicd-libs (however, we can automatically patch this upon releases of cicd-libs, so that should be finer).

also: please add a release-note :-)

.ci/build_python_packages
```diff
@@ -55,6 +55,8 @@ python3 "${repo_dir}/setup.py" bdist_wheel
 rm -rf "${repo_dir}/build"
 python3 "${repo_dir}/setup.whd.py" bdist_wheel
 rm -rf "${repo_dir}/build"
+python3 "${repo_dir}/setup.cfg_mgmt.py" bdist_wheel
```
8R0WNI3, line 58: Not directly related to this pull request, but it this script actually required anymore? I thought this functionality was completely moved into GH-action 🤔

.github/workflows/build-and-test.yaml
```diff
@@ -514,6 +518,29 @@ jobs:
                 network_exposure: protected
                 user_interaction: gardener-operator
 
+      - name: Build OCI Image "job-image-cfg-mgmt" (using ocm-oci-build-action)
```
8R0WNI3, line 521: The actual difference to the usual "job-image" (containing of `cfg_mgmt`-package) is still missing, isn't it?

.github/workflows/build-and-test.yaml
```diff
@@ -514,6 +518,29 @@ jobs:
...
+        uses: ./.github/actions/ocm-oci-build
+        id: image-build
```
8R0WNI3, line 523: I guess it might cause trouble to have the same id as the step above.

.ci/build_python_packages
zkdev: it is not required for GHA; thought this was kept intentionally, e.g. local builds / debugging

ccwienk: @zkdev : hmmm...... I think this is rather a left-over. might consider adjusting it and deduplicate w/ gh-actions

requirements.txt
```diff
@@ -21,18 +17,9 @@ ensure
...
 github3.py<5.0.0
-google-api-core<3
-google-api-python-client<3
-google-auth
-google-cloud-storage<3
-google-crc32c>1.1.4
 html2text
 jsonschema
 jsonpath-ng
-kubernetes>=26.0.0
```
8R0WNI3, line 32: Just a reminder: We'll still need to adjust the deployments which require the kubernetes client to be available in the job-image.

setup.cfg_mgmt.py
```diff
@@ -0,0 +1,40 @@
...
+setuptools.setup(
+    name='gardener-cicd-cfg-mgmt',
+    version=version(),
+    description='Gardener CI/CD Config Management',
+    long_description='Gardener CI/CD Config Management',
+    long_description_content_type='text/markdown',
+    python_requires='>=3.12',
+    packages=[
+        'cfg_mgmt',
+        'concourse.steps',
```
8R0WNI3, line 33: Why is `concourse.steps` required here? 🤔

ccwienk: very valid point! any code from concourse.steps should *never* be used outside concourse-template (I know there are some deviations from that - however, we should use this opportunity to get rid of those)

requirements.txt
ccwienk: as in: those should also use the (new) image?

setup.cfg_mgmt.py
```diff
@@ -0,0 +1,40 @@
...
+    ],
+    package_data={
+        'concourse.steps':['*.mako'],
```
ccwienk, line 36: ditto: why do we need concourse-template? this should already be contained in `gardener-cicd-libs`, which should be a dependency of the new package
-> drop

ccwienk (changes requested):
see inline comments

requirements.txt
zkdev: Yes, k8s deployment pipelines need to use the new image.

setup.cfg_mgmt.py
zkdev: see https://github.com/gardener/cc-utils/pull/1145#discussion_r1971184416

setup.cfg_mgmt.py
zkdev: as discussed:
do not include `concourse.steps` package in cfg-mgmt package and add corresponding error handling to cicd-libs.

8R0WNI3 (approved):
/lgtm
However, please remove https://github.com/gardener/cc-utils/pull/1145/commits/8377c7e5639cef54ed39cf631e22d841eed20fed from this PR (again) prior to merging.

8R0WNI3 (approved):

PR #1146, opened 2025-02-03
GitHubActions - fixes and cleanups

ccwienk:
**Release note**:

```other developer
- install gardener-gha-libs from sources (rather than pypi)
- fix for base-component-descriptor-workflow (parameter-checking)
```

### DIFF
--- a/.github/actions/install-gardener-gha-libs/action.yaml
+++ b/.github/actions/install-gardener-gha-libs/action.yaml
@@ -2,6 +2,9 @@ name: Install Gardener-GHA-Libs
 description: |
   Install Python3 and gardener-gha-libs
 
+  As GitHub-Actions in https://github.com/gardener/cc-utils still sees frequent development,
+  install from sources (checking out head of `master`), rather than from PYPI.
+
 runs:
   using: composite
   steps:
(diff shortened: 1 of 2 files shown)


.github/actions/install-gardener-gha-libs/action.yaml
```diff
@@ -2,20 +2,49 @@ name: Install Gardener-GHA-Libs
...
         python3 --version
-        if ! pip install gardener-gha-libs > /tmp/pip-install.log; then
-          echo 'An error occured while tring to run pip install:'
-          cat /tmp/pip-install.log
-          exit 1
+
+        pip install setuptools >/dev/null
+
+        version=$(.ci/read-version)
+        if [ -n "$(echo "${version}" | cut -d- -f2)" ]; then
```
TuanAnh17N, line 29: can we simplify this? maybe something like this:

```bash
if [[ "$version" == *-* ]]; then
  version="${version%%-*}-dev0"
  echo "$version" | .ci/write-version
fi
```

ccwienk: considering this pullrequest will also switch to bash: sure

PR #1147, opened 2025-02-04
add cfg-option for post-import-operation

ccwienk:
Specifically for use-case of importing "bump-commit", (hard-coded) rebase after import of commit will not work (unless special-case, where release-commit and bump-commit can be created in sequence). Make import-action more flexible to support this usecase.

**Release note**:

```other developer
make import-commit action more flexible (allow configuring post-import action)
```


8R0WNI3 (approved):
/lgtm

PR #1148, opened 2025-02-05
add release and version gha-actions

ccwienk:
**Release note**:

```other developer
add github-action `version`, as successor to `version` trait (concourse)
```

### DIFF
--- a/.github/actions/version/action.yaml
+++ b/.github/actions/version/action.yaml
@@ -1,0 +1,168 @@
+name: Version
+description: |
+  performs version-operations for release- and non-release pipelines. This action assumes version
+  is kept track of within repository (either in a regular versionfile, or using callbacks). It also
+  assumes versions adhere to "relaxed semver" (relaxed in that an optional v-prefix is allowed, and
+  `patch` may be omitted (e.g. v1.2)).
+
+  The action defines some defaults w.r.t versionfile/callback-lookup such to reduce amount of
+  explicit configuration for convenience.
+
+  Specifically, the following files are honoured (all paths relative to repository-root):
+    - VERSION # textfile containing version
+    - .ci/read-version # executable; callback (version read from stdout)
+    - .ci/write-version # executable; callback (version written to stdin)
+
+  If exactly one of `VERSION`, or `.ci/read-version`, `.ci/write-version` is present, those are
+  used (without the need to set inputs accordingly). If both are present, explicit inputs are
+  required.
+
+  The action will read the current version (repository-version), and calculate the "effective
+  version" from it, by performing the specified `version-operation`.
+
+  The action offers different options of how to convey created diff from  writing effective version,
+  controlled via the `repository-operation`-input:
+
+  - leave-dirty: do not create a commit, leave diff in repository
+  - commit-to-head: create a commit, update head (regular result of git commit)
+  - capture-commit: create a commit, but capture using `capture-commit` action (no change to head)
+
+inputs:
+  versionfile:
+    required: false
+    type: string
+  read-callback:
+    required: false
+    type: string
+  write-callback:
+    required: false
+    type: string
+  version:
+    required: false
+    type: string
+    description: |
+      pass-in version for more control, if needed. versionfile/callbacks will not be used for
+      reading (however, they will be used for writing). In conjunction w/ setting version-operation
+      to `noop`, this allows for full control over effective-version.
+  commit-message:
+    required: false
+    type: string
+    description: |
+      commit-message-template to use if the action should create a commit.
+      currently only supported template-var: ${version} (bash-syntax).
+    default: Set version to ${version}
+  version-operation:
+    required: true
+    type: choice
+    default: commit-digest-as-prerelease
+    description: |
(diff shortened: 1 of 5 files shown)


.github/actions/version/action.yaml
```diff
@@ -0,0 +1,162 @@
...
+      - bump-patch: increment patch-version by 1
+
+      Note to `commit-digest-as-prelease`: The commit-digest that will be used is the one on which
+      pipeline-run is based (not, e.g. the commit-digest of version-change-commit)
+    options:
+      - noop
+      - finalise
+      - commit-digest-as-prerelease
+      - bump-minor
+      - bump-patch
```
8R0WNI3, line 73: Maybe only for the future, but possible options should also be `bump-major` as well as `timestamp-as-prerelease`, right?

.github/actions/version/action.yaml
```diff
@@ -0,0 +1,162 @@
...
+    required: false
+    type: string
+    description: |
+      commit-message-template to use if the action should create a commit.
+      currently only supported template-var: ${version} (bash-syntax).
+    default: Set version to ${version}
+  version-operation:
+    required: true
+    type: choice
+    default: commit-digest
```
8R0WNI3, line 57: nit: I suppose this is a development leftover and should adhere to `commit-digest-as-prerelease`?

.github/actions/version/action.yaml
```diff
@@ -0,0 +1,162 @@
...
+          --version="${{ inputs.version }}" \
+          --operation="${{ inputs.version-operation }}" \
+          --extra-version-outfile /tmp/effective-version
+
+        if [ "${{ inputs.repository-operation }}" == leave-dirty ]; then
+            restore_stash
+            exit 0
+        fi
+
+        export version="$(cat /tmp/effective-version)"
```
8R0WNI3, line 145: Why storing as environment variable?

.github/actions/version/action.yaml
```diff
@@ -0,0 +1,162 @@
...
+        touch /tmp/timestamp-ref
+
+        "${GITHUB_ACTION_PATH}/version_cli.py" \
+          --versionfile="${{ inputs.versionfile }}" \
+          --read-callback="${{ inputs.read-callback }}" \
+          --write-callback="${{ inputs.write-callback }}" \
+          --root-dir="${PWD}" \
+          --version="${{ inputs.version }}" \
+          --operation="${{ inputs.version-operation }}" \
+          --extra-version-outfile /tmp/effective-version
```
8R0WNI3, line 138: I think passing of the commit digest is still missing 🤔

.github/actions/version/version_cli.py
```diff
@@ -0,0 +1,276 @@
...
+    versionfile = parsed.versionfile
+    read_callback = parsed.read_callback
+    write_callback = parsed.write_callback
+
+    if versionfile and (read_callback or write_callback):
+        logger.error('either --versionfile, or both --read-callback, --write-callback must be set')
+        exit(1)
+
+    if bool(read_callback) ^ bool(write_callback):
+        logger.error('--read-callback and --write-callback must both be set')
```
8R0WNI3, line 134: Shouldn't we also exit with an error in that case?

.github/actions/version/version_cli.py
```diff
@@ -0,0 +1,276 @@
...
+def process_version(
+    version: str,
+    operation: VersionOperation,
+    commit_digest: str,
+):
+    if operation is VersionOperation.NOOP:
+        return version
+    parsed_version = version_mod.parse_to_semver(version)
+    if operation is VersionOperation.FINALISE:
+        return semver.finalize_version(str(parsed_version))
```
8R0WNI3, line 192: nit: You might use `parsed_version.finalize_version()` instead since you already parsed it to semver (ditto for other operations below).

.github/workflows/build-and-test.yaml
```diff
@@ -77,9 +73,6 @@ jobs:
           setuptools_version=${effective_version}
           # hardcode to bumping "minor" for now
           next_minor=$(( $minor + 1 ))
-          next_version="${major}.${next_minor}.${patch}-dev"
```
8R0WNI3, line 84: nit: You might also drop
```
# hardcode to bumping "minor" for now
next_minor=$(( $minor + 1 ))
```

.github/actions/version/action.yaml
ccwienk: good catch, and: yes

.github/actions/version/action.yaml
ccwienk: I wanted to make it slightly more obvious it will be made accessible to templating of commit-message. as this is not technically necessary, will reduce to just defining a variable

.github/actions/version/action.yaml
ccwienk: dang! again: good catch!
done

.github/actions/version/action.yaml
ccwienk: done

PR #1149, opened 2025-02-06
Fix install gha action

ccwienk:
**Release note**:

```bugfix developer
install-gardener-gha-libs action will no longer leave copy of cc-utils in workdir
```


8R0WNI3 (approved):
/lgtm

PR #1151, opened 2025-02-07
Gha fork handling

ccwienk:
**Release note**:

```other developer
check for current repository being a fork. if so, always publish to fork-specific ghcr.io-oci-repositories, rather then gardener's google-artifact-registry to avoid interference.
```

### DIFF
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -63,6 +63,10 @@ jobs:
             oci_repository=${snapshots_repo}
           fi
 
+          echo "ocm_releases_repository=${releases_repo}"
+          echo "ocm_repository=${ocm_repository}"
+          echo "oci_repository=${oci_repository}"
+
           echo "ocm_releases_repository=${releases_repo}" >> "${GITHUB_OUTPUT}"
           echo "ocm_repository=${ocm_repository}" >> "${GITHUB_OUTPUT}"
           echo "oci_repository=${oci_repository}" >> "${GITHUB_OUTPUT}"
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -31,7 +31,27 @@ jobs:
       - name: params
         id: params
         run: |
-          repo_base=europe-docker.pkg.dev/gardener-project
+          set -eu
+          set -o pipefail
+
+          is_fork=$(
+            curl -qL -H 'Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}' \
+            https://api.github.com/repos/${{ github.repository }} \
+            | jq .fork
+          )
+          echo "repo is a fork: ${is_fork}"
+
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-test.yaml
```diff
@@ -31,9 +31,22 @@ jobs:
...
+            curl -qL -H 'Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}' \
+            https://api.github.com/repos/${{ github.repository }} \
+            | jq .fork
+          )
+          echo "repo is a fork: ${is_fork}"
           repo_base=europe-docker.pkg.dev/gardener-project
-          snapshots_repo="${repo_base}/snapshots"
-          releases_repo="${repo_base}/releases"
+
+          if [ $is_fork ]; then
```
TuanAnh17N, line 42: fork status is checked using `if [ $is_fork ]; then`. In bash, any non-empty string (including false) evaluates as `true`. since the GitHub API returns the fork field as a boolean, `jq .fork` extracts a `true` or `false` value—but when used in bash, even `false` is non-empty. would it be safer to explicitly checkif is_fork is `"true"`, like this?

```bash
if [ "$is_fork" = "true" ]; then
  ...
fi
```
simple test:

```bash
is_fork=false
if [ $is_fork ]; then
  echo "condition is true"
else
  echo "condition is false"
fi

condition is true
```

.github/workflows/build-and-test.yaml
```diff
@@ -31,9 +31,22 @@ jobs:
...
+            # always use snapshots-repository for forks to avoid publishing releases from forks.
+            snapshots_repo="${repo_base}/snapshots"
```
TuanAnh17N, line 44: `snapshots_repo` is assigned the same value in both if and else. to reduce redundancy, you could move it out of the if statement. smth like this:
```bash
snapshots_repo="${repo_base}/snapshots"
if [ "$is_fork" = "true" ]; then
  releases_repo="${repo_base}/snapshots"
else
  releases_repo="${repo_base}/releases"
fi
```

.github/workflows/build-and-test.yaml
ccwienk: @TuanAnh17N : good catch.

.github/workflows/build-and-test.yaml
ccwienk: yeah, that's a matter of taste. I kept it redundantly, as that felt easier to read to me, w/ acceptable redundancy.

as @8R0WNI3 pointed out out-of-bands, I think we should change this anyhow (I suppose to publishing to `ghcr.io/${{ github.repository }}`), as our existing lookups will also honour release-versions from snapshot-repositories.

8R0WNI3 (approved):
/lgtm
We should adjust the release note to recent changes though.

ccwienk:
> /lgtm We should adjust the release note to recent changes though.

good point. will do before merging

PR #1152, opened 2025-02-10
use version-action in build-and-test-workflow

ccwienk:
pass-in params for version-action into workflow; create second workflow (in parallel to `release.yaml`) to define default-values specific for pipeline-flavour (to handle differences between release and non-release-runs whilst avoiding redundancies).

**Release note**:

```other developer
- define separate workflow for triggering non-release runs
- re-use version-action some more
```

### DIFF
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -7,6 +7,24 @@ on:
         required: false
         type: boolean
         default: false
+      version-operation:
+        required: true
+        type: string
+        description: |
+          version-operation to pass to `version` action (cannot validate allowed values here)
+          one of:
+          - noop
+          - finalise
+          - commit-digest-as-prerelease
+          - timestamp-as-prerelease
+          - bump-major
+          - bump-minor
+          - bump-patch
+      version-commit-message:
+        required: false
+        type: string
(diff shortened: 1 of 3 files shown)


.github/workflows/build-and-test.yaml
```diff
@@ -7,6 +7,23 @@ on:
         required: false
         type: boolean
         default: false
+      version-operation:
+        required: true
+        type: string
+        description: |
+          version-operation to pass to `version` action (cannot validate allowed values here)
```
8R0WNI3, line 14: Why is `type: choice` not possible here?

.github/workflows/build-and-test.yaml
```diff
@@ -7,6 +7,23 @@ on:
...
+          one of:
+          - noop
+          - finalise
+          - commit-digest-as-prerelease
```
8R0WNI3, line 18: What about `timestamp-as-prerelease`?

ccwienk: will add. actually, I was a bit uncertain, whether it might be better to not duplicate documentation, but reference instead :-/

8R0WNI3 (approved):
/lgtm

PR #1153, opened 2025-02-10
Reduce K8S python client version

mliepold:
In order to unblock the dev pipeline, currently affected by
https://github.com/kubernetes-client/python/issues/2344

**Release note**:

```other operator
Restrict K8S python client version to <32.0.0
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -29,7 +29,7 @@ google-crc32c>1.1.4
 html2text
 jsonschema
 jsonpath-ng
-kubernetes>=26.0.0
+kubernetes<32.0.0,>=26.0.0
 msal<2
 openstacksdk
 oss2


ccwienk (approved):

ccwienk:
@mliepold : if this helps short-term, we can downgrade. however, as discussed out-of-bands, I think we should rather drop this dependency from cc-utils / gardener-cicd-libs, and re-declare it, as needed, in downstream tools-image (mind the immensely heavy turnaround currently in-place through this indirect consumption)

zkdev:
@ccwienk how do we proceed here?

ccwienk:
@zkdev : I already approved the pullrequest (I think pinning the version should be fine for as). However, I think we typically leave it to the PullRequest-Authors to merge their own pullrequests (*bump* @mliepold ).

however, I think this change will not have to be relased in a timely manner, as there is already a downstream workaround (force-downgrade in other image-build in place).

mid-term, this change will become obsolete, once we drop kubernetes-package from gardener-cicd-libs's dependencies.

mliepold:
Closing this PR, because the issue has been mitigated in the downstream image.

/close

PR #1154, opened 2025-02-11
Remove ruleset version from diki finding key

AleksandarSavchev:
This PR removes the ruleset version from `DikiFinding`'s key property. The ruleset version is not needed for an identifier since all findings of a single scan are from the same ruleset version.

This change will cause a recreating of current GitHub issues and will prevent future recreations caused by using a newer ruleset version.

**Release note**:

```breaking operator
Remove the ruleset version from `DikiFinding`'s key property.
```

### DIFF
--- a/dso/model.py
+++ b/dso/model.py
@@ -418,7 +418,7 @@ class DikiFinding(Finding):
 
     @property
     def key(self) -> str:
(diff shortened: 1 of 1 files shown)


8R0WNI3 (approved):
/lgtm

PR #1157, opened 2025-02-15
fix: prevent duplicate pr creation

TuanAnh17N:
**Release note**:

```other operator
Prevent duplicate PR creation when a pull request already exists.
```

### DIFF
--- a/.github/workflows/update-dockerfile.yaml
+++ b/.github/workflows/update-dockerfile.yaml
@@ -9,6 +9,8 @@ on:
 jobs:
   update-dockerfile:
     runs-on: ubuntu-latest
+    env:
+      BRANCH: update-dockerfile
     permissions:
       contents: write
       pull-requests: write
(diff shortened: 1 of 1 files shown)


.github/workflows/update-dockerfile.yaml
```diff
@@ -76,12 +76,24 @@ jobs:
...
         run: |
-          gh pr create \
-            -B master \
-            -H update-dockerfile \
-            --title "Update Dockerfile with latest Docker and Buildx versions" \
-            --body "Updates the Dockerfile to use the latest versions:
-              - Docker: ${{ steps.get_docker_version.outputs.docker_version }}
-              - Buildx: ${{ steps.get_buildx_version.outputs.buildx_version }}"
+          # Check if a pull request for the branch "update-dockerfile" already exists
+          pr_number=$(gh pr list --head update-dockerfile --state open --json number --jq '.[].number')
```
ccwienk, line 80: it might be nice to pull-up branch-name to a shared variable (I think doing this via and env-var would be the idiomatic way to do that in github-actions)
(not strictly related to the scope of this pullrequest, admittedly)

ccwienk (approved):

ccwienk:
btw @TuanAnh17N : considering e.g. https://github.com/gardener/cc-utils/pull/1156, I think it would be reasonable to also filter for final versions (we do not typically want to upgrade to release-candidates - unless special cases maybe, where we urgently want to consume a new feature)

ccwienk (approved):

PR #1159, opened 2025-02-19
introduce rescoringclass for SAST

TuanAnh17N:
**Release note**:

```other operator
NONE
```

### DIFF
--- a/dso/model.py
+++ b/dso/model.py
@@ -403,6 +403,16 @@ class SastFinding(Finding):
 
 
 @dataclasses.dataclass(frozen=True)
+class RescoreSastFinding:
+    sast_status: SastStatus
+    sub_type: SastSubType
+
+    @property
+    def key(self) -> str:
+        return _as_key(self.sast_status, self.sub_type)
+
+
+@dataclasses.dataclass(frozen=True)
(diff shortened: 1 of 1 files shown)


dso/model.py
```diff
@@ -457,7 +467,7 @@ class CustomRescoring:
         RescoringVulnerabilityFinding
         | RescoringLicenseFinding
         | MalwareFindingDetails
-        | SastFinding
+        | RescoreSastFinding
```
ccwienk, line 470: is it correct to _replace_ instead of add?

8R0WNI3: I think this will also require us to drop the `severity` property from the automatically created rescorings, will it not?

8R0WNI3 (approved):
/lgtm

dso/model.py
ccwienk: @8R0WNI3 : so what's the resolution? is replacing the correct way to go + is not dropping severity also okay?

8R0WNI3: Yes, the replacement here is intentional and @TuanAnh17N already removed the `severity` from the automatically created rescorings in the delivery-service.

PR #1162, opened 2025-02-20
Extent compliance diff script

heldkat:
**Release note**:

```other operator

Include a new feature  in the compliance Diff script that is required for the ComponentTest reporting

### DIFF
--- a/cli/gardener_ci/compliance.py
+++ b/cli/gardener_ci/compliance.py
@@ -48,13 +48,13 @@ class DiffArguments:
     right_name: str
     left_version: str
     right_version: str
+    outfile_prefix: str
     ocm_repo_urls: list[str]
     exclude_component_names: list[str] = None
     exclude_component_resource_names: list[ComponentResourceNames] = None
     resource_types: list[str] = None
     name_template: str = None
     name_template_expr: str = None
-    outfile_prefix: str = 'resource-diff'
 
 
 def diff(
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: cli/gardener_ci/compliance.py [hunk 1]
```diff
@@ -227,12 +229,14 @@ def resource_as_dict(component, resource, resource_id):
         else:
             name = resource.name
 
-        return {
+        name2 = f"{component.name.split('/')[-1]}_{component.version}"
```
ccwienk, line 232: can you choose a more expressive name that makes it easier to understand what `name2` should hold?

also, could you elaborate (preferably in commit-message) what is the general idea of what you are trying to achieve?

The way I understand this diff, you want to extend the `resource_as_dict`-function such that will return both the dict on the resource it returned earlier, and in addition also return a shortened part of the component-name; where you consume the latter in a second place in the code.

Could you elaborate for which purpose you collect (shortened) component-names?
also, it would appear to me to be easier to either include component-names in resource-dict, or to define a dedicated datastructure to return (as opposed to returning a two-tuple); the name-mangling I assume should rather be done on the consuming-side than inside the function yielding the resource-dicts

also, it should be noted that the way you collect component-name-versions, you will end up w/ duplicate component-name-versions (depending on inspectected components, of course) - is this intended? Also, by discarding all but the last path-segment of component-names, you might run into ambiguities - what's the rationale behind shortening component-names?

lastly, the conventional form of serialising component-versions reads `<name>:<version>`, as opposed to you using an underscore (`_`) character. while this is of course up to you, I also would like to ask whether there is a specific reason to choose this as output format (if there is: fine w/ me - if there is not, I would suggest to keep to conventional format).

ccwienk (changes requested):

PR #1163, opened 2025-02-21
Bump commit prerelease

ccwienk:
**Release note**:

```bugfix developer
set prerelease (again) for bumped versions (github-actions)
```

### DIFF
--- a/.github/actions/version/action.yaml
+++ b/.github/actions/version/action.yaml
@@ -54,13 +54,11 @@ inputs:
   version-operation:
     required: true
     type: choice
-    default: commit-digest-as-prerelease
+    default: set-prerelease
     description: |
       how to calculate effective version. Possible values:
       - noop: keep version as-is (useful in conjunction w/ passing-in `version`)
-      - finalise: make version final (remove build/prerelease-suffix)
-      - commit-digest-as-prerelease: set commit-digest as prerelease-suffix
-      - timestamp-as-prerelease: set unix-epoch-timestamp as prerelease-suffix
+      - set-prerelease: replace / set version's prerelease (e.g. 1.2.3-${prerelease})
       - bump-major: increment major-version by 1
       - bump-minor: increment minor-version by 1
       - bump-patch: increment patch-version by 1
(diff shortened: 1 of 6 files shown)


ccwienk:
action fails, because this change introduces an incompatible change. -> safe to ignore

.github/actions/version/action.yaml
```diff
@@ -142,6 +150,7 @@ runs:
           --root-dir="${PWD}" \
           --version="${{ inputs.version }}" \
           --operation="${{ inputs.version-operation }}" \
+          --bump-prerelease="${{ inputs.version-bump-prerelease }}" \
```
8R0WNI3, line 153: You'll also have to add this new parameter to the argument parser.

.github/actions/version/action.yaml
```diff
@@ -75,6 +75,14 @@ inputs:
       - bump-major
       - bump-minor
       - bump-patch
+  version-bump-prerelease:
```
8R0WNI3, line 78: Hmm, what do you think about generalising this a little bit more: Allowing to specify an arbitrary prerelease here (also allow selection of "meta-prereleases" `commit` and `timestamp`), that would make this less of a special single purpose parameter. Also, it would allow us to drop `commit-digest-as-prerelease` and `timestamp-as-prerelease` operations as those are basically `finalise` operation + the respective "meta-prerelease".

.github/actions/version/action.yaml
ccwienk: 🤦🏻 thx

.github/actions/version/action.yaml
ccwienk: you mean we should offer to pass `prerelease` either as a str-literal, or as special-value (e.g. `${timestamp}`)?
We might do that - albeit at the cost of being more low-level, i.e:

```
"bump-minor with dev-prerelase" would mean:
with:
  version-operation: bump-minor
  prerelease: dev

whereas finalise would mean:
with:
  version-operation: noop
  prerelease: ''

whereas "commit-digest-as-prerelease" would become:
with:
  version-operation: noop
  prerelease: ${{ github.sha }} # pass-in from github-ctx
OR
 prerelease: "${commit_digest}" # version-action would resolve this to current head-commit

whereas "timestamp-as-prerelease" would become:
with:
 version-operation: noop
 prerelease: "${timestamp}" # passed in from our action
OR:
 prerelease: "$(date +%s)" # let user specify a command we will evaluate
```

considering that, we would probably set `version-operation` to default to `noop`.

My rationale behind the suggested API was to keep the API less complex (although the introduction of the new parameter already conflicts w/ this target) :-/

8R0WNI3 (approved):
/lgtm

PR #1165, opened 2025-02-24
Add model classes for ODG crypto extension

8R0WNI3:
**Release note**:

```other operator

```


dso/model.py
```diff
@@ -431,6 +441,134 @@ def key(self) -> str:
         return _as_key(self.provider_id, self.ruleset_id, self.rule_id)
 
 
+class AssetTypes(enum.StrEnum):
```
ccwienk, line 444: should we prefix those new enums w/ `Crypto` or `CBOM`?

PR #1166, opened 2025-02-24
dso: add models for findings from gardener/inventory

dnaeon:
This PR adds models to support the integration between OCM Gear and [gardener/inventory
](https://github.com/gardener/inventory)

**Release note**:

```other operator
Add models for gardener/inventory findings
```

### DIFF
--- a/dso/model.py
+++ b/dso/model.py
@@ -67,9 +67,10 @@ class Datasource:
     DELIVERY_DASHBOARD = 'delivery-dashboard'
     DIKI = 'diki'
     FALCO = 'falco'
+    INVENTORY = 'inventory'
 
     @staticmethod
-    def datasource_to_datatypes(datasource: str) -> tuple[str]:
+    def datasource_to_datatypes(datasource: str) -> tuple[str, ...]:
         return {
             Datasource.ARTEFACT_ENUMERATOR: (
                 Datatype.COMPLIANCE_SNAPSHOTS,
@@ -112,6 +113,9 @@ class Datasource:
             ),
             Datasource.FALCO: (
                 Datatype.FALCO_FINDING,
+            ),
+            Datasource.INVENTORY: (
+                Datatype.INVENTORY_FINDING,
             ),
         }[datasource]
 
@@ -279,6 +283,7 @@ class Datatype:
     CRYPTO_ASSET = 'crypto_asset'
     CRYPTO = 'finding/crypto'
     FALCO_FINDING = 'finding/falco'
+    INVENTORY_FINDING = 'finding/inventory'
 
     @staticmethod
     def datatype_to_datasource(datatype: str) -> str:
@@ -294,6 +299,7 @@ class Datatype:
             Datatype.CRYPTO_ASSET: Datasource.CRYPTO,
             Datatype.CRYPTO: Datasource.CRYPTO,
             Datatype.FALCO_FINDING: Datasource.FALCO,
+            Datatype.INVENTORY_FINDING: Datasource.INVENTORY,
         }[datatype]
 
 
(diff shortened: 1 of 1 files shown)


dso/model.py
```diff
@@ -431,6 +438,29 @@ def key(self) -> str:
         return _as_key(self.provider_id, self.ruleset_id, self.rule_id)
 
 
+@dataclasses.dataclass(frozen=True)
```
ccwienk, line 441: nit: I suppose you added `frozen=True` for consistency w/ the other classes in this module. however, as it has a mutable attr (resources), it cannot truly be frozen (which is why I suggest to drop `frozen=True` param

dso/model.py
```diff
@@ -431,6 +438,29 @@ def key(self) -> str:
...
+
+    # Short summary of the finding
+    summary: str
+
+    # Resources identified by gardener/inventory as being orphan
+    resources: list[dict]
+
+    @property
+    def key(self) -> str:
+        return _as_key(self.provider_name, self.resource_kind)
```
zkdev, line 461: The `key` will be necessary to construct stable IDs for findings.
I am not familiar with gardener/inventory, however can you confirm that it is intended to only track one finding per e.g. `gcp-vm`?

From a gut feeling, I would have expected that we want to track each resource individually, e.g. `gcp-vm-testvm1`.

This will be crucial if we want to create github issues for orphaned resources.

dnaeon: Hey @zkdev ,

An example for orphan GCP Virtual Machines with the current key set to `provider_name` and `resource_kind` would look like this (simplified):

``` json
{
    "provider_name": "gcp",
    "resource_kind": "virtual_machine",
    "summary": "Orphan GCP Instances",
    "resources": [
        {
            "name": "vm-01",
            "project": "project-1",
        },
        {
            "name": "vm-02",
            "project": "project-2"
        }
    ]
}
```

> This will be crucial if we want to create github issues for orphaned resources.

Correct, we want to have Github issues created for orphan resources.

In the example above only a single Github issue should be created, and the issue details would list two virtual machines as being orphan.

Let me know if you have other questions, thanks!

ccwienk: I think it should work like this. However, note, that if we do not create individual entries for each orphaned resource, we will have no means to track those individually. In case of vulnerabilities (where we also started by grouping - in this case by oci-image - we had the issue that tracking-issues were never closed in some cases (because of constant ingress of new findings).

That being said, I think it might be preferable to _not_ group at this level, but do the grouping in delivery-gear-extension.

dnaeon:
> Hi @dnaeon, as discussed out-of-bands, please rather move the attributes which are used to identify the resources to the `artefact: ComponentArtefactId` attribute, for example something like:
> 
> ```yaml
> artefact:
>   component_name: <name-of-the-landscape-component>
>   component_version: ~
>   artefact_kind: runtime
>   artefact:
>     artefact_name: <hyperscaler-resource-uuid>
>     artefact_version: ~
>     artefact_type: gcp/virtual_machine
>     artefact_extra_id:
>       name: vm-01
>       project: project-1
> meta:
>   type: finding/orphan-resource
>   datasource: inventory
> data:
>   severity: <severity>
>   summary: Orphan GCP Instances
> discovery_date: <yyyy-mm-dd>
> ```
> 
> The criteria, which of those attributes should be used for grouping in the GitHub issues, will then be configured specifically for this finding type.

Thanks, @8R0WNI3. Will get back to this one soon. I'll move it to a draft state until then.

ccwienk:
@dnaeon : I see this pullrequest is stalled for ~2w now - are you waiting for additional input from our end?

dnaeon:
Hey @ccwienk ,

> I see this pullrequest is stalled for ~2w now - are you waiting for additional input from our end?

No need for any input from your side, I just need to get to this one, but have been busy with other stuff, that's why I moved it to draft.

Will be coming back to this one soon enough, and then I'll rebase it.

ccwienk (approved):

PR #1173, opened 2025-02-27
Rm ocm retention policy

zkdev:
Google Artifact Registry has a built-in mechanism.

**Release note**:

```breaking developer
OCM clean-up logic (based on version-policies) is removed (we use oci-registry built-ins instead).
```

### DIFF
--- a/cnudie/purge.py
+++ b/cnudie/purge.py
@@ -1,4 +1,3 @@
-import collections.abc
 import logging
 import traceback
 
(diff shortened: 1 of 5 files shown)


ccwienk (changes requested):
removing this seems like a good idea. however, there are quite a few pipelines that configured this explicitly. I suggest to remove existing configurations prior to dropping this codepath

zkdev:
@ccwienk 
I kept the attribute spec exactly due to this, so we can clean up afterwards.

ccwienk:
@zkdev : I understood that. Still, I do not think we should just silently change pipeline-contract w/o rolling this out explicitly (there might be pipelines that actually rely on this cleanup-code).

PR #1174, opened 2025-02-27
add oci-ocm-multiarch workflow

ccwienk:
**Release note**:

```feature developer
add reusable workflow `oci-ocm-multiarch` that will allow building multi-arch OCI Images, running the build-jobs on runners w/ matching platform, if available.
```

### DIFF
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -352,7 +352,7 @@ jobs:
             --file component-descriptor.yaml
 
           echo "adding image-resource"
-          echo "${{ needs.images.outputs.ocm_resources }} " | \
+          echo "${{ needs.images.outputs.ocm-resource }} " | \
           python3 -m ocm append resource \
             --file component-descriptor.yaml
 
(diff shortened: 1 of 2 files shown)


ccwienk:
there are still a couple of open TODOs (which I plan to add to this pullrequest):
- [ ] re-use in cc-utils + concourse-repository
- [ ] upload ocm-resource-fragments as artefacts (to support more than one image - thanks GHA for making such simple things so cumbersome)
- [ ] add prebuild-hook (required by both cc-utils, and concourse-repository)
- [ ] maybe: add option to import commit
- [ ] maybe (probably separate pullrequest): export/merge-ocm-fragment-actions

.github/workflows/oci-ocm-multiarch.yaml
```diff
@@ -0,0 +1,300 @@
...
+  workflow_call:
+    inputs:
+      oci-registry:
+        description: |
+          The (optional) oci-registry as part of image-push-target.
+          If both oci-registry, and oci-repository are present, push-target is calculated as:
+          {oci-registry}/{oci-repository}
+          If only one of both is passed, the missing parameter will be ignored.
+        type: string
+        required: true
```
8R0WNI3, line 19: The description states this property is optional, but this flag says differently (ditto for `oci-repository` input).

.github/workflows/oci-ocm-multiarch.yaml
```diff
@@ -0,0 +1,300 @@
...
+        type: string
+        default: 'linux/arm64,linux/amd64'
+      tag:
+        description: |
+          The image-tag to set. Occurrences of `{version}` will be replaced by `version` input
+          (which is the default).
+        type: string
+        default: '{version}'
+        required: false
+      target:
```
8R0WNI3, line 41: nit: Add description

.github/workflows/oci-ocm-multiarch.yaml
```diff
@@ -0,0 +1,300 @@
...
+          args = []
+
+          for platform_name in platform_names:
+            if not platform_name in runners_by_platform:
+              print(f'WARNING: no runner for {platform_name=}')
+            platform_tag = f'${{ inputs.tag }}-{platform_name.replace("/", "_")}'
+            args.append({
+              'platform-name': platform_name,
+              'runner': runners_by_platform.get(platform_name, 'ubuntu-latest'),
+              'tag': f'${{ inputs.tag }}-{platform_name.replace("/", "_")}',
```
8R0WNI3, line 155: nit: Reuse already defined `platform_tag`

.github/workflows/oci-ocm-multiarch.yaml
ccwienk: done

.github/workflows/oci-ocm-multiarch.yaml
ccwienk: you are correct, of course. However, I wonder: _should_ we actually make both inputs optional?

My rationale behind splitting this into two is to follow common practise we see within gardener-org (probably not only there), where registry and repository are often split. this will also work well w/ our approach of using different registries (snapshots vs releases).

If we keep those inputs (both) as mandatory, GHA will do a (much) earlier validation of wrong usage (we can fail only later, at runtime, otherwise). GHA unfortunately does not allow specification of complex conditions ("at least one of both must be given, and non-empty").

So I would currently tend (different than orginally stated in documentation) to make both inputs mandatory. WDYT?

.github/workflows/oci-ocm-multiarch.yaml
ccwienk: yet again: good catch - done

.github/workflows/build-and-test.yaml
```diff
@@ -467,65 +470,40 @@ jobs:
...
-                authentication_enforced: true
-                availability_requirement: low
-                confidentiality_requirement: high
-                integrity_requirement: high
-                network_exposure: protected
-                user_interaction: gardener-operator
+    uses: ./.github/workflows/oci-ocm-multiarch.yaml
+    with:
+      name: job-image
+      oci-registry: ${{ needs.params.outputs.oci_repository }}
```
8R0WNI3, line 485: Hmm, can we consistently use `oci-registry` != `oci-repository`? I don't like that `oci_repository` has to be passed here to `oci-registry` although there is another input called `oci-repository` which expects `cicd/job-image`...

.github/workflows/build-and-test.yaml
```diff
@@ -467,65 +470,40 @@ jobs:
...
+      oci-repository: cicd/job-image
+      oci-platforms: 'linux/arm64,linux/amd64'
```
8R0WNI3, line 487: Shouldn't this be read from `needs.params.outputs.oci_platforms`?

.github/workflows/build-and-test.yaml
```diff
@@ -467,65 +470,40 @@ jobs:
...
 
   images:
-    name: Build OCI Images
+    name: Build Job-Image
     needs:
       - version
-      - package
+      - package # needed indirectly (distribution-packages-artefact)
       - params
-    outputs:
```
8R0WNI3, line 475: Does this jobnot have to define any outputs anymore? Like `oci-reference` or`ocm-resource`?

.github/workflows/build-and-test.yaml
ccwienk: yep, I also did not like this whilst creating this diff (but did not yet add a refactoring-commit). will do so

.github/workflows/build-and-test.yaml
ccwienk: If I am not mistaken, outputs do not need to be "re-declared" in case of a job that calls a reusable workflow

ccwienk:
note to self: remember to autosquash prior to merging

8R0WNI3 (approved):
/lgtm

PR #1176, opened 2025-02-27
add release-notes as inline-resource to component-descriptors

ccwienk:
Add release-notes as inline-resource (i.e. store as inline-blob) to component-descriptors published upon releasing. This will make it easier to retrieve them from scripts and tooling (such as the OpenDelivery-Gear).

**Release note**:

```other operator
add release-notes as inline-resource to component-descriptors
```


8R0WNI3 (approved):
/lgtm

PR #1177, opened 2025-02-27
Drop `correlation_id` from snapshots and add utility function

8R0WNI3:
**Release note**:

```other developer
Dropped the `correlation_id` property from compliance snapshots (ODG)
```


ccwienk (approved):
you _do_ like ternary operator :-)

PR #1178, opened 2025-02-28
also attach release-notes for release-GHA

ccwienk:
As done for `release`-trait in concourse-pipeline, also attach release-notes to component-descriptor as inlined resource as part of `release`-GitHub-Action.

**Release note**:

```other developer
release-GitHub-Action will now attach release-notes as inline artefact into published OCM-Component-Descriptors
```

### DIFF
--- a/.github/actions/release/action.yaml
+++ b/.github/actions/release/action.yaml
@@ -69,7 +69,7 @@ inputs:
     description: |
       the release-notes to publish as body for GitHub-Release (release-notes action might be
       handy to collect those)
-    required: true
+    required: false
   github-token:
     description: |
       the github-auth-token to use for authenticating against GitHub.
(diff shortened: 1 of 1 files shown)


.github/actions/release/action.yaml
```diff
@@ -98,18 +98,76 @@ runs:
       shell: bash
       run: |
         # pass to next steps
-        echo "${{ inputs.component-descriptor }}" > /tmp/component-descriptor
```
8R0WNI3, line 101: Why is this dropped? It is tried to consume the component version from the tmp file in the next line 🤔

.github/actions/release/action.yaml
```diff
@@ -98,18 +98,76 @@ runs:
...
         version=$(yq .component.version /tmp/component-descriptor)
         echo "version=${version}" >> $GITHUB_OUTPUT
         tag_name="${version}"
         tag_ref="refs/tags/${tag_name}"
         echo "tag-ref=${tag_ref}" >> $GITHUB_OUTPUT
         echo "tag-name=${tag_name}" >> $GITHUB_OUTPUT
 
+    - name: attach-release-notes-to-component-descriptor
+      if: ${{ inputs.release-notes }}
```
8R0WNI3, line 109: nit: The `release-notes` input is marked as "required", so there should be no case where it is unset, no?

.github/actions/release/action.yaml
ccwienk: oh. dang it. I suppose it will make sense to merge this step into the next one, then (I wanted to deduplicate dereferentiation of `component-descriptor`-input

.github/actions/release/action.yaml
ccwienk: I am not sure (need to check). I _think_, marking an input as `required` means that one hast to explicitly pass a value for it. However, the passed value might be "falsy" (empty string, null, `false`). In this case, it makes sense to skip this step.

However, even with that being said, it might also make sense to demote release-notes to an optional input

ccwienk:
merging, as all review-comments are addressed

PR #1180, opened 2025-03-05
Add ruleset and rule names to DikiFinding

AleksandarSavchev:
**Release note**:

```other operator
NONE
```

### DIFF
--- a/dso/model.py
+++ b/dso/model.py
@@ -422,8 +422,10 @@ class DikiCheck:
 class DikiFinding(Finding):
     provider_id: str
     ruleset_id: str
+    ruleset_name: str
     ruleset_version: str
     rule_id: str
+    rule_name: str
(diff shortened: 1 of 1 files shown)


dso/model.py
```diff
@@ -422,8 +422,10 @@ class DikiCheck:
 class DikiFinding(Finding):
     provider_id: str
     ruleset_id: str
+    ruleset_name: str
```
8R0WNI3, line 425: As already existing Diki findings do not contain these two attributes (yet), I think we must declare them as optional `str | None` first, until all Diki findings have been adjusted to contain these attributes as well. Otherwise, deserialisation into this model class will fail.

8R0WNI3 (changes requested):

8R0WNI3 (approved):
/lgtm

PR #1182, opened 2025-03-06
Fix ocm fragments export action quoting + documentation

ccwienk:
**Release note**:

```bugfix developer
- improve quoting for export-ocm-fragments-action for yaml-documents passed as inputs
- correction to documented artefact-name-template
```

### DIFF
--- a/.github/actions/export-ocm-fragments/action.yaml
+++ b/.github/actions/export-ocm-fragments/action.yaml
@@ -15,11 +15,11 @@ description: |
 
   The aforementioned directory-tree will be TARed into a file named
   [<ctx>-]<hexdigest>.ocm-artefacts.tar.gz and uploaded into a artefact named
-  [<ctx>-]<hexdigest>-ocm-artefacts
+  [<ctx>-]<hexdigest>.ocm-artefacts
 
   Therefore, it is possible to use `download-artifact`'s globbing to retrieve _all_ artefacts:
-  `*-ocm-artefacts` # regardless of ctx, or if ctx is not set
-  `<ctx>-*-ocm-artefacts` # specific for given ctx
+  `*.ocm-artefacts` # regardless of ctx, or if ctx is not set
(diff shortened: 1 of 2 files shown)


.github/actions/export-ocm-fragments/action.yaml
```diff
@@ -106,15 +112,21 @@ runs:
...
+          ocm_resources = f.read()
+
+        with open('/tmp/ocm-sources') as f:
+          ocm_sources = f.read()
+
         resources = list(
           export_ocm.iter_artefacts(
-            artefacts_yaml='''${{ inputs.ocm-resources }}''',
+            artefacts_yaml=ocm_resources,
             artefacts_file='${{ inputs.ocm-resources-file }}',
```
8R0WNI3, line 124: Since `ocm_resources` (same for `ocm_sources` below) are now stored in a file anyways, might it make sense to just pass the file reference here so that `export_ocm.iter_artefacts` does not have to differentiate between passed-in yaml-strs or files?

ccwienk: very valid point!

.github/actions/export-ocm-fragments/action.yaml
```diff
@@ -106,16 +117,34 @@ runs:
...
 
+        with open('/tmp/ocm-resources') as f:
+          ocm_resources = f.read()
+
+        with open('/tmp/ocm-sources') as f:
+          ocm_sources = f.read()
+
+        resources_files = [
+          f for f in
+          '/tmp/ocm-resources',
```
8R0WNI3, line 128: I think this is invalid syntax, no? 🤔

ccwienk: you mean because of missing parantheses or brackets? yes....

8R0WNI3 (approved):
/lgtm

PR #1185, opened 2025-03-07
Improve release action

ccwienk:
**Release note**:

```bugfix developer
- do not publish component-descriptor prior to release-job (cc-utils pipeline)
- improve robustness of loading YAML documents passed as input
- authenticate against target oci-registry prior to publishing ocm-component-descriptor from release-job
```


8R0WNI3 (approved):
/lgtm

PR #1187, opened 2025-03-10
Switch to ocm export and import actions

ccwienk:
**Release note**:

```other developer
Use `export-ocm-fragments` and `merge-ocm-fragments` actions to collect effective component-descriptor from multiple jobs from build- and test-pipeline. Add some functionality (mostly convenience-related) along the way.
```


.github/actions/export-ocm-fragments/action.yaml
```diff
@@ -87,7 +94,7 @@ runs:
   using: composite
   steps:
     - name: install-gardener-gha-libs
-      uses: gardener/cc-utils/.github/actions/install-gardener-gha-libs@master
+      uses: ccwienk/cc-utils/.github/actions/install-gardener-gha-libs@master
```
ccwienk, line 97: note to self: switch back to `gardener/cc-utils`

.github/actions/export-ocm-fragments/action.yaml
```diff
@@ -161,34 +170,82 @@ runs:
...
               continue
 
+            fhash = hashlib.sha256()
+            with open(fpath, 'rb') as f:
+              while chunk := f.read(4096):
+                fhash.update(chunk)
+
+            # symlink, unless file already is named according to its content-hash-digest
+            fhashname = f'sha256:{fhash.hexdigest()}'
+            if fhash != fhashname:
```
zkdev, line 196: I am wondering why this check is necessary.
How can these values become equal if `fhashname` is derived from `fhash`?

.github/actions/export-ocm-fragments/action.yaml
```diff
@@ -161,34 +170,82 @@ runs:
...
+          if local_ref.startswith('sha256:'):
+            continue
+          if not os.path.exists(local_fpath := os.path.join(blobs_dir, local_ref)):
+            continue
+          # local file exists, and it does not meet content-addresssing scheme (
+          # <alg>:<hexdigest>). We should find it from previous step where we calculated
+          # digests, and created symlinks.
+          hashname = orig_name_to_digest_names.get(local_ref, None)
+          if not hashname:
+            print(f'WARNING: did not find local blobfile for {artefact=}')
```
zkdev, line 226: Shouldn't we rather raise?

.github/actions/export-ocm-fragments/action.yaml
ccwienk: good catch. I actually wanted to compare fname w/ fhashname!
as to your question: this check I added, as the caller (of this action) might already name the blobfile accordingly (in which case no symlinking is required (or possible..))

.github/actions/export-ocm-fragments/action.yaml
ccwienk: I think no. Rationale: this action is intended to be fairly low-level. Callers might decide to already upload blobs into target-oci-artefact (where ocm-component-descriptor will be uploaded). As an alternative, missing files might also be added at a later point-in-time (theoretically).

Hence, I did on purpose not raise here. We might consider adding a "strict" mode, though (where we would consider this an error).

.github/actions/release/action.yaml
```diff
@@ -19,6 +19,11 @@ inputs:
       Effective Component-Descriptor to publish. Any local-blobs are assumed to already be
       present in target-OCI-Repository.
     required: true
+  component-descriptor-blobs-dir:
+    description: |
+      Directory containing blobfiles referenced as local-blobs in component-descriptor
+      (name-scheme: <digets>:<hexdigest>).
```
8R0WNI3, line 25: typo: `digets` -> `digests`. However, shouldn't this name `<algorithm>:<hexdigest>`?

.github/actions/export-ocm-fragments/action.yaml
```diff
@@ -161,34 +170,82 @@ runs:
...
+          access['size'] = os.stat(local_fpath).st_size
+          print(f'INFO: patched {artefact["name"]}\'s access to {hashname=} (was: {local_ref=})')
+
+          if not 'type' in artefact:
+            artefact['type'] = magic.detect_from_filename(
+              os.path.join(blobs_dir, local_ref),
+            ).mime_type
+
+        ocm_artefacts_yaml_bytes = yaml.safe_dump(ocm_artefacts).encode('utf-8')
+        ocm_artefacts_digest = hashlib.sha1(ocm_artefacts_yaml_bytes).hexdigest()
```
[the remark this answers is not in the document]
ccwienk: yeah, we might. as mentioned in the documentation, the digest calculated for artefact- and filenames, are, however, intended to be opaque, and they serve a different purpose (as compared to how we use sha256 for content-addressing blobs in OCI-registries).

namely, I wanted to ensure that multiple artefacts created in parallel result in different file- and artefact-names, without the user having to ensure this. consider, e.g. a case where multiple resources are emitted, that have the same name (but differ in either version, or extra-identity). We might of course use a mangled version of full identity. However, I wanted to keep things simple (in particular, we do not need this process to be reversible).

I chose sha1 mainly because I reckoned it might be nicer for users inspecting those contents to have shorter names (also for sha1, there is virtually no risk of collisions, considering we will only ever likely see low two-digit amounts of artefacts / files, methinks). Thinking about it some more, using `uuid4` might also be a good option (although using a content-digest will have the theoretical advantage of deduplicating identical artefacts, should they be emitted).

so wdyt? I understand your concern w.r.t. consistency. however, my main reason for sha1 is shorted names (+ different purpose). WDYT? apart from shorter names (so a bit of convenience for debugging / shortening log-outputs), I have not strong preference for sha1

8R0WNI3 (approved):
/lgtm

PR #1188, opened 2025-03-12
Declare OCM dependency towards bdba-client

zkdev:
**Release note**:

```other dependency
Declare OCM dependency towards "ocm.software/ocm-gear/bdba-client"
```
```feature developer
component-references can now be declared via definition file in repository
```

### DIFF
--- a/.github/workflows/base-component-descriptor.yaml
+++ b/.github/workflows/base-component-descriptor.yaml
@@ -26,6 +26,24 @@ on:
         required: false
         type: string
         default: SAP SE
+      component-references-file:
+        required: false
+        type: string
+        description: |
+          relative path (to repo root) to YAML-form file containing definition of
+          component-references added to base-component-descriptor, defaults to
+          `.ocm/component-references.yaml`.
+          action will only fail on absence of file if an explicit value is given.
+
+          Example contents of '.ocm/component-references.yaml'
+          ```
+          - componentName: github.com/gardener/gardener
+            version: v1.110.0
+            name: gardener
+          ```
+          for full specification, see:
+          https://github.com/open-component-model/ocm-spec/blob/main/doc/01-model/02-elements-toplevel.md#references
+
       labels:
         required: false
         type: string
@@ -33,18 +51,23 @@ on:
           Labels to set for the component in YAML-form (caveat: need to quote). May either be a
           single object, or an array.
           Example:
+
             # single label
+            ```
             name: cloud.gardener.cnudie/responsibles
             value:
               - type: githubTeam
                 teamname: gardener/maintainers
                 github_hostname: github.com
+            ```
 
             # list of labels
+            ```
             - name: label1
               value: value1
             - name: label2
               value: value2
+            ```
       src-labels:
         required: false
         type: string
@@ -137,6 +160,27 @@ jobs:
             commit: ${commit}
             ref: ${{ github.ref }}
           EOF
+
+          fail_if_absent=false
+          component_references_path='.ocm/component-references.yaml'
+
+          if [ -n "${{ inputs.component-references-file }}" ]; then
+            fail_if_absent=true
+            component_references_path="${{ inputs.component-references-file }}"
+          fi
+
+          if [ -f "$component_references_path" ]; then
+            echo "adding component-references from ${component_references_path}"
+            cat $component_references_path | python3 -m ocm append component-reference \
+              --file component-descriptor.yaml
+
+          else
+            if [ "$fail_if_absent" = true ]; then
+              echo "error: ${{ inputs.component-references-file }} not found"
+              exit 1
+            fi
+
+          fi
 
           echo "Component-Descriptor:"
           cat component-descriptor.yaml
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -197,6 +197,7 @@ jobs:
           value: |
             we use bandit (linter) for SAST scans
             see: https://bandit.readthedocs.io/en/latest/
+      component-references-file: '.ocm/component-references.yaml'
 
   update-draft-relase:
     if: ${{ inputs.mode != 'release' &&
--- a/ocm/__main__.py
+++ b/ocm/__main__.py
@@ -6,6 +6,8 @@ import json
 import os
 import sys
 
+import dacite
+
 import cnudie.retrieve
 import ocm
 import ocm.upload
@@ -124,41 +126,37 @@ def append(parsed):
 
     if parsed.type in ('r', 'resource'):
         attr = component['resources']
+        ocm_cls = ocm.Resource
+
     elif parsed.type in ('s', 'source'):
         attr = component['sources']
+        ocm_cls = ocm.Source
+
     elif parsed.type in ('c', 'component-reference'):
         attr = component['componentReferences']
+        ocm_cls = ocm.ComponentReference
 
     if _have_yaml:
         obj = yaml.safe_load(sys.stdin)
     else:
         obj = json.load(sys.stdin)
 
-    if not (labels := parsed.labels):
-        labels = []
-    else:
-        labels = [dataclasses.asdict(l) for l in _iter_parsed_labels(labels=labels)]
+    obj = obj if isinstance(obj, list) else [obj]
 
-    def inject_labels(artefact: dict, labels):
-        if not 'labels' in artefact:
-            artefact['labels'] = labels
-        else:
-            artefact['labels'].extend(labels)
-
-    if isinstance(obj, list):
-        for o in obj:
-            inject_labels(
-                artefact=o,
-                labels=labels,
-            )
-
-        attr.extend(obj)
-    else:
-        inject_labels(
-            artefact=obj,
-            labels=labels,
+    artefacts = [
+        dacite.from_dict(
+            data_class=ocm_cls,
+            data=o
         )
-        attr.append(obj)
+        for o in obj
+    ]
+
+    labels = list(_iter_parsed_labels(labels=labels)) if (labels := parsed.labels) else []
+    for artefact in artefacts:
+        artefact.labels = list(artefact.labels) + labels
+
+    attr: list
+    attr.extend([dataclasses.asdict(a) for a in artefacts])
 
     with open(parsed.file, 'w') as f:
         if _have_yaml:
--- a/.ocm/component-references.yaml
+++ b/.ocm/component-references.yaml
@@ -1,0 +1,3 @@
+- componentName: ocm.software/ocm-gear/bdba-client
+  version: 0.2.0
+  name: bdba


ocm/__main__.py
```diff
@@ -247,9 +249,16 @@ def main():
...
         aliases=('a',),
-        help='appends resources or sources to component-descriptor',
+        help='appends resources, sources, or component-references to component-descriptor',
+    )
+    add_parser.add_argument(
+        'type',
+        choices=(
+            'r', 'resource',
+            's', 'source',
+            'cref', 'component-reference',
```
ccwienk, line 259: `c` is still free - preferably stick to single-letter short-forms for the sake of consistency.

ccwienk (commented):
in case of component-references (which have a fixed attr-set), we might also expose a dedicated command, like so:

```
ocm add component-ref \
--name <name> \ # might default to component-name's last part
--component-name <cname> \ # might optionally contain `:<version>`-suffix for convenience
--component-version <version> \
[--label ...] # see existing labels-handling

```

.github/workflows/build-and-test.yaml
```diff
@@ -359,6 +359,20 @@ jobs:
...
+          cat << EOF > bdba.ocm-component-reference
+          name: bdba
+          componentName: ocm.software/ocm-gear/bdba-client
+          version: 0.2.0
+          digest: null
+          extraIdentity: {}
+          labels: []
+          EOF
+
+          cat bdba.ocm-component-reference | \
```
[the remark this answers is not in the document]
zkdev: I have no strong opinion here, however we should keep it consistent with other `ocm append` calls (they are using the cat approach)

zkdev:
@ccwienk `ocm add component-ref` added, not sure about invoking in base component-descriptor workflow (will this workflow be re-used?)

ocm/__main__.py
```diff
@@ -234,6 +236,37 @@ def upload(parsed):
     )
 
 
+def add_cref(parsed):
+    raw = _parse_yaml_or_json(parsed.file)
+    component = raw['component']
+    component_references: list = component['componentReferences']
```
ccwienk, line 242: you should not assume `componentReferences` already exists.

-> check if absent, and initialise as empty list

ocm/__main__.py
```diff
@@ -234,6 +236,37 @@ def upload(parsed):
...
+
+    name = parsed.name or parsed.cname.split('/')[-1]
+    version = parsed.version
+    if not version:
+        parts = parsed.cname.split(':')
+        if len(parts) != 2:
+            print('error: unable to extract `:<version>`-suffix from component-name')
```
ccwienk, line 249: nit: I think it is better to hint caller about the expected input, like so:
"expected `<component-name>:<version>`, but saw {actual-input}`

ocm/__main__.py
```diff
@@ -234,6 +236,37 @@ def upload(parsed):
...
 
+def add_cref(parsed):
+    raw = _parse_yaml_or_json(parsed.file)
+    component = raw['component']
+    component_references: list = component['componentReferences']
+
+    name = parsed.name or parsed.cname.split('/')[-1]
+    version = parsed.version
+    if not version:
+        parts = parsed.cname.split(':')
```
ccwienk, line 247: why not split like so:
```
name, version = parsed.cname.split(':')
```

ocm/__main__.py
```diff
@@ -255,20 +288,64 @@ def main():
...
+    )
+
+    add_cref_parser.add_argument('--file', '-f', required=True)
+    add_cref_parser.add_argument(
+        '--name', '-n',
+        help='name of the component-reference, defaults to last part of component-name',
+    )
+    add_cref_parser.add_argument(
+        '--component-name', '-c',
+        dest='cname',
```
ccwienk, line 309: preferably do not abbreviate. we commonly use argparse's default-naming-scheme (with some exceptions, e.g. pluralising if action is set to `append`);

ocm/__main__.py
```diff
@@ -255,20 +288,64 @@ def main():
...
+        required=True,
+        help='might optionally contain `:<version>`-suffix',
+    )
+    add_cref_parser.add_argument(
+        '--component-version', '-v',
+        dest='version',
+    )
+    add_cref_parser.add_argument(
```
ccwienk, line 317: can't we merge this into the `append`-subcommand?

.github/workflows/base-component-descriptor.yaml
```diff
@@ -138,6 +138,12 @@ jobs:
             ref: ${{ github.ref }}
           EOF
 
+          echo "adding bdba component-reference"
+          python3 -m ocm add component-ref \
```
ccwienk, line 142: as we discussed this last week (out-of-bands): This is _not_ a good place to add repository-specific customations.

-> base-component-descriptor-workflow is intended to be re-usable. We _might_ add an input to pass-in component-references (which would make it more convenient for us to do pass-in via stdin, btw), or create a `.ci/component-descriptor`-callback (which base-component-descriptor will need to honour, of course).

ccwienk (changes requested):

.github/workflows/build-and-test.yaml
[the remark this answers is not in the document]
ccwienk: Input redirection might have the benefit of avoiding to spawn a subprocess (for cat).

That being said, and knowing bash does feature input-redirection, at least for me, it is friendlier (from code-reading-perspective) to put the input before the command into which it is passed 😅 

Also, consistency is a good argument (so if we want to use a different style, we should do so consistently.

ocm/__main__.py
zkdev: well, you asked for a "dedicated command" and gave an example with `ocm add component-ref`

we could also incorporate into `append` command

.ci/component_descriptor-callback
```diff
@@ -0,0 +1,50 @@
...
+
+
+def main(parsed):
+    raw = _parse_yaml_or_json(parsed.file)
+    component_references = raw['component'].get('componentReferences', [])
+
+    component_references.append(dataclasses.asdict(ocm.ComponentReference(
+        name='bdba-client',
+        componentName='ocm.software/ocm-gear/bdba-client',
+        version='0.2.0',
```
ccwienk, line 34: this is not a nice place to maintain dependency-versions. I suggest to put this into a dedicated file (either a plain text file containing just the version, or a yaml file, as we already have been doing for the past years in other repositories..)

Also, I think it would be preferable to use a shellscript - this should mean a lot less boilerplate.

I think it should be preferable to keep contract towards component-descriptor-callback-scripts as small as possible. While it is obviously not a strong hurdle for us to offer python + gha-libs, writing such callbacks like this will make it a lot more difficult to locally run the script, esp. considering other repositories.

ccwienk (commented):
can you already bring the first commit (the one adding the new subcommand for adding crefs) to master? I think that one is already fine

.github/workflows/base-component-descriptor.yaml
```diff
@@ -26,6 +26,23 @@ on:
         required: false
         type: string
         default: SAP SE
+      component-references:
+        required: false
+        type: string
+        description: |
+          relative path (to repo root) to YAML/Json file containing definition of
```
ccwienk, line 33: two nits:
- JSON-documents are _always_ also YAML-documents (so there is no actually added information by mentioning JSON explicitly)
- JSON is an acronym (JavaScript Object Notation), so by convention it should also be set in uppercase

.github/workflows/base-component-descriptor.yaml
```diff
@@ -26,6 +26,23 @@ on:
...
+          component-references added to base-component-descriptor, defaults to
+          `.ocm/component-references.yaml`.
+          action will only raise on absence of file if an explicit value is given.
```
ccwienk, line 36: nit: I think the term `raise` is common only in Python - in other contexts, I think it is more adequate to use terms like `fail` or `error` (also, there is technically no means to raise an exception from a GitHub-Action..)

.github/workflows/base-component-descriptor.yaml
```diff
@@ -26,6 +26,23 @@ on:
...
+
+          Example contents:
+          - componentName: github.com/gardener/gardener
+            version: v1.110.0
+            name: gardener/gardener
```
ccwienk, line 41: I think it is a great idea to add an example, however consider:
- putting it as a code-block (\`\`\`...\`\`\`)
- maybe add a comment indicating the filepath (to make the example even easier to understand/follow)
- `/` character is not allowed in cref-names as of OCM-spec (I think)

.github/workflows/base-component-descriptor.yaml
```diff
@@ -26,6 +26,23 @@ on:
...
+            # following key-values pairs are optional
+            digest: foo
```
ccwienk, line 43: at least `foo` is not a valid digest - I would maybe rather drop it (also to make the example shorter), and rather put a link to OCM-Spec (about crefs) for further reading.

.github/workflows/base-component-descriptor.yaml
```diff
@@ -26,6 +26,23 @@ on:
...
+            extraIdentity:
+              foo: bar
```
ccwienk, line 45: likewise - while this extra-identity is syntactically correct, I think it will not help too much. Consider especially that within the entire gardener-bom we have not a single occurrance of a cref w/ an extra-id (also, I doubt there will very likely be any such usage _ever_ - but that's a different story). So, also to keep things simple, and make the example less confusing, I would suggest to drop it (people should _understand_ that they can add additional attributes into a YAML file :-)

.github/workflows/base-component-descriptor.yaml
```diff
@@ -26,6 +26,23 @@ on:
         required: false
         type: string
         default: SAP SE
+      component-references:
```
ccwienk, line 29: naming: should we rather call it component-references-path (or: -file)? I know: will lead to more typing efforts + if checking a concrete example (with a value), it should be obvious it is a filepath, so it is also fine w/ me to not add this suffix.

OTOH, we _might_ see cases / the need to pass-in crefs (from another action... - likely not as a value from calling the workflow). So adding a `-file` (or `-path`) suffix will leave us this name open

.github/workflows/base-component-descriptor.yaml
```diff
@@ -138,6 +155,27 @@ jobs:
             ref: ${{ github.ref }}
           EOF
 
+          if [ -n "${{ inputs.component-references }}" ]; then
```
ccwienk, line 158: I think we might deduplicate the "catting", like so:

- set `fail_if_absent` (or similar) if input is given
- check existence (fail if absent and fail_if_absent (`exit 1`), else `exit 0

Admittedly, at least for the exit0-part, this would need to be put into a function (or guarded w/ another if), as we will still need to keep going.

However, this way, we could combine catting w/ appending (and thus avoid having to call append w/ empty list - which I think is a but clumsy)

ocm/__main__.py
```diff
@@ -122,17 +122,20 @@ def append(parsed):
...
         attr = component['sources']
     elif parsed.type in ('c', 'component-reference'):
         attr = component['componentReferences']
-
-    if _have_yaml:
-        obj = yaml.safe_load(sys.stdin)
-    else:
-        obj = json.load(sys.stdin)
+        for o in obj if isinstance(obj, list) else [obj]:
+            o.setdefault('digest', None)
```
ccwienk, line 137: oh dear - what have I started. Guess we should switch to using `ocm`'s model-classes 🙈 
background: i initially wanted to keep dependencies to an absolute minimum. then I figured it is a pattern to have an action installing some libs (install-gardener-gha-libs). Ever since then, there is no longer the strong need to avoid using code from cc-utils - also I think I overshot a bit 🙈 

also: do we _have_ to initialse those attrs? Probably, we might also allow absent attrs / define default-nones in ocm/__init__.py, wdyt?

ccwienk (changes requested):

ocm/__main__.py
zkdev: agree w.r.t. using `ocm` model classes
didn't consider it as I was not aware of the rational 😅 

---

ocm/init provides sane default values already 🙂

ccwienk (approved):

PR #1189, opened 2025-03-17
cleanup: drop obsolete backwards-compatibility cruft

ccwienk:
- cc-utils-versions is accessible either via
 - meta-step / job-image-tag
 - pip show gardener-cicd-libs # and friends
- HELM_V3_VERSION is outdated - nobody seems to care
- HELM_ARCH - no known usages
- placing helm into /cc/utils/bin: nobody should hardcode such paths
- helm3: let's see..

**Release note**:

```other developer
drop unused backwards-compatibility-cruft from cc-job-image:
- cc-utils-versions is accessible either via
 - meta-step / job-image-tag
 - pip show gardener-cicd-libs # and friends
- HELM_V3_VERSION is outdated - nobody seems to care
- HELM_ARCH - no known usages
- placing helm into /cc/utils/bin: nobody should hardcode such paths
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,24 +18,12 @@ FROM $BASE_IMAGE
 
 ARG TARGETARCH
 
-COPY --from=builder /pkgs/usr /usr
 COPY --from=ocm-cli /bin/ocm /bin/ocm
-COPY --from=builder /cc/utils/bin/component-cli /bin/component-cli
+
+COPY --from=builder /pkgs/usr /usr
+COPY --from=builder /cc/utils/bin /cc/utils/bin
 COPY --from=builder /usr/lib/libmagic.so.1 /usr/lib/libmagic.so.1
 COPY --from=builder /usr/lib/libmagic.so.1.0.0 /usr/lib/libmagic.so.1.0.0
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -28,14 +28,3 @@ COPY --from=builder /usr/share/misc/magic.mgc /usr/share/misc/magic.mgc
...
 ENV PATH=$PATH:/cc/utils/bin
-
-# place version file into container's filesystem to make it easier to
-# determine the image version during runtime
-COPY VERSION /metadata/VERSION
-
-ENV HELM_V3_VERSION=v3.12.2
-ENV HELM_ARCH="${TARGETARCH}"
-# copy to where helm is expected
-COPY --from=builder /cc/utils/bin/helm /usr/local/bin/helm
```
8R0WNI3, line 39: (Why) don't we require helm in the resulting image anymore?

ccwienk: hmm, valid ask. We do still need helm (at least for helmchart-handling in `publish`-trait). I forgot this is actually just a shellscript that will retrieve helm.

However, as `/cc/utils/bin` is added to PATH, for consistency reasons, I think we should keep the entire `bin`-directory. WDYT?

8R0WNI3 (approved):
/lgtm

PR #1194, opened 2025-03-19
Improve export ocm fragments action

ccwienk:
**Release note**:

```other developer
- signal unresolveable local blobfilerefs as error
- allow including blobs-dir in local blobfilerefs
```


8R0WNI3 (approved):
/lgtm

PR #1196, opened 2025-03-20
add helmchart-action (with OCM-support)

ccwienk:
**Release note**:

```other developer
add helmchart-action (with OCM-support)
```

### DIFF
--- a/.github/actions/helmchart/action.yaml
+++ b/.github/actions/helmchart/action.yaml
@@ -1,0 +1,241 @@
+name: helmchart
+description: |
+  Packages a Helm-Chart in a localisable fashion, publishing the helmchart as an OCI-Artefact,
+  a mapping (for localisation) as an OCM-side-artefact, and an OCM-Resource-Fragment. The latter
+  two are published using the `export-ocm-fragments`-action.
+
+  Localisation means that metadata in the form of mappings between the Chart's values, and
+  OCM-Resources (typically OCI-Images) are declared (and exposed) in machine-readable form, such
+  that it is possible to generate a full `values.yaml` from a (localised) OCM-Component-Descriptor
+  (containing a reference to the localised helmchart), so that all external references of the
+  chart are explicitly defined (in other words: the helmchart can be used w/o any of its hardcoded
+  default-references).
+
+inputs:
+  name:
+    required: true
+    type: string
+    description: |
+      The Helm-Chart's name. This value will be injected into the Helm-Chart's `Chart.yaml` file
+      as `.name` attribute (if not identical). It will also be used as OCM-Resource-Name, and be
+      appended to the target-OCI-repository (the latter is mandated by Helm).
+
+  dir:
+    required: true
+    type: string
+    description: |
+      Relative path to Chart-Directory (directory containing `Chart.yaml`).
+
+  oci-registry:
+    required: true
+    type: string
+    description: |
+      The OCI-Registry to publish the Helm-Chart to.
+
+  oci-repository:
+    required: false
+    type: string
+    description: |
+      If passed, this value is concatenated to `oci-registry`-input. Note that also in this case,
+      the `name` input will be appended to the end of the target-oci-reference.
+
+  component-descriptor:
+    required: true
+    type: string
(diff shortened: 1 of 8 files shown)


ccwienk:
note to reviewers: I am aware this is largely copy-paste from concourse/steps/helmchart.mako.
Considering we will (eventually) phase-out concourse, I think it is not a good investment to try to deduplicate at this point; also, the API we implement against (both uploading helmcharts and ocm-mappings) is defined externally, w/o foreseeable changes.

.github/actions/helmchart/action.yaml
```diff
@@ -0,0 +1,245 @@
...
+            yaml.safe_load(f),
+          )
+        component = component_descriptor.component
+
+        version = component.version
+        oci_registry = '${{ inputs.oci-registry }}'
+        oci_repository = '${{ inputs.oci-repository }}'
+        oci_target_ref = oci_registry
+        if oci_repository:
+          oci_target_ref = f'{oci_target_ref.strip("/")}/{oci_repository}'
```
8R0WNI3, line 128: nit: Also strip `oci_repository` here?

.github/actions/helmchart/action.yaml
```diff
@@ -0,0 +1,245 @@
...
+            exit(1)
+
+          mappings = [
+            dacite.from_dict(
+              data=mapping,
+              data_class=helm.HelmchartValueMapping,
+            ) for mapping in mappings
+          ]
+
+        if not isinstance(mappings, list):
```
8R0WNI3, line 145: I think you can drop this check here since it is already checked above if `mappings` is an instance of `list`.

.github/actions/helmchart/action.yaml
```diff
@@ -0,0 +1,245 @@
...
+        )
+
+        helm.patch_values_yaml(
+          component=component,
+          values_yaml_path=(values_yaml_path := os.path.join(dir, 'values.yaml')),
+          mappings=mappings,
+        )
+
+        outdir = 'helmchart.d'
+        os.makedirs(outdir, exist_ok=False)
```
8R0WNI3, line 165: What's the benefit of using `makedirs` instead of just `mkdir` in this case?

.github/actions/helmchart/helm.py
```diff
@@ -0,0 +1,288 @@
...
+        '../../..',
+    ),
+)
+
+# flake8: noqa: E402
+import collections
+import dataclasses
+import datetime
+import json
+import hashlib
```
8R0WNI3, line 19: nit: `hashlib` before `json`

.github/actions/helmchart/helm.py
```diff
@@ -0,0 +1,288 @@
...
+
+import jsonpath_ng
+import yaml
+
+import oci.model
+import oci.client
```
8R0WNI3, line 25: ditto

.github/actions/helmchart/helm.py
```diff
@@ -0,0 +1,288 @@
...
+        return
+
+    print(f'{chart_yaml_path}\'s `name` differs from {name=} - patching')
+    chart['name'] = name
+    with open(chart_yaml_path, 'w') as f:
+        yaml.safe_dump(chart, f)
+
+
+def main():
+    pass
```
8R0WNI3, line 284: What's the purpose of this? 😅

.github/actions/helmchart/action.yaml
ccwienk: good catch

.github/actions/helmchart/action.yaml
ccwienk: there is none. I think I chose makedirs because it was the first that came to my mind (from memory). changed to mkdir

.github/actions/helmchart/helm.py
ccwienk: m(
I think I originally wanted to add a CLI for testing (however, I then decided to write a unittest instead). but I forgot to drop the (useless) main-function. will also rm executable-bit

PR #1199, opened 2025-04-02
Rm elasticsearch usages

zkdev:
**Release note**:

```breaking developer
Elasticsearch usages are purged from job-image and gardener-cicd provided python distribution packages
```


cfg_mgmt/reporting.py
```diff
@@ -86,107 +84,6 @@ def evaluate_cfg_element_status(
     )
 
 
-def cfg_element_statuses_responsible_summaries(
```
ccwienk, line 89: how is removal of elasticsearch related to removal of this piece of code? is this a left-over that we kept for the sole purpose of also uploading to elasticsearch?

whd/dispatcher.py
```diff
@@ -79,11 +76,6 @@ def concourse_clients(
     def dispatch_create_event(
         self,
         create_event,
-        delivery_id: str,
```
ccwienk, line 82: no longer passing elasticsearch-client makes sense, but why also drop the other parameters? were those only used for reporting into ES (in which case, it is of course reasonable to drop)?

question also applies to subsequent similar code-changes

whd/pull_request.py
```diff
@@ -420,41 +399,15 @@ def ensure_pr_resource_updates(
...
-    es_client: ccc.elasticsearch.ElasticSearchClient,
     retries=10,
     sleep_seconds=3,
 ):
     time.sleep(sleep_seconds)
 
     retries -= 1
     if retries < 0:
-        try:
-            log_outdated_resources(
```
ccwienk, line 432: will it make sense to keep logging (to stdout/stderr) for potential debugging?

ccwienk (commented):
see inline comments. lgtm if you can confirm my questions

cfg_mgmt/reporting.py
zkdev: yes, there are some dedicated functions for cfg-status upload towards elasticsearch.
this has **nothing** to do with github compliance issues for cfg elements

whd/dispatcher.py
zkdev: yes, additional parameters have only been used for elasticsearch metrics

whd/pull_request.py
zkdev: I have never used this metric at all, even though it was introduced back in 2019 with https://github.com/gardener/cc-utils/commit/dd4ba15e68 😅 

I don't see value in keeping them.

PR #1203, opened 2025-04-14
Update OCM-spec according to upstream schema

8R0WNI3:
See https://github.com/open-component-model/ocm/blob/main/resources/component-descriptor-v2-schema.yaml for reference.

**Release note**:

```other developer
Updated OCM-spec according to upstream schema
```

### DIFF
--- a/ocm/__init__.py
+++ b/ocm/__init__.py
@@ -211,10 +211,19 @@ class ResourceRelation(enum.StrEnum):
     EXTERNAL = 'external'
 
 
+@dc
+class MergeSpec:
+    algorithm: str | None = None # OCM schema requires /^[a-z][a-z0-9/_-]+$/
+    config: str | int | float | bool | dict | list | None = None
+
+
 @dc(frozen=True)
 class Label:
     name: str
     value: str | int | float | bool | dict | list
+    version: str | None = None # OCM schema requires /^v[0-9]+$/
+    signing: bool = False
+    merge: MergeSpec | None = None
 
 
 _no_default = object()
@@ -257,6 +266,7 @@ class LabelMethodsMixin:
 class NormalisationAlgorithm(enum.StrEnum):
     JSON_NORMALISATION = 'jsonNormalisation/v1'
     OCI_ARTIFACT_DIGEST = 'ociArtifactDigest/v1'
+    GENERIC_BLOB_DIGEST = 'genericBlobDigest/v1'
 
 
 @dc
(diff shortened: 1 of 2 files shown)


ocm/__init__.py
```diff
@@ -299,11 +308,18 @@ class SignatureSpec:
     mediaType: str
 
 
+@dc
+class TimestampSpec:
+    value: str | None = None
+    time: str | None = None
```
zkdev, line 314: add hint w.r.t. allowed format, see [jsonspec](https://github.com/json-schema-org/json-schema-spec/blob/8af8969ca011c9369b8da4d00415cab93b0c0ad9/specs/jsonschema-validation.md?plain=1#L367)

ocm/ocm-component-descriptor-schema.yaml
```diff
@@ -165,6 +193,10 @@ definitions:
         $ref: '#/definitions/digestSpec'
       signature:
         $ref: '#/definitions/signatureSpec'
+      timestamp:
+        oneOf:
+        - type: 'null'
```
zkdev, line 198: is this required to stay compatible with our existing component-descriptors?

8R0WNI3: Actually, no it's not required, as it's already stated by not adding this property to the list of required properties, that it is optional. Will adjust accordingly.

ocm/__init__.py
```diff
@@ -299,11 +309,18 @@ class SignatureSpec:
     mediaType: str
 
 
+@dc
+class TimestampSpec:
+    value: str | None = None
+    time: str | None = None # expected format date-time: %Y-%m-%dT%H:%M:%SZ
```
ccwienk, line 315: might add `RFC3339 rounded to seconds` (taken from ocm-spec)

ocm/__init__.py
```diff
@@ -211,10 +211,19 @@ class ResourceRelation(enum.StrEnum):
     EXTERNAL = 'external'
 
 
+@dc
+class LabelMerge:
```
ccwienk, line 215: nit: would it be more expressive to name this `LabelMergeSpec` (or just MergeSpec) - would also match the term used in [ocm-spec](https://github.com/open-component-model/ocm-spec/blob/main/doc/01-model/03-elements-sub.md#labels)

ccwienk (commented):
some naming/docstr-nits, otherwise lgtm

zkdev (approved):
/lgtm

ocm/__init__.py
```diff
@@ -211,10 +211,19 @@ class ResourceRelation(enum.StrEnum):
...
+    algorithm: str | None = None # OCM schema requires /^[a-z][a-z0-9/_-]+$/
+    config: str | int | float | bool | dict | list | None = None
+
+
 @dc(frozen=True)
 class Label:
     name: str
     value: str | int | float | bool | dict | list
+    version: str | None = None # OCM schema requires /^v[0-9]+$/
```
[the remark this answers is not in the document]
8R0WNI3: Actually, it's even stricter than relaxed SemVer, hence I decided to directly refer to the underlying regex.

ccwienk (approved):

PR #1204, opened 2025-04-14
dso: `normalise_artefact_extra_id` dict may have `any` values

dnaeon:
Some models may have values other than `dict[str, str]`, which specify their extra identity.

This commit ensures that any values passed to `normalise_artefact_extra_id` are normalized as string values before generating the stable identity.

**Release note**:

```other operator
`normalise_artefact_extra_id` normalizes values as strings
```

### DIFF
--- a/dso/model.py
+++ b/dso/model.py
@@ -125,7 +125,7 @@ class Datasource:
 
 
 def normalise_artefact_extra_id(
-    artefact_extra_id: dict[str, str],
+    artefact_extra_id: dict[str, any],
 ) -> str:
     '''
     generate stable representation of `artefact_extra_id`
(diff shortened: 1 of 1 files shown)


dnaeon:
Quick testing shows that the proposed changes to `normalise_artefact_extra_id` work as expected, however when used with the Delivery Service API and trying to delete artefacts the items are not deleted. I suppose, because the Delivery Service DELETE API uses a different logic for normalization, so that needs to be fixed as well.

ccwienk:
@dnaeon : changing allowed values for extra-id conflicts w/ [ocm-spec](https://github.com/open-component-model/ocm-spec/blob/main/doc/01-model/03-elements-sub.md#identifiers), which only allows str:str-mappings

you should rather mv some of those attributes into the `data`-attribute of your extension's artefact-metadata-types (there you are completely free w.r.t. your schema), and include attribute relevant for deriving a unique ID into a `key`-propery (see dso/model.py for examples).

there is no need for artefact-metadata-entries to be unique already by their ocm-coordinates

PR #1206, opened 2025-04-14
add bdba package installation to linter

TuanAnh17N:
**Release note**:

```other operator
NONE
```

### DIFF
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -216,6 +216,7 @@ jobs:
     runs-on: ubuntu-latest
     needs:
       - package
+      - params
     container:
       image: python:3.12-alpine
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/build-and-test.yaml
```diff
@@ -256,6 +257,15 @@ jobs:
           cat /tmp/pip3-install.log
         fi
 
+        echo 'downloading bdba package'
+        bdba_version=$(sed -n '/componentName: ocm.software\/ocm-gear\/bdba-client/{n;s/^[[:space:]]*version:[[:space:]]*//p;}' .ocm/component-references.yaml)
+        ocm_repo='${{ needs.params.outputs.ocm-repository }}'
+        python -m ocm download resource \
+          --ocm-repository '${ocm_repo}' \
+          --component 'ocm.software/ocm-gear/bdba-client:${bdba_version}' \
```
TuanAnh17N, line 265: in this implementation, the component name for the bdba dependency is hardcoded. ideally, we should extract this dynamically from the cc-utils component descriptor. This would prevent inconsistencies if the component name ever changes upstream.

do we already have an action or utility that does this?

ccwienk: @TuanAnh17N : I think hardcoding should be acceptable for now; we do have python-APIs (ocm package + cnudie.retrieve) that could be used for that; actually, we might extract this information from base-component-descriptor, I think.

.github/workflows/build-and-test.yaml
```diff
@@ -256,6 +257,17 @@ jobs:
           cat /tmp/pip3-install.log
         fi
 
+        pkg_dir=dist
+        mkdir -p ${pkg_dir}
+        echo "downloading bdba package"
+        bdba_version=$(sed -n "/componentName: ocm.software\/ocm-gear\/bdba-client/{n;s/^[[:space:]]*version:[[:space:]]*//p;}" .ocm/component-references.yaml)
```
ccwienk, line 263: it is also okay to read this from component-references; I think reading from base-component-descriptor would be slightly better, as this would not rely on this implementation-detail, but I think this is good enough / also a pragmatic choice.

PR #1207, opened 2025-04-15
Add normalisation to OCM model

8R0WNI3:
Adds utilities to create a normalised version of a component descriptor matching the `jsonNormalisation/v1` of the OCM-cli.

**Release note**:

```feature developer
Added implementation of the `jsonNormalisation/v1` algorithm to the OCM model.
```

### DIFF
--- a/ocm/__init__.py
+++ b/ocm/__init__.py
@@ -211,10 +211,19 @@ class ResourceRelation(enum.StrEnum):
     EXTERNAL = 'external'
 
 
+@dc
+class MergeSpec:
+    algorithm: str | None = None # OCM schema requires /^[a-z][a-z0-9/_-]+$/
+    config: str | int | float | bool | dict | list | None = None
+
+
 @dc(frozen=True)
 class Label:
     name: str
     value: str | int | float | bool | dict | list
+    version: str | None = None # OCM schema requires /^v[0-9]+$/
+    signing: bool = False
+    merge: MergeSpec | None = None
 
 
 _no_default = object()
@@ -257,6 +266,7 @@ class LabelMethodsMixin:
 class NormalisationAlgorithm(enum.StrEnum):
     JSON_NORMALISATION = 'jsonNormalisation/v1'
     OCI_ARTIFACT_DIGEST = 'ociArtifactDigest/v1'
+    GENERIC_BLOB_DIGEST = 'genericBlobDigest/v1'
 
 
 @dc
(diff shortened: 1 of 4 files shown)


ocm/__init__.py
```diff
@@ -225,6 +227,16 @@ class Label:
     signing: bool = False
     merge: LabelMerge | None = None
 
+    def normalise(self) -> list[dict]:
```
ccwienk, line 230: codestyle/nit: how about making this a property, and call it `normalised`? Rationale: a method `normalise` might be expected to actually change the underlying object (which it does not).

Edit: I still would vote for calling it `[as]_normalised`, but making it a property will make it inconsistent w/ other, similar, methods where we do need to pass-in parameters

ocm/__init__.py
```diff
@@ -459,6 +471,42 @@ class ComponentReference(Artifact, LabelMethodsMixin):
...
+            ))
+
+            digest = component_descriptor.digest(
+                component_descriptor_lookup=component_descriptor_lookup,
+                image_ref_to_digest_lookup=image_ref_to_digest_lookup,
+            )
+
+            raw['digest'] = dataclasses.asdict(DigestSpec(
+                hashAlgorithm='SHA-256',
+                normalisationAlgorithm=NormalisationAlgorithm.JSON_NORMALISATION,
```
ccwienk, line 499: should we make the normalisation-algorithm a parameter? we could define the (now) only supported value as default-value.
same for hashAlgorithm, I suppose

ccwienk (commented):
Admittedly, the OCM-Spec also calls this process `normalisation` (this term is even reflected in attributes, such as `normalisationAlgorithm`). So this is not in-scope for this pullrequest. Still, I have to say I think normalisation is misleading, considering this process involves dropping information, such as accesses.

Anyhow, I am not too certain it is a good fit to mix normalisation-methods into our dataclasses (as opposed to making them regular functions - which might be reasonably collected in a dedicated module (e.g. ocm/sign or similar).

here is why I think so: thus-far our dataclasses had no involvement into operations like retrieval (we have a separate module for that - cnudie.retrieve); there are functions for (de)serialisation from/to dict, admittedly. Apart from that, those dataclasses were mere bindings to (a subset of) OCM's model-elements.

Those new `normalise`-functions are slightly inconsistent to the, in that they:
- involve functionality of retrieving OCM-Component-Descriptors (admittedly nicely  abstracted by callback-api)
- recursively yield (normalised) raw dicts ("pre-serialised" data, if you will)

To me, adding this normalisation-code brings in the new aspect of serialisation, but mixes it to the aspects covered before that (just declaring the data-structures) - which is inconsistent w/ how we handle other aspects in our codebase (I already mentioned cnudie.retrieve, there is also cnudie.iter, ..). So from consistency-perspective, I would also argue separation into a dedicated module might be better.

wdyt?

PR #1212, opened 2025-04-17
drop obsolete / unused dependencies

ccwienk:
**Release note**:

```other developer
drop some unused dependencies
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,6 @@ GitPython
 Mako<2.0.0
 Sphinx
 awesomeversion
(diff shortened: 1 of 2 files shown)


zkdev (approved):
/lgtm

let's hope there are no downstream usages we are not aware of 😅

ccwienk:
> /lgtm
> 
> let's hope there are no downstream usages we are not aware of 😅

yes, let us hope for the best. + add dependencies where needed again, should we find such a usage.

PR #1213, opened 2025-04-17
Add `verify` flag to OCM component descriptor normalisation

8R0WNI3:
**Release note**:

```other operator
Add `verify` flag to OCM component descriptor normalisation
```

### DIFF
--- a/ocm/sign.py
+++ b/ocm/sign.py
@@ -11,6 +11,10 @@ import json
 import ocm
 
 
+class DigestMismatchException(Exception):
+    pass
+
+
 def normalise_obj(
     obj: dict,
 ) -> list[dict]:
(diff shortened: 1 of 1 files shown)


ocm/sign.py
```diff
@@ -174,17 +218,24 @@ def normalise_component_descriptor(
...
 
+    @param component_descriptor:
+        the component descriptor of which a normalised representation should be created
     @param component_descriptor_lookup:
         lookup for component descriptors by their identity
-    @param image_ref_to_digest_lookup:
-        used to retrieve the digest of an oci-resource by its image reference
+    @param access_to_digest_lookup:
+        used to retrieve the digest of an resource by its access specification
+    @param verify:
```
ccwienk, line 235: nit: there are different kinds of verification (we might, e.g. also check compliance to json-schema..)
-> how about naming this `verify_digests` instead?

ccwienk (approved):

PR #1214, opened 2025-04-17
Use `alpine:edge` for alpine-sap-image

zkdev:
Following edge allows use to consume patches earlier, especially relevant for security related patches which have not been ported back to alpine release branches.

open-delivery-gear is the only known consumer. If this changes in future and following edge is not possible anymore, consider creating dedicated odg-alpine-(base-)image.

**Release note**:

```other operator
NONE
```


8R0WNI3 (approved):
/lgtm

PR #1216, opened 2025-04-22
add status description to osidfinding dataclass

TuanAnh17N:
**Release note**:

```other operator
NONE
```

### DIFF
--- a/dso/model.py
+++ b/dso/model.py
@@ -460,6 +460,24 @@ class OsIdFinding(Finding):
     def key(self) -> str:
         return _as_key(self.osid.ID)
 
+    @property
+    def status_description(self) -> str:
+        if self.os_status is OsStatus.BRANCH_REACHED_EOL:
+            return 'Branch has reached end-of-life'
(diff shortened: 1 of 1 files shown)


dso/model.py
```diff
@@ -460,6 +460,24 @@ class OsIdFinding(Finding):
...
+            return 'Image is more than one patchlevel behind'
+        elif self.os_status is OsStatus.AT_MOST_ONE_PATCHLEVEL_BEHIND:
+            return 'Image is at most one patchlevel behind'
+        elif self.os_status in (
+            OsStatus.EMPTY_OS_ID,
+            OsStatus.NO_BRANCH_INFO,
+            OsStatus.NO_RELEASE_INFO,
+            OsStatus.UNABLE_TO_COMPARE_VERSION,
+        ):
+            return f'No valid OS scan result ({self.os_status.name})'
```
8R0WNI3, line 477: Why did you explicitly decide to use `.name` here (instead of the value which is chosen per default)?

TuanAnh17N: true :(

8R0WNI3 (approved):

PR #1218, opened 2025-04-22
factor-out reusable release workflow

ccwienk:
**Release note**:

```other developer
factor-out reusable release workflow
```


.github/workflows/release.yaml
```diff
@@ -1,110 +1,91 @@
...
+          app-id: ${{ vars.GARDENER_GITHUB_ACTIONS_APP_ID }}
+          private-key: ${{ secrets.GARDENER_GITHUB_ACTIONS_PRIVATE_KEY }}
+      - uses: gardener/cc-utils/.github/actions/release@master
         with:
           component-descriptor: ${{ steps.component-descriptor.outputs.component-descriptor }}
           component-descriptor-blobs-dir: /tmp/ocm/blobs.d
-          release-commit-objects: ${{ needs.build.outputs.version-commit-objects }}
-          release-commit-digest: ${{ needs.build.outputs.version-commit-digest }}
-          release-commit-target: branch
-          version-read-callback: .ci/read-version
```
8R0WNI3, line 53: Why don't we have to pass the version-callback anymore? 🤔

ccwienk: actually, this become obsolete already some weeks ago, as those callbacks specifically have precedence:

from actions/version: :-)
```
The action defines some defaults w.r.t versionfile/callback-lookup such to reduce amount of
explicit configuration for convenience.

Specifically, the following files are honoured (all paths relative to repository-root):
  - VERSION # textfile containing version
  - .ci/read-version # executable; callback (version read from stdout)
  - .ci/write-version # executable; callback (version written to stdin)
```

ccwienk: -> As this release-workflow is supposed to be opinionated, I figured it might be a good start to only add the bare minimum of needed configuration-options (inputs in GHA's terms), and extend those, as needed.

8R0WNI3 (approved):
/lgtm

PR #1219, opened 2025-04-23
Also add `allowed_processing_time` (+ `due_date`) to findings + rescorings model

8R0WNI3:
To not honour already existing findings + rescorings in case existing allowed processing times change, also store the `allowed_processing_time` (+ `due_date`) which is valid upon creation time. This is also a prerequisite to allow rescorings with a manual `due_date` input (e.g. in case of exceptions).

**Release note**:

```other operator

```

### DIFF
--- a/dso/model.py
+++ b/dso/model.py
@@ -695,6 +695,12 @@ class MetaRescoringRules(enum.StrEnum):
 
 @dataclasses.dataclass(frozen=True)
 class CustomRescoring:
+    '''
+    The `allowed_processing_time` is stored relatively to allow the rescoring to apply to findings
+    with different discovery dates, i.e. in case the rescoring is of scope "global" or "component".
+    Alternatively, the explicit `due_date` can be set in case the `due_date` is independent of the
+    individual discovery dates (for example, this might be the case if exceptions apply).
+    '''
     finding: (
         RescoringVulnerabilityFinding
         | RescoringLicenseFinding
@@ -712,6 +718,8 @@ class CustomRescoring:
     )
     matching_rules: list[str] = dataclasses.field(default_factory=list)
     comment: str | None = None
+    allowed_processing_time: str | None = None
+    due_date: datetime.date | None = None
 
     @property
     def key(self) -> str:
@@ -721,6 +729,8 @@ class CustomRescoring:
             self.user.key,
             self.comment,
             self.finding.key,
+            self.allowed_processing_time,
+            self.due_date.strftime('%Y-%m-%d') if self.due_date else None,
         )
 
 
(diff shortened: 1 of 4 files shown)


dso/model.py
```diff
@@ -712,6 +712,7 @@ class CustomRescoring:
     )
     matching_rules: list[str] = dataclasses.field(default_factory=list)
     comment: str | None = None
+    latest_processing_date: datetime.date | None = None
```
ccwienk, line 715: how about `due_date`?

- shorter
- less ambiguous (latest_processing_date might also be interpreted as the point in time where the last processing took place)

8R0WNI3: I agree, however I chose `latest_processing_date` for consistency reasons with our existing codebase. I'll take a look if we can easily adjust the existing references of `latest_processing_date` to `due_date` as well and if that's the case, will change it here accordingly.

dso/model.py
```diff
@@ -721,6 +729,8 @@ def key(self) -> str:
             self.user.key,
             self.comment,
             self.finding.key,
+            str(self.allowed_processing_time),
```
ccwienk, line 732: hmmm.... I think we should either define the unit (e.g. allowed_processing_time_days), and _not_ convert to str, OR define something more flexible, such as `20d` (for 20 days) - in which case we would not allow int.

or, asking differently: why convert to str here?

ccwienk (changes requested):
see inline comment

dso/model.py
```diff
@@ -712,6 +718,8 @@ class CustomRescoring:
     )
     matching_rules: list[str] = dataclasses.field(default_factory=list)
     comment: str | None = None
+    allowed_processing_time: str | None = None
+    due_date: datetime.date | None = None
```
ccwienk, line 722: should we add a (maybe in post-init) validation that will raise if either both or none are set (i.e. XOR)

dso/model.py
```diff
@@ -721,6 +729,8 @@ def key(self) -> str:
             self.user.key,
             self.comment,
             self.finding.key,
+            self.allowed_processing_time,
+            self.due_date.isoformat() if self.due_date else None,
```
ccwienk, line 733: how about using epoch (w/o milliseconds) instead (rationale: shorter / easier, as format does not have to be kept stable - or rather: is trivial to keep stable).

ccwienk (commented):
I am somewhat concerned about using timestamp as part of key (w.r.t. stability) - apart from that: lgtm

dso/model.py
8R0WNI3: As discussed, will explicitly format date to `YYYY-MM-DD`.

ccwienk (approved):

PR #1221, opened 2025-04-25
add base-component-descriptor-action

ccwienk:
Generalise contract introduced from
`base-component-descriptor`-workflow. Honour a `base-component-file`.

Note that `merge-ocm-fragments`-action still needs to be enabled to merge artefacts specified from base-component-descriptor (this is left for another commit).

**Release note**:

```other developer
add base-component-descriptor-action
```


.github/actions/base-component-descriptor/base_component_descriptor.py
```diff
@@ -0,0 +1,119 @@
...
+    else:
+        if absent_ok:
+            raw = {}
+        else:
+            print(f'Error: not an existing file {path=}')
+            exit(1)
+
+    for forbidden_attr in ('version', 'creationTime'):
+        if forbidden_attr in raw:
+            print('Error: must not specify {forbidden_attr=} in base-component')
```
8R0WNI3, line 64: Missing `f` so that it is a formatted string.

.github/actions/base-component-descriptor/base_component_descriptor.py
```diff
@@ -0,0 +1,119 @@
...
+            exit(1)
+
+    if 'main-source' in raw: # also allow kebap-case
+        raw['main_source'] = raw.pop('main-source')
```
8R0WNI3, line 68: If allowing multiple forms, I'd also vote to allow camelCase as the other properties are defined as camelCase as well.

.github/actions/base-component-descriptor/base_component_descriptor.py
```diff
@@ -0,0 +1,119 @@
...
+
+    return dacite.from_dict(
+        data_class=BaseComponent,
+        data=raw,
```
8R0WNI3, line 72: I'm not sure here, don't we require an enum cast as well?

.github/actions/base-component-descriptor/base_component_descriptor.py
```diff
@@ -0,0 +1,119 @@
...
+
+    if not component.repositoryContexts:
+        component.repositoryContexts = [
+            ocm.OciOcmRepository(baseUrl=ocm_repo),
+        ]
+
+    if not component.provider:
+        component.provider = provider
+
+    if not component.creationTime:
```
8R0WNI3, line 95: As it's forbidden to configure the `creationTime` in the `.ocm/base-component.yaml`, this will always be the case no?

.github/actions/base-component-descriptor/base_component_descriptor.py
```diff
@@ -0,0 +1,119 @@
...
+        component.creationTime = creation_time.strftime('%Y-%m-%dT%H:%M:%SZ')
+
+    main_source_raw = ci.util.merge_dicts(
+        dataclasses.asdict(main_source),
+        component.main_source or {},
```
8R0WNI3, line 100: nit: If we'd set an empty `dict` as default value for `main_source`, we wouldn't need this extra handling here (and below for the "pop"-operation.

.github/actions/base-component-descriptor/action.yaml
```diff
@@ -0,0 +1,146 @@
...
+  version:
+    required: true
+    type: string
+    description: |
+      the effective version to use
+  ocm-repository:
+    required: true
+    type: string
+    description: |
+      the OCM-Repository the Component-Descriptor is intended to be published to. Will be added
```
8R0WNI3, line 46: nit: Will be only added as first (and only) entry in case no other ocm-repository has been specified in the `.ocm/base-component.yaml` file.

.github/actions/base-component-descriptor/action.yaml
```diff
@@ -0,0 +1,146 @@
...
+        base_component_dict = base_component_descriptor.as_component_descriptor_dict(
+          component=base_component,
+        )
+        component_descriptor_str = yaml.dump(
+            data=base_component_dict,
+            Dumper=ocm.EnumValueYamlDumper,
+        )
+
+        with open(os.environ['GITHUB_OUTPUT'], 'a') as f:
+          f.write(component_descriptor_str)
```
8R0WNI3, line 142: Don't you have to specify the output name here as well?

.github/actions/base-component-descriptor/base_component_descriptor.py
ccwienk: the way BaseComponent is currently defined, this should not be necessary.

- OciOcmRepository allows `str` as value for `type`
- ComponentReference does not have any enums
- Label does not have any enums

All other attributes consist of JSON-native types.

.github/actions/base-component-descriptor/base_component_descriptor.py
ccwienk: that's correct. I figured it would be nice for the function's merge-semantics to be consistent (for a hypothetical case where the function might be used in a different context

.github/actions/base-component-descriptor/base_component_descriptor.py
ccwienk: yeah...... it will be slightly more flexible this way (for example, this way, we would properly handle a case where main_source is set to `null`). Anyhow, changed it. However, I think this is a less robust style of coding, so I am not too sure I like that.

ditto for guarding the mentioned pop (which I would prefer to leave in guarded mode)

.github/actions/base-component-descriptor/action.yaml
ccwienk: good catch (not just a nit IMO). documentation should be correct

PR #1223, opened 2025-04-28
extend for GHAS secret findings

Sallyan:
Extend the model for new extension of GHAS (Github Advanced Security) secret findings

**Release note**:

```other operator

```


dso/model.py
```diff
@@ -508,6 +514,21 @@ class DikiFinding(Finding):
...
+    line: int
+    state: str
+    created_at: str
+    updated_at: str
+    repository: str
+
+    # every finding has a key with repo+path+line+secret type
+    @property
+    def key(self) -> str:
+        return _as_key(self.repository, self.path, str(self.line), self.secret_type)
```
ccwienk, line 530: I think it should be fine to not include github-host, considering we also have ocm-component + source (although we _might_ consider including it, just to be sure).

However, I think it might be worthwhile considering to include finding-ID (I think I remember such an ID from the example you showed last time) into `key`. Rationale: it might happen that a different secret will be found in the same path, at the same line. In which (likely very rare) case I think it will be better to be able to tell those apart.
WDYT?

Sallyan: Yes, I agree with you. 
Adding the html_url and secret to the findings. Actually this is the first draft, the new findings format is not clear, let's see. Thanks for taking time to look at this. 😃 
```
"html_url": "https://github.com/kubernetes/demo/security/secret-scanning/123"
```

dso/model.py
```diff
@@ -516,18 +516,20 @@ def key(self) -> str:
...
     created_at: str
     updated_at: str
     repository: str
 
-    # every finding has a key with repo+path+line+secret type
+    # every finding has a key with repo + alert url + secret name + secret type
     @property
     def key(self) -> str:
-        return _as_key(self.repository, self.path, str(self.line), self.secret_type)  
+        return _as_key(self.repository, self.html_url, self.secret, self.secret_type)
```
ccwienk, line 532: I think, considering the html_url contains host + org + repository, you might even drop the (now) redundant `self.repository` part (probably even `secret` and secret_type - if the id at the end of html-url is unique per github-repository). The only requirement we (/ ODG) have for `key` is to be globally unique and stable/reproducible - which seems to be met by just the html_url.

dso/model.py
```diff
@@ -509,6 +516,20 @@ class DikiFinding(Finding):
...
 
+@dataclasses.dataclass(frozen=True)
+class GitHubSecretFinding(Finding):
+    github_instance: str
+    repository: str
+    secret: str
+    secret_type: str
+    secret_type_display_name: str
+    number: int
+    state: str
```
8R0WNI3, line 527: What was the rational behind not including the `path` + `line` here anymore? I think without it, one cannot correlate from these information to the actual finding and thus always has to lookup in the GHAS UI, no?

Sallyan: In the response of secret-scanning alert, it doesn't contain the problematic secret path and line. But 
there is a location_url, could get the path and line with another call. [(github API)](https://docs.github.com/en/enterprise-server@3.13/rest/secret-scanning/secret-scanning?apiVersion=2022-11-28#list-locations-for-a-secret-scanning-alert)
So here I add the path and line, set the html_url as the unique key.

8R0WNI3:
Due to migration of `dso` package to delivery-service, this pull request has been superseded by https://github.com/open-component-model/delivery-service/pull/441.

PR #1224, opened 2025-04-28
refactor: pass (by default) base-component-descriptor as artefact

ccwienk:
Avoid some more boilerplate: Let `merge-ocm-fragments`-action by default fallback to retrieve base-component-descriptor from artefact (also, let base-component-descriptor-action publish such artefacts), effectively allowing to drop explicit passing of base-component-descriptor from build-action to callers.

**Release note**:

```other developer
Let `merge-ocm-fragments`-action by default fallback to retrieve base-component-descriptor from artefact (also, let base-component-descriptor-action publish such artefacts), effectively allowing to drop explicit passing of base-component-descriptor from build-action to callers.
```

### DIFF
--- a/.github/actions/import-commit/action.yaml
+++ b/.github/actions/import-commit/action.yaml
@@ -12,10 +12,17 @@ description: |
 
 inputs:
   commit-objects:
-    required: true
+    required: false
     description: |
       a base64-encoded tarfile containing the objects to import into git-repository. The expected
       format matches the one output from `capture-commit` action.
+
+      Either this input, or `commit-objects-artefact` must be passed.
+  commit-objects-artefact:
+    required: false
+    description: |
+      name of artefact to use as alternative to directly passing objects as input via the
+      `commit-objects` input.
   commit-digest:
     required: false
     description: |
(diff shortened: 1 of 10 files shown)


.github/actions/merge-ocm-fragments/action.yaml
```diff
@@ -66,6 +77,29 @@ runs:
...
+      if: ${{ inputs.component-descriptor-artefact }}
+      uses: actions/download-artifact@v4
+      with:
+        name: ${{ inputs.component-descriptor-artefact }}
+    - name: extract-component-descriptor-artefact
+      if: ${{ inputs.component-descriptor-artefact }}
+      shell: bash
+      run: |
+        archive=component-descriptor.tar.gz
+        if [ -f "${archive}" ]; then
```
TuanAnh17N, line 90: shouldn't this be negated? right now, it checks if the file exists, but based on the error message, i think we want to check if it does not exist

.github/actions/merge-ocm-fragments/action.yaml
```diff
@@ -66,6 +77,29 @@ runs:
...
+          echo "Error: expected file not present: ${archive}"
+          exit 1
+        fi
+        tar xf ${archive}
+        component_descriptor_path='component-descriptor.yaml'
+        if [ -f "${component_descriptor_path}" ]; then
```
TuanAnh17N, line 96: same here

.github/actions/merge-ocm-fragments/action.yaml
ccwienk: yes, of course m(
-> thx

.github/actions/import-commit/action.yaml
```diff
@@ -60,10 +67,25 @@ runs:
         if which git &>/dev/null; then exit 0; fi
         apt-get install -y git
       shell: bash
+    - name: import-commit-objects-from-input
+      if: ${{ inputs.commit-objects != '' }}
+      run: |
+        set -eu
+        echo "${{ inputs.commit-objects }}" | base64 -d | tar x
+    - name: import-commit-objects-from-artefact
+      if: ${{ inputs-commit-objects-artefact != '' }}
```
TuanAnh17N, line 76: i think this is a typo

`inputs.commit-objects-artefact`

.github/actions/import-commit/action.yaml
```diff
@@ -60,10 +67,25 @@ runs:
...
+      uses: actions/download-artifact@v4
+      with:
+        name: ${{ inputs.commit-objects-artefact }}
+    - name: extract-commit-objects-from-artefact
+      if: ${{ inputs-commit-objects-artefact != '' }}
```
TuanAnh17N, line 81: ditto

.github/actions/import-commit/action.yaml
ccwienk: thanks for catching this!
done

PR #1226, opened 2025-04-29
add resuable workflow for building helmcharts w/ OCM

ccwienk:
Add a convenience-wrapper around helmchart-action to avoid boilerplate for pipelines relying on the defaulting offered from `prepare` workflow.

Concretely-speaking, this flow will avoid having to collect ocm-resource-fragments containing images referenced in helmchart-mappings (which are needed in order to validate mappings).

**Release note**:

```other developer
Add a convenience-wrapper around helmchart-action to avoid boilerplate for pipelines relying on the defaulting offered from `prepare` workflow.
```


8R0WNI3 (approved):
/lgtm

PR #1228, opened 2025-05-05
Move `dso` package to delivery-service

8R0WNI3:
**Release note**:

```
NONE
```

### DIFF
--- a/cnudie/iter.py
+++ b/cnudie/iter.py
@@ -3,13 +3,24 @@ import dataclasses
 import enum
 
 import cnudie.retrieve
-import dso.labels
 import ocm
+
+
+@dataclasses.dataclass
+class ExtraComponentReference:
+    component_reference: ocm.ComponentIdentity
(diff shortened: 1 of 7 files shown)


zkdev:
Since os-id extension was rolled-out and checkmarx got replaced with SAST scans, can't we move `dso.model` to delivery-service repository?

If so, I suggest to combine with this PR (rather than extending `dso.model` in cc-utils)

cnudie/iter.py
```diff
@@ -3,13 +3,24 @@
 import enum
 
 import cnudie.retrieve
-import dso.labels
 import ocm
 
 
+@dataclasses.dataclass(frozen=True)
+class ExtraComponentReference:
```
ccwienk, line 10: do ECR and RCRL need to be frozen (esp: hashable)? If not, we might use lists and would not need dacite to cast..

zkdev (approved):
/lgtm

PR #1229, opened 2025-05-07
Don't raise upon 404 on DELETE manifest operations per default

8R0WNI3:
**Release note**:

```other developer
OCI client does not raise upon 404 per default for DELETE manifest operations
```

### DIFF
--- a/oci/client.py
+++ b/oci/client.py
@@ -814,6 +814,7 @@ class Client:
         image_reference: om.OciImageReference | str,
         purge: bool=False,
         accept: str=om.MimeTypes.prefer_multiarch,
+        absent_ok: bool=False,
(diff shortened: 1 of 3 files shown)


oci/client.py
```diff
@@ -814,6 +814,7 @@ def delete_manifest(
         image_reference: om.OciImageReference | str,
         purge: bool=False,
         accept: str=om.MimeTypes.prefer_multiarch,
+        absent_ok: bool=True,
```
zkdev, line 817: I think we should default to `False` (or adjust all current usages) to ensure no break in current behaviours.

zkdev (approved):
/lgtm

PR #1230, opened 2025-05-07
Allow oci-client without credential-lookup (anonymous auth)

zkdev:
Useful when working with public images, as credential handling boilerplate can be omitted.

**Release note**:

```other developer
gardener-cicd-libs oci-client will not use credentials by default (useful for public images)
```


8R0WNI3 (approved):
/lgtm

PR #1231, opened 2025-05-08
Drop dependency towards bdba client

TuanAnh17N:
**Release note**:

```other operator
removed bdba-client from component references and build workflow as it is now handled externally.
```

### DIFF
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -197,20 +197,6 @@ jobs:
           cat /tmp/pip3-install.log
         fi
 
-        pkg_dir=/tmp/dist
-        echo "downloading bdba package"
-        bdba_version=$(
-          yq \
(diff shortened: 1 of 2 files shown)


TuanAnh17N:
linter failure is expected — cfg_mgmt still exists in the repo until the migration pr is merged

.github/workflows/build-and-test.yaml
```diff
@@ -199,20 +199,6 @@ jobs:
           cat /tmp/pip3-install.log
         fi
 
-        pkg_dir=/tmp/dist
-        echo "downloading bdba package"
-        bdba_version=$(
-          yq \
```
8R0WNI3, line 205: You can also drop the installation of `yq` above :-)

8R0WNI3 (approved):
/lgtm

PR #1232, opened 2025-05-14
default to org-secret

ccwienk:
Do not require calling workflows to explicitly pass-in org-secret to release-workflow. Instead, have it default to reading org-secret. This will make usage of additional secrets more convenient.

**Release note**:

```other developer
reusable `release.yaml` workflow now no longer needs to be passed org-secrets explicitly, if using `secrets: inherit`.
```


8R0WNI3 (approved):
/lgtm

PR #1235, opened 2025-05-15
fix: Pass cfg-factory if provided

zkdev:
Otherwise fallback to secrets-server is performed which will always fail for replicate-secrets pipelines.

**Release note**:

```bugfix developer
delivery-client factory will now pass "ConfigFactory" to auth-token-lookup
```


8R0WNI3 (approved):
/lgtm

ccwienk (approved):

PR #1236, opened 2025-05-16
Resolve reference to `cnudie` package in `ocm download` command

8R0WNI3:
**Release note**:

```other developer
The `gardener-ocm` Python package no longer requires the `cnudie` package to be installed
```


zkdev (approved):
/lgtm

PR #1238, opened 2025-05-20
workaround: improve handling of OCI-Manifests w/o mediaType

ccwienk:
Some few (we currently know of _one_) OCI-Manifest do not feature a `mediaType` toplevel-attribute (as recommended by OCI-Spec). This will lead to deserialisation errors, if the underlying manifest is not a single image-manifest, but a Manifest-List / OCI-Image-Index.

To improve / make oci package more robust, honour Content-Type conveyed from OCI-Registries via the `Content-Type` header in those cases.

**Release note**:

```other developer
workaround: improve handling of OCI-Manifests w/o mediaType

Some OCI-Manifests do not seem to contain a `mediaType`-attribute. This would lead to deserialisation errors in case those manifests contained either image-lists, or oci-image-indexes. As Docker seems to accept such manifests, improve `oci` package to follow suit. Honour Content-Type-header as fallback for such manifests.
```


cli/gardener_ci/_oci.py
```diff
@@ -307,9 +307,17 @@ def manifest(
...
+
+    media_type = manifest_dict.get(
+        'mediaType',
+        manifest_raw.headers.get('Content-Type')
+    )
+
     if pretty and not print_expr:
         manifest = oci.model.as_manifest(
             manifest=manifest_raw.text,
+            media_type=media_type,
```
8R0WNI3, line 320: I think the `media_type` should be passed into the below `.as_manifest()` call as well 🤔

ccwienk: as always: very good catch - ty

PR #1246, opened 2025-05-26
Honour imagevector in base-component-descriptor action

ccwienk:
**Release note**:

```other developer
honour imagevector in base-component-descriptor GitHub-Action
```

### DIFF
--- a/ocm/gardener.py
+++ b/ocm/gardener.py
@@ -1,0 +1,149 @@
+'''
+this module contains code specific to how Gardener-Components (https://github.com/gardener)
+define cross-references to resources to other (Gardener-)Components.
+
+Code in this module is not intended for re-use.
+'''
+
+import copy
+
+import yaml
+
+import oci.model
+import ocm
+
+
+def add_resources_from_imagevector(
+    component: ocm.Component,
+    images_yaml_path: str,
+    component_prefixes: list[str],
+) -> ocm.Component:
+  # images_yaml_path, as found e.g. at github.com/gardener/gardener charts/images.yaml
+  def iter_images():
+    with open(images_yaml_path) as f:
+      for part in yaml.safe_load_all(f):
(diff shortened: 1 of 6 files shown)


ocm/gardener.py
```diff
@@ -5,25 +5,46 @@
...
-  def iter_images():
+) -> collections.abc.Generator[dict, None, None]:
     with open(images_yaml_path) as f:
       for part in yaml.safe_load_all(f):
         yield from part['images']
 
+
+def add_resources_from_imagevector(
+    component: ocm.Component,
+    image_dicts: collections.abc.Iterable[dict],
```
8R0WNI3, line 45: nit: Rather name this `images` or something like that? That those are of type `dict` is already expressed via type hint.

ocm/gardener.py
```diff
@@ -0,0 +1,170 @@
...
+
+  for image_dict in image_dicts:
+    name = image_dict['name']
+    resource_id = image_dict.get('resourceId', {'name': name})
+    source_repo = image_dict.get('sourceRepository', None)
+    img_repo = image_dict['repository']
+    extra_identity = image_dict.get('extraIdentity', {})
+    labels = copy.copy(image_dict.get('labels', []))
+    tag = image_dict.get('tag', None)
+    target_version = image_dict.get('targetVersion', None)
```
zkdev, line 66: Why not introduce a dataclass to describe `image_dict`?
Especially useful considering the default values.

ccwienk: we might do that in a separate change. However:

- this pullrequest is foremostly focussed around re-using the existing implementation also for base-component-descriptor-action
- the contract is defined on gardener-side, and is not fully known to us / not explicitly defined (to my knowledge). consuming as raw dict makes this slightly more obvious

PR #1267, opened 2025-06-26
Require configuration of targets in CTT `processing.cfg`

8R0WNI3:
**Release note**:

```breaking operator
The OCI replication now requires configuration of `targets` in the `processing.cfg` + `PrefixUploader` was changed to `RepositoryUploader`
```


ccwienk (changes requested):
as discussed: let us make this change backwards-compatible and hence make rolling-out easier.

PR #1271, opened 2025-06-30
allow (re-)setting of extra-tags for OCI-OCM-Workflow

ccwienk:
**Release note**:

```other developer
oci-ocm reusable workflow allows to (re-)set extra-tags
```

### DIFF
--- a/.github/workflows/oci-ocm.yaml
+++ b/.github/workflows/oci-ocm.yaml
@@ -37,6 +37,12 @@ on:
         type: string
         default: '{version}'
         required: false
+      extra-tags:
+        type: string
+        required: false
+        description: |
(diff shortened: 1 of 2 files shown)


.github/workflows/oci-ocm.yaml
```diff
@@ -336,6 +342,7 @@ jobs:
               src_image_refs=src_image_refs,
               tgt_image_ref=tgt_image_ref,
               oci_client=oci_client,
+              extra_tags='${{ inputs.extra-tags }}'.split(','),
```
8R0WNI3, line 345: nit: We might add an additional `strip` or replace whitespaces to prevent wrong inputs, wdyt?

PR #1272, opened 2025-06-30
Create "replication plan" for CTT processing

8R0WNI3:
**Release note**:

```other developer
Refactored CTT to create a "replication plan" prior to actual replication to ease development and error analysis
```


ctt/process_dependencies.py
```diff
@@ -523,6 +472,146 @@ def create_backwards_compatible_cfg(
...
+    tgt_ocm_repository: str,
+    oci_client: oci.client.Client,
+    delivery_service_client: delivery.client.DeliveryServiceClient,
+    replication_mode: oci.ReplicationMode=oci.ReplicationMode.PREFER_MULTIARCH,
+    component_filter: collections.abc.Callable[[ocm.Component], bool] | None=None,
+    reftype_filter: collections.abc.Callable[[cnudie.iter.NodeReferenceType], bool] | None=None,
+    remove_label: collections.abc.Callable[[str], bool]=None,
+) -> ctt.model.ReplicationPlanStep:
+    tgt_ocm_repo = ocm.OciOcmRepository(
+        baseUrl=tgt_oci_registry,
```
ccwienk, line 573: nit: I think the discrimination between baseUrl and subPath for OciOcm-Repositories is de-facto deprecated; hence, I suggest to merge this function's `tgt_oci_registry` and `tgt_ocm_repository` into just one parameter `ocm_repository`

ctt/process_dependencies.py
```diff
@@ -523,6 +472,146 @@ def create_backwards_compatible_cfg(
...
+        subPath=tgt_ocm_repository,
+    )
+
+    tgt_component_descriptor_lookup = cnudie.retrieve.create_default_component_descriptor_lookup(
```
ccwienk, line 577: pass-in (so that caller can, for example, opt to re-use existing lookups)?

ctt/process_dependencies.py
```diff
@@ -523,6 +472,146 @@ def create_backwards_compatible_cfg(
...
+        ocm_repository_lookup=cnudie.retrieve.ocm_repository_lookup(tgt_ocm_repo.oci_ref),
+        oci_client=oci_client,
+        delivery_client=delivery_service_client,
+        fallback_to_service_mapping=False,
+    )
+
+    component_descriptors = tuple(determine_changed_components(
```
ccwienk, line 584: we might - if this is more convenient - keep references to `ocm.Component` (rather than ComponentDescriptor), and create the `ComponentDescriptor`- envelope ad-hoc during upload.

ctt/process_dependencies.py
```diff
@@ -523,6 +472,146 @@ def create_backwards_compatible_cfg(
...
+        reftype_filter=reftype_filter,
+    ))
+
+    components = tuple(iter_replication_plan_components(
+        component_descriptors=component_descriptors,
+        tgt_ocm_repo=tgt_ocm_repo,
+        remove_label=remove_label,
+    ))
+
+    resources = list(iter_replication_plan_resources(
```
ccwienk, line 599: why use a tuple for components, but a list for resources?

ctt/process_dependencies.py
```diff
@@ -577,96 +673,93 @@ def process_images(
...
+            remove_label=remove_label,
+        )
+        replication_plan.steps.append(replication_plan_step)
+
+    logger.info(replication_plan)
+
+    for replication_plan_step in replication_plan.steps:
+        yield from process_replication_plan_step(
+            replication_plan_step=replication_plan_step,
+            component_descriptor_v2=component_descriptor_v2,
```
ccwienk, line 699: can we drop the `_v2`-suffix, while we are refactoring (if you remember, this v2-suffix dates back to very olden times, where not even cnudie was coined as a name)? :-)

ctt/process_dependencies.py
```diff
@@ -577,96 +673,93 @@ def process_images(
...
         )
 
-        if extra_tags := processing_job.extra_tags:
-            target_ref = om.OciImageReference(processing_job.upload_request.target_ref)
-            target_repo = target_ref.ref_without_tag
+        if not oci_manifest_digest:
+            raise RuntimeError(f'No digest returned for {replication_plan_resource=}')
+
+        if processing_mode is ProcessingMode.DRY_RUN:
+            extra_tags = []
```
ccwienk, line 754: should it not be sufficient / better to skip pushing of extra-tags, rather than making them an empty list?

ctt/process_dependencies.py
```diff
@@ -677,122 +770,84 @@ def process_job(processing_job: processing_model.ProcessingJob):
...
             ):
                 digest.value = oci_manifest_digest.removeprefix('sha256:')
+                replication_plan_resource.target.digest = digest
+
+        if replication_plan_resource.convert_to_relative_ref:
+            # remove host from target ref
+            replication_plan_resource.target.access = ocm.RelativeOciAccess(
+                reference=om.OciImageReference(
+                    image_reference=replication_plan_resource.tgt_ref,
+                    normalise=False, # don't inject docker special handlings
```
ccwienk, line 788: is it actually necessary to have this special-handling (I know this is not new code - nonetheless, now might be a good time to cleanup..)?

ccwienk (changes requested):

ctt/model.py
```diff
@@ -0,0 +1,160 @@
...
+@dataclasses.dataclass
+class ReplicationPlan:
+    '''
+    Contains all information required for replication. The described `steps` can be processed in
+    parallel. Each `step` describes the replication to a different target registry.
+    '''
+    steps: list[ReplicationPlanStep] = dataclasses.field(default_factory=list)
+
+    def __str__(self) -> str:
+        return f'''\
```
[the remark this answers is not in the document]
8R0WNI3: I think `textwrap.dedent` is not a good choice here because it would require all `steps` to include that indentation as well to work as expected.

ctt/process_dependencies.py
8R0WNI3: Well, by making them an empty list the pushing of extra-tags is skipped (as there are none), but I agree it might be better to explicitly skip pushing in that case.

ctt/process_dependencies.py
8R0WNI3: Yes, it is still required... Otherwise, `registry-1.docker.io` would be prepended :-(

ctt/model.py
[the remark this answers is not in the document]
ccwienk: valid point.

ctt/process_dependencies.py
ccwienk: so you use the fact that list is mutable (and hence its elements might be subject to being re-ordered), as a means to express the order is irrelevant? hmmmm, that might be slightly too subtle to most readers (including me...).

asking differently: is there the _need_ to mutate the list? if not, I would suggest to use tuples in both cases.

8R0WNI3: No, will use tuple instead.

ctt/model.py
```diff
@@ -0,0 +1,160 @@
...
+    def __str__(self) -> str:
+        return (
+            f'{self.source.name}:{self.source.version} '
+            f'[{self.src_ref} -> {self.preliminary_tgt_ref}]'
+        )
+
+
+@dataclasses.dataclass
+class ReplicationPlanComponent:
+    source: ocm.ComponentDescriptor
```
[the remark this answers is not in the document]
8R0WNI3: The `ocm.gardener.UpgradeVector` only comprises the component _identity_, we require the full OCM component descriptor here to be able to modify it prior to replication.

ctt/model.py
```diff
@@ -0,0 +1,158 @@
...
+    def tgt_ref(self) -> oci.model.OciImageReference:
+        if self.target.access.type is ocm.AccessType.OCI_REGISTRY:
+            return oci.model.OciImageReference(self.target.access.imageReference)
+        elif self.target.access.type is ocm.AccessType.RELATIVE_OCI_REFERENCE:
+            return oci.model.OciImageReference(self.target.access.reference, normalise=False)
+        else:
+            raise ValueError(self.source.access.type)
+
+    @property
+    def preliminary_tgt_ref(self) -> oci.model.OciImageReference:
```
ccwienk, line 42: very nice!

ctt/model.py
```diff
@@ -0,0 +1,158 @@
...
+            return self.tgt_ref
+
+        tgt_ref = self.tgt_ref
+
+        if self.digest:
+            digest = self.digest
+        elif tgt_ref.has_digest_tag:
+            digest = self.tgt_ref.tag
+        else:
+            digest = 'sha256:<placeholder-digest>'
```
ccwienk, line 59: nit: should we make this slightly more self-describing, like `"<placeholder-if-you-see-me-this-indicates-a-bug>"`?

ccwienk (approved):

PR #1273, opened 2025-07-01
Support multiple values with same resource in image mapping

8R0WNI3:
Add support if the image of the same resource should be used for multiple Helm values with distinct top level value references. This is especially useful if the image is used in multiple sub charts.

**Release note**:

```other operator

```


zkdev:
Is this a breaking change? @8R0WNI3

ccwienk (approved):
good catch!

8R0WNI3:
> can you please update the corresponding test to cover this scenario?

Done.

> also I am not too happy with passing around values via splitted string in dict keys :-/
maybe a different data structure will help (I have no concrete suggestion yet, will think a bit about it)

I updated it to a nested dictionary instead, wdyt?

PR #1281, opened 2025-07-07
Add `component-diff` GitHub workflow (+ action)

8R0WNI3:
**Release note**:

```feature developer
Added a (re-usable) workflow to determine the component-diff
```

### DIFF
--- a/.github/actions/component-diff/action.yaml
+++ b/.github/actions/component-diff/action.yaml
@@ -1,0 +1,129 @@
+name: component-diff
+description: |
+  An action to determine the component-diff between the current version (read from passed-in
+  component-descriptor) and the greatest published release-version. The diff is structured according
+  to `cnudie.util.ComponentDiff`.
+
+inputs:
+  component-descriptor:
+    required: true
+    type: string
+    description: |
+      An OCM-Component-Descriptor in YAML form.
+  ocm-repositories:
+    required: true
+    type: string
+    description: |
+      A (comma-separated) list of ocm-repository-URLs where component-versions are to be looked up
+      (order is honoured).
+
+      Caveat: this action will only honour final versions. Hence, at least one ocm-repository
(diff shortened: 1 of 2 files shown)


.github/actions/component-diff/action.yaml
```diff
@@ -0,0 +1,111 @@
...
+        current_component_descriptor = ocm.ComponentDescriptor.from_dict(
+          yaml.safe_load('''${{ steps.component-descriptor.outputs.component-descriptor}}'''),
+        )
+        component_name = current_component_descriptor.component.name
+        component_version = current_component_descriptor.component.version
+
+        ocm_repositories = '${{ inputs.ocm-repositories }}'.split(',')
+        ocm_repository_lookup = cnudie.retrieve.ocm_repository_lookup(*ocm_repositories)
+
+        oci_client = oci.client.Client()
```
ccwienk, line 56: I think we should rather use an OCI-Client what will honour "docker-auth", to at least prepare _some_ means of authentication, where required.

btw, there is also [this](https://github.com/gardener/cc-utils/pull/1282) related pullrequest, which I forgot to publish some days ago, which might come in handy

.github/actions/component-diff/action.yaml
```diff
@@ -0,0 +1,111 @@
+name: component-diff
+description: |
+  An opinionated action to determine the component-diff between the current version (read from
```
ccwienk, line 3: I think it would fit in better if this action would need the (resolved / merged) component-descriptor as an input, and if the convenience-workflow wrapping it would deal w/ collecting the component-descriptor from `prepare.yaml` workflow. This way, the action could be instrumented differently (i.e. w/o implemanting against how `prepare.yaml` exposes base-component-descriptor), while still being convenient (for users of wrapping workflow).

PR #1283, opened 2025-07-07
refactor: relocate helmchart-imagemap processing

TuanAnh17N:
**Release note**:

```other operator
Moved helmchart-imagemapping to make it reusable
```

### DIFF
--- a/ocm/helm.py
+++ b/ocm/helm.py
@@ -1,0 +1,97 @@
+import jsonpath_ng
+
+import oci.client
+import oci.model as om
+import ocm
+import ocm.util
+
+
+def localised_helmchart_values(
+    component: ocm.Component,
+    oci_client: oci.client.Client,
+    resource_name: str,
+    resource_version: str | None=None,
+    resource_extra_id: dict[str, str] | None=None,
+    resource_type: str='helmchart-imagemap',
+) -> dict:
+    '''
+    Resolves image references from a helmchart-imagemap resource and returns a dictionary
+    with helm values
+    '''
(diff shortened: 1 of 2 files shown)


helm_image_mapping.py
```diff
@@ -0,0 +1,91 @@
...
+            continue
+        if resource_version and resource_version != resource.version:
+            continue
+        if mapping_type != resource.type:
+            continue
+        if resource_extra_id and resource.extraIdentity != resource_extra_id:
+            continue
+        break
+    else:
+        raise ValueError(f'Error: did not find {mapping_type=} for {resource=}')
```
8R0WNI3, line 30: nit: `Error:` is obsolete since we already raise a `ValueError` :-)

Also, logging `resource` here does not really make sense because it is just a (random) resource of the component. Instead, we should rather print the passed-in values for name, version, ...

helm_image_mapping.py
```diff
@@ -0,0 +1,91 @@
...
+        resource_name = image_mapping['resource']['name']
+        for resource in component.resources:
+            if resource.name != resource_name:
+                continue
+            if not resource.type is ocm.ArtefactType.OCI_IMAGE:
+                continue
+            break # found it
+        else:
+            print(f'Error: did not find oci-image w/ {resource_name=} in {component=}')
+            exit(1)
```
8R0WNI3, line 65: ditto: Raise instead of print + exit.

8R0WNI3 (approved):
/lgtm

PR #1286, opened 2025-07-10
Add support for AWS Elastic Container Registry in OCI package

8R0WNI3:
**Release note**:

```feature developer
The Python `oci` package now supports AWS Elastic Container Registries
```


oci/client.py
```diff
@@ -1072,8 +1122,24 @@ def _put_blob_chunked(
             method='POST',
             headers={
                 'content-length': '0',
-            }
+            },
+            raise_for_status=False,
         )
+
+        if (
```
ccwienk, line 1129: :-/

firstly, I think a brief comment explaining this codepath (aws-oci-registry requires repositories to be initialised) would be very helpful for future readers.

secondly, I think we could/should make this slightly more generic, as there are (or were) other oci-registries requiring initialisation (such as AliYun's OCI-Registry).

How about, for example, adding a function (/decorator) that has a semantics of "initialise repository if needed", and have it (for now) contain special-case-code for AWS? I feel this could make the code more expressive, and also easier to extend for other registries.

If taking the decorator-path, maybe this would also help to deduplicate the second codepath below.

oci/client.py
```diff
@@ -1193,3 +1276,19 @@ def _put_blob_single_post(
 
         res.raise_for_status()
         return res
+
+    def _create_aws_repository(
```
ccwienk, line 1280: shouldn't this be part of oci/aws.py?

oci/aws.py
```diff
@@ -0,0 +1,225 @@
...
+
+    res = request(
+        action=AwsAction.GET_AUTHORIZATION_TOKEN,
+        body=body,
+        image_reference=image_reference,
+        credentials=credentials,
+        session=session,
+    )
+
+    token = res.json()['authorizationData'][0]['authorizationToken']
```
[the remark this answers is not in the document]
8R0WNI3: There are only multiple authorisation tokens in case multiple registry-ids are specified in the request parameters. However, as both aspects are hard-coded in this function, I would keep it as-is + add a comment explaining this :-)

ccwienk (approved):
very nice; I especially like the initialiser-dectorator

PR #1304, opened 2025-07-18
Switch to ocm release notes handling

ccwienk:
**Release note**:

```other developer
Switch to ocm release notes handling
```

### DIFF
--- a/.github/actions/release-notes/action.yaml
+++ b/.github/actions/release-notes/action.yaml
@@ -85,20 +85,32 @@ runs:
           ${draft_arg:-} \
           ${ocm_repositories_arg:-} \
           --full-release-notes full-release-notes.md \
-          --local-release-notes release-notes.md
+          --local-release-notes release-notes.md \
+          --subcomponent-release-notes subcomponent-release-notes.md
 
         echo 'Full Release-notes written to `full-release-nodes.md`'
         echo 'Local Release-notes written to `release-notes.md`'
+        echo 'Subcomponent Release-notes written to `subcomponent-release-notes.md`'
 
         echo 'release-notes<<EOF' >> ${GITHUB_OUTPUT}
         cat release-notes.md >> ${GITHUB_OUTPUT}
         echo EOF >> ${GITHUB_OUTPUT}
 
         echo 'full-release-notes<<EOF'
-        cat full-release-notes.md > ${GITHUB_OUTPUT}
+        cat full-release-notes.md >> ${GITHUB_OUTPUT}
         echo EOF >> ${GITHUB_OUTPUT}
 
-        cat << EOF > ${GITHUB_STEP_SUMMARY}
+        echo 'subcomponent-release-notes<<EOF'
+        cat subcomponent-release-notes.md >> ${GITHUB_OUTPUT}
+        echo EOF >> ${GITHUB_OUTPUT}
+
+        cat << EOF >> ${GITHUB_STEP_SUMMARY}
         ## Release-Notes (local)
         $(cat release-notes.md)
         EOF
+
+        if [ $( stat -c%s subcomponent-release-notes.md) -gt 0]; then
+        cat << "EOF" >> ${GITHUB_STEP_SUMMARY}
+        ## Release-Notes (subcomponents)
+        $(cat subcomponent-release-notes.md)
+        fi
--- a/.github/actions/release-notes/release_notes_cli.py
+++ b/.github/actions/release-notes/release_notes_cli.py
@@ -86,6 +86,11 @@ def main():
         '--local-release-notes',
         default=None,
     )
+    parser.add_argument(
+        '--subcomponent-release-notes',
+        default=None,
+        help='release-notes from sub-components',
+    )
 
     parsed = parser.parse_args()
     print(pprint.pformat(parsed), file=sys.stderr)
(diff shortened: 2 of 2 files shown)


release_notes/ocm.py
```diff
@@ -105,3 +114,41 @@ def release_notes_range(
...
         logger.info(f'found {len(notes)=} characters of release-notes for {component_id=}')
         yield component_id, notes
+
+
+def release_notes_range_recursive(
+    version_vector: ocm.gardener.UpgradeVector,
+    component_descriptor_lookup: ocm.ComponentDescriptorLookup,
+    version_lookup: ocm.VersionLookup,
+    oci_client: oci.client.Client,
+    version_filter=None,
```
8R0WNI3, line 124: nit: You might pass a default like `lambda: True` to reduce the conditional branches in the function :-)

concourse/steps/release.py
```diff
@@ -154,6 +157,29 @@ def collect_release_notes(
...
+            version_vector=version_vector,
+            component_descriptor_lookup=component_descriptor_lookup,
+            version_lookup=version_lookup,
+            oci_client=oci_client,
+            version_filter=version.is_final,
+        )
+    ))
+
+    if sub_component_release_notes:
+        release_notes_markdown = f'{release_notes_markdown}\n{sub_component_release_notes}'
```
8R0WNI3, line 181: This will cause all recursive release notes to be included in the "current" component's release notes (including the one uploaded as local blob), no? If it does, this will mean we have redundant release notes if we fetch them recursively 🤔

.github/actions/release-notes/release_notes_cli.py
```diff
@@ -168,18 +169,37 @@ def github_api_lookup(repo_url):
...
+            component_descriptor_lookup=component_descriptor_lookup,
+            version_lookup=ocm_version_lookup,
+            oci_client=oci_client,
+            version_filter=version.is_final,
+        )
+    ))
 
-    append_newline = not release_notes_md.endswith('\n')
+    if sub_component_release_notes:
+        release_notes_md = f'{release_notes_md}\n{sub_component_release_notes}'
```
8R0WNI3, line 196: ditto?

.github/actions/ocm-upgrade/ocm_upgrade.py
```diff
@@ -155,19 +155,23 @@ def create_upgrade_pullrequest_diff(
...
-    release_notes = '\n'.join((rn for _, rn in rno.release_notes_range(
-        version_vector=upgrade_vector,
-        versions=versions,
-        oci_client=oci_client,
-        component_descriptor_lookup=component_descriptor_lookup,
-        absent_ok=True,
-    )))
+    release_notes = '\n'.join((
+        rno.release_notes_markdown_with_heading(cid, rn)
+        for cid, rn in rno.release_notes_range_recursive(
```
8R0WNI3, line 167: I think what is still missing is to fetch the release notes from OCM for the Concourse-based upgrade-PRs as well.

ccwienk: yes, you are correct (I actually assumed this might be okay do omit, as their bodies are (as of this change) no longer honoured. However, for the sake of consistency, you have a point. Will add

.github/actions/release-notes/release_notes_cli.py
ccwienk: you mean I should also implement this for draft-release-jobs on cc?

concourse/steps/release.py
ccwienk: I of course pondered that, and figured there should not yet be release-notes for current version, as the component-descriptor should not yet have been published. If I am not mistaken, the release-notes collected in this step will subsequently be uploaded

concourse/steps/release.mako
```diff
@@ -524,8 +526,8 @@ release_tag = tags[0].removeprefix('refs/tags/')
 draft_tag = f'{version_str}-draft'
 
 if release_notes_md is not None:
-  release_notes_md, is_full_release_notes  = github.release.body_or_replacement(
-    body=release_notes_md,
+  fulle_release_notes_md, is_full_release_notes  = github.release.body_or_replacement(
```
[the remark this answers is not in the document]
ccwienk: thx

.github/actions/release-notes/action.yaml
```diff
@@ -79,15 +84,21 @@ runs:
...
         cat release-notes.md >> ${GITHUB_OUTPUT}
         echo EOF >> ${GITHUB_OUTPUT}
 
+        echo 'full-release-nodes<<EOF'
+        cat full-release-notes.md > ${GITHUB_OUTPUT}
+        echo EOF >> ${GITHUB_OUTPUT}
+
         cat << EOF > ${GITHUB_STEP_SUMMARY}
-        ## Release-Notes
+        ## Release-Notes (local)
```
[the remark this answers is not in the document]
ccwienk: yeah, I also figured that. and then, I figured, it might be even better to explicitly write the extra-part. hmmm... which we could actually calculate doing str-manipulation, now that I think of it.

.github/actions/release-notes/action.yaml
```diff
@@ -79,15 +84,33 @@ runs:
...
+        echo 'subcomponent-release-notes<<EOF'
+        cat subcomponent-relase-notes.md >> ${GITHUB_OUTPUT}
+        echo EOF >> ${GITHUB_OUTPUT}
+
+        cat << EOF >> ${GITHUB_STEP_SUMMARY}
+        ## Release-Notes (local)
         $(cat release-notes.md)
         EOF
+
+        if [ $( stat -c%s subcomponent-release-notes.md) == 0]; then
```
8R0WNI3, line 112: Mustn't this check for `!= (or >) 0`?

ccwienk: dang! of course. hits me all the time <.<

PR #1313, opened 2025-07-24
split oci-image-building and -pushing into two steps

ccwienk:
For long-running builds, it may happen that auth-token retrieved via OIDC expires during build, leading to failing attempts for uploading built images.

Hence, split building and pushing into two steps, and perform another oidc-authentication after build, before attempting to push. Keep initial authorisation such that read-access (for base-images) during build is possible, if authentication is required.

**Release note**:

```other developer
split oci-image-building and -pushing into two steps
```

### DIFF
--- a/.github/workflows/oci-ocm.yaml
+++ b/.github/workflows/oci-ocm.yaml
@@ -330,12 +330,12 @@ jobs:
           extra-auths: ${{ steps.extra-auth.outputs.extra-auths }}
 
       - name: 'push ${{ inputs.name }} / ${{ matrix.args.platform-name }}'
-        uses: docker/build-push-action@v6
         if: ${{ needs.preprocess.outputs.push == 'true' }}
+        shell: bash
         id: push
-        with:
-          push: ${{ needs.preprocess.outputs.push }}
-          tags: ${{ matrix.args.image-reference }}
+        run: |
+          set -eu
+          docker push ${{ matrix.args.image-reference }}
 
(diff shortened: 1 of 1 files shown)


.github/workflows/oci-ocm.yaml
```diff
@@ -314,6 +314,7 @@ jobs:
         uses: docker/build-push-action@v6
         id: build
         with:
+          load: true
```
8R0WNI3, line 317: I think we have to remove the `push` parameter below now.

.github/workflows/oci-ocm.yaml
```diff
@@ -322,6 +323,22 @@ jobs:
...
+          gh-token: ${{ secrets.GITHUB_TOKEN }}
+          extra-auths: ${{ steps.extra-auth.outputs.extra-auths }}
+
+      - name: 'push ${{ inputs.name }} / ${{ matrix.args.platform-name }}'
+        uses: docker/build-push-action@v6
+        if: ${{ needs.preprocess.outputs.push == 'true' }}
+        id: push
+        with:
+          platforms: ${{ matrix.args.platform-name }}
+          push: ${{ needs.preprocess.outputs.push }}
```
8R0WNI3, line 339: nit: We might just set this to `true` as a constant as it will be always `true` anyways if this step is executed.

8R0WNI3 (approved):
/lgtm

PR #1314, opened 2025-07-25
fail if localBlob's file is not present

ccwienk:
If export-ocm-fragment action is used to export a resource using local-blob from a file, signal it as an error if referenced file does not exist (as this typically indicates a user error).

**Release note**:

```other developer
fail if localBlob's file is not present
```


8R0WNI3 (approved):
/lgtm

PR #1317, opened 2025-07-28
Add new release notes model and tarutils

8R0WNI3:
**Release note**:

```other developer

```


ccwienk (approved):
I think (when it comes to instrumentation), we should
- be forgiving w/ unexpected titles/categories for release-notes from ocm-component-descriptors + local repo (during release)
- be strict during validation (when inspecting _unmerged_ pullrequests)

PR #1319, opened 2025-07-29
Add machine-readable release notes archive

TuanAnh17N:
**Release note**:

```other developer
Release notes documents are now used during the release and attached to the OCM component descriptor
```


release_notes/archive.py
```diff
@@ -0,0 +1,69 @@
...
+
+def read_release_notes_from_dir(source_dir: str) -> list:
+    docs = []
+    for fname in os.listdir(source_dir):
+        if not fname.endswith('.yaml'):
+            continue
+        with open(os.path.join(source_dir, fname)) as f:
+            raw_data = yaml.safe_load(f)
+
+            rn_data = raw_data.get('release_notes') or raw_data.get('release-notes') or []
```
8R0WNI3, line 19: What about directly using `dacite` for deserialisation, i.e.
```python
release_notes_doc = dacite.from_dict(
    data_class=rnm.ReleaseNotesDoc,
    data=raw_data,
    config=dacite.Config(
        cast=[enum.Enum],
    ),
)
```

release_notes/archive.py
```diff
@@ -0,0 +1,69 @@
...
+import os
+
+import dacite
+import yaml
+
+import release_notes.tarutil as rnt
+import release_notes.model as rnm
+
+
+def read_release_notes_from_dir(source_dir: str) -> list:
```
8R0WNI3, line 11: I cannot find any usages of this function, where is it intended to be used?

release_notes/archive.py
```diff
@@ -0,0 +1,69 @@
...
+                    release_notes=release_notes_entries,
+                )
+            )
+    return docs
+
+
+def persist_release_notes(
+    notes: list[rnm.ReleaseNotesDoc],
+    base_path: str = '.',
+    archive_name: str = 'release-notes-archive.tar.gz',
```
8R0WNI3, line 47: Right now the tarstream is not gzip-compressed, so we should not state it in the file-ending (or we should add the gzip-compression).

release_notes/archive.py
```diff
@@ -0,0 +1,69 @@
...
+    archive_path = os.path.join(base_path, archive_name)
+
+    with open(archive_path, 'wb') as f:
+        for chunk in rnt.release_notes_docs_into_tarstream(notes):
+            f.write(chunk)
+
+    return archive_path
+
+
+def extract_release_notes(
```
8R0WNI3, line 58: I can't find any usages of this function as well 🤔

release_notes/ocm.py
```diff
@@ -154,3 +155,28 @@ def release_notes_range_recursive(
             component_descriptor_lookup=component_descriptor_lookup,
             absent_ok=True,
         )
+
+
+def add_release_notes_resource(
+    component: ocm.Component,
+    archive_path: str,
+    resource_name: str = 'release-notes-archive',
+    media_type: str = 'application/gzip',
```
8R0WNI3, line 164: ditto: We don't use gzip-compression (yet). Also, since it is a tar-archive, the media-type should rather be `application/tar+gzip`.

release_notes/ocm.py
```diff
@@ -154,3 +155,28 @@ def release_notes_range_recursive(
...
+):
+    logger.info(f'adding release-notes {resource_name=} from {archive_path}')
+
+    resource = ocm.Resource(
+        name=resource_name,
+        version=component.version,
+        type=ocm.ArtefactType.BLOB,
+        relation=ocm.ResourceRelation.LOCAL,
+        access=ocm.LocalBlobAccess(
+            localReference=os.path.abspath(archive_path),
```
8R0WNI3, line 174: Unfortunately, specifying a path to a local file is not working here (I know with GHA and the `export-ocm-fragments` actions we kind-of have these semantics). 🙈 In this case, the `localReference` refers to the digest of the OCI blob (which has to be uploaded separately). See for example the implementation of the "old" release notes resource ([ref](https://github.com/gardener/cc-utils/blob/33950cf7c88d4c00ecd02ea2c5980fd86773ac18/concourse/steps/release.mako#L494)) (which we should also remove when we switch to this "new" resource).

release_notes/ocm.py
```diff
@@ -154,3 +155,28 @@ def release_notes_range_recursive(
...
+            mediaType=media_type
+        ),
+    )
+    existing = [r for r in component.resources if r.name == resource_name]
```
8R0WNI3, line 178: I think we should not only check for the name to identify a resource, however, why is this even required?

8R0WNI3 (changes requested):

release_notes/archive.py
TuanAnh17N: you are right 🤔 this function is currently unused and i think not needed (anymore). 

i likely just forgot to delete it

release_notes/model.py
```diff
@@ -415,54 +415,19 @@ def _source_component(
...
-
-    return ReleaseNote(
-        source_block=source_block,
-        author=author,
-        reference=ref,
-        source_component=source_component,
-        is_current_repo=is_current_repo,
-        from_same_github_instance=from_same_github_instance
+) -> ReleaseNoteEntry:
+    return ReleaseNoteEntry(
```
ccwienk, line 421: since this function now almost exclusively wraps the initialiser, how about, well, using the initialiser (inline)? As an alternative, you might mv this either to SourceBlock (`as_release_note_entry`), or as a factory-method to RNE

wdyt?

release_notes/model.py
```diff
@@ -415,54 +415,19 @@ def _source_component(
...
+        mimetype='text/markdown',
+        contents=source_block.note_message,
+        category=ReleaseNotesCategory(source_block.category.lower()),
+        audience=ReleaseNotesAudience(source_block.target_group.lower()),
+        author=ReleaseNotesAuthor(
+            hostname='github.com',
```
ccwienk, line 427: let us not hardcode this, but instead pass it in. In our intended usages, we should always have means to learn about the github-url

release_notes/utils.py
```diff
@@ -272,3 +276,30 @@ def request_pull_requests_from_api(
                            f'{pending.keys()} is/are either not closed or cannot be found')
 
     return result
+
+
+def release_notes_files_into_docs(
+    repo_dir: str,
```
ccwienk, line 282: you might default this to `.` / PWD - this would fit into how we built our github-actions..

release_notes/utils.py
```diff
@@ -272,3 +276,30 @@ def request_pull_requests_from_api(
...
+    rel_path: str='.ocm/release-notes',
+) -> list[rnm.ReleaseNotesDoc]:
+    docs_dir = os.path.join(
+        repo_dir,
+        rel_path,
+    )
+    return read_release_notes_from_dir(docs_dir)
```
ccwienk, line 289: since this function, again, basically just wraps rrnfd, why not drop it.

we could simply keep accepting `release_notes_dir` as single parameter, and default it to `.ocm/release-notes`; I think there is limited value from accepting two parameters. It should be okay to make it the caller's responsibility to tell us the directory (-> separation of concern)

release_notes/utils.py
```diff
@@ -272,3 +276,30 @@ def request_pull_requests_from_api(
...
+
+
+def read_release_notes_from_dir(
+    release_notes_docs_dir: str
+) -> collections.abc.Generator[rnm.ReleaseNotesDoc, None, None]:
```
ccwienk, line 294: nit: use Iterable as typehint

release_notes/utils.py
```diff
@@ -272,3 +276,30 @@ def request_pull_requests_from_api(
...
+    for fname in os.listdir(release_notes_docs_dir):
```
ccwienk, line 295: I think - depending on discussion in @8R0WNI3 's pullrequest - we might rather:
- define a suffix (e.g. `.release-notes.yaml`)
- os.walk (i.e.: recursively process all matching regular files)

ccwienk (changes requested):

.github/actions/release/action.yaml
```diff
@@ -250,30 +252,34 @@ runs:
           version=component.version,
         )
 
+        local_release_notes_docs = list(
```
ccwienk, line 255: nit: is it actually needed to create a list?

concourse/steps/release.mako
```diff
@@ -469,31 +468,31 @@ except:
...
-    data=release_notes_octets,
+    digest=tar_digest,
+    octets_count=len(tar_stream),
+    data=tar_stream,
   )
 
   component.resources.append(
     ocm.Resource(
-      name=release_notes.ocm.release_notes_resource_name,
+      name='release-notes-archive',
```
ccwienk, line 489: I think we should keep publishing "old" release-notes in parallel for the time being - at least for a short transitioning phase - wdyt?

test/release_notes.py
```diff
@@ -0,0 +1,177 @@
...
+
+import yaml
+
+import ocm
+import concourse.steps.release as csr
+import release_notes.model as rnm
+import release_notes.tarutil as rnt
+
+
+class DummyOciClient:
```
ccwienk, line 14: [magic-mock](https://docs.python.org/3/library/unittest.mock.html) might be useful..

test/release_notes.py
```diff
@@ -0,0 +1,177 @@
...
+    assert len(oci_client.uploads) == 1
+
+    assert len(component.resources) == 1
+
+
+def test_collect_release_notes_merges_sources(monkeypatch):
+    component = DummyComponent()
+    dummy_component_descriptor = DummyComponentDescriptor(component)
+
+    with tempfile.TemporaryDirectory() as tmpdir:
```
ccwienk, line 121: use  [tempdir fixture](https://docs.pytest.org/en/6.2.x/fixture.html)

test/release_notes.py
```diff
@@ -0,0 +1,177 @@
...
+            lambda **kwargs: [fetched_doc]
+        )
+        monkeypatch.setattr(
+            'release_notes.ocm.release_notes_range_recursive',
+            lambda **kwargs: [
+                (ocm.ComponentIdentity(name='github.com/example/dep', version='0.1.0'),
+                 'Subcomponent fix')
+            ]
+        )
+        monkeypatch.setattr(
```
ccwienk, line 148: I think mocking so many functions is not a good pattern. To me, this is an indication that we should either re-organise our code (e.g. do dependency-injection where missing), extend the testing-setup, or re-organise the tests

ccwienk (changes requested):

release_notes/utils.py
TuanAnh17N: thx :)

release_notes/model.py
```diff
@@ -132,9 +129,47 @@ def as_markdown(self) -> str | None:
...
+
+            for entry in entries:
+                entry: ReleaseNoteEntry
+                author = f'@{entry.author.username}'
+                audience = entry.audience
+                pullrequest = entry.pullrequest
+                block_lines.append(
+                    f'```{category} {audience} {pullrequest} {author}\n'
+                    f'{entry.contents.strip()}\n'
+                    f'```'
```
8R0WNI3, line 167: I think (for now) we should re-use existing format of release-notes' markdown, that being said:
- `header` would become a level-1 header -> leading `#`
- `title` would become a level-2 header (instead of level-3 like above) -> leading `##` and a trailing extra `\n`
- the block-line is structured like this:
```python
f'- `[{audience.name}]` {entry.contents.strip()} by {author} [{pullrequest}]'
```

release_notes/fetch.py
```diff
@@ -302,29 +302,29 @@ def fetch_draft_release_notes(
...
     component: ocm.Component,
-    component_descriptor_lookup: cnudie.retrieve.ComponentDescriptorLookupById,
     version_lookup: cnudie.retrieve.VersionLookupByComponent,
     git_helper: gitutil.GitHelper,
     github_api_lookup: rnu.GithubApiLookup,
     version_whither: str | None = None,
     version_whence: str | None = None,
     version_whither_ref_commit: git.Commit | None = None,
-) -> set[rnm.ReleaseNote]:
+) -> rnm.ReleaseNotesDoc | None:
```
8R0WNI3, line 327: In the cases were an empty set `set()` is returned, we should now return `None` instead.

release_notes/fetch.py
```diff
@@ -251,12 +251,11 @@ def _determine_blocks_to_include(
 
 def fetch_draft_release_notes(
     component: ocm.Component,
-    component_descriptor_lookup: cnudie.retrieve.ComponentDescriptorLookupById,
     version_lookup: cnudie.retrieve.VersionLookupByComponent,
     git_helper: gitutil.GitHelper,
     github_api_lookup: rnu.GithubApiLookup,
     version_whither: str,
-) -> set[rnm.ReleaseNote]:
+) -> rnm.ReleaseNotesDoc | None:
```
8R0WNI3, line 258: I this function is not a case where `None` is returned 🤔

concourse/steps/release.py
```diff
@@ -141,66 +140,40 @@ def collect_release_notes(
...
-) -> tuple[str, str]:
-    release_note_blocks = release_notes.fetch.fetch_release_notes(
+) -> tuple[list[rnm.ReleaseNotesDoc], list[rnm.ReleaseNotesDoc]]:
+    repo_path = git_helper.repo_path
+    local_release_notes_path = os.path.join(repo_path, '.ocm/release-notes')
+
+    local_release_notes_docs = []
+    if os.path.isdir(local_release_notes_path):
+        local_release_notes_docs = list(
+            rnu.read_release_notes_from_dir(local_release_notes_path)
```
8R0WNI3, line 151: nit: `rnu.read_release_notes_from_dir` already returns an empty iterable in case the directory does not exist, hence there is no need to do this extra check right here.

concourse/steps/release.py
```diff
@@ -141,66 +140,40 @@ def collect_release_notes(
...
+        )
+
+    fetched_release_notes_docs = release_notes.fetch.fetch_release_notes(
```
8R0WNI3, line 154: `fetch_release_notes` (now) only returns a single `ReleaseNotesDoc` (or `None`) and not an iterable.

concourse/steps/release.py
```diff
@@ -141,66 +140,40 @@ def collect_release_notes(
...
-    else:
-        full_release_notes_markdown = release_notes_markdown
-
-    if (component_resources_markdown := release_notes.markdown.release_note_for_ocm_component(
-        component=component,
-    )):
-        full_release_notes_markdown += '\n\n' + component_resources_markdown
+    component_release_notes_docs = [
+        doc for doc in grouped_release_notes_docs
+        if doc.component_id == component_id
```
8R0WNI3, line 169: There _must_ only ever be exactly one `ReleaseNotesDoc` now with matching `component_id` (because we grouped just before), so rather not use a list right here and we might also raise in case multiple `ReleaseNotesDoc` match the `component_id` (this indicates a bug).

concourse/steps/release.py
```diff
@@ -141,66 +140,40 @@ def collect_release_notes(
...
-
-    sub_component_release_notes_docs = list(rno.release_notes_for_subcomponents(
-        whence_component=whence_component,
-        whither_component=component,
-        component_descriptor_lookup=component_descriptor_lookup,
-        version_lookup=version_lookup,
-        oci_client=oci_client,
-        version_filter=version.is_final,
-    ))
+    all_release_notes_docs = local_release_notes_docs + fetched_release_notes_docs
```
8R0WNI3, line 162: For the `local_release_notes_docs`, it might be the case that the `ocm` property is not filled (or only partially filled -> no version) because those might be created by a developer without specifying the `ocm` properties. In both cases (not filled or partially filled), we should set the `ocm` property to the current component-identity (because for subcomponents it is always filled by the update-deps jobs).

concourse/steps/release.py
```diff
@@ -141,66 +140,40 @@ def collect_release_notes(
     git_helper: GitHelper,
     release_version: str,
     component: ocm.Component,
-    component_descriptor_lookup: ocm.ComponentDescriptorLookup,
     version_lookup: ocm.VersionLookup,
-    oci_client: oci.client.Client,
-) -> tuple[str, str]:
-    release_note_blocks = release_notes.fetch.fetch_release_notes(
+) -> tuple[list[rnm.ReleaseNotesDoc], list[rnm.ReleaseNotesDoc]]:
```
8R0WNI3, line 144: As mentioned, the return value might rather be `tuple[rnm.ReleaseNotesDoc, list[rnm.ReleaseNotesDoc]]`.

concourse/steps/draft_release.mako
```diff
@@ -102,16 +96,15 @@ git_helper = gitutil.GitHelper(
     ),
 )
 try:
-    release_note_blocks = release_notes.fetch.fetch_draft_release_notes(
+    release_notes_docs = release_notes.fetch.fetch_draft_release_notes(
```
8R0WNI3, line 99: We also have to consider the "local" release notes (from the `.ocm/release-notes` directory here as well. We might just be able to reuse `concourse.steps.release.collect_release_notes` (need to check) 🤔

concourse/steps/release.mako
```diff
@@ -469,13 +470,35 @@ except:
   logger.warning('release will continue')
   traceback.print_exc()
 % else:
-release_notes_md = None
-full_release_notes_md = None
+release_notes_docs = []
+subcomponent_release_notes_docs = []
```
8R0WNI3, line 474: nit: What about moving this initialisation prior to `% if process_release_notes:`? That way, we don't have to take care of this initialisation twice.

concourse/steps/release.mako
```diff
@@ -469,13 +470,35 @@ except:
...
 % endif
 
 tgt_ref = cnudie.util.target_oci_ref(component=component)
 
-if release_notes_md:
+<%
```
8R0WNI3, line 479: Why did you place this within a `<% ... %>` block?

concourse/steps/release.mako
```diff
@@ -469,13 +470,35 @@ except:
...
+    release_notes_docs,
+  )
+
+  component_resources_markdown = release_notes.markdown.release_note_for_ocm_component(
+    component=component,
+  )
+  if component_resources_markdown:
+    release_notes_md = f'{release_notes_md}\n\n{component_resources_markdown}'
+
+  subcomponent_md = rno.release_notes_docs_as_markdown(
```
8R0WNI3, line 490: `rno` is not defined here, either use full `release_notes.ocm` or introduce `rno` alias.

concourse/steps/release.mako
```diff
@@ -469,13 +470,35 @@ except:
...
+<%
+  release_notes_md = release_notes.ocm.release_notes_docs_as_markdown(
+    release_notes_docs,
+  )
+
+  component_resources_markdown = release_notes.markdown.release_note_for_ocm_component(
+    component=component,
+  )
+  if component_resources_markdown:
+    release_notes_md = f'{release_notes_md}\n\n{component_resources_markdown}'
```
8R0WNI3, line 488: Prior to this change, the `component_resources_markdown` was only added to `full_release_notes_md` and also _after_ adding the release notes for the subcomponents. I think we should keep this stable.

concourse/steps/release.mako
```diff
@@ -469,13 +470,35 @@ except:
...
+    prepend_title=False,
+  )
+
+  if subcomponent_md:
+    full_release_notes_md = f'{release_notes_md}\n{subcomponent_md}'
+  else:
+    full_release_notes_md = release_notes_md
+%>
+
+% if release_notes_md:
```
8R0WNI3, line 501: I think if you decided to use the mako `% if` (which will cause this block to be conditionally rendered), you must not indent it afterwards (as opposed to using a traditional Python `if`).

concourse/steps/release.mako
```diff
@@ -497,8 +520,37 @@ if release_notes_md:
...
+  oci_client.put_blob(
+    image_reference=tgt_ref,
+    digest=tar_digest,
+    octets_count=len(tar_stream),
+    data=tar_stream,
+  )
+
+  component.resources.append(
+    ocm.Resource(
+      name='release-notes-archive',
```
8R0WNI3, line 542: Rather use `release_notes.ocm.release_notes_resource_name` to have a central place to configure the resource name.

.github/actions/release/action.yaml
```diff
@@ -243,6 +247,7 @@ runs:
       if: ${{ inputs.release-notes }}
       shell: python
       run: |
+        import base64
```
8R0WNI3, line 250: We must also update the `if` statement of this step to also run if only the release-notes-tar is provided (+ add handling for one of both values to be not set within this step) or instead use separate steps for the legacy and the new release notes handling.

.github/actions/release/action.yaml
```diff
@@ -292,6 +297,29 @@ runs:
...
+        oci_client.put_blob(
+            image_reference=tgt_oci_ref,
+            digest=tar_digest,
+            octets_count=len(tar_bytes),
+            data=tar_bytes,
+        )
+
+        component.resources.append(
+            ocm.Resource(
+                name='release-notes-archive',
```
8R0WNI3, line 313: ditto: Use `release_notes.ocm.release_notes_resource_name`

.github/actions/release/action.yaml
```diff
@@ -292,6 +297,29 @@ runs:
...
+                version=component.version,
+                type='vnd.gardener/release-notes-archive',
```
8R0WNI3, line 315: We should use a common type here between Concourse and GHA.

.github/actions/release-notes/release_notes_cli.py
```diff
@@ -253,12 +258,19 @@ def github_api_lookup(repo_url):
...
     if parsed.subcomponent_release_notes:
         with open(parsed.subcomponent_release_notes, 'w') as f:
             f.write(sub_component_release_notes)
 
+    if parsed.tar_output:
+        tar_stream = b''.join(rnt.release_notes_docs_into_tarstream(
+            release_notes_docs + grouped_sub_component_release_notes_docs
+        ))
+        with open(parsed.tar_output, 'wb') as f:
+            f.write(tar_stream)
```
ccwienk, line 272: I suggest to use streaming, considering we get a generator in the first place:

```
  for chunk in ...:
    f.write(chunk)
```

It will very likely not be an issue to read release-notes into memory, it is not a nice pattern to do so IMO.

.github/actions/release-notes/action.yaml
```diff
@@ -96,6 +100,7 @@ runs:
           --full-release-notes full-release-notes.md \
           --local-release-notes release-notes.md \
           --subcomponent-release-notes subcomponent-release-notes.md ${extra_args:-}
+          --tar-output release-notes.tar
```
8R0WNI3, line 103: missing `\` at the end of the above line

8R0WNI3 (changes requested):

release_notes/model.py
```diff
@@ -132,9 +129,47 @@ def as_markdown(self) -> str | None:
...
 
         header = f'[{self.ocm.component_name}:{self.ocm.component_version}]'
-        release_notes = '\n'.join(release_note.contents for release_note in self.release_notes)
 
-        return f'{header}\n{release_notes}'
+        prerendered_entries = []
+        categorized_entries = collections.defaultdict(list)
+
+        for entry in self.release_notes:
+            if entry.type == ReleaseNotesType.PRERENDERED:
```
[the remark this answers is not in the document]
TuanAnh17N: ffs 😂

concourse/steps/release.mako
```diff
@@ -497,8 +520,37 @@ if release_notes_md:
...
+      version=component.version,
+      type=ocm.ArtefactType.BLOB,
+      access=ocm.LocalBlobAccess(
+        localReference=tar_digest,
+        size=len(tar_stream),
+        mediaType='application/tar',
```
ccwienk, line 548: should we consider to compress (e.g. gzip) release-notes? I think release-notes will be fairly small; otoh it is a quite common practise 🤔

release_notes/fetch.py
```diff
@@ -333,7 +336,7 @@ def fetch_release_notes(
     :param version_whence: Optional argument to retrieve release notes starting at a specific \
         version. If not given, the closest version to `version_whither` is used.
 
-    :return: A set of ReleaseNote objects for the specified component.
+    :return: A set of ReleaseNotesDoc objects for the specified component.
```
8R0WNI3, line 339: nit: This is not a set anymore.

Also, please return `None` instead in the places where still an empty set is returned.

concourse/steps/release.py
```diff
@@ -141,66 +140,52 @@ def collect_release_notes(
...
-    if sub_component_release_notes:
-        full_release_notes_markdown = f'{release_notes_markdown}\n{sub_component_release_notes}'
-    else:
-        full_release_notes_markdown = release_notes_markdown
+    component_matches = [
+        doc for doc in grouped_release_notes_docs
+        if doc.component_id == component_id
+    ]
+    if len(component_matches) > 1:
+        raise RuntimeError(f'Multiple ReleaseNotesDoc found for {component_id=}')
```
8R0WNI3, line 180: nit: You might state in the error that this indicates a programming bug (as we grouped before).

concourse/steps/release.mako
```diff
@@ -451,11 +452,11 @@ for validation_error in cnudie.validate.iter_violations(
...
   logger.warning(f'{validation_error=}')
 
+release_notes_docs = []
+subcomponent_release_notes_docs = []
 % if process_release_notes:
-release_notes_md = None
-full_release_notes_md = None
 try:
-  release_notes_md, full_release_notes_md = collect_release_notes(
+  release_notes_docs, subcomponent_release_notes_docs = collect_release_notes(
```
8R0WNI3, line 459: `release_notes_docs` is not a list but only the single `release_notes_doc` for the "current" component.

concourse/steps/release.py
```diff
@@ -141,66 +140,52 @@ def collect_release_notes(
...
+            not doc.ocm
+            or not doc.ocm.component_name
+            or not doc.ocm.component_version
+        ):
+            doc.ocm = rnm.ReleaseNotesOcmRef(
+                component_name=component.name,
+                component_version=component.version,
+            )
+
+    fetched_release_notes_doc = release_notes.fetch.fetch_release_notes(
```
8R0WNI3, line 162: `fetch_release_notes` might also return `None`.

.github/actions/release-notes/release_notes_cli.py
```diff
@@ -168,20 +174,18 @@ def github_api_lookup(repo_url):
         return github_api
 
     try:
-        release_notes_md = ''
+        release_notes_docs = []
         if parsed.draft:
-            release_note_blocks = release_notes.fetch.fetch_draft_release_notes(
+            release_notes_docs = release_notes.fetch.fetch_draft_release_notes(
```
8R0WNI3, line 179: What is actually the difference between `fetch_draft_release_notes` and `fetch_release_notes`? Do we still need both? /cc @ccwienk

.github/actions/release-notes/release_notes_cli.py
```diff
@@ -168,20 +174,18 @@ def github_api_lookup(repo_url):
...
                 component=component,
-                component_descriptor_lookup=component_descriptor_lookup,
                 version_lookup=ocm_version_lookup,
                 git_helper=git_helper,
                 github_api_lookup=github_api_lookup,
                 version_whither=component.version,
             )
         else:
-            release_note_blocks = release_notes.fetch.fetch_release_notes(
+            release_notes_docs = release_notes.fetch.fetch_release_notes(
```
8R0WNI3, line 187: Both `fetch_*` functions (now) return a single `release_notes_doc` or `None`.

.github/actions/release-notes/release_notes_cli.py
```diff
@@ -241,9 +246,9 @@ def github_api_lookup(repo_url):
         sub_component_release_notes = None
 
     if sub_component_release_notes:
```
8R0WNI3, line 248: We also have to consider the release-notes from local directory here instead of retrieving them again via `release_notes_for_subcomponents`(similar to how it is done in `concourse.steps.release.collect_release_notes`. We might just be able to move said function to `release_notes` module and re-use it for both Concourse and GHA.

8R0WNI3 (changes requested):

release_notes/fetch.py
```diff
@@ -417,14 +355,79 @@ def fetch_release_notes(
...
+        version_whither_ref_commit=version_whither_ref_commit,
+        is_draft=is_draft,
+    ):
+        release_notes_docs.append(fetched_release_notes_doc)
+
+    grouped_release_notes_docs = rno.group_release_notes_docs(release_notes_docs)
+
+    component_id = component.identity()
+
+    component_matches = [
```
ccwienk, line 414: naming-nit: is is more expressive to e.g. name this matching_release_notes (xxx_matches will rather imply a boolean value - or a list of matching components)

PR #1320, opened 2025-07-29
Omit padding if not required

zkdev:
tar spec: https://www.gnu.org/software/tar/manual/html_node/Standard.html

A tar archive contains a sequence of blocks with a fixed amount of bytes per block (BLOCKSIZE). Thus, we add empty bytes as padding to fill the blocks.
The expression `if (missing := tarfile.BLOCKSIZE - (uploaded_bytes % tarfile.BLOCKSIZE)):` also evaluates to true if `uploaded_bytes` is already of BLOCKSIZE length (or a multiple of it), ultimately creating a block of empty bytes.

This *should* not cause problems, because the end of the tar archive is marked with *two* blocks of empty bytes. However, manual testing unveiled that at least the bsdtar (version 3.5.3) implementation terminates the tar archive already after *one* block of empty bytes.

This leads to a situation where adding an unnecessary padding creates a tar archive that, if extracted using some tar implementations, does not create the expected files.

**Release note**:

```other operator

```


oci/__init__.py
```diff
@@ -556,6 +556,11 @@ def image_layers_as_tarfile_generator(
 
         # need to pad full blocks w/ NUL-bytes
         if (missing := tarfile.BLOCKSIZE - (uploaded_bytes % tarfile.BLOCKSIZE)):
+
+            if missing == tarfile.BLOCKSIZE:
```
8R0WNI3, line 560: Great catch! :-)

I think we will also have to add this case to `tarutil.concat_blobs_as_tarstream` as well as recently introduced `release_notes.tarutil.release_notes_docs_into_tarstream`.

ccwienk: I think this would be nicer to read like so (also, no additional if required):

```
if (remainder := uploaded_bytes % tarfile.BLOCKSIZE):
  missing = tarfile.BLOCKSIZE - remainder
```

If it is needed elsewhere, we could consider factoring it into a function (but I guess this is too much overhead if it is just two locations).

zkdev:
> how about taking the first commit (the fix), and already release it, and split the refactoring into a separate activity?

wanted to avoid patching multiple occasions just to refactor directly after

PR #1322, opened 2025-07-30
let trusted-checkout honour "ok-to-test"-label

ccwienk:
For security-reasons, pullrequests from untrusted sources (such as forked repositories) are either run with limited permissions (if using `on.pull_request` as event-source), or will by default not checkout changes from pullrequest.

Trusted-Checkout-Action was introduced to allow acceptably secure privileged runs for pullreqeust from trusted authors.

However, this does not work well in cases of e.g. new contributors. Hence, implement (again, as in place for concourse) a label `reviewed/ok-to-test`, which, if set, will have the same effect (trust the pullrequest).

For this option to work reasonably well, event-trigger should be changed like so:

```
on:
  pull_request_target:
    types:
      - labeled
```

This will restrict triggering to labelling-events. In order to avoid runs being triggered from labels different than `reviewed/ok-to-test`, jobs need to be fitted w/ a filtering-condition, in addition (this cannot be handled on event-trigger-level, unfortunately).

**Release note**:

```feature developer
let trusted-checkout honour "ok-to-test"-label
```

### DIFF
--- a/.github/actions/trusted-checkout/action.yaml
+++ b/.github/actions/trusted-checkout/action.yaml
@@ -25,6 +25,19 @@ inputs:
       passed to `actions/checkout`'s as `fetch-depth` input.
 
       useful for cases where longer history is required.
+  trusted-label:
+    type: string
+    default: 'reviewed/ok-to-test'
+    description: |
+      a label, that can be used to mark pullrequests as trusted, regarless of author's association.
+
+      For security-reasons, this action will remove the label, such that it will have to be
+      added again (adding the label will work as trigger).
+  needs-trusted-label:
+    type: string
+    default: 'needs/ok-to-test'
+    description: |
+      a label that is added as a marker in order to notify lack of "trusted-label"
 
 runs:
   using: composite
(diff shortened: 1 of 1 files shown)


.github/actions/trusted-checkout/action.yaml
```diff
@@ -25,6 +25,19 @@ inputs:
...
+    type: string
+    default: 'reviewed/ok-to-test'
+    description: |
+      a label, that can be used to mark pullrequests as trusted, regarless of author's association.
+
+      For security-reasons, this action will remove the label, such that it will have to be
+      added again (adding the label will work as trigger).
+  needs-trusted-label:
+    type: string
+    default 'needs/ok-to-test'
```
TuanAnh17N, line 38: missing colon

.github/actions/trusted-checkout/action.yaml
```diff
@@ -49,6 +62,16 @@ runs:
         author_association = '${{ github.event.pull_request.author_association }}'
         allowed_to_checkout = author_association in allowed_author_associations
 
+        # if pullrequest is labelled, it is also considered trusted, regardless of
+        # author's association (rationale: only codeowners can add labels)
+        trusted_label = '${{ inputs.trusted-label }}'
+
+        if '${{ github.event.action }}' == 'labeled':
+          if '${{ github.event.action.label.name }}' == trusted_label:
```
TuanAnh17N, line 70: is this valid? shouldn't this be `github.event.label.name`

.github/actions/trusted-checkout/action.yaml
```diff
@@ -75,3 +98,27 @@ runs:
           ),
           check=True,
         )
+    - if: ${{ steps.calc.outputs.remove-label }}
+      uses: gardener/cc-utils/.github/actions/install-gardener-gha-libs@master
+    - name: remove-trusted-and-add-needs-label
+      if: ${{ steps.calc.outputs.remove-label }}
+      uses: gardener/cc-utils/.github/actions/install-gardener-gha-libs@master
```
TuanAnh17N, line 105: we install `install-gardener-gha-libs` twice under the same condition. is this intentional?

.github/actions/trusted-checkout/action.yaml
ccwienk: I think so:

https://github.com/ccwienk-org/test/actions/runs/16621814737/job/47027804279#step:4:618

.github/actions/trusted-checkout/action.yaml
ccwienk: dang! nope, that's copy-pasta i missed to clenaup - thx

PR #1326, opened 2025-08-01
mv ocm postprocessing

ccwienk:
**Release note**:

```bugfix developer
mv ocm postprocessing into ocm-merge action
```


.github/actions/base-component-descriptor/action.yaml
```diff
@@ -149,7 +157,12 @@ runs:
           creation_time=datetime.datetime.now(),
         )
 
-        if imagevector_file := ocm.gardener.find_imagevector_file(repo_root=os.getcwd()):
+        if '${{ inputs.post-process }}' == 'component-cli':
```
TuanAnh17N, line 160: is it safe to use `${{ inputs.* }}`directly in `shell: python` steps? it doesn't feel quite right to me 🤔

PR #1329, opened 2025-08-04
add "pullrequest-trusted-helper"

ccwienk:
For running privileged pipelines against pullrequests, we use labelling to control which pullrequests are actually allowed to run, in order to prevent execution of malicious pullrequests.

Add a helper-action, and -workflow, that can be used to indicate such label needs to be set, or automatically set it (if the pullrequest-author is trusted).

**Release note**:

```other developer
add "pullrequest-trusted-helper"
```


.github/actions/pullrequest-trust-helper/action.yaml
```diff
@@ -0,0 +1,122 @@
...
+
+        for label in pull_request.labels():
+          if label.name == trusted_label:
+            print(f'{trusted_label=} already present - exiting')
+            exit(0)
+          if label.name == needs_trusted_label:
+            has_needs_trusted_label = False
+
+        if not has_needs_trusted_label:
+          pull_request.add_labels(needs_trusted_label)
```
8R0WNI3, line 94: Is it intended that if `okay_to_add_trusted_label` evaluates to `True` afterwards, both the `needs_trusted_label` as well as the `trusted_label` will be set?

.github/workflows/pullrequest-trust-helper.yaml
```diff
@@ -0,0 +1,55 @@
...
+        default: 'needs/ok-to-test'
+      trusted-teams:
+        description: |
+          an optional, comma-separated list of team-names. Members of any of those teams
+          are considered to be trusted.
+        type: string
+        default: ''
+
+jobs:
+  runs-on: ubuntu-latest
```
TuanAnh17N, line 47: Is runs-on valid without a job name?

.github/actions/pullrequest-trust-helper/action.yaml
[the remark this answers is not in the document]
TuanAnh17N: i think there is a small logic issue. `has_needs_trusted_label` is initially `True`, but it's set to `False` when the label is found. that leads to the label being added when it's already present - which seems unintended

ccwienk: mhh, good point. I also pondered that, and reckoned, it might not hurt to always set the "needed"-label. OTOH, after spending some more thoughts, I think it might be confusing / perceived as a bug.

-> will remove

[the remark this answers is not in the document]
ccwienk: 💡 you are (/ were) correct - thx

PR #1333, opened 2025-08-05
expose published component-descriptor as output

ccwienk:
needed (/helpful) for gardenlogin and gardenctl-v2, where additional publishing-steps are to be done (which will be easiest to be done in separate jobs/workflows).

**Release note**:

```other developer
expose published component-descriptor as output
```

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -143,6 +143,10 @@ on:
           published as part of component-descriptor, as well as via slack, if `slack-channel-id`
           input is specified).
         value: ${{ jobs.release-and-bump.outputs.release-notes }}
+      component-descriptor:
+        description: |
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -153,6 +157,7 @@ jobs:
 
     outputs:
       release-notes: ${{ steps.release-notes.outputs.release-notes }}
+      component-descriptor: ${{ steps.component-descriptor.outputs.component-descriptor }}
```
8R0WNI3, line 160: If we use the component descriptor output of the `component-descriptor` step right here, it won't contain changes made during release action (e.g. attachment of release-notes artefact).

ccwienk: you are correct - I switched to using the release-action's output (and added the modified component-descriptor as an output)

PR #1335, opened 2025-08-06
Fixate commit for pullrequests

ccwienk:
**Release note**:

```other developer
fixate commit for pullrequests in `trusted-checkout` action to avoid race-conditions in conjunction with "label-based trust".
```

### DIFF
--- a/.github/actions/trusted-checkout/action.yaml
+++ b/.github/actions/trusted-checkout/action.yaml
@@ -103,11 +103,14 @@ runs:
           ),
           check=True,
         )
+
+        checkout_ref = '${{ github.event.pull_request.head.sha }}'
+
         subprocess.run(
           (
             'git',
             '-C',  repo_path,
-            'checkout', '-B', '${{ github.head_ref }}', 'FETCH_HEAD',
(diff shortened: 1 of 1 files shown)


.github/actions/trusted-checkout/action.yaml
```diff
@@ -103,11 +103,14 @@ runs:
...
         )
+
+        checkout_ref = '${{ github.event.pull_request.head.sha }}'
+
         subprocess.run(
           (
             'git',
             '-C',  repo_path,
-            'checkout', '-B', '${{ github.head_ref }}', 'FETCH_HEAD',
+            'checkout', '-B', '${{ github.head_ref }}', "${checkout_ref}",
```
8R0WNI3, line 128: This is still evaluated as Python and not as bash, so it should rather be `f'{checkout_ref}'` (or just `checkout_ref`).

ccwienk: you are right- I mixed up the used shell <.<
thx

8R0WNI3 (approved):
/lgtm

PR #1336, opened 2025-08-06
render diff as codeblock

ccwienk:
**Release note**:

```other operator
render version-commit-diff as codeblock
```


matthias-horne (approved):
Thank you!

PR #1338, opened 2025-08-07
Fix docker cfg

ccwienk:
**Release note**:

```other operator

```

### DIFF
--- a/.github/workflows/oci-ocm.yaml
+++ b/.github/workflows/oci-ocm.yaml
@@ -279,7 +279,7 @@ jobs:
             exit 0
           fi
           echo 'extra-auths<<EOF' >> "${GITHUB_OUTPUT}"
-          cat <<EOF >> ${GITHUB_OUTPUT}
+          cat <<EOF >> "${GITHUB_OUTPUT}"
           {
             "registry-1.docker.io": {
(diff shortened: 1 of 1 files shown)


.github/workflows/oci-ocm.yaml
```diff
@@ -278,9 +278,24 @@ jobs:
...
+            },
+            "index.docker.io": {
+              "auth": "${{ secrets.DOCKERHUB_RO_AUTH }}"
+            },
+            "https://index.docker.io/v1/": {
+              "auth": "${{ secrets.DOCKERHUB_RO_AUTH }}"
+            }
+          }
           EOF
+          echo EOF >> "${GITHUB_OUTPUT}"
```
TuanAnh17N, line 298: do we need this extra `echo EOF`? the `cat << EOF` block already handles the clsoing marker, no?

ccwienk: hehe, yes, this is needed (I acknowledge, this is a bit confusing)

the "inner" EOF is for the `cat`-command, whereas the "outer" EOF (the one you commented on) is for GitHub-Actions to understand our multiline-output is done

TuanAnh17N: 💡

PR #1341, opened 2025-08-08
fix: do not attempt to remove label if absent

ccwienk:
trusted-checkout will attempt to remove "trusted-label". However, removal requires pull-requests:write-permission. Avoid requiring each caller to grant this permission in cases where label was already removed (by an earlier job).

**Release note**:

```other operator

```


8R0WNI3 (approved):
/lgtm

PR #1351, opened 2025-09-02
Use recommended Bot accounts for GHA based on GH instance

8R0WNI3:
**Release note**:

```other operator

```

### DIFF
--- a/.github/actions/setup-git-identity/action.yaml
+++ b/.github/actions/setup-git-identity/action.yaml
@@ -8,11 +8,11 @@ inputs:
   user_name:
     description: The Git user name
     required: false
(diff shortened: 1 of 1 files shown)


.github/actions/setup-git-identity/action.yaml
```diff
@@ -23,6 +23,32 @@ runs:
...
+              user_id="41898282"
+              ;;
+              github.tools*)
+              user_id="33697"
+              ;;
+              github.wdf*)
+              user_id="127534"
+              ;;
+              *)
+              echo "don't know how to handle GitHub host ${github_host}"
```
ccwienk, line 45: should we add a graceful fallback instead?

PR #1358, opened 2025-09-05
feat(release-notes): Add optional filter to control what is posted to slack

zkdev:
release-notes are categorised and intended for different audiences. As the slack posting is user facing, it makes sense to offer filter options here.

By default, release-notes will not be filtered and only release-notes from the released component itself will be posted.

**Release note**:

```feature developer
Offer optional filters for release-notes posting towards slack
```


release_notes/ocm.py
```diff
@@ -380,3 +381,54 @@ def release_note_for_ocm_component(component: ocm.Component) -> str:
...
+            oci_client=oci_client,
+        ):
+            ocm_ref = release_notes_docs.ocm
+            component_id = ocm.ComponentIdentity(
+                name=ocm_ref.component_name,
+                version=ocm_ref.component_version,
+            )
+
+            if not component_id == rnode.component_id:
+                # release-notes can contain contents for subcomponents, ignore them we will visit
```
8R0WNI3, line 430: If we collect the release notes of subcomponents by retrieving the release notes archive of the subcomponents directly, we will loose changes made in downstream components, e.g. if the release-notes of cc-utils are modified/corrected in the delivery-service, those modifications will be lost.
Also (though neglectable), downloading the release notes archive for each subcomponent individually will require more OCI downloads.

Wdyt about only using the release notes archive of the current component (available as GHA-artefact anyways) (this also includes all recursive release notes of subcomponents), then iterate over the subcomponents via `cnudie.iter.iter` and read the respective release notes for each subcomponent out of this archive?

release_notes/utils.py
```diff
@@ -245,3 +246,38 @@ def request_pull_requests_from_api(
...
     return result
+
+
+def filter_release_notes(
+    release_notes_docs: collections.abc.Iterable[rnm.ReleaseNotesDoc],
+    audiences: collections.abc.Container[rnm.ReleaseNotesAudience] = [],
+    categories: collections.abc.Container[rnm.ReleaseNotesCategory] = [],
+) -> collections.abc.Generator[rnm.ReleaseNotesDoc, None, None]:
+    '''
+    Filter out single release-notes, or entire release-notes documents.
```
8R0WNI3, line 257: nit: As of now, this function won't filter out entire release-notes documents (even if they don't contain any release-notes).

.github/workflows/release.yaml
```diff
@@ -143,6 +143,43 @@ on:
           passed as same-named input to release-notes action. Controls whether copies of
           release-notes of sub-components are included.
 
+      categories:
+        type: string
+        default: ''
+        description: |
+          comma separated list of release-note categories to include in slack posting.
```
ccwienk, line 150: I think it would be nice to mention all categories. for readability, I think it will be preferable to allow linebreak-separates values.

Also, it would be easier to understand the offerered contract by explicitly mentioning it can be configured to select a subset of categories.

Also, `release`-action does not only post release-notes to slack + it accepts a wide range of inputs. By convention (in other such cases), we prefix parameters that are passed to other actions (in this case, e.g. with `release-notes-<name>`) - I think we should keep this pattern.

.github/workflows/release.yaml
```diff
@@ -143,6 +143,43 @@ on:
...
+          defaults to include *all* categories.
+
+          for all known categories, see:
+          https://github.com/gardener/cc-utils/blob/master/release_notes/model.py
+
+          example:
+          value: "breaking,feature"
+
+      audiences:
```
ccwienk, line 159: naming + not only relevant for posting to slack (see above)

.github/workflows/release.yaml
```diff
@@ -143,6 +143,43 @@ on:
...
+          comma separated list of release-note audiences to include in slack posting.
+          defaults to include *all* audiences.
+
+          for all known audiences, see:
+          https://github.com/gardener/cc-utils/blob/master/release_notes/model.py
+
+          example:
+          value: "developer,operator"
+
+      subcomponent-recursion-depth:
```
ccwienk, line 172: ditto

.github/workflows/release.yaml
```diff
@@ -217,12 +254,56 @@ jobs:
...
 
-      - name: write-release-notes-to-file
+      - name: write-filtered-release-notes-to-file
         if: ${{ inputs.slack-channel-id != '' }}
+        shell: python
         run: |
-          cat <<"EOF" > release-notes.md
-          ${{ steps.release-notes.outputs.full-release-notes }}
-          EOF
+          import yaml
```
ccwienk, line 261: generally-speaking, I think the filtering should be done in `release-notes`-action.

-> pass filtering-arguments to release-notes-action, propagate whatever release-notes action returns. This will also allow us to convey such parameters differently (e.g. via cfg-file in repository, or via cfg-artefact)

release_notes/ocm.py
```diff
@@ -380,3 +381,54 @@ def release_note_for_ocm_component(component: ocm.Component) -> str:
         component_release_notes += category_markdown
 
     return component_release_notes
+
+
+def iter_release_notes_docs(
```
ccwienk, line 386: there is already `release_notes_for_subcomponents`. I think it will make more sense to re-use this function, and filter the returned ReleaseNotesDoc-iterable (recursion-depth is already prepared as a non-exposed parameter, there).

ccwienk (changes requested):

release_notes/utils.py
```diff
@@ -245,3 +246,38 @@ def request_pull_requests_from_api(
...
+    Follows include and exclude semantics, whereas absence of filter values disables filtering.
+    It is sufficient if any matches, however exclude has precedence.
+    Filter list values allow regex and use `fullmatch` semantics.
+    '''
+    for release_notes_doc in release_notes_docs:
+        filtered_release_notes = []
+
+        for release_note in release_notes_doc.release_notes:
+            category_filter = reutil.re_filter(
+                include_regexes=categories,
```
ccwienk, line 268: why use regular expressions, when we are actually searching for exact strings / StrEnum-Values even?

release_notes/utils.py
```diff
@@ -245,3 +246,38 @@ def request_pull_requests_from_api(
                            f'{pending.keys()} is/are either not closed or cannot be found')
 
     return result
+
+
+def filter_release_notes(
+    release_notes_docs: collections.abc.Iterable[rnm.ReleaseNotesDoc],
+    audiences: collections.abc.Container[rnm.ReleaseNotesAudience] = [],
```
ccwienk, line 253: do not use mutable values as default values

release_notes/utils.py
zkdev: agree, we don't need regex (anymore)
this stems from first approach where filtering component names was supported

release_notes/ocm.py
8R0WNI3: > how is that possible?
release-notes are attached to the corresponding OCM, which is immutable.

That's correct, the release-notes which are stored in the OCM component descriptor of subcomponents cannot be updated in the OCM component descriptor of said subcomponent after release. However, _it is possible_ to update release-notes of subcomponents in a downstream component (prior to the release of that downstream component) which will result in those updated release-notes to be included in the release-notes archive of the downstream component.

> Planned on doing that first, however this is limited as only direct sub components are included in the archive. Thus, for recursion-depth > 1, we would have to go via OCI anyways.

No, the release-notes archive must always included _all_ release-notes of _all_ recursive subcomponents.

zkdev: we had a quick discussion and clarified how recursive release-notes are transported via release-notes archive without fetching OCI, also the modification part is now clear 👍

.github/workflows/release.yaml
zkdev: > not only relevant for posting to slack

We discussed that release-notes filter should only affect the slack posting, not the release-notes published via OCM. Thus the inpus must only affect the action(s) posting to slack.

What am I missing here?

release_notes/utils.py
```diff
@@ -245,3 +245,35 @@ def request_pull_requests_from_api(
...
+    release_notes_doc: rnm.ReleaseNotesDoc,
+    audiences: collections.abc.Container[rnm.ReleaseNotesAudience] = (),
+    categories: collections.abc.Container[rnm.ReleaseNotesCategory] = (),
+) -> rnm.ReleaseNotesDoc:
+    '''
+    Filter out single release-notes.
+
+    Follows include and exclude semantics, whereas absence of filter values disables filtering.
+    It is sufficient if any matches, however exclude has precedence.
+    Filter list values allow regex and use `fullmatch` semantics.
```
8R0WNI3, line 260: This comment is not accurate anymore :-)

.github/actions/release-notes/action.yaml
```diff
@@ -98,38 +109,108 @@ runs:
...
+        import cnudie.retrieve
+        import ocm
+        import oci.auth
+        import oci.client
+        import release_notes.ocm as rno
+        import release_notes.utils as rnu
+
+        component_descriptor_content = '${{ inputs.component-descriptor }}\n'
+        with open('/tmp/component-descriptor.yaml', 'w') as f:
+          f.write(component_descriptor_content)
```
8R0WNI3, line 147: What is the advantage of using a temporary file here?

.github/actions/release-notes/action.yaml
```diff
@@ -98,38 +109,108 @@ runs:
...
+        release_notes_docs_dir = '/tmp/release-notes-dir'
+
+        with tarfile.open('/tmp/release-notes.tar', 'r') as tar:
+          tar.extractall(path=release_notes_docs_dir)
+
+        all_release_notes_docs = rno.read_release_notes_from_dir(
+          release_notes_docs_dir=release_notes_docs_dir,
+        )
+
+        for variant in ('github', 'slack'):
```
8R0WNI3, line 177: Shouldn't we iterate over all specified variants, and not just these hardcoded ones?

.github/actions/release-notes/action.yaml
```diff
@@ -98,38 +109,108 @@ runs:
...
-          if [ $( stat -c%s subcomponent-release-notes.md) -gt 0 ]; then
-            echo '## Release-Notes (subcomponents)' >> ${GITHUB_STEP_SUMMARY}
-            cat subcomponent-release-notes.md >> ${GITHUB_STEP_SUMMARY}
-          fi
-        fi
     - uses: gardener/cc-utils/.github/actions/upload-artifact@master
       with:
-        name: release-notes
-        path: release-notes.tar
+        name: github-release-notes
```
8R0WNI3, line 210: We could also write all variants into an archive again, that way we would not require an individual upload action for each variant and we could also dynamically configure additional variants.

.github/actions/release-notes/action.yaml
```diff
@@ -34,16 +34,27 @@ inputs:
     default: false
     description: |
       if true, will collect draft-release-notes
-outputs:
```
8R0WNI3, line 37: I suppose we can also drop the `include-subcomponent-release-notes` input now, no?

.github/actions/release/action.yaml
```diff
@@ -274,7 +265,9 @@ runs:
           credentials_lookup=oci.auth.docker_credentials_lookup(),
         )
 
-        release_notes_markdown = '''${{ inputs.release-notes }}'''
+        with open('/tmp/release-notes.md') as f:
```
8R0WNI3, line 268: This will require download of `legacy-release-notes` artifact first I think 🤔

.github/workflows/release.yaml
```diff
@@ -193,6 +271,16 @@ jobs:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           draft: false
           include-subcomponent-release-notes: ${{ inputs.include-subcomponent-release-notes }}
+          release-notes-variants: |
```
8R0WNI3, line 274: I think we should also migrate the `post-build.yaml` workflow to these variants, wdyt?

zkdev: wasn't aware of that one, will adjust. thanks.

.github/actions/release-notes/action.yaml
```diff
@@ -34,16 +34,27 @@ inputs:
...
+      Used to create different subsets of release-notes.
+      A variant configuration is provided as flat yaml object, whereas the toplevel attribute is
+      considered the variant-name.
+      For each variant, the filtered and rendered markdown is written to
+      `/tmp/<variant-name>-release-notes.md`.
+
+      Expected structure (see release workflow inputs for allowed values):
+      ```
+      my-variant:
+        audiences: 'developer,operator'
```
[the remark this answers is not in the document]
zkdev: What do you mean with "downstream actions/workflows"?
cc-utils action and workflows will be adjusted consistently with this PR.

.github/actions/release-notes/release_notes_cli.py
```diff
@@ -237,14 +287,67 @@ def github_api_lookup(repo_url):
...
+            lookup=component_descriptor_lookup,
+            recursion_depth=variant_cfg.recursion_depth,
+            node_filter=cnudie.iter.Filter.components,
+        ):
+            for rn in all_release_note_docs:
+                if rn.component_id == cnode.component_id:
+                    filtered_release_notes_docs.append(rn)
+
+        release_notes_docs = [
+            rnu.filter_release_notes(
```
8R0WNI3, line 319: nit: You might add this filter function call into the above "append" statement, that way we don't have to iterate twice :-)

.github/actions/release-notes/release_notes_cli.py
```diff
@@ -237,14 +287,67 @@ def github_api_lookup(repo_url):
...
+                categories=variant_cfg.categories,
+            )
+            for release_notes_doc in filtered_release_notes_docs
+        ]
+
+        release_notes_markdown = rno.release_notes_docs_as_markdown(
+            release_notes_docs=release_notes_docs,
+        )
+
+        if not release_notes_markdown.endswith('\n'):
```
8R0WNI3, line 331: I think we have some subsequent steps, which check for release-notes markdown to be true/falsy. As this will never return an empty markdown, those checks will always evaluate to true. So rather extend this check to `if release_notes_markdown and not release_notes_markdown.endswith('\n')`

.github/workflows/release.yaml
8R0WNI3: This is still required before merging as `release-notes` action now defines a new interface.

.github/actions/release/action.yaml
```diff
@@ -243,14 +234,20 @@ runs:
       uses: gardener/cc-utils/.github/actions/download-artifact@master
       with:
         name: release-notes
-        path: /tmp
+        path: /tmp/release-notes.tar
+
+    - name: extract-release-notes
+      shell: base
```
8R0WNI3, line 240: bash?

.github/actions/release-notes/action.yaml
```diff
@@ -33,16 +33,27 @@ inputs:
...
+        categories: 'bugfix,feature'
+        recursion_depth: 0
+      - name: your-variant
+        audiences: 'operator'
+        categories: 'feature'
+        recursion_depth: -1
+      ```
+
+      Only name is required.
+      Audiences and categories default to "include all", whereas recursion-depth defauls to 0.
```
8R0WNI3, line 56: typo: `defauls` -> `defaults`

.github/actions/release-notes/release_notes_cli.py
```diff
@@ -42,6 +48,14 @@ def ensure_trailing_newline(text: str) -> str:
     return f'{text}\n'
 
 
+@dataclasses.dataclass
+class ReleaseNotesVariantCfg:
+    name: str
+    audiences: tuple[str, ...] = dataclasses.field(default_factory=tuple)
+    categories: tuple[str, ...] = dataclasses.field(default_factory=tuple)
```
8R0WNI3, line 55: We could also use `release_notes.model.ReleaseNotesAudience` and `release_notes.model.ReleaseNotesCategory` here instead of `str`. That way, we also have a validation in place that the specified cfg is valid.

.github/actions/release-notes/action.yaml
```diff
@@ -96,37 +126,9 @@ runs:
...
-
-        echo 'release-notes<<EOF' >> ${GITHUB_OUTPUT}
-        cat release-notes.md >> ${GITHUB_OUTPUT}
-        echo EOF >> ${GITHUB_OUTPUT}
-
-        echo 'full-release-notes<<EOF' >> ${GITHUB_OUTPUT}
-        cat full-release-notes.md >> ${GITHUB_OUTPUT}
-        echo EOF >> ${GITHUB_OUTPUT}
-
-        echo '## Release-Notes (local)' >> ${GITHUB_STEP_SUMMARY}
```
8R0WNI3, line 115: Let us keep writing release-notes (at least local ones) to the summary, that was already useful for me multiple times during error analysis :-)

zkdev: the artefact is available, however I agree that github-summary is more user friendly plus artefacts are cleaned-up eventually.

.github/workflows/release.yaml
```diff
@@ -232,4 +281,4 @@ jobs:
           payload: |
             channel_id: ${{ inputs.slack-channel-id }}
             title: '[${{ steps.component-descriptor.outputs.component-version }} released]'
-            file: ./release-notes.md
+            file: /tmp/release-notes/slack-release-notes.md
```
8R0WNI3, line 284: Why don't we have to download the release-notes artifact before anymore?

.github/workflows/post-build.yaml
```diff
@@ -181,4 +181,3 @@ jobs:
         with:
           version: ${{ needs.component-descriptor.outputs.version }}
           github-token: ${{ secrets.GITHUB_TOKEN }}
-          release-notes: ${{ steps.drn.outputs.full-release-notes }}
```
8R0WNI3, line 184: Don't we also have to pass `release-notes-variants` here to `release-notes` action?

.github/actions/release-notes/release_notes_cli.py
```diff
@@ -119,8 +134,39 @@ def main():
...
+
+    release_notes_variants_cfg = []
+    if parsed.release_notes_variants_cfg_path:
+        with open(parsed.release_notes_variants_cfg_path, 'r') as f:
+            release_notes_variants_cfg = [
+                dacite.from_dict(
+                    data_class=ReleaseNotesVariantCfg,
+                    data=raw,
+                    config=dacite.Config(
+                        cast=[tuple],
```
8R0WNI3, line 164: I think we also require `enum` cast now as well.

.github/workflows/release.yaml
zkdev: Steps within a workflow job share a filesystem, we wouldn't even need the artifact-based sharing in the first place.
However, I think it is more explicit and thus prefer to keep it 🙂.

8R0WNI3 (approved):
/lgtm

PR #1361, opened 2025-09-09
Fix `ocm-upgrade` GHA behaviour in case of multiple componentversions + upgrades

8R0WNI3:
**Release note**:

```other operator

```


ccwienk (commented):
won't we need to group crefs and use greatest version as whence-version?

PR #1367, opened 2025-09-11
Add image update for provider extensions

wpross:
This PR adds a GitHub Action including a python script to be called from a re-usable workflow that is added as well. The workflow can be called from a workflow in the various `gardener-extension-provider-<provider>` repo's in order to update their `imagevector/images.yaml` file. A corresponding PR for azure is [PR1287](https://github.com/gardener/gardener-extension-provider-azure/pull/1287). The logic only updates image versions and adds new versions as explained in the doc string of the python script. Removal of image versions with deprecated K8S versions still have to be manually removed.

The resulting changes in `imagevector/images.yaml` are commited and a PR is created automatically. Within the PR, a list of all the updated image versions is given with links to the respective release notes. This list includes intermediate releases. All the links have to be checked by a developer/reviewer for possible breaking changes. To ensure this, a checklist is added to the PR.

**Release note**:

```feature developer
Add GHA and corresponding Workflow that are to be called from a `gardener-extension-provider-<provider>`  repo's GitHub Workflow in order to update the `imagevector/images.yaml` file, e.g. https://github.com/gardener/gardener-extension-provider-azure/blob/master/imagevector/images.yaml
```


.github/actions/update-extension-provider-images/update-images.py
```diff
@@ -0,0 +1,537 @@
...
+
+    for (major, minor, patch), tag in available_versions:
+        if (major, minor) in current_minor_versions:
+            max_current_patch = max(p for p, _ in current_minor_versions[(major, minor)])
+            if patch > max_current_patch:
+                newer_patch.append(tag)
+        elif (major, minor, patch) > highest_current:
+            newer_major_or_minor.append(tag)
+
+    return {"patch": newer_patch, "minor": newer_major_or_minor}
```
ccwienk, line 181: consider using `semver` package

ccwienk (changes requested):

.github/actions/update-extension-provider-images/update-images.py
```diff
@@ -0,0 +1,537 @@
...
+def check_regctl_available() -> bool:
+    """Check if regctl command is available."""
+    try:
+        subprocess.run(["regctl", "--help"], capture_output=True, check=True)  # nosec: B603, B607
+        return True
+    except (subprocess.CalledProcessError, FileNotFoundError):
+        return False
+
+
+def run_regctl_command(repository: str) -> List[str]:
```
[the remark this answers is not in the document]
wpross: done

.github/actions/update-extension-provider-images/update-images.py
```diff
@@ -0,0 +1,537 @@
...
+    Returns:
+        Optional[Tuple[int, int, int]]: A tuple of (major, minor, patch) integers, otherwise None.
+    """
+    match = re.match(r"^v(\d+)\.(\d+)\.(\d+)$", tag)
+    if match:
+        return (int(match.group(1)), int(match.group(2)), int(match.group(3)))
+    return None
+
+
+def get_clean_tag(tag: str) -> Optional[str]:
```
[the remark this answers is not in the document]
wpross: done

.github/actions/update-extension-provider-images/update-images.py
```diff
@@ -0,0 +1,537 @@
...
+    """
+    Filter a list of tags to include only valid 'vX.Y.Z' semantic version tags.
+
+    Args:
+        tags (List[str]): A list of tag strings from the registry.
+
+    Returns:
+        List[str]: A filtered list of valid tags.
+    """
+    return [tag for tag in tags if get_clean_tag(tag)]
```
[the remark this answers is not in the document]
wpross: done

.github/actions/update-extension-provider-images/update-images.py
```diff
@@ -0,0 +1,537 @@
...
+
+
+# --- Core Logic Functions ---
+def find_newer_versions(current_tags: List[str], available_tags: List[str]) -> Dict[str, List[str]]:
```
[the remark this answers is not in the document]
wpross: done

.github/actions/update-extension-provider-images/update-images.py
wpross: done

wpross:
I've refactored the script to use standard `cc-utils` libraries:
- Replace `regctl` subprocess calls with the native `oci.client`.
- Replace custom version parsing with the `version` and `semver` modules.

New image existence check is introduced to improve robustness:
- Before updating a tag, the script now verifies the image manifest exists in the container registry.
- This check includes a retry mechanism with exponential backoff to handle transient errors like rate limiting (`429`
) that I got when testing with the `aws` provider.
- It also correctly handles multi-arch manifests from registries like `registry.k8s.io` by sending a permissive `Acce
pt` header, relevant for the `openstack` provider.
- Images that cannot be verified are reported in the PR notes as `missing images` instead of being updated, preventing broken builds.

.github/actions/update-extension-provider-images/update-images.py
```diff
@@ -0,0 +1,626 @@
...
+reusable workflow.
+
+Manual Intervention:
+This script handles version *updates*. Manually editing `images.yaml` is still
+required to remove entire image groups for deprecated Kubernetes versions that
+are no longer supported.
+Image Updater Script.
+"""
+
+import argparse
```
ccwienk, line 37: codestyle: according to [pep-8](https://peps.python.org/pep-0008/#imports), imports should be grouped (stdlib, 3rdparty, local).

In cc-utils (+ repositories using cc-utils), we additionally discriminate between local code and code used from cc-utils ("2.5rd"-party, if you will :-)), and order imports alphabetically.

Concretely-speaking, you should group:

argparse
collections
re
sys
time
typing

semver
yaml

oci
version

.github/actions/update-extension-provider-images/update-images.py
```diff
@@ -0,0 +1,626 @@
...
+from oci.model import OciImageNotFoundException
+from version import parse_to_semver, is_final
+
+
+# --- Type Definitions ---
+class Update(TypedDict):
+    """A structured representation of a single image update."""
+
+    image_name: str
+    old_tag: Optional[str]
```
ccwienk, line 55: According to [documentation](https://typing.python.org/en/latest/spec/typeddict.html#required-and-notrequired), you should use Required/NotRequired for that. I think.

Note that we typically use [dataclasses](https://docs.python.org/3/library/dataclasses.html) in cc-utils (there, `Optional` is appropriate to use :-))

.github/actions/update-extension-provider-images/update-images.py
```diff
@@ -0,0 +1,626 @@
...
+        repository (str): The image repository to query.
+
+    Returns:
+        List[str]: A list of tag strings, e.g., ['v1.2.3', 'v1.2.4'].
+                   Returns an empty list on error.
+    """
+    try:
+        return oci_client.tags(image_reference=repository)
+    except Exception as e:
+        print(f"Error retrieving tags for {repository}: {e}", file=sys.stderr)
```
ccwienk, line 76: do you actually expect errors to occur often? To my experience, ignoring errors, and emitting a warning is often a bad pattern, especially for pipelines that run continuously (-> logs are rarely checked, so errors will go unnoticed).

Also, as there is only one usage, I would suggest to inline this function, and do try-except (if deemed necessary) at point of usage, where log-message could be made more specifically (with current code, in case of error, there will be both an error-indication _and_ log-entry about "no tags found").

That being said: can it ever happen that there is no tag? I would assume such a case would indicate an erroneous configuration.

Also, consider also printing stacktrace (`traceback.print_exc()`). This makes debugging _a lot_ easier.

.github/actions/update-extension-provider-images/update-images.py
```diff
@@ -0,0 +1,626 @@
...
+    """A structured representation of a single image update."""
+
+    image_name: str
+    old_tag: Optional[str]
+    new_tag: str
+    update_type: str  # 'patch', 'minor', or 'singleton'
+
+
+# --- Helper Functions ---
+def get_tags_from_registry(oci_client: oci_client.Client, repository: str) -> List[str]:
```
ccwienk, line 61: nit: the correct(er) term would be `[oci]_repository` (you correctly named the parameter).

[see oci-distribution-spec](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#definitions) for reference:

```
Registry: a service that handles the required APIs defined in this specification
Repository: a scope for API calls on a registry for a collection of content (including manifests, blobs, and tags).
```

.github/actions/update-extension-provider-images/update-images.py
```diff
@@ -0,0 +1,626 @@
...
+    Returns:
+        bool: True if the image exists, False otherwise.
+    """
+    image_ref = f"{repository}:{tag}"
+    retries = 3
+    delay = 2  # seconds
+
+    for i in range(retries):
+        try:
+            oci_client.manifest(image_reference=image_ref, accept="*/*")
```
ccwienk, line 98: consider using `absent_ok=True`. In this case, absence of requested image will not result in an exception. Also, preferably use `head_manifest` - this does exactly what you need, I think (+ avoids unnecessary transfer of data)

Lastly, you should not have to set the `accept`-header. oci-client's default behaviour should match what you try to do by passing */* (also, some OCI-Registries might not like */*. which is probably not a concern for _this_ script, though).

.github/actions/update-extension-provider-images/update-images.py
```diff
@@ -0,0 +1,626 @@
...
+            return True
+        except OciImageNotFoundException:
+            print(f"  ✗ Manifest for {image_ref} not found.", file=sys.stderr)
+            return False
+        except Exception as e:
+            # Check if the error is a rate-limiting error
+            if "429" in str(e) and "Too Many Requests" in str(e):
```
ccwienk, line 105: will you run queries against gardener-project's GAR exclusively, or will you also target other OCI-Registries, such as docker-hub, quay.io, and the like?

In either way, preferably use the more specific exception raised from `requests` (requests.exceptions.HTTPError, I think?). This exception contains an int w/ status-code, so you do not have to parse / guess. Also, it should be sufficient to only check for status-code.

Consider running `oci-auth`-action prior to calling this script, such as to allow the oci-client to run authenticated requests (this might be helpful, especially if issueing requests against docker-hub - readonly-credentials for DH are implicitly included from said action..)

.github/actions/update-extension-provider-images/update-images.py
```diff
@@ -0,0 +1,626 @@
...
+        except ValueError:
+            continue
+
+    available_versions = sorted(available_versions_map.keys())
+
+    current_versions = []
+    for tag in current_tags:
+        try:
+            current_versions.append(parse_to_semver(tag))
+        except ValueError:
```
ccwienk, line 157: why not use `parse_to_semver`'s `invalid_semver_ok`? that way, this could be shortened to just:

```
current_versions = [
  pv for v in ('1.2.3', '1.2.2',)
  if (pv:= version.parse_to_semver(v))
]
```

However, my main point is to avoid doing too much error-silencing, as this could hide actually unexpected errors.

.github/actions/update-extension-provider-images/update-images.py
```diff
@@ -0,0 +1,626 @@
...
+            continue
+    current_versions.sort()
+
+    if not current_versions:
+        existing_minor_tags = []
+        missing_minor_tags = []
+        for ver in available_versions:
+            tag = available_versions_map[ver]
+            if image_exists(oci_client, repository, tag):
```
ccwienk, line 166: I do not quite get why this check is required. To my understanding, the tags returned from OCI-Repository are guaranteed to exist. The only corner-case might be that a tag might have been removed immediately after you requested list-tags. Which is a corner-case that (IMO) does not really suggest (to me at least) this should be double-checked (also, even then, the tag could still be removed after your check..)

.github/actions/update-extension-provider-images/update-images.py
```diff
@@ -0,0 +1,626 @@
...
+        if name not in new_versions_by_name:
+            continue
+
+        new_entries_for_this_name = []
+        newer = new_versions_by_name[name]
+        has_target_version = any("targetVersion" in img for img in image_list)
+
+        if not has_target_version:
+            # Handle images without 'targetVersion' (singleton components)
+            all_newer_tags = newer["patch"] + newer["minor"]
```
ccwienk, line 240: nit: as mentioned before, using semver-arithmetic, you can only tell if a version is _greater_ or _smaller_. You have no means to determine whether it is newer or older (in a chronological sense)

.github/actions/update-extension-provider-images/update-images.py
```diff
@@ -0,0 +1,626 @@
...
+                            "image_name": name,
+                            "old_tag": old_tag,
+                            "new_tag": latest_tag,
+                            "update_type": "singleton",
+                        }
+                    )
+            else:
+                print(
+                    f"Warning: Found {len(image_list)} entries for singleton image '{name}' "
+                    f"but expected 1. Skipping update for this image.",
```
ccwienk, line 261: shouldn't you rather signal this as an error, and bail out?

.github/actions/update-extension-provider-images/update-images.py
```diff
@@ -0,0 +1,626 @@
...
+                )
+
+        all_new_entries_global.extend(new_entries_for_this_name)
+
+    # Apply changes to the main data structure
+    images_data["images"].extend(all_new_entries_global)
+    return updates
+
+
+def find_intermediate_versions(
```
ccwienk, line 365: this's somewhat resembling `version.iter_upgrade_path`, wdyt? :-)

ccwienk (changes requested):

.github/actions/update-extension-provider-images/update-images.py
wpross: I remove the `image_exists` fctn for now, see [this comment](https://github.com/gardener/cc-utils/pull/1367#discussion_r2362759276)

.github/actions/update-extension-provider-images/update-images.py
wpross: I remove the `image_exists` fctn for now, see [this comment](https://github.com/gardener/cc-utils/pull/1367#discussion_r2362759276)

.github/actions/update-extension-provider-images/action.yaml
```diff
@@ -0,0 +1,64 @@
...
+      id: run_script_step
+      shell: bash
+      run: |
+        set -euo pipefail
+
+        summary=$(python3 ${{ github.action_path }}/update-images.py \
+          --images-yaml-path ${{ inputs.images-yaml-path }} \
+          --release-notes-path ${{ inputs.release-notes-path }})
+        
+        exit_code=$?
```
ccwienk, line 52: as you set `-euo pipefail` above, this line will not be reached in case of exitcode !== 0. Which is fine (I suggest to simply drop unnecessary boilerplate)

.github/actions/update-extension-provider-images/update-images.py
```diff
@@ -0,0 +1,701 @@
...
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "--images-yaml-path",
+        default=DEFAULT_IMAGES_PATH,
+        help="Path to the images.yaml file.",
+    )
+    parser.add_argument(
+        "--release-notes-path",
+        default=DEFAULT_RELEASE_NOTES_PATH,
```
ccwienk, line 613: ditto

.github/workflows/update-extension-provider-images.yaml
```diff
@@ -0,0 +1,124 @@
...
+            gh pr create \
+              --head ${{ inputs.branch-name }} \
+              --base master \
+              --title "🤖 Chore: Auto-update container image versions" \
+              --body "$pr_body"
+          fi
+
+      - name: Clean up release notes file
+        if: always()
+        run: rm -f release-notes.md
```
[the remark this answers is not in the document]
wpross: I'm only commiting the `imagevector/images.yaml` file (see [here](https://github.com/gardener/cc-utils/pull/1367/files#diff-d7112002425bb21b081e55dd7715151ef7eac4f6c66d1dcf8a37abaf4668c18aR64)). The `release-notes.md` is only used for collecting the links to the release-notes that are added to the PR's `Component Changelogs` section so that the reviewer can follow the links and check for breaking changes.

.github/workflows/update-extension-provider-images.yaml
```diff
@@ -0,0 +1,124 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+name: Reusable Image Version Updater
+
+on:
+  workflow_call:
+    inputs:
+      branch-name:
+        description: 'The name of the branch for the pull request.'
```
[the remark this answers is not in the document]
wpross: The branch is created by the action and there's no rule in place in the `gardener-extension-provider-` repos that would apply branch protection rules to it. So I guess it's fine.

.github/actions/update-extension-provider-images/update-images.py
```diff
@@ -0,0 +1,691 @@
...
+            tag = available_versions_map[ver]
+            existing_minor_tags.append(tag)
+        return {
+            "patch": [],
+            "minor": existing_minor_tags,
+        }
+
+    highest_current_ver = current_versions[-1]
+
+    highest_patch_for_minor = defaultdict(lambda: semver.Version(0))
```
ccwienk, line 137: I think that's slightly hacky.... how about using `version.find_greatest_version_with_matching_minor` instead? I think this should do the same

.github/actions/update-extension-provider-images/update-images.py
```diff
@@ -0,0 +1,691 @@
...
+            return (image["name"], parse_to_semver(image["tag"]))
+        except ValueError:
+            return (image["name"], semver.Version(0))
+
+    images_data["images"].sort(key=sort_key)
+
+
+def write_yaml_with_formatting(data: Dict[str, Any], filename: str) -> None:
+    """
+    Write the dictionary to a YAML file using the project's style.
```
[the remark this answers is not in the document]
wpross: yes, `yaml.safe_dump` doess not provide the required formatting options. 
Anyways, `ruamel` would be really nice to use in a future optimization to e.g. allow for comments like `#keep` to fix an image to a specific version or similar.

.github/workflows/update-extension-provider-images.yaml
```diff
@@ -0,0 +1,124 @@
...
+          images_path="${{ inputs.images-yaml-path || 'imagevector/images.yaml' }}"
+          git add "$images_path"
+          git commit -m "Chore: Auto-update image versions"
+          git checkout -B ${{ inputs.branch-name }}
+          git push origin ${{ inputs.branch-name }} --force
+
+      - name: Create or Update Pull Request
+        if: steps.check_changes.outputs.changes_detected == 'true'
+        env:
+          GITHUB_TOKEN: ${{ steps.app-token.outputs.token }}
```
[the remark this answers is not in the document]
wpross: done

.github/actions/update-extension-provider-images/update-images.py
```diff
@@ -0,0 +1,694 @@
...
+    """A structured representation of a single image update."""
+
+    image_name: str
+    new_tag: str
+    update_type: str  # 'patch', 'minor', or 'singleton'
+    old_tag: Optional[str] = None
+
+
+# --- Helper Functions ---
+def validate_images_data(images_data: dict[str, Any]) -> None:
```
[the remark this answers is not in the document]
wpross: I've added dataclasses for the data in the `images.yaml` now and also added according type hints. I've also transferred some of the functionality from functions to class methods.

.github/actions/update-extension-provider-images/update-images.py
```diff
@@ -0,0 +1,694 @@
...
+            greater_minor_tags.append(original_tag)
+
+    return {
+        "patch": sorted(greater_patch_tags, key=parse_to_semver),
+        "minor": sorted(greater_minor_tags, key=parse_to_semver),
+    }
+
+
+def update_singleton_image(
+    image_list: list[dict[str, Any]], all_greater_tags: list[str], name: str
```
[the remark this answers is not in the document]
wpross: I'm using `ruff` to autoformat the file and unfortunately it's not configurable to consider this, but I've made sure the format matches.

.github/actions/update-extension-provider-images/update-images.py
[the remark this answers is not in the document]
wpross: see [this comment](https://github.com/gardener/cc-utils/pull/1367#issuecomment-3327502332).
I've switched to using `yaml.safe_dump` as suggested after checking if the resulting format would be an issue.

.github/workflows/update-extension-provider-images.yaml
```diff
@@ -0,0 +1,124 @@
...
+
+      - name: Create GitHub App Token
+        id: app-token
+        uses: actions/create-github-app-token@v2.1.1
+        with:
+          app-id: ${{ vars.GARDENER_GITHUB_ACTIONS_APP_ID }}
+          private-key: ${{ secrets.GARDENER_GITHUB_ACTIONS_PRIVATE_KEY }}
+
+      - name: Setup Git Identity
+        uses: ./.github/actions/setup-git-identity
```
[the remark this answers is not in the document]
wpross: done

.github/workflows/update-extension-provider-images.yaml
```diff
@@ -0,0 +1,124 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@v5.0.0
+
+      - name: Run image update action
+        id: image_update
+        uses: ./.github/actions/update-extension-provider-images
+
+      - name: Read release notes file
+        id: release_notes
+        continue-on-error: true
```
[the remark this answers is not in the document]
wpross: I check for an empty `release-notes.md` file now explicitly.

.github/actions/update-extension-provider-images/action.yaml
```diff
@@ -0,0 +1,56 @@
...
+    - name: Validate inputs
+      shell: bash
+      run: |
+        if [[ ! -f "${{ inputs.images-yaml-path }}" ]]; then
+          echo "Error: images.yaml file not found at ${{ inputs.images-yaml-path }}"
+          exit 1
+        fi
+
+    - name: Install Gardener GHA Libs
+      uses: ./.github/actions/install-gardener-gha-libs
```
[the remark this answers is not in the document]
wpross: done

.github/actions/update-extension-provider-images/action.yaml
```diff
@@ -0,0 +1,56 @@
...
+
+    - name: Run image update script
+      id: run_script_step
+      shell: bash
+      run: |
+        set -euo pipefail
+
+        summary=$(python3 ${{ github.action_path }}/update-images.py \
+          --images-yaml-path ${{ inputs.images-yaml-path }} \
+          --release-notes-path ${{ inputs.release-notes-path }})
```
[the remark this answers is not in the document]
wpross: done

.github/actions/update-extension-provider-images/requirements.txt
```diff
@@ -0,0 +1,2 @@
+PyYAML
+semver
```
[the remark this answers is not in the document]
8R0WNI3: The removal is not pushed yet :-)

.github/actions/update-extension-provider-images/update-images.py
```diff
@@ -37,54 +37,151 @@
...
+    name: str
+
+
+@dataclass
+class ImageEntry:
+    name: str
+    sourceRepository: str
+    repository: str
+    tag: str
+    labels: list[Label]
```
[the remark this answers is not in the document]
8R0WNI3: I see, however, the `ocm.Label` class is the same which is also used to parse the image-data within `ocm.gardener`, so it is actually allowed to add additional labels (!= CVE categorisation). Also, my motivation was that I don't like to maintain multiple definitions of the same label model (e.g. the label model itself but also specific instances like the CVE categorisation which is already defined [here](https://github.com/open-component-model/delivery-service/blob/1572b7c0eba9e9f13cc46f2835076f28856615a5/odg/cvss.py#L138C7-L138C24)).

ccwienk (approved):

PR #1373, opened 2025-09-17
Checkout Submodules using github-app-authtoken

ccwienk:
**Release note**:

```other developer
Allow checking-out of submodules from repositories on same github(-enterprise)-instance, also if authentication is required, using default Github-App-issued auth-tokens.
```


8R0WNI3 (approved):
/lgtm

PR #1375, opened 2025-09-18
create authtoken in trusted-checkout

ccwienk:
**Release note**:

```other developer
trusted-checkout-action now optionally creates github-app-authtokens
```


8R0WNI3 (approved):
/lgtm

PR #1382, opened 2025-09-19
GHA release workflow supports multiple tags via `github-tag-templates` input field

### LINKED ISSUE #1381: Github actions should support creating release with multiple tags
# Context / Motivation

In [etcd-druid](https://github.com/gardener/etcd-druid), we have multiple go modules, and upon every release, we need to create multiple tags for each version, like [this](https://github.com/gardener/etcd-druid/blob/c7d1e687148ec2bdfcab54cd2e35e6dd6deb271a/.ci/pipeline_definitions#L115-L118). Currently cc-utils's `release.yaml` workflow only supports single tag via the `github-tag-template` input, which is insufficient for our usecase.

# Implementation Proposal

I would like to add `github-tag-templates`, which is a list of strings instead of a single string. If this is the case, then we need to decide how to deprecate/remove `github-tag-template`.

shreyas-s-rao:
GHA release workflow supports multiple tags via `github-tag-templates` input field. I have replaced the previous `github-tag-template` field with `github-tag-templates`, since I nobody is using `github-tag-template` in gardener org yet, and I wanted to avoid the complexity of handling the same input from multiple input fields and deprecation overhead.

Also, this field is technically a string (comma-separated values validated in `release/action.yaml`), since the schema does not allow `array` or `list` type in reusable workflows - [ref discussion](https://github.com/orgs/community/discussions/11692).

I also split `push release-tag` step in release action into `push tag-refs` and `push release-commit` for clear separation of responsibilities.

Please let me know if the approach is correct. Also, majority of the changes are AI generated, so please let me know if there's anything I need to correct here as well.

Fixes #1381 

**Release note**:

```other developer
GHA `release` workflow now supports multiple tags via `github-tag-templates` input field which takes comma-separated values of target ref templates.
```
```breaking developer
GHA `release` workflow input field `github-tag-template` has now been replaced by `github-tag-templates` which takes comma-separated values of target ref templates. While the input name needs to be adapted, no change is needed to the existing value.
```

### DIFF
--- a/.github/actions/release/action.yaml
+++ b/.github/actions/release/action.yaml
@@ -140,13 +140,14 @@ inputs:
           os: linux
           arch: x86_64
       ```
-  github-tag-template:
+  github-tag-templates:
     default: '${version}'
     type: string
     description: |
-      GitHub tag template to use for the created release tag. Currently only supported template-var:
+      Comma-separated list of GitHub tag templates to use for the created release tags. Each template can use
       ${version} (bash-syntax): Version read from local component descriptor (i.e. previous version
       operations are honoured).
+      The first tag in the list will be used to create the GitHub release.
   on-existing-component-descriptor:
     default: fail
     type: choice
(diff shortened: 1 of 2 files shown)


.github/actions/release/action.yaml
```diff
@@ -201,10 +202,42 @@ runs:
...
-        tag_ref="refs/tags/${tag_name}"
-        echo "tag-ref=${tag_ref}" >> $GITHUB_OUTPUT
-        echo "tag-name=${tag_name}" >> $GITHUB_OUTPUT
+        # Parse github-tag-templates string into array
+        IFS=',' read -ra tag_templates <<< "${{ inputs.github-tag-templates }}"
+        tag_names=()
+        tag_refs=()
+        for template in "${tag_templates[@]}"; do
+          template=$(echo "$template" | xargs) # trim whitespace
+          tag_name=$(echo "$template" | sed "s/\\${version}/$version/g")
```
8R0WNI3, line 211: This shouldn't be necessary as the `version` is already inserted in this statement: `"${{ inputs.github-tag-templates }}"`

.github/actions/release/action.yaml
```diff
@@ -352,16 +385,21 @@ runs:
         path: /tmp/component-descriptor.tar.gz
         name: component-descriptor
 
-    - name: push release-tag
+    - name: push tag-refs
       shell: bash
       run: |
         set -eu
+        for tag_ref in ${{ steps.preprocess.outputs.tag-refs }}; do
+          push_spec="@:$$tag_ref"
```
8R0WNI3, line 393: typo: duplicate `$$`

.github/actions/release/action.yaml
```diff
@@ -140,13 +140,14 @@ inputs:
           os: linux
           arch: x86_64
       ```
-  github-tag-template:
+  github-tag-templates:
     default: '${version}'
     type: string
     description: |
-      GitHub tag template to use for the created release tag. Currently only supported template-var:
+      Comma-separated list of GitHub tag templates to use for the created release tags. Each template can use
```
ccwienk, line 147: how about using a _newline-separated_ list?

Admittedly, for the "older" actions and workflows in cc-utils, I (unfortunately) followed "docker-build-push"-action's approach of using comma-separated lists (with the downside of resulting in long lines, and bad readability). Later, I learnt that other actions seem to prefer newline-separated inputs, which IMO have a far better readability.

.github/actions/release/action.yaml
```diff
@@ -140,13 +140,14 @@ inputs:
...
       ${version} (bash-syntax): Version read from local component descriptor (i.e. previous version
       operations are honoured).
+      The first tag in the list will be used to create the GitHub release.
```
ccwienk, line 150: how about defining an additional input:
- keep github-tag-template
- add: git-extra-tag-templates (optional)

This makes will make the contract more obvious, even w/o reading the documentation

ccwienk (commented):
generally-speaking, it is a reasonable ask to allow creation of multiple tags (thanks for taking the time to raise a pullrequest, rather than just a feature-request :-)).

However, I would prefer this change to focus on the actual feature (one tag -> list of (extra-)tags). Refactoring and adding of validation code may also be adequate. However, I would prefer such changes to be done in separate commits.

Also, with such increased amount of code, I would prefer to mv majority of code into a separate script (any file within an action's directory is available at `${{ github.action_path }}`).

.github/actions/release/action.yaml
shreyas-s-rao: I felt it would be more natural defining all the tags in one single list, rather than splitting into a single template and another list of templates

.github/workflows/release.yaml
```diff
@@ -96,12 +96,14 @@ on:
         description: |
           If set to `true`, the workflow will create a new GitHub release (or convert an existing
           draft release to an actual release).
-      github-tag-template:
+      github-tag-templates:
```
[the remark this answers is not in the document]
shreyas-s-rao: I observed that no other repo in g/g uses `github-tag-template`, hence took the liberty of changing it incompatibly. 

> but also from other orgs, some not visible in public

Ah, I was not aware of this.

Please check and let me know if a breaking change can be made. If you still feel having two separate fields makes sense, I'll make the change.

PR #1385, opened 2025-09-26
Refactor OCI-Auth action

8R0WNI3:
**Release note**:

```other developer
The `oci-auth` action now implements the OIDC flow against AWS, Azure and GCP itself (no external action used anymore)
```
```feature developer
The `oci-auth` action now supports authentication for multiple OCI image references
```


.github/actions/oci-auth/oidc.py
```diff
@@ -0,0 +1,418 @@
...
+    def from_dict(raw: dict):
+        registry_type = oci.model.OciRegistryType(raw['type'])
+
+        if registry_type is oci.model.OciRegistryType.AWS:
+            data_class = AwsOidcConfiguration
+
+        elif registry_type is oci.model.OciRegistryType.AZURE:
+            data_class = AzureOidcConfiguration
+
+        elif registry_type is oci.model.OciRegistryType.GAR:
```
zkdev, line 38: how about using a dict instead?

.github/actions/oci-auth/oidc.py
```diff
@@ -0,0 +1,418 @@
...
+    image_reference: oci.model.OciImageReference,
+) -> AwsOidcConfiguration | AzureOidcConfiguration | GarOidcConfiguration:
+    github_server_url = os.environ['GITHUB_SERVER_URL']
+    github_org = os.environ['GITHUB_REPOSITORY_OWNER']
+
+    if github_server_url.startswith('https://github.tools'):
+        github_host = 'gh-tools'
+    elif github_server_url.startswith('https://github.wdf'):
+        github_host = 'gh-wdf'
+    elif github_server_url == 'https://github.com':
```
zkdev, line 98: same here

.github/actions/oci-auth/oidc.py
```diff
@@ -0,0 +1,418 @@
...
+        session=session,
+        method='POST',
+        data=body,
+        headers={
+            'Content-Type': 'application/x-www-form-urlencoded',
+        },
+    )
+    refresh_token = res.json()['refresh_token']
+
+    username = '00000000-0000-0000-0000-000000000000'
```
zkdev, line 269: what is the rational behind this dummy username?

.github/actions/oci-auth/oidc.py
```diff
@@ -0,0 +1,418 @@
...
+
+    token = base64.b64encode(f'{username}:{refresh_token}'.encode()).decode()
+
+    return {
+        'auth': token,
+    }
+
+
+def authenticate_against_gar(
```
zkdev, line 278: if available, please add links to documentation of hyperscaler specific authentications

.github/actions/oci-auth/oidc.py
8R0WNI3: These conditions are using `startswith` function, so I think a dictionary is not working here 🤔

.github/actions/oci-auth/oidc.py
```diff
@@ -0,0 +1,418 @@
...
+
+    prefixes = set()
+    for oidc_cfg in oidc_cfgs:
+        if oidc_cfg.type is not image_reference.registry_type:
+            continue
+
+        if oidc_cfg.github_host != github_host:
+            continue
+
+        prefixes.update(oidc_cfg.oci_repository_prefixes)
```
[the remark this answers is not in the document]
8R0WNI3: Because we only want to display the known prefixes for the given host. I'll update the error message accordingly to state this as well.

.github/actions/oci-auth/oidc.py
8R0WNI3: Because this is required by Azure in case no actual username exists (see for example [this documentation](https://learn.microsoft.com/en-us/azure/container-registry/container-registry-authentication?tabs=azure-cli#az-acr-login-with---expose-token) -> it does not really handle this authentication case, but it also shows the username...).

zkdev (approved):
/lgtm

PR #1389, opened 2025-10-02
Add OWNERS files

dimityrmirchev:
This PR prepares the repository for migration to prow.

Part of https://github.com/gardener/org/issues/3

Please review and adjust assignments if necessary.


zkdev:
@dimityrmirchev what about CODEOWNERS, github-teams, component responsibles and direct repository access?

I'd like to avoid to introduce yet another location where ownership has to be maintained

PR #1390, opened 2025-10-06
Remove default `cluster-cfg` input for `kubernetes-auth` action

8R0WNI3:
As the `kubernetes-auth` action is intended to be used in more generic scenarios, drop this specific default value which is used for integrationtests only. Instead, users should consider using the `run-testmachinery-tests.yaml` workflow to run integrationtests which already defines reasonable defaults.

**Release note**:

```other operator

```


zkdev (approved):
/lgtm

do we need to raise PRs towards all repositories currently relying on this default?

8R0WNI3:
I checked repositories using GitHub search and it seems there are no usages relying on this default value.

PR #1393, opened 2025-10-07
Fix multi-line output handling in workflow and action

wpross:
This fixes the encoding of multi-line statements in the workflow and the action so that we get a clean output in the resulting PR.
**Release note**:

```bugfix developer
Fix multi-line encoding in update-extension-provider-images workflow.
```


zkdev (approved):
/lgtm

PR #1401, opened 2025-10-13
Adapt upgrade PR logic to cope with component reference alias

voelzmo:
This PR adapts the upgrade PR logic to cope with component reference `name`, which we introduce as additional property to the `ComponentIdentity` as `alias` (as there is already a property called `name`, referring to the component name).

If the component name is `github.com/gardenlinux/gardenlinux`, we switch to a different schema for upgrade PR titles. Instead of `[ci:component:<component_name>:<version_from>-><version_to>]`, we switch to `[ci:alias:<component_alias>:<version_from>-><version_to>]`. This way, we can distinguish upgrade PRs for `gardenlinux` and `gardenlinux-test` on landscape-dev.

part of https://github.tools.sap/kubernetes/landscape-stability-backlog/issues/468

**Release note**:

```other operator
Adapt upgrade PR logic to cope with component reference alias
```

### DIFF
--- a/.github/actions/ocm-upgrade/ocm_upgrade.py
+++ b/.github/actions/ocm-upgrade/ocm_upgrade.py
@@ -13,7 +13,7 @@ try:
     import ocm
 except ImportError:
     # make local development more convenient
-    repo_root = os.path.join(os.path.dirname(__file__), '../../..')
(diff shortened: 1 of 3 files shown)


voelzmo:
replaced by #1404

PR #1410, opened 2025-10-20
Antispam

ccwienk:
**Release note**:

```other developer
limit length of step-summary when printing version-commit
```


.github/actions/base-component-descriptor/action.yaml
```diff
@@ -247,7 +247,7 @@ runs:
             ```
             {textwrap.indent(component_descriptor_str, '  ')}
             ```
-          '''))
+            '''))
```
8R0WNI3: Wasn't this correct before?

ccwienk: image
well, the first line is not being indented correctly (this has bugged me for quite a while).

PR #1417, opened 2025-10-23
Include `signingserver` module in GHA libs

8R0WNI3:
**Release note**:

```other operator

```

### DIFF
--- a/signingserver.py
+++ b/signingserver.py
@@ -3,10 +3,12 @@ A client for Signing-Server
 '''
 
 import dataclasses
+import enum
 import hashlib
 import io
 import logging
 import re
+import typing
 import urllib.parse
 
 import cryptography.x509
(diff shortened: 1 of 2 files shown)


signingserver.py
```diff
@@ -15,12 +17,28 @@
 import urllib3
 
 import ci.util
-import model.signing_server as ms
 
 
 logger = logging.getLogger(__name__)
 
 
+class SigningAlgorithm(enum.StrEnum):
```
ccwienk: can we deduplicate this w/ model/signing_server.py? OTOH, since the latter will be dissolved at some point not too far, probably this is not worth the efforts

8R0WNI3: Yes, as soon as the consumers are adjusted to use this enum, the one in the `model.signing_server` module can be dropped.

PR #1431, opened 2025-10-30
fix: always ignore version-downgrades

ccwienk:
If following an upstream-component, if using `strictly-follow` mode, ocm-upgrade action should force thus-configured upgrade-pipelines to ignore any versions (such as hotfix-versions) that are not referenced from (greatest) version of upstream-component. However, it should never pick-up downgrades.

**Release note**:

```other operator

```

### DIFF
--- a/.github/actions/ocm-upgrade/ocm_upgrade.py
+++ b/.github/actions/ocm-upgrade/ocm_upgrade.py
@@ -396,9 +396,9 @@ def create_upgrade_pullrequests(
                     logger.info(f'already at target {cref.componentName} {target=} (skip)')
                     continue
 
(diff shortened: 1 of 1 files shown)


.github/actions/ocm-upgrade/ocm_upgrade.py
```diff
@@ -396,7 +396,7 @@ def create_upgrade_pullrequests(
                     logger.info(f'already at target {cref.componentName} {target=} (skip)')
                     continue
 
-                if tv < cv and upstream_update_policy is not UpstreamUpdatePolicy.STRICTLY_FOLLOW:
+                if tv < cv:
```
8R0WNI3: nit: We should also adjust the logging message below

PR #1435, opened 2025-11-03
Consider `.ocm/branch-info.yaml` file for draft release-note creation

8R0WNI3:
**Release note**:

```feature developer
Draft release-notes are also created for branches matching the `release-branch-template` configured via `.ocm/branch-info.yaml` file
```
```feature developer
Added input `create-release-branch` to release workflow to automatically create a release-/hotfix-branch upon release
```

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -230,6 +230,13 @@ on:
         description: |
           passed as same-named input to release-notes action. Controls whether copies of
           release-notes of sub-components are included.
+      create-release-branch:
+        type: boolean
+        default: false
+        description: |
+          controls whether a release-/hotfix-branch should be created for the current release.
+          Honours the `.ocm/branch-info.yaml`'s `release-branch-template` setting if any, otherwise
+          defaults to "release-v$major.$minor" as template.
 
     outputs:
       release-notes:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -334,3 +341,60 @@ jobs:
             channel_id: ${{ inputs.slack-channel-id }}
             title: '[${{ steps.component-descriptor.outputs.component-version }} released]'
             file: /tmp/release-notes/slack-release-notes.md
+
+      - name: reset-repository-to-release-commit
```
ccwienk: we _could_ also "import" the release-commit, as done in release-action (which would save us a git-fetch-operation / work even if release-commit would not be tagged); I would deem this to be preferable (-> less dependencies against remote repository).

.github/workflows/release.yaml
```diff
@@ -334,3 +341,60 @@ jobs:
...
+        if: ${{ inputs.create-release-branch == 'true' && inputs.next-version != 'bump-patch' }}
+        uses: gardener/cc-utils/.github/actions/version@master
+        with:
+          versionfile: ${{ inputs.version-versionfile }}
+          read-callback: ${{ inputs.version-read-callback }}
+          write-callback: ${{ inputs.version-write-callback }}
+          callback-action-path: ${{ inputs.next-version-callback-action-path }}
+          commit-message: "next version: ${version}"
+          version-operation: bump-patch
+          prerelease: dev
```
ccwienk: should we also make this configurable via branch-info (or another cfg-file in .ocm)?

.github/workflows/release.yaml
```diff
@@ -334,3 +341,60 @@ jobs:
...
+          version="${{ steps.component-descriptor.outputs.version }}"
+
+          version_no_v="${version#v}"
+
+          export major=$(echo "$version_no_v" | cut -d. -f1)
+          export minor=$(echo "$version_no_v" | cut -d. -f2)
+          export patch=$(echo "$version_no_v" | cut -d. -f3)
+
+          # default values if not present in .ocm/branch-info.yaml file
+          release_branch_template="release-v$major.$minor"
```
ccwienk: I suggest we should honour existing `VERSION` in that we should either keep v-prefix, or omit it (see remark above)

.github/workflows/release.yaml
```diff
@@ -334,3 +341,60 @@ jobs:
...
+          if [ -f .ocm/branch-info.yaml ]; then
+            release_branch_template_overwrite=$(cat .ocm/branch-info.yaml | yq .release-branch-template)
+
+            if [ $release_branch_template_overwrite != "null" ]; then
+              echo "INFO: Found .ocm/branch-info.yaml file with branch template $release_branch_template_overwrite"
+              release_branch_template=$release_branch_template_overwrite
+            fi
+          fi
+
+          branch="$(echo $release_branch_template | envsubst)" # fill in values of major/minor/patch
```
ccwienk: nice! did not know about envsubst. I suppose we should consider switching other places in code (e.g. in `version`-action)

.github/workflows/post-build.yaml
```diff
@@ -210,22 +210,63 @@ jobs:
...
-            github.event_name == 'push'
-        }}
+    # never run for forks and only for push events
+    if: ${{ ! fromJSON(needs.repo-metadata.outputs.fork) && github.event_name == 'push' }}
     runs-on: ${{ vars.DEFAULT_RUNNER || 'ubuntu-latest' }}
     permissions:
       contents: write
       id-token: write
     steps:
+      - uses: actions/checkout@v4
```
ccwienk: I assume you do not use `trusted-checkout`, because this is not required for push-triggered runs?

.github/workflows/post-build.yaml
```diff
@@ -210,22 +210,63 @@ jobs:
...
+          release_branch_template=$(cat .ocm/branch-info.yaml | yq .release-branch-template)
+
+          if [ $release_branch_template == "null" ]; then
+            echo "INFO: Not on default branch and no release-branch-template found, skipping draft release."
+            echo "create-draft-release=false" >> ${GITHUB_OUTPUT}
+            exit 0
+          fi
+
+          # replace $major/$minor/$patch with regex patterns and mark start/end of line
+          regex="${release_branch_template//\$major/[0-9]+}"
```
ccwienk: would be nice to have this as a function (accompanied w/ some few unittests)

.github/workflows/release.yaml
8R0WNI3: I'm not sure if there is any need for this. As of now, it's also hardcoded in the `release` action and this value is not used for the "snapshot"-releases as well, so I figure we might rather wait and see if there is any more configuration needed. Wdyt?

ccwienk (commented):
lgtm, ~apart from last inline-comment~

PR #1441, opened 2025-11-06
Fix upgrade-PRs not being closed in case of a stale "whence"-version

8R0WNI3:
**Release note**:

```bugfix developer
Upgrade-PRs are now closed (again) if their "whence"-version is behind the currently referenced component version
```

### DIFF
--- a/github/pullrequest.py
+++ b/github/pullrequest.py
@@ -208,6 +208,7 @@ def iter_obsolete_upgrade_pull_requests(
     upgrade_pull_requests: collections.abc.Iterable[UpgradePullRequest],
     pr_naming_pattern: str='component-name',
     keep_hotfix_versions: bool=True,
+    reference_component: ocm.Component | None=None,
 ) -> collections.abc.Generator[UpgradePullRequest, None, None]:
     grouped_upgrade_pull_requests = collections.defaultdict(list)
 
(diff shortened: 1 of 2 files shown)


TuanAnh17N (approved):

github/pullrequest.py
```diff
@@ -53,10 +53,15 @@ def is_obsolete(
         reference_refs = sorted(
             [
                 rc for rc in reference_component.componentReferences
-                if rc.componentName == self.upgrade_vector.component_name
-                and rc.name == self.component_reference_name
+                if (
+                    rc.componentName == self.upgrade_vector.component_name
+                    and (
+                        not self.component_reference_name
```
zkdev: How does a stale "from" version lead to reading an upgrade-pr in a way so it has no component-ref-name ?

zkdev: But shouldn't we rather set it for the concourse case then, instead of introducing a special handling which will become obsolete (and thus polluting GHA code) once Concourse is shut down?

8R0WNI3: This is not a new special handling for GHA but just honouring the existing dataclass model correctly: `component_reference_name: str | None`. Tbh I don't want to refactor the Concourse upgrade logic to differentiate between component references having the same `componentName` but a distinct `name` if it is meant to be dropped/not used anyways.

ccwienk: factoring this matcher into a (local) function and adding a docstr might help, considering this
a) was subject to a regression
and 
b) apparently was hard to understand, even in the context of this pullrequest

wdyt?

ccwienk (approved):

PR #1449, opened 2025-11-13
Support `automatic` version operation if `.ocm/branch-info.yaml` exists

8R0WNI3:
**Release note**:

```feature developer
Added `automatic` version operation option to release workflow which uses the `.ocm/branch-info.yaml` file to determine the version part to bump
```

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -74,6 +74,7 @@ on:
         description: |
           How the next version (from bump-commit) should be calculated. Will be passed to `version`
           action. (some) allowed values are:
+            - automatic (use `.ocm/branch-info.yaml` to determine the version part to bump)
             - bump-major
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -288,6 +289,67 @@ jobs:
...
+          import os
+
+          import yaml
+
+          import ocm.branch_info
+
+
+          next_version = '${{ inputs.next-version }}'
+
+          if next_version != 'automatic':
```
ccwienk: how about putting this check into an `if` (such that GHA will not have to go through the lengths of starting a python-interpreter (admittedly, that comes at the price of adding a condition for reading next-version) :-))

ccwienk: agreed

PR #1454, opened 2025-11-18
Update Dockerfile with latest Docker and Buildx versions

ghost:
Updates the Dockerfile to use the latest versions:
      - Docker: 29.0.4
      - Buildx: v0.30.1


ccwienk:
@TuanAnh17N : I suppose there is a bug in updating workflow (empty buildx-version); IIRC, we saw this already some weeks ago. no big deal, but I reckon we should fix this eventually.

PR #1457, opened 2025-11-19
fix(distribution): Resolve implicit dependency

zkdev:
`gardener-oci` imports these modules already on package init but does not ship them (neither directly, nor indirectly -> therefore is unusable without `gardener-cicd-libs`). As `gardener-cicd-libs` is already referencing `gardener-oci` package, move required modules to `gardener-oci`.

**Release note**:

```bugfix developer
`gardener-oci` python distribution package no longer implicitly requires `gardener-cicd-libs` to be installed.
```

### DIFF
--- a/oci/__init__.py
+++ b/oci/__init__.py
@@ -5,19 +5,17 @@ import hashlib
 import io
 import json
 import logging
-import tarfile
 import zlib
 
 import dacite
 import requests
 
-import tarutil
 import oci.auth as oa
 import oci.client as oc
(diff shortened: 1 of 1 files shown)


8R0WNI3 (approved):
/lgtm

ccwienk (changes requested):
let us not increase footprint of gardener-oci-package, please

frewilhelm:
Hey folks, any update on this?

PR #1463, opened 2025-11-26
Add `github_app_api_lookup` helper function

8R0WNI3:
**Release note**:

```other operator

```

### DIFF
--- a/github/__init__.py
+++ b/github/__init__.py
@@ -2,13 +2,16 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+import collections.abc
+import dataclasses
 import os
+import re
 import typing
 
 import github3
 
 RepoUrl: typing.TypeAlias = str
-GithubApiLookup = typing.Callable[[RepoUrl], github3.GitHub]
+GithubApiLookup = collections.abc.Callable[[RepoUrl], github3.GitHub | None]
 
 
 def host_org_and_repo(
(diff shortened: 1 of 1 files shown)


github/__init__.py
```diff
@@ -99,3 +101,52 @@ def github_app_api(
...
+                continue
+
+            return github_app_api(
+                github_app_private_key=creds.private_key,
+                github_app_id=creds.app_id,
+                repo_url=repo_url,
+            )
+
+        if absent_ok:
+            return None
```
zkdev: nit: not sure, should we reflect the `None` also in return typehint?

github/__init__.py
```diff
@@ -99,3 +101,52 @@ def github_app_api(
...
+
+        raise RuntimeError(f'no matching GitHub-App credentials for {repo_url=}')
```
zkdev: probably `ValueError` is more suited

github/__init__.py
```diff
@@ -99,3 +101,52 @@ def github_app_api(
...
+
+@dataclasses.dataclass
+class GitHubAppCredentials:
+    private_key: str | bytes
+    app_id: int
+    host: str | None = None
+    org: str | None = None
+    repo: str | None = None
+
+    def matches(self, repo_url: str) -> bool:
```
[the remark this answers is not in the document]
8R0WNI3: No, no particular reason. Will adjust to allow RegEx instead.

PR #1554, opened 2026-04-09
Improve pinning

ccwienk:
**Release note**:

```other operator
cc-util's "install-gardener-gha-libs" action and instrumentations of it now also pin pythoncode installed from cc-utils by commit-digest (along with pinning cross-references to actions and workflows).
```

### DIFF
--- a/.github/actions/install-gardener-gha-libs/action.yaml
+++ b/.github/actions/install-gardener-gha-libs/action.yaml
@@ -2,8 +2,8 @@ name: Install Gardener-GHA-Libs
 description: |
   Install Python3 and gardener-gha-libs
 
-  As GitHub-Actions in https://github.com/gardener/cc-utils still sees frequent development,
-  install from sources (checking out head of `master`), rather than from PYPI.
+  Sources are installed from the `v1` branch of cc-utils (pinned to a specific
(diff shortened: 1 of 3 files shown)


.github/actions/install-gardener-gha-libs/action.yaml
```diff
@@ -39,7 +40,7 @@ runs:
       run: |
         set -euo pipefail
         # fallback if running on GHE
-        git clone https://github.com/gardener/cc-utils
+        git clone --branch v1 https://github.com/gardener/cc-utils
```
TuanAnh17N, line 43: the new pinning logic rewrites `ref:` for the `actions/checkout` path, but the GHE fallback still uses `git clone --branch v1 ...`, so that path is still branch-based rather than digest-based. it that intentional?

ccwienk: yep, you are correct - will have a talk w/ the co-author about that :-)

ccwienk: went for another approach: https://github.com/gardener/cc-utils/pull/1559/changes

ccwienk:
superseded by https://github.com/gardener/cc-utils/pull/1559/changes

Issue #1467, opened 2025-11-26
Please provide SPDX SBOMs for all Gardener Container Images


vlerenc:
# Context / Motivation
In various security-sensitive industries, the ask was raised to automatically/out-of-the-box have SPDX SBOMs right within all Gardener container images. This will make it easy to attach standard tools that can check Gardener container images without additional integration effort.

# Implementation Proposal
As discussed out-of-band with @ccwienk, having Gardener's GitHub actions automatically include SPDX SBOMs when building container images would be of great help to our adopters.

cc @majst01 (please take the chance to chime in with more details, if you like)

ccwienk:
Implemented (forgot to close) :-)

@vlerenc : in case I missed something, please reopen / follow-up.

vlerenc:
Thank you @ccwienk ! cc @majst01 .

majst01:
Thanks !

PR #1556, opened 2026-04-09
oci-ocm: generate SBOM and push as OCI referrer

ccwienk:
https://github.com/gardener/cc-utils/issues/1467

**Release note**:

```feature operator
The oci-ocm build pipeline now generates an SPDX SBOM (via anchore/sbom-action) and pushes it as an OCI 1.1 referrer manifest alongside built images.
```

### DIFF
--- a/oci/sbom.py
+++ b/oci/sbom.py
@@ -1,0 +1,120 @@
+'''
+Utilities for pushing SBOM documents as OCI referrers (OCI 1.1).
+
+An SBOM is stored as a single-layer OCI manifest whose `subject` field points
+to the image it describes.  The referrer manifest is pushed by digest; registries
+that implement the OCI 1.1 referrers API expose it via GET /v2/<repo>/referrers/<digest>.
+'''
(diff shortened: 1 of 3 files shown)


oci/sbom.py
```diff
@@ -0,0 +1,122 @@
+'''
+Utilities for pushing SBOM documents as OCI referrers (OCI 1.1).
+
+An SBOM is stored as a single-layer OCI manifest whose `subject` field points
+to the image it describes.  Registries that implement the OCI 1.1 referrers
+API will expose it via GET /v2/<repo>/referrers/<digest>; others fall back to
```
TuanAnh17N, line 6: tthe docstring says registries wihtout referrers API support “fall back to a synthetic tag”, but I don’t see that fallback implemented 🤔

ccwienk: good catch, dropped comment :-)

majst01:
Nice, thanks

PR #1557, opened 2026-04-10
Implement tag-listing pagination per OCI Distribution Spec

8R0WNI3:
Follow the Link header for pagination when listing tags, as specified in https://github.com/opencontainers/distribution-spec/blob/main/spec.md#listing-tags

Both the sync (client.py) and async (client_async.py) clients now iterate through all pages and return the full list of tags.

**Release note**:

```bugfix developer
The OCI client now supports pagination for listing tags (required for (at least) GHCR)
```

### DIFF
--- a/oci/client.py
+++ b/oci/client.py
@@ -230,6 +230,26 @@ def _scope(image_reference: str | om.OciImageReference, action: str):
     return scope
 
 
+def _next_paginated_url(
+    url: str,
+    headers: dict,
+) -> str | None:
+    # follow pagination as per OCI Distribution Spec
+    # Link header format: <url>; rel="next"
+    if 'rel="next"' in (link_header := headers.get('Link', '')):
+        next_url = link_header.split(';')[0].strip().strip('<>')
+        # resolve relative URLs against the base API url
+        if next_url.startswith('/'):
+            parsed = urllib.parse.urlparse(url)
+            url = f'{parsed.scheme}://{parsed.netloc}{next_url}'
+        else:
+            url = next_url
+    else:
+        url = None
+
+    return url
+
+
 def initialise_repository_if_required(func):
     '''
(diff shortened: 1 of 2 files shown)


oci/client.py
```diff
@@ -893,34 +893,55 @@ def to_digest_hash(
...
-
-        # Google-Artifact-Registry (maybe also others) will return http-200 + HTML in certain
-        # error cases (e.g. if image_reference contains a pipe (|) character)
-        try:
-            tags = res.json()['tags']
-        except json.decoder.JSONDecodeError as jde:
-            if not (content_type := res.headers['Content-Type']) == 'application/json':
-                jde.add_note(f'unexpected Content-Type: {content_type=}')
+        while url:
+            res = self._request(
```
ccwienk: can you factor-out this code into a (private or local) function? Also, it might be nicer to add `iter_tags` as additional method (which `tags` could re-use).

ccwienk (approved):
very nice - thx!

ccwienk:
@8R0WNI3 : an additional question: is there an option to choose from where to start fetching? I imagine, that, depending on use-case, and, assuming there are assumptions about ordering that can be taken, this could also be helpful (only fetch first chunk of oldest tags to determine first (smallest version, if parseable), or only fetch first chunk of newset tags to determine last (smallest version, ..)?

8R0WNI3:
At least I am not aware of any such options. From what I could see (regarding GHCR), the returned tags are ordered by creation timestamp, so only fetching a subset of chunks would at least not be sufficient to determine the greatest/smallest n-versions (according to SemVer).

ccwienk:
> At least I am not aware of any such options. From what I could see (regarding GHCR), the returned tags are ordered by creation timestamp, so only fetching a subset of chunks would at least not be sufficient to determine the greatest/smallest n-versions (according to SemVer).

I see. figured as much :-(

PR #1558, opened 2026-04-13
Improve GitHub auths resilience

ccwienk:
**Release note**:

```other operator
github-auth (used by trusted-checkout) is now more resilient in case of transient networking issues
```

### DIFF
--- a/.github/actions/github-auth/get-token.py
+++ b/.github/actions/github-auth/get-token.py
@@ -142,7 +142,7 @@ def main():
 
     token_server = args.token_server
     if '://' not in token_server:
-        token_server = f'http://{token_server}'
+        token_server = f'https://{token_server}'
 
     id_token = get_oidc_token(request_url, request_token, args.audience)
     token = exchange_token(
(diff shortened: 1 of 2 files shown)


.github/actions/github-auth/get-token.py
```diff
@@ -0,0 +1,196 @@
...
+    4xx responses are treated as systematic failures and are not retried.
+    '''
+    url = req.full_url
+    for attempt in range(retries + 1):
+        try:
+            with _opener.open(req, timeout=timeout_seconds) as resp:
+                return json.load(resp)
+        except urllib.error.HTTPError as e:
+            body = e.read().decode()
+            if e.code < 500:
```
TuanAnh17N: this treats all `<500` HTTP responses as non-retriable, but that also includes `429 Too Many Requests` (and arguably `408`), which are usually transient and worth retrying, no?

ccwienk: valid point. will update

PR #1562, opened 2026-04-17
Install `delivery` package via `odg-client` PyPI package

### LINKED ISSUE #38: Move delivery-client to OCM organistion
### Context
[delivery-client](https://github.com/gardener/cc-utils/blob/master/delivery/client.py) lives in cc-utils.
We should move it to OCM organisation, so ODG maintainers have full control over the codebase (e.g. to introduce new routes)

### Implementation Proposal
- [ ] move (or copy) delivery-client to delivery-service repository
- [ ] established new pipeline publishing pypi package for delivery-service-client
- [ ] install newly published packages whereever needed:
  - [ ] (move only) Gardener (LSS, ...?) (sync with colleagues)
  - [ ] ODG extensions

8R0WNI3:
For reference, see:
- https://github.com/open-component-model/delivery-service/pull/737
- https://github.com/open-component-model/delivery-service/pull/739
- https://github.com/open-component-model/delivery-service/pull/756

Fixes https://github.com/open-component-model/open-delivery-gear/issues/38

**Release note**:

```other developer
Removes the `delivery` package from cc-utils and instead installs it via `odg-client` PyPI package
```


ccwienk:
@8R0WNI3 : I think there is some convention on PYPI (which admittedly existing packages also do not largely follow) that distribution-package-name should reflect python-package-names. for current example, this would mean to use something along the lines of `odg[.client]`. Should we use this opportunity to also go for different package-name(s)? I think `delivery` never was a well-chosen name, at least if publishing to PYPI (for local packages, such generic terms are of course sine).

8R0WNI3:
Generally speaking, I agree, I was thinking in a very similar direction. However, I don't think we have to couple this directly to this activity as we do not yet have to adjust any usages/imports. I'd suggest to introduce a new (duplicate) client upstream alongside the existing one, and once that is in place, we can start migrating the usages to the new package. Wdyt?

ccwienk:
> Generally speaking, I agree, I was thinking in a very similar direction. However, I don't think we have to couple this directly to this activity as we do not yet have to adjust any usages/imports. I'd suggest to introduce a new (duplicate) client upstream alongside the existing one, and once that is in place, we can start migrating the usages to the new package. Wdyt?

well, technically, we can choose either approach. My rationale for suggesting to prepone renaming was to avoid exposing such disturbances to the upstream package in PYPI (at this point, it might still be an option to pull old versions)

8R0WNI3:
With https://github.com/open-component-model/delivery-service/pull/756, the `odg-client` PyPI package will also include a `odg_client` package. The `delivery` package will still be included as well during a migration period.

8R0WNI3:
@ccwienk the above mentioned change is now published, so from my PoV we are good to proceed here :-)

8R0WNI3:
@ccwienk @TuanAnh17N could you please take over this PR? Apparently, I lack permissions to update my branch anymore or the re-trigger the failed linter job.

ccwienk (changes requested):
~dependency must be declared by/in setup.gha.py (-> install-gardener-gha-libs action)~ (missed requirements.txt is honoured)

PR #1563, opened 2026-04-17
All in one updates

ccwienk:
**Release note**:

```other operator
upgrade-dependencies.yaml reusable workflow now creates all upgrade-pullrequests in one run
```

### DIFF
--- a/oci/client.py
+++ b/oci/client.py
@@ -220,6 +220,15 @@ class OciRoutes:
         )
 
 
+def _retry_after_seconds(res: requests.models.Response, fallback: float) -> float:
+    try:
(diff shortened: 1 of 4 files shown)


oci/client.py
```diff
@@ -490,15 +507,8 @@ def _authenticate(
             )
 
             if res.status_code == 429 and remaining_retries > 0:
```
TuanAnh17N, line 509: _request() now retries on 5xx, but _authenticate() still retries only on 429. that means token-service 5xx responses will still fail immediately, i think.

ccwienk: saw your comment slightly too late (sry for that)

PR #1567, opened 2026-04-20
Don't prepend docker registry to image-ref with scheme

8R0WNI3:
**Release note**:

```other operator

```

### DIFF
--- a/oci/util.py
+++ b/oci/util.py
@@ -9,8 +9,8 @@ def normalise_image_reference(image_reference: str):
     if not isinstance(image_reference, str):
         raise ValueError(image_reference)
 
-    if image_reference.startswith('/'):
-        # keep relative reference as is
+    if image_reference.startswith('/') or '://' in image_reference:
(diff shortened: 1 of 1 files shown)


ccwienk:
this change would allow OCI-Image-References that violate the [Spec](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#pulling-manifests)

> <name> refers to the namespace of the repository. <reference> MUST be either (a) the 
> digest of the manifest or (b) a tag. The <reference> MUST NOT be in any other format. 
> Throughout this document, <name> MUST match the following regular expression:
> 
> [a-z0-9]+((\.|_|__|-+)[a-z0-9]+)*(\/[a-z0-9]+((\.|_|__|-+)[a-z0-9]+)*)*

Considering `oci` package intends to follow OCI-Spec, I do not think this change is a good fit.

ccwienk:
https://github.com/gardener/cc-utils/pull/1568

PR #1568, opened 2026-04-20
ocm-schema: forbid scheme-prefixed OCI image references

ccwienk:
**Release note**:

```other operator
ocm-spec contained on `ocm` package now features strict OCI-Image-Reference schema.
```


8R0WNI3 (approved):
/lgtm, thanks!

PR #1572, opened 2026-04-21
Enhance update-images.py with directives and comment preservation

wpross:
- Add 'freeze' directive to prevent updates on specific entries
- Add 'max-supported-k8s' directive to prevent new k8s version entries
- Add 'version-mapping' directive for GCP CCM version format
- Preserve custom comments in images.yaml output
- Handle manual k8s constraints (e.g., csi-provisioner '>= 1.34')
- Fix comment indentation in YAML output

**Release note**:

```other operator
Add directives for update-images.py: `freeze` (prevent updates), `max-supported-k8s` (limit k8s
  version entries), `version-mapping` (GCP CCM version format). Custom comments above image entries
  are now preserved. Images with manual k8s constraints (e.g., `>= 1.34` unrelated to image version)
   are updated while preserving targetVersion.
```

### DIFF
--- a/.github/actions/update-extension-provider-images/update-images.py
+++ b/.github/actions/update-extension-provider-images/update-images.py
@@ -36,7 +36,7 @@ Comment Directives (one per line, placed above image entries):
 
 Comments not recognized as directives are preserved in the output.
 
-Manual k8s Constraints:
+Explicit K8s Constraints:
 Images with targetVersion like '>= 1.34' that don't correlate with the image version
 (e.g., csi-provisioner v6.x with '>= 1.34') are detected automatically. These entries
 are updated to the latest available version while preserving the original targetVersion.
@@ -52,11 +52,13 @@ are no longer supported.
 """
 
 import argparse
+import math
 import re
 import sys
 import traceback
 from collections import defaultdict
 from dataclasses import dataclass, asdict, replace
+from io import StringIO
 from typing import Optional, Any
 
 import dacite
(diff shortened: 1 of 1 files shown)


.github/actions/update-extension-provider-images/update-images.py
```diff
@@ -77,14 +98,61 @@ def update_target_version(self, new_target_version: str) -> None:
...
+
         return replace(
                 self,
                 tag=new_tag,
-                targetVersion=new_target_version
+                targetVersion=new_target_version,
+                _directives=filtered_directives if filtered_directives else None
         )
 
+    def is_frozen(self) -> bool:
```
ccwienk, line 115: as a stylistic nit, such getter-methods (any method w/o parameter) can be decorated with the `@property` decorator, like so:

```
  @property
  def foo(self): return 42

# usage
my_obj.foo # saves the `()` boilerplate
```

.github/actions/update-extension-provider-images/update-images.py
```diff
@@ -64,6 +84,7 @@ class ImageEntry:
     labels: list[OcmLabel]
     targetVersion: Optional[str] = None
     resourceId: Optional[ResourceId] = None
+    _directives: Optional[dict[str, Any]] = None  # Internal field for directives
```
ccwienk, line 87: I think adding such a "private" field to a dataclass is not a nice choice. for two reasons:
a) dataclasses are typically (/ in this repostory) used as plain datacontainers, with the occasional convenience-method
b) this slightly violates separation-of-concern. thus-far, ImageEntry is created via deserialisation using dacite

you might consider going for module-level helper functions (outside of dataclass), like so:

```
# standalone helpers instead of methods
def is_frozen(entry: ImageEntry, directives: DirectivesMap) -> bool:
    return directives.get(_key(entry), {}).get('freeze', False)


def _key(entry: ImageEntry) -> tuple[str, str, str]:
    return (entry.name, entry.repository, entry.tag)

# callers thread directives through explicitly
def update_images_data(
    images_data: ImagesData,
    new_versions_by_name_repo: dict[tuple[str, str], dict[str, list[str]]],
    directives: DirectivesMap,
) -> list[ImageUpdate]:
    ...
    for img in frozen_check_list:
        if is_frozen(img, directives):
            ...


```

.github/actions/update-extension-provider-images/update-images.py
```diff
@@ -77,14 +98,61 @@ def update_target_version(self, new_target_version: str) -> None:
...
+    def is_frozen(self) -> bool:
+        return bool(self._directives and self._directives.get('freeze', False))
+
+    def should_skip_minor_update(self) -> bool:
+        return bool(self._directives and self._directives.get('max_supported_k8s', False))
+
+    def has_version_mapping(self) -> bool:
+        return bool(self._directives and self._directives.get('version_mapping', False))
+
+    def has_manual_k8s_constraint(self) -> bool:
```
ccwienk, line 124: naming-nit: wouldn't "explicit" constraint fit better, maybe (rationale: manual to me implies actual (human) interaction, which I hope is not actually the case :-))?

.github/actions/update-extension-provider-images/update-images.py
```diff
@@ -77,14 +98,61 @@ def update_target_version(self, new_target_version: str) -> None:
...
+            target_version = parse_to_semver(target_version_str)
+            image_version = parse_to_semver(self.tag)
+
+            # If target version doesn't match image version, it's a manual constraint
+            return (target_version.major, target_version.minor) != \
+                   (image_version.major, image_version.minor)
+        except ValueError:
+            return False
+
+    def set_directives(self, directives: dict[str, Any]) -> None:
```
ccwienk, line 153: as mentioned before, _directives i somewhat a misfit for dataclasses, IMO. so are setters and getter (those are more common on enterprise Java :-))

.github/actions/update-extension-provider-images/update-images.py
```diff
@@ -133,38 +207,174 @@ class ImageUpdate:
...
     Returns:
-        Validated ImagesData object
+        Tuple of (Validated ImagesData object, raw YAML data for comment parsing)
 
     Raises:
         ValueError: If the file cannot be read, parsed, or validated
     """
+    yaml = YAML()
+    yaml.preserve_quotes = True
+    yaml.width = 4096
```
ccwienk, line 229: i think ruamel also accepts `float('inf')` (I think that was used before), as well as `math.inf`, which I think is better choice, as it expresses the intent better than just a large number

.github/actions/update-extension-provider-images/update-images.py
```diff
@@ -661,20 +1124,26 @@ def main() -> None:
     )
     args = parser.parse_args()
 
+    # Read the raw YAML content first (for comment parsing)
     try:
-        images_data = load_and_validate_images_data(args.images_yaml_path)
-    except (FileNotFoundError, yaml.YAMLError, ValueError) as e:
+        with open(args.images_yaml_path, 'r') as f:
+            yaml_content = f.read()
+        images_data, original_yaml_data = load_and_validate_images_data(args.images_yaml_path)
```
ccwienk, line 1131: nit: after you already read the yaml-file, why not pass _it_ rather than have load-and-validate read it again?

.github/actions/update-extension-provider-images/update-images.py
```diff
@@ -133,38 +207,174 @@ class ImageUpdate:
...
         ValueError: If the file cannot be read, parsed, or validated
     """
+    yaml = YAML()
+    yaml.preserve_quotes = True
+    yaml.width = 4096
+
     with open(images_yaml_path, "r") as f:
-        raw_data = yaml.safe_load(f)
+        raw_yaml_data = yaml.load(f)
+        raw_yaml_data['images']
```
ccwienk, line 233: in case this is a check for key's existence, I would prefer it to be more explicit (if 'images' not in ... raise)

.github/actions/update-extension-provider-images/update-images.py
```diff
@@ -133,38 +207,174 @@ class ImageUpdate:
...
+    lines = yaml_content.split('\n')
+    image_directives = {}
+
+    for line_num, directive_info in comment_directives.items():
+        # Find the corresponding image entry by matching name and position
+        image_name = directive_info['name']
+
+        # Count which occurrence of this image name this is
+        occurrence = 0
+        for i in range(int(line_num)):
```
ccwienk, line 363: aren't the keys (line_num) already ints?

PR #1573, opened 2026-04-22
fix(send-mail): add setup python to ensure proper python version

klocke-io:
**What this PR does:**
Add `setup-python@v5` step to the `send-mail` composite action to ensure Python 3.14+ is available.

**Which issue(s) this PR fixes:**
https://github.tools.sap/kubernetes/documentation/actions/runs/32615993/job/154188090

**Release note**:
```bugfix operator
Fix send-mail action, ensuring the proper python version is installed. 
```

### DIFF
--- a/.github/actions/send-mail/action.yaml
+++ b/.github/actions/send-mail/action.yaml
@@ -71,6 +71,11 @@ inputs:
 runs:
   using: composite
   steps:
+    - name: Setup Python
+      uses: actions/setup-python@v5
+      with:
+        python-version: '3.14'
+
     - name: Send Mail
       shell: bash
       run: |


ccwienk:
@klocke-io considering python 3.9 already is [eol](https://endoflife.date/python), and cc-utils in general does not guarantee compatibility with python < 3.11, I suggest we instead use `setup-python` action to install the python-version we need. as a value-add, this makes us independent from pre-installed python, which I deem an advantage in general. wdyt?

klocke-io:
@ccwienk PR updated, thanks for the suggestion :)

.github/actions/send-mail/action.yaml
```diff
@@ -71,6 +71,11 @@ inputs:
 runs:
   using: composite
   steps:
+    - name: Setup Python
+      uses: actions/setup-python@v5
+      with:
+        python-version: '3.11'
```
ccwienk: I suggest to pin to either 3.14, or even latest (if setup-python offers this)

ccwienk (commented):
thx - please remove first commit (for compatibility w/ python < 3.10, as this is no longer needed)

.github/actions/send-mail/action.yaml
klocke-io: Done :)

ccwienk (approved):

PR #1577, opened 2026-04-24
capture/import-commit: improve robustness, guard against darwin

ccwienk:
Re-applies cfcf2ac23, which was reverted in 1563637c2 due to a regression.

## Root cause

`find -newermt "${ref}"` treats its argument as a date/time string (not a
file path), causing GNU find to exit with an error when a file path is
passed. This broke `capture-commit`, propagating as exit 128 in downstream
`import-release-commit` steps.

## Fix

Use `find -newer "${ref}"` instead — standard GNU find behaviour: compares
mtime of found files against mtime of the reference file.

All other changes from the original commit are preserved:
- `set -euo pipefail` in `capture-commit`
- darwin/macOS guard in both `capture-commit` and `action.yaml`
- `validate-inputs` step in `import-commit/action.yaml`
- explicit `-C "${GITHUB_WORKSPACE}"` on `tar x`

### DIFF
--- a/.github/actions/capture-commit/action.yaml
+++ b/.github/actions/capture-commit/action.yaml
@@ -72,6 +72,14 @@ runs:
     - name: capture-commit
       id: capture
       run: |
+        # darwin/macOS is not supported. differences that would need addressing:
+        # - `stat --format=%s` (GNU) -> `stat -f %z` (BSD)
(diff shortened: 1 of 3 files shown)


TuanAnh17N (approved):
/lgtm

PR #1578, opened 2026-04-24
Drop usage of Delivery-Service client in CTT

### LINKED ISSUE #38: Move delivery-client to OCM organistion

8R0WNI3:
Because of the existing fallbacks in the OCM repository lookup which are integrated into the Delivery-Service, it is not a good fit for usages within the CTT where no fallbacks should be used to do qualified existence checks.
For reference, see https://github.com/open-component-model/delivery-service/pull/745.

**Release note**:

```bugfix developer
Dropped usage of Delivery-Service client in CTT
```

### DIFF
--- a/ctt/process_dependencies.py
+++ b/ctt/process_dependencies.py
@@ -16,7 +16,6 @@ import jsonschema
 import logging
 import os
 import threading
-import typing
 
 import dacite
 import yaml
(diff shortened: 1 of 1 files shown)


TuanAnh17N (approved):

ccwienk:
@8R0WNI3 : as discussed out-of-bands, I would prefer to entirely drop the delivery_service_client parameter in CTT (after having adjusted all callers).

8R0WNI3:
Yes, but that is also what this pull request is suggesting (albeit in a multi-step approach):
```python
delivery_service_client=None, # unused, drop eventually once callers have been adjusted
```

8R0WNI3:
Technically, this change would not be required anymore with https://github.com/gardener/cc-utils/pull/1592 -> leaving it up to you whether to merge or to close this PR then.

ccwienk:
> Technically, this change would not be required anymore with #1592 -> leaving it up to you whether to merge or to close this PR then.

As said, I think my preference will be do drop this parameter (incompatible change). I think for CTT, the involved complexity is not justified by the (small) benefit of reducing load against involved OCI-Registries

TuanAnh17N (approved):

PR #1586, opened 2026-04-27
integrate act-based action tests into build-and-test workflow

ccwienk:
## Summary

- Moves the `test-actions` job from the standalone `test-actions.yaml` workflow into `build-and-test.yaml` as a peer job alongside `unittests` and `lint`
- Removes the now-redundant `test-actions.yaml`
- Act tests now gate `post-build` along with all other build-and-test jobs, rather than running as a separate workflow that doesn't block release


TuanAnh17N (approved):
/lgtm

PR #1587, opened 2026-04-28
fix(github): handle missing avatar_url in release author

ccwienk:
**Release note**:

```other operator
handle absence of avatar_url in github-api-response
```


TuanAnh17N (approved):
/lgtm

PR #1591, opened 2026-04-29
fix(ocm): reject extraIdentity with non-string values via schema

ccwienk:
## Summary

- YAML \`safe_load\` parses unquoted integers (e.g. \`architecture: 386\`) as \`int\`; downstream code calling \`ComponentDescriptor.from_dict()\` then crashes with a confusing \`dacite.WrongTypeError\`
- Root cause fix: add \`additionalProperties: {type: string}\` to \`identityAttribute\` in the OCM schema so documents with non-string \`extraIdentity\` values are rejected at validation time with a clear error
- The primary fix is in the upstream repository (quoting the GHA expression), this PR ensures the schema enforces the invariant for any caller

## Changes

- \`ocm/ocm-component-descriptor-schema.yaml\`: \`identityAttribute\` now requires all values to be strings
- \`test/ocm/componentmodel_test.py\`: regression test verifies \`ComponentDescriptor.validate()\` rejects a descriptor with \`architecture: 386\` (integer)

## Test plan

- [ ] \`pytest test/ocm/\` — all 17 tests pass

### DIFF
--- a/ocm/ocm-component-descriptor-schema.yaml
+++ b/ocm/ocm-component-descriptor-schema.yaml
@@ -61,6 +61,7 @@ definitions:
   identityAttribute:
     type: 'object'
     propertyNames: { $ref: '#/definitions/identityAttributeKey' }
+    additionalProperties: { type: 'string' }
 
   repositoryContext:
     type: 'object'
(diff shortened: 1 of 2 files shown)


Michael5601 (approved):
lgtm :)

PR #1592, opened 2026-04-30
Allow disabling of Delivery-Service's OCM repository lookup

8R0WNI3:
When specifying `fallback_to_service_mapping=False` in the `delivery_service_component_descriptor_lookup`, it is expected that the built-in OCM repository lookup in the Delviery-Service is not used, and instead the (local) explicitly configured OCM repositories are to be exclusively used. To adhere to this behaviour, make use of the (new) `ignore_lookup` parameter.

**Release note**:

```other operator

```

### DIFF
--- a/cnudie/retrieve.py
+++ b/cnudie/retrieve.py
@@ -334,6 +334,7 @@ def delivery_service_component_descriptor_lookup(
                     name=component_id.name,
                     version=component_id.version,
                     ocm_repo_url=ocm_repo.oci_ref if ocm_repo else None,
+                    ignore_lookup=ocm_repo is not None, # don't use lookup if `ocm_repo` is specified
                 )
 
                 if component_descriptor:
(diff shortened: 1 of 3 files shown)


cnudie/retrieve.py
```diff
@@ -334,6 +334,7 @@ def lookup(
                     name=component_id.name,
                     version=component_id.version,
                     ocm_repo_url=ocm_repo.oci_ref if ocm_repo else None,
+                    ignore_lookup=bool(ocm_repo), # don't use lookup if `ocm_repo` is specified
```
ccwienk: `bool` will also be False for empty str, hence, I think explicit "is not None" would be better.

can you specify the first upstream version of `odg-client` that introduces this new parameter, and add a version-constraint into requirements.txt accordingly?

ccwienk (approved):
/lgtm

PR #1594, opened 2026-05-07
add github-artifact as asset type

IndritFejza:
Add github-artifact as asset type so we can use the release wf to also publish github artifacts to releases

**Release note**:

```other operator
NONE
```


IndritFejza:
Not relevant for now. Will check other approaches

PR #1595, opened 2026-05-08
fix(params): Use correct event name `pull_request` (omit `_target`)

8R0WNI3:
GitHub does not emit an event called `pull_request_target` but only a `pull_request` one. This one should be used instead to retrieve the respective author association (consistently like how it is done in the trusted-checkout action). Refs:
- https://docs.github.com/en/webhooks/webhook-events-and-payloads#pull_request
- https://github.com/gardener/cc-utils/blob/7dd669e3ca119ba3a0de8ce49815928dd25b5a3e/.github/actions/trusted-checkout/action.yaml#L255

**Release note**:

```bugfix developer
Correctly determine author association in params action to determine "can-push" status later
```

### DIFF
--- a/.github/actions/params/action.yaml
+++ b/.github/actions/params/action.yaml
@@ -134,7 +134,7 @@ runs:
             event_name='${{ github.event_name }}',
             repo_owner='${{ github.repository_owner }}',
             head_owner='${{ github.event.pull_request.head.repo.owner.login }}',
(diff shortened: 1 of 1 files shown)


ccwienk (approved):
Fix is correct. `github.event.pull_request_target` is not a valid context path; author association is at `github.event.pull_request.author_association`.

PR #1596, opened 2026-05-11
fix(ocm-upgrade): use greatest upstream ref version, not first match

ccwienk:
## Summary

When resolving the `whither-version` from an upstream component descriptor,
the previous code stopped at the first `componentReference` whose `componentName`
matched (`break`). If the upstream landscape has multiple references to the same
component (e.g. two `gardenlinux` entries at different versions for different seed
architectures), the result depended on YAML order rather than semantics.

The fix mirrors what `iter_greatest_component_references` already does for the
downstream component: scan all matching upstream refs and keep the greatest version.

## Test plan

- [ ] Verify that a landscape with duplicate upstream component-refs (same `componentName`, different versions) now gets an upgrade PR targeting the greatest upstream version rather than the first one encountered.

### DIFF
--- a/.github/actions/ocm-upgrade/ocm_upgrade.py
+++ b/.github/actions/ocm-upgrade/ocm_upgrade.py
@@ -421,15 +421,10 @@ def create_upgrade_pullrequests(
                     version=upstream_version,
                 )
             )
-            upstream_target_version = None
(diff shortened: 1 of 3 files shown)


TuanAnh17N (approved):
/lgtm

PR #1597, opened 2026-05-13
handle only ocm related files/folders in merge-ocm-fragments action

IndritFejza:
handle only ocm related resources in the merge-ocm-fragments action

**Release note**:

```other operator
NONE
```

### DIFF
--- a/.github/actions/merge-ocm-fragments/action.yaml
+++ b/.github/actions/merge-ocm-fragments/action.yaml
@@ -163,10 +163,7 @@ runs:
         cd "${{ inputs.outdir }}"
 
         # workaround: download-artifact@v3 does not allow merging. hence, flatten directories
-        for d in $(find -type d); do
-          if [ "$d" == "." ]; then
-            continue
-          fi
+        for d in $(find -type d -name "${{ steps.preprocess.outputs.artefact-name }}"); do
           echo "debug: sub-directory $d contains:"
           ls $d
           mv $d/* .


IndritFejza:
If you don't agree with the proposal what would be another way to skip non-ocm related directories from being processed by the action? 😊 

In general, the script seems to not like folders that contain subfolders. 
So, regardless of testrunner, if we have an artifact that is a folder that contains subfolders, the script fails.


```
debug: sub-directory ./A contains:
B
C.json
debug: sub-directory ./A/B contains:
ls: cannot access './A/B': No such file or directory
```

.github/actions/merge-ocm-fragments/action.yaml
```diff
@@ -167,6 +167,10 @@ runs:
           if [ "$d" == "." ]; then
             continue
           fi
+          if [[ "$d" == "./testrunner"* ]]; then
+            echo "skipping testrunner directory $d"
+            continue
+          fi
```
IndritFejza:
```suggestion
          if [ ! -d "$d" ]; then
            continue
          fi
          if [[ "$d" == "./testrunner"* ]]; then
            echo "skipping testrunner directory $d"
            continue
          fi
```

TuanAnh17N:
thanks for the pr!
the `[ ! -d "$d" ]` guard fixes the immediate crash, but i'd suggest a different approach that avoids hardcoding caller-specific directory names in this shared action.

the root cause seems to be that the flatetning loop iterates over all subdirectories, including non-ocm artifacts. since ocm fragments are always `.tar.gz` files (that's the contract of export-ocm-fragemnts), we can just move those directly instead of flattening everything, i think 🤔

```bash
find . -mindepth 2 -name "*.tar.gz" -exec mv {} . \;

echo "extracting files with pattern *.tar.gz"
for tarffile in $(find . -maxdepth 1 -name "*.tar.gz"); do
  echo "extracting ${tarfile} into $PWD"
  tar xf "${tarfile}"
  unlnk "${tarfile}"
done

```
since the flattening loop can't be removed entirely - download-artifact@v3 (used on GHE) doesn't support merge-multiple (according to the comment above) - scoping it to .tar.gz files, seems right to me

IndritFejza:
> thanks for the pr! the `[ ! -d "$d" ]` guard fixes the immediate crash, but i'd suggest a different approach that avoids hardcoding caller-specific directory names in this shared action.
> 
> the root cause seems to be that the flatetning loop iterates over all subdirectories, including non-ocm artifacts. since ocm fragments are always `.tar.gz` files (that's the contract of export-ocm-fragemnts), we can just move those directly instead of flattening everything, i think 🤔
> 
> ```shell
> find . -mindepth 2 -name "*.tar.gz" -exec mv {} . \;
> 
> echo "extracting files with pattern *.tar.gz"
> for tarffile in $(find . -maxdepth 1 -name "*.tar.gz"); do
>   echo "extracting ${tarfile} into $PWD"
>   tar xf "${tarfile}"
>   unlnk "${tarfile}"
> done
> ```
> 
> since the flattening loop can't be removed entirely - download-artifact@v3 (used on GHE) doesn't support merge-multiple (according to the comment above) - scoping it to .tar.gz files, seems right to me

sound good. Will you follow up? I can close the PR

.github/actions/merge-ocm-fragments/action.yaml
```diff
@@ -167,6 +167,13 @@ runs:
           if [ "$d" == "." ]; then
             continue
           fi
+          if [ ! -d "$d" ]; then
+            continue
```
ccwienk: dead code (find is called with `-type d`, so will only ever yield directories in the first place)

ccwienk:
Thanks for the fix! A few thoughts:

**Why do `testrunner` directories end up in `outdir` at all?**
The `import-ocm-fragments` step downloads artifacts matching `*.ocm-artefacts` — so either a `testrunner`-related artifact matches that glob (naming issue?), or something else is placing that directory into `outdir` before this step runs. Worth understanding the root cause.

**Alternative: drop the flatten loop entirely**

Each exported artifact (from `export-ocm-fragments`) contains exactly one `*.tar.gz` file. After `download-artifact` places them into per-artifact subdirectories, the structure is:

```
outdir/
  <artifact-name>/
    <hash>.ocm-artefacts.tar.gz
```

The subsequent `find -name "*.tar.gz"` loop already recurses — so it will find those files without any flattening. The blobs end up in the right place because they are *inside* the tar, not alongside it.

That would reduce the `extract-ocm-fragments` step to:

```sh
set -eu
cd "${{ inputs.outdir }}"

echo "extracting ocm-fragment archives"
for tf in $(find . -name "*.tar.gz"); do
  echo "extracting ${tf}"
  tar xf "${tf}"
  unlink "${tf}"
done
```

No guards, no hardcoded names. The `testrunner` directory (and any other unrelated subdirectory) is simply never touched — as long as it contains no stray `*.tar.gz` files.

@IndritFejza — does the `testrunner` directory ever contain any `*.tar.gz` files?

IndritFejza:
@ccwienk 

>Why do testrunner directories end up in outdir at all?

I think it has to do with the fact that we use [download-artifact@v3](https://github.com/IndritFejza/cc-utils/blob/7dd669e3ca119ba3a0de8ce49815928dd25b5a3e/.github/actions/merge-ocm-fragments/action.yaml#L165)

It downloads **all** artifacts that exist in the wf-artifacts. It ignored `pattern`. So not only testrunner artifects end up in the outdir, every other artifact is.

v4 supports the `pattern` but that's not used in our GHE.

IndritFejza:
To not be blocked by this issue, we decided to just tar the artifact that we publish in the wf-artifacts.

As I stated [before](https://github.com/gardener/cc-utils/pull/1597#issuecomment-4440697961), seems that the existing logic here doesn't work with having `folders that contain subfolders`

---

But I believe as soon as another party/or-us decide to publish other wf artifacts, we would be blocked to work only with tar/zip folders and not just plain folders.

IndritFejza:
Maybe an easier solution is to just do the filtering at the beginning based on the pattern `*.ocm-artefacts` again.

So handle only files that end with `ocm-artefacts` and ignore others. Basically re-writing the same logic as what `pattern` is supposed to do in `download-artifact`

Will this new proposal work?

ccwienk:
> Maybe an easier solution is to just do the filtering at the beginning based on the pattern `*.ocm-artefacts` again.
> 
> So handle only files that end with `ocm-artefacts` and ignore others. Basically re-writing the same logic as what `pattern` is supposed to do in `download-artifact`
> 
> Will this new proposal work?

yes, I think explicitly choosing files of which we know the naming convention would be preferable to hardcoding exceptions for some "random" (from pipeline-template's pov) fnames.

also, please cleanup commits (probably squash into just one)

.github/actions/merge-ocm-fragments/action.yaml
```diff
@@ -163,10 +163,7 @@ runs:
...
 
         # workaround: download-artifact@v3 does not allow merging. hence, flatten directories
-        for d in $(find -type d); do
-          if [ "$d" == "." ]; then
-            continue
-          fi
+        for d in $(find -type d -name "${{ steps.preprocess.outputs.artefact-name }}"); do
           echo "debug: sub-directory $d contains:"
           ls $d
           mv $d/* .
```
IndritFejza, line 169: still not super sure if this PR/change is fine. 

Maybe we should put first only ocm related files/artifacts in a temp/different folder and not play around directly on the ${{ inputs.outdir }} ... because (in GHE) there might be other non ocm-related files/folders in it and we might extract them for no reason.

ccwienk:
Superseded by #1626, which takes a slightly different approach: instead of filtering directories by name in the flatten step, the flatten step is dropped entirely — `find . -name "*.ocm-artefacts.tar.gz"` recurses into subdirectories naturally, and the filename pattern already excludes non-OCM artifacts. Thanks for the analysis and root-cause investigation!

PR #1598, opened 2026-05-27
fix(oci): propagate replication mode in recursive `replicate_artifact` calls

fabianburth:
Without this, nested indexes fail because the recursive call defaults to `REGISTRY_DEFAULTS`, losing the Accept header needed to resolve sub-indexes.

**Note for the reviewer:**
I confirmed the fix with a manual script run.

**Release note**:

```bugfix user
Support replication of nested index manifests
```

### DIFF
--- a/oci/__init__.py
+++ b/oci/__init__.py
@@ -203,6 +203,7 @@ def replicate_artifact(
                     src_image_reference=src_reference,
                     tgt_image_reference=tgt_reference,
                     oci_client=client,
+                    mode=mode,
                     annotations=annotations,
                 )
 
--- a/oci/__init__.py
+++ b/oci/__init__.py
@@ -199,12 +199,17 @@ def replicate_artifact(
 
                 logger.info(f'replicating to {tgt_reference=}')
 
+                # only propagate PREFER_MULTIARCH (preserves nested indices); never pass
+                # NORMALISE_TO_MULTIARCH as it would wrap sub-manifests in spurious index layers
+                recursive_mode = ReplicationMode.REGISTRY_DEFAULTS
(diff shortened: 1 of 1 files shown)


oci/__init__.py
```diff
@@ -203,6 +203,7 @@ def replicate_artifact(
                     src_image_reference=src_reference,
                     tgt_image_reference=tgt_reference,
                     oci_client=client,
+                    mode=mode,
```
fabianburth: if we support recursing deeper, I'm wondering whether we should also pass the `platform_filter` option along

oci/__init__.py
zkdev, line 269: what about the second recursive invocation?
I think we should be consistent here.

oci/__init__.py
zkdev: agree, without passing the platform filter the transferred artefact could become inconsistent

oci/__init__.py
fabianburth: I did think about it, but since it is behind
```
        elif media_type in (
            om.OCI_MANIFEST_SCHEMA_V2_MIME,
            om.DOCKER_MANIFEST_SCHEMA_V2_MIME,
        ):
```
I believe this being an index manifest into which we could traverse deeper is not valid.

zkdev: what if in future one would introduce a mode / filter which in fact is relevant also for single manifests?

oci/__init__.py
fabianburth: Went back on that and opted for the most conservative approach possible to avoid any unexpected side effects. Feel free to object!

TuanAnh17N (approved):

zkdev (approved):
/lgtm

PR #1599, opened 2026-05-27
fix(ci): Avoid collisions of capture-commit archive

zkdev:
When using the optional release-branch feature, the capture-commit archive is not cleaned up. As both steps will run on the same worker (with the very same filesystem), there can be collisions if the archive names match.

Clean up after archive upload is safe, as subsequent steps only consume the archive via explicit artifact download.

**Release note**:

```bugfix developer
NONE
```

### DIFF
--- a/.github/actions/capture-commit/action.yaml
+++ b/.github/actions/capture-commit/action.yaml
@@ -154,3 +154,11 @@ runs:
       with:
         name: ${{ inputs.to-artefact }}
         path: commit-objects.tar.gz
+
+    - name: clean-up-capture-commit-archive
+      shell: bash
+      run: |
+        set -euo pipefail
+
+        # archive was uploaded already, avoid local collisions in filesystem
+        unlink commit-objects.tar.gz


.github/actions/capture-commit/action.yaml
```diff
@@ -154,3 +154,10 @@ runs:
       with:
         name: ${{ inputs.to-artefact }}
         path: commit-objects.tar.gz
+
+    - name: clean-up-capture-commit-archive
+      shell: bash
+      id: cleanup
```
ccwienk: nit: id is never used

.github/actions/capture-commit/action.yaml
```diff
@@ -154,3 +154,10 @@ runs:
...
+      run: |
+        # archive was uploaded already, avoid local collisions in filesystem
+        unlink "commit-objects.tar.gz"
```
ccwienk: two nits:
- quotes are not needed
- should at least `set -e` (or: `set -euo pipefail`) as general best-practice and to detect regressions (e.g. inconsistent renaming)

ccwienk (commented):
some nits, otherwise: lgtm

ccwienk (approved):

PR #1600, opened 2026-05-29
remove github/compliance package

TuanAnh17N:
the github.compliance module was only used to create cfg-related issues on concourse, which is no longer in use. compliance issues are now managed by odg, which does not depend on this module.

additionally, the odg-client dropped the sprint model in a recent change, causing lint failures in milestone.py which would require adapting dead code.

**Release note**:

```other operator

```


8R0WNI3 (approved):
/lgtm, thank you :-)

zkdev (approved):
/lgtm

I am also not aware of any remaining usages for this package

PR #1603, opened 2026-06-08
fix(oci): Only check content-type in case of error in tag listing

8R0WNI3:
Apparently, AWS ECR is using the `Content-Type: text/plain` even for valid JSON responses in the tag listing route. Therefore, don't fixate the content-type during deserialisation already, but only in case an error occurs (that part was already implemented).

**Release note**:

```bugfix developer
The async OCI client is not enforcing a content-type as part of the deserialisation of tags (AWS ECR is using `text/plain` instead of `application/json`)
```

### DIFF
--- a/oci/client_async.py
+++ b/oci/client_async.py
@@ -610,7 +610,9 @@ class Client:
         # Google-Artifact-Registry (maybe also others) will return http-200 + HTML in certain
         # error cases (e.g. if image_reference contains a pipe (|) character)
         try:
-            tags_res = await res.json()
+            tags_res = await res.json(
+                content_type=None, # AWS ECR is using `text/plain` instead of `application/json`
+            )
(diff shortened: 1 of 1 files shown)


TuanAnh17N (approved):

oci/client_async.py
```diff
@@ -610,7 +610,7 @@ async def _tags_single_request(
         # Google-Artifact-Registry (maybe also others) will return http-200 + HTML in certain
         # error cases (e.g. if image_reference contains a pipe (|) character)
         try:
-            tags_res = await res.json()
+            tags_res = await res.json(content_type=None)
```
ccwienk: might be worth adding a comment explaining the motivation behind disabling content-type-check.

ccwienk (approved):

PR #1604, opened 2026-06-10
add SPDX documents for external OCI images

ccwienk:
Extends the existing SPDX generation to cover external OCI image resources
(`relation: external`) in addition to local ones.

SPDX documents are cached in the component's own OCI registry
(`<registry>/spdx-cache/<repo>@<digest>`) to avoid redundant scans and
to work around the read-only nature of upstream registries.

### DIFF
--- a/.github/actions/merge-ocm-fragments/action.yaml
+++ b/.github/actions/merge-ocm-fragments/action.yaml
@@ -72,6 +72,32 @@ inputs:
     description: |
       the ocm-component-prefixes to use in case post-processing is set to `component-cli`.
 
+  oci-registry:
+    required: false
+    type: string
+    description: |
+      If passed, SPDX documents are generated for all external OCI resources and stored in
+      a cache repository under this registry. OCI authentication is performed against the
+      registry automatically (OIDC preferred; the calling job must have `id-token: write`).
+
+      When omitted, the action behaves exactly as before — no SPDX generation, no OCI auth.
+
+  sbom-cache-repo-prefix:
+    required: false
+    type: string
+    default: sbom-cache
+    description: |
+      Repository-name prefix used when storing cached SBOM documents inside `oci-registry`.
+      Defaults to `sbom-cache`.
+
+  spdx-force-rescan:
+    required: false
+    type: boolean
+    default: false
+    description: |
+      When set to `true`, the SBOM cache is ignored and all external images are re-scanned.
+      Useful when a newer syft version is desired. Defaults to `false`.
+
 outputs:
   component-descriptor:
     description: |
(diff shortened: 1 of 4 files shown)


.github/actions/merge-ocm-fragments/spdx_external.py
```diff
@@ -0,0 +1,771 @@
...
+    syft_ver: str | None,
+    tmpdir: str,
+) -> tuple[_ImageInfo, bytes, bytes] | None:
+    info = miss.info
+    resource_name = info.resource['name']
+    source_ref = om.OciImageReference.to_image_ref(info.digest_ref)
+    cache_repo = _cache_repo(cache_registry, cache_prefix, source_ref)
+
+    try:
+        with tempfile.TemporaryDirectory() as tmp:
```
TuanAnh17N: nit: i'm not sure on this. should this be `TemporaryDirectory(dir=tmpdir)` to keep output files on the same filesystem as syft's working space? the module docstring says TMPDIR is forwarded so everything lands on the same filesystem, but this doesn't follow through 🤔

ccwienk: good catch!

.github/actions/merge-ocm-fragments/spdx_external.py
```diff
@@ -0,0 +1,772 @@
...
+    oci_client: oc.Client,
+    cache_registry: str,
+    cache_prefix: str,
+) -> tuple[_ImageInfo, bytes, bytes] | None:
+    source_ref = om.OciImageReference.to_image_ref(hit.info.digest_ref)
+    cref = _cache_ref(cache_registry, cache_prefix, source_ref, hit.info.source_digest)
+    cache_repo = _cache_repo(cache_registry, cache_prefix, source_ref)
+    try:
+        manifest_res = oci_client.manifest_raw(cref)
+        manifest = yaml.safe_load(manifest_res.content)
```
TuanAnh17N: i remember we used `json.loads` for parsing OCI manifests (prob. in oci/client.py). we could use it here aswell

ccwienk: indeed. technically, json is always yaml-compliant, but we will safe some cpu-load by using json-parser :-)

.github/actions/merge-ocm-fragments/action.yaml
```diff
@@ -312,11 +338,46 @@ runs:
...
+
     - name: output-component-descriptor
       id: output
       shell: bash
       run: |
         set -eu
         echo "component-descriptor<<EOF" >> "${GITHUB_OUTPUT}"
         cat "${{ inputs.outdir }}/component-descriptor.yaml" >> "${GITHUB_OUTPUT}"
-        echo EOF >> ${GITHUB_OUTPUT}
+        echo EOF >> ${GITHUB_OUTPUT}
```
TuanAnh17N: nit: missing newline at end of file

ccwienk: wow! that's something i really hate.... have to tell claude about this :-)

TuanAnh17N (approved):
/lgtm

PR #1606, opened 2026-06-11
feat(sbom): add CycloneDX 1.6 SBOM generation alongside SPDX

ccwienk:
Extends SBOM generation in the \`oci-ocm\` workflow to produce both SPDX and CycloneDX 1.6 documents from a single syft scan.

**Changes:**
- Replace \`anchore/sbom-action\` with a direct \`syft scan\` invocation using two \`-o\` outputs — one scan, no redundant work
- Push both SPDX and CycloneDX 1.6 documents as OCI referrer manifests (OCI 1.1)
- Inline SBOM resources use the payload's own name; \`sbom-format\` in \`extraIdentity\` (e.g. \`spdx-2.3\`, \`cyclonedx-1.6\`) distinguishes variants — more idiomatic OCM than encoding format in the resource name
- OCI referrer resources use \`{name}-sbom-ref\` suffix to avoid needing an access-kind extraIdentity key
- New label schema \`gardener.cloud/sbom\` with structured \`data-source\` (tool, version, kind) and \`format\` fields, replacing the old \`gardener.cloud/sbom/syft\` label
- \`oci.sbom.push_sbom_referrer\`: rename \`syft_version\` → \`tool_version\`; align OCI annotation key to \`gardener.cloud/sbom/tool-version\`

No existing consumers of the old resource names / label — schema change is intentional.

**Release note**:
\`\`\`feature operator
SBOM generation now produces both SPDX 2.3 and CycloneDX 1.6 documents per image, attached as OCI referrer manifests and OCM resources.
\`\`\`

### DIFF
--- a/.github/workflows/oci-ocm.yaml
+++ b/.github/workflows/oci-ocm.yaml
@@ -217,8 +217,9 @@ on:
         default: true
         required: false
         description: |
-          Whether to generate an SPDX SBOM for each built image and push it as an OCI referrer
-          manifest alongside the image. Requires the image to have been pushed (`push != never`).
+          Whether to generate SPDX and CycloneDX SBOMs for each built image and push them as OCI
+          referrer manifests alongside the image. Requires the image to have been pushed
+          (`push != never`).
     secrets:
       auth-token:
         required: false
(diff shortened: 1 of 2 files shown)


.github/workflows/oci-ocm.yaml
```diff
@@ -582,56 +592,83 @@ jobs:
...
+              extra_identity = {
+                  'version': version,
+                  'os': os_name,
+                  'architecture': architecture,
+                  'sbom-format': sbom_format,
+              }
+              label = make_label(sbom_format)
+              labels = [label] if label else []
+              inline = ocm.Resource(
+                  name=name,
```
TuanAnh17N, line 624: the pr descirption says inline resources are named `{name}-sbom`, but here it's just `name` 🤔

.github/workflows/oci-ocm.yaml
```diff
@@ -507,29 +508,26 @@ jobs:
...
-          format: spdx-json
-          output-file: sbom.spdx.json
-          upload-artifact: false
-          upload-release-assets: false
+        uses: anchore/sbom-action/download-syft@v0
 
-      - name: 'get syft version / ${{ matrix.args.platform-name }}'
+      - name: 'generate SBOM / ${{ matrix.args.platform-name }}'
         if: ${{ inputs.generate-sbom && needs.preprocess.outputs.push == 'true' }}
         id: syft-version
```
TuanAnh17N: nit: step now does both scan and version extrction - `id: sbom-generate` or simlar might read more clealry

.github/workflows/oci-ocm.yaml
ccwienk: good catch! that was a last-minute-change ;-)

.github/workflows/oci-ocm.yaml
ccwienk: also: good catch

TuanAnh17N (approved):
/lgtm

PR #1608, opened 2026-06-12
install syft to ~/.local/bin to fix GHE runner permission error

ccwienk:
## Summary

- GHE runners run as a non-root user without write access to `/usr/local/bin`
- The existing syft install used `curl ... | sh -s -- -b /usr/local/bin`, which fails with `Permission denied`
- Fix: install to `~/.local/bin` (user-writable) and append to `GITHUB_PATH`

Affects three places:
- `.github/actions/merge-ocm-fragments/action.yaml` (the reported regression)
- `.github/workflows/build-and-test.yaml` (same pattern)
- `.github/workflows/oci-ocm.yaml` (previously used `anchore/sbom-action/download-syft@v0`, replaced with inlined equivalent that works on GHE)

Validated on a GHE runner (install + `syft scan alpine:latest`).

### DIFF
--- a/.github/actions/merge-ocm-fragments/action.yaml
+++ b/.github/actions/merge-ocm-fragments/action.yaml
@@ -342,8 +342,11 @@ runs:
       if: ${{ inputs.oci-registry != '' }}
       shell: bash
       run: |
+        install_dir="${HOME}/.local/bin"
+        mkdir -p "${install_dir}"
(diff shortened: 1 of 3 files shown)


IndritFejza (approved):
/lgtm

TuanAnh17N (approved):

PR #1621, opened 2026-06-17
sbom-upload: add reusable workflow for uploading SBOMs to GCS

ccwienk:
Adds a reusable `workflow_call` workflow that fetches SBOM documents for all
resources in an OCM component tree and uploads them to a GCS bucket.

All caller-specific parameters are explicit inputs — no environment-specific
defaults are hardcoded.

**Features**
- Adaptive upload concurrency (AIMD, 1–8 workers) to avoid rate-limiting
- Incremental: skips resources/blobs already present in the bucket
- 429 retried up to 20× (with `Retry-After` + jitter); 5xx retried 3×; other 4xx fail immediately
- OCI fetches parallelised (16 workers); uploads governed by adaptive throttle
- Token exchange via the `oidc-gcp-token-exchange` composite action (also in this repo)

**Release note**:
```feature operator
Reusable workflow `sbom-upload.yaml` for uploading OCM component SBOM documents to a GCS bucket.
```

### DIFF
--- a/.github/workflows/sbom-upload.yaml
+++ b/.github/workflows/sbom-upload.yaml
@@ -1,0 +1,115 @@
+name: Upload SBOMs to GCS
+
+# Reusable workflow: fetches SBOM documents for all resources in an OCM component
+# tree and uploads them to a GCS bucket via the Google Cloud Storage JSON API.
+#
+# Designed for self-hosted runners that have access to a GCP instance-metadata
+# endpoint (needed by the oidc-gcp-token-exchange action for the dual-token auth).
+#
+# Caller example:
+#
+#   jobs:
+#     sbom-upload:
+#       uses: gardener/cc-utils/.github/workflows/sbom-upload.yaml@v1
+#       with:
+#         ocm-component: 'github.com/my-org/my-component:1.2.3'
+#         ocm-repositories: |
+#           europe-docker.pkg.dev/my-gcp-project/releases
+#         oci-registry: europe-docker.pkg.dev/my-gcp-project
+#         gcs-bucket: my-cumulus-pipeline-id
+#         token-exchange-api-url: 'https://api.example.com'
+#         token-exchange-audience: 'api://MyService'
+#         token-exchange-pipeline-id: 'my-pipeline'
+#         token-exchange-pipeline-group-id: 'my-org/my-project'
+#         token-exchange-scope: 'my-scope'
+#       permissions:
+#         id-token: write
+
+on:
+  workflow_call:
+    inputs:
+      ocm-component:
+        description: 'Root OCM component in name:version format'
+        required: true
+        type: string
+      ocm-repositories:
+        description: 'Newline-separated list of OCM repository base URLs'
+        required: true
+        type: string
+      oci-registry:
+        description: >
+          OCI registry host (or image-reference prefix) to authenticate against,
+          passed to the oci-auth action. Must cover all OCM repositories.
+        required: true
+        type: string
+      gcs-bucket:
+        description: 'GCS bucket name (pipeline ID) to upload SBOMs into'
+        required: true
+        type: string
+      token-exchange-api-url:
+        description: 'Base URL of the token exchange API'
+        required: true
+        type: string
+      token-exchange-audience:
+        description: 'OIDC audience for both the GitHub and GCP identity tokens'
+        required: true
+        type: string
+      token-exchange-pipeline-id:
+        description: 'Pipeline identifier passed to the token exchange API'
+        required: true
+        type: string
+      token-exchange-pipeline-group-id:
+        description: 'Pipeline group identifier passed to the token exchange API'
+        required: true
+        type: string
+      token-exchange-scope:
+        description: 'Scope to request from the token exchange API (e.g. "cumulus")'
+        required: true
+        type: string
+      runner:
+        description: 'Runner label(s) as a JSON array string, e.g. "[\"self-hosted\"]"'
+        required: false
+        type: string
+        default: '["self-hosted"]'
+
+jobs:
+  sbom-upload:
+    runs-on: ${{ fromJSON(inputs.runner) }}
+    permissions:
+      id-token: write
+
+    steps:
+      - name: Exchange tokens
+        uses: gardener/cc-utils/.github/actions/oidc-gcp-token-exchange@master
+        with:
+          api-url: ${{ inputs.token-exchange-api-url }}
+          oidc-audience: ${{ inputs.token-exchange-audience }}
+          pipeline-id: ${{ inputs.token-exchange-pipeline-id }}
+          pipeline-group-id: ${{ inputs.token-exchange-pipeline-group-id }}
+          systems: ${{ inputs.token-exchange-scope }}
+          token-env-prefix: 'GCS_'
+        # exports GCS_<SCOPE_UPPER>_TOKEN into $GITHUB_ENV
+
+      - name: Export GCS token
+        id: gcs-token
+        shell: bash
+        env:
+          SCOPE: ${{ inputs.token-exchange-scope }}
+        run: |
+          upper=$(echo "$SCOPE" | tr '[:lower:]' '[:upper:]')
+          token_var="GCS_${upper}_TOKEN"
+          echo "gcs-token=${!token_var}" >> "$GITHUB_OUTPUT"
+
+      - name: Authenticate to OCI registry
+        uses: gardener/cc-utils/.github/actions/oci-auth@master
+        with:
+          gh-token: ${{ secrets.GITHUB_TOKEN }}
+          oci-image-reference: ${{ inputs.oci-registry }}
+
+      - name: Fetch and upload SBOMs
+        uses: gardener/cc-utils/.github/actions/sbom-upload@sbom-upload-workflow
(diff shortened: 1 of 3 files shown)


.github/workflows/sbom-upload.yaml
```diff
@@ -0,0 +1,338 @@
...
+        type: string
+      runner:
+        description: 'Runner label(s) as a JSON array string, e.g. "[\"self-hosted\"]"'
+        required: false
+        type: string
+        default: '["self-hosted"]'
+
+jobs:
+  sbom-upload:
+    runs-on: ${{ fromJSON(inputs.runner) }}
```
Michael5601, line 77: **question** **(non-blocking):** Is it relevant for the caller to decide on which runner to perform the workflow? Can this be made simpler by hardcoding any capable runner?

.github/workflows/sbom-upload.yaml
```diff
@@ -0,0 +1,338 @@
...
+      id-token: write
+      contents: read
+
+    env:
+      GCS_BUCKET: ${{ inputs.gcs-bucket }}
+      OCM_COMPONENT: ${{ inputs.ocm-component }}
+      OCM_REPOSITORIES: ${{ inputs.ocm-repositories }}
+
+    steps:
+      - uses: actions/checkout@v4
```
Michael5601: **suggestion (blocking):** Should be changed to v6 like the rest of the repository already uses.

.github/workflows/sbom-upload.yaml
```diff
@@ -0,0 +1,338 @@
...
+          echo "GCS_TOKEN=${!token_var}" >> "$GITHUB_ENV"
+
+      - name: Authenticate to OCI registry
+        uses: gardener/cc-utils/.github/actions/oci-auth@v1
+        with:
+          gh-token: ${{ secrets.GITHUB_TOKEN }}
+          oci-image-reference: ${{ inputs.oci-registry }}
+
+      - name: Install cc-utils
+        uses: gardener/cc-utils/.github/actions/install-gardener-gha-libs@master
```
Michael5601: **suggestion (blocking):** The action points on master `(@master)` which means a push on master from anyone changes this workflows behavior. It should be changed to `@v1` like the other actions in this workflow.

.github/workflows/sbom-upload.yaml
```diff
@@ -0,0 +1,338 @@
...
+          errors = 0
+
+          def _process(item):
+              node, mapping, fmt_id, blob = item
+              try:
+                  data = si.fetch_sbom_document(mapping, oci_client)
+              except Exception as e:
+                  print(f'error: {node.resource.name} ({fmt_id}): {e}', file=sys.stderr)
+                  return 'error'
+              name_out = _gcs_upload(data, blob)
```
Michael5601: **suggestion (blocking):** The method `_gcs_upload` can raise but is not wrapped in try-except like the method `fetch_sbom_document` directly above.

.github/workflows/sbom-upload.yaml
```diff
@@ -0,0 +1,338 @@
...
+
+      - name: Fetch and upload SBOMs
+        env:
+          GCS_TOKEN: ${{ env.GCS_TOKEN }}
+        run: |
+          python3 - <<'EOF'
+          import json, os, sys, urllib.request, urllib.parse
+
+          import cnudie.retrieve
+          import oci.auth
```
Michael5601: **Thought:** It is rather hard to review, lint, debug or change an inlined python script. Especially if it is very big. Could this be outsourced in a separate file like it's done in the actions?

Michael5601 (changes requested):
Most looks good to me, I only have some smaller suggestions.

.github/workflows/sbom-upload.yaml
TuanAnh17N: Is the chekcout step actualyl needed here? i don't see any step reading from the working tree 🤔  and if not could `contents: read` be dropped as well?

.github/workflows/sbom-upload.yaml
```diff
@@ -0,0 +1,338 @@
...
+        with:
+          gh-token: ${{ secrets.GITHUB_TOKEN }}
+          oci-image-reference: ${{ inputs.oci-registry }}
+
+      - name: Install cc-utils
+        uses: gardener/cc-utils/.github/actions/install-gardener-gha-libs@master
+
+      - name: Fetch and upload SBOMs
+        env:
+          GCS_TOKEN: ${{ env.GCS_TOKEN }}
```
TuanAnh17N: `GCS_TOKEN` was already written to `$GITHUB_ENV` by the prior step, so it should be available to this step automatically

.github/workflows/sbom-upload.yaml
```diff
@@ -0,0 +1,338 @@
...
+                  if k not in ('sbom-format', 'version')
+              ]
+              extra = ('-' + '-'.join(extra_vals)) if extra_vals else ''
+              return (
+                  f'{_mangle(component.name)}:{component.version}'
+                  f'_{_mangle(resource.name)}:{resource.version}'
+                  f'{extra}.sbom.{fmt_id}'
+              )
+
+          import concurrent.futures, threading, time, random
```
TuanAnh17N: why the late import? 🤔

.github/workflows/sbom-upload.yaml
```diff
@@ -0,0 +1,338 @@
...
+
+          class _UploadThrottle:
+              '''Adaptive concurrency for GCS uploads; shrinks on 429, grows on success.'''
```
TuanAnh17N: looking at the code is this actually AIAD? 🤔 (both sides do += 1 / -= 1)

.github/workflows/sbom-upload.yaml
```diff
@@ -0,0 +1,338 @@
...
+                          name = json.load(r)['name']
+                      throttle.release(grew=True)
+                      return name
+                  except urllib.error.HTTPError as e:
+                      if e.code == 429:
+                          n429 += 1
+                          throttle.release(shrank=True)
+                          if n429 >= MAX_429:
+                              raise
+                          wait = int(e.headers.get('Retry-After', min(2 ** n429, 120)))
```
TuanAnh17N: i'm not to familiar with this but is it safe to use int(...) here? i remember darkly that this can be a date aswell

.github/workflows/sbom-upload.yaml
ccwienk: good catch. actually, it should rather use our trusted-checkout...

ccwienk: indeed - that's something i also stumbled upon when reviewing another workflow. seems to be a claude-artefact.....

.github/workflows/sbom-upload.yaml
ccwienk: If I am not mistaken, we configure all references (to other actions from this here repository) to point to `master`. they are rewritten ("pinned") to v1 as part of release-workflow.

.github/workflows/sbom-upload.yaml
ccwienk: that's correct. I think that's a case of "explicit is better than implicit", but does not adhere to our conventions thus-far.

.github/workflows/sbom-upload.yaml
ccwienk: it could. that's always a trade-off:

Unfortunately, different to actions, workflows do not feature the option to implicitly take files from their directory with them, so we would have to retrieve a non-inlined script, e.g. using install-gardener-gha-libs (and find a reasonable location for the script). Depending on the length of the script, and the amount of inputs to be passed, this may make quite an ugly interface.

-> will check

However, after inspecting this script some more, i also do not like the pattern of using environment-variables for passing values. this also makes things rather cumbersome to test locally. I tend towards making this another action, and keep the workflow just as a means to call it.

.github/workflows/sbom-upload.yaml
ccwienk: I think only claude should know ^^

.github/workflows/sbom-upload.yaml
ccwienk: what's AIAD?

.github/workflows/sbom-upload.yaml
ccwienk: we should definitively guard against unexpected values / parsing-errors

.github/workflows/sbom-upload.yaml
ccwienk: again: good catch

.github/workflows/sbom-upload.yaml
TuanAnh17N: additive increase, additive decrease - boht sides change the target by 1 (as opposed to AIMD where the decrease would be multiplicative e.g. havling) 🙈

.github/workflows/sbom-upload.yaml
ccwienk: reworked script into a standalone-script w/ CLI into a (new) action

.github/workflows/sbom-upload.yaml
ccwienk: right - updated accordingly

Michael5601 (approved):

PR #1622, opened 2026-06-18
release: fix Asset.matches() to reject resources with extra extraIdentity keys

ccwienk:
## Problem

SBOM resources (added by the oci-ocm.yaml workflow) share `name`, `os`, and
`architecture` with the binary resource they describe — as required by the
naming convention in `sbom/README.md`. They are distinguished by an additional
`sbom-format` key in `extraIdentity`.

`Asset.matches()` did a subset check: it only verified fields present in
`asset.id`, silently ignoring any extra `extraIdentity` keys on the resource.
This caused both the binary and the SBOM inline resource to match, producing:

```
Error: asset=… is ambiguous (more than one matching OCM-Resource)
```

## Fix

After the existing field-value loop, reject a resource if its `extraIdentity`
contains keys not covered by `asset.id`. OCM identity is `name + extraIdentity`
in full — an extra key makes it a distinct resource, not the same one.

## Effect

- Binary resources (no `sbom-format` in extraIdentity) continue to match as before.
- SBOM resources no longer match when callers don't specify `sbom-format` in `id`.
- No changes required on the gardener/gardener side.

### DIFF
--- a/.github/actions/release/release.py
+++ b/.github/actions/release/release.py
@@ -35,7 +35,10 @@ class Asset:
             if not resource_value == v:
                 return False
 
-        return True
+        # reject if resource carries extraIdentity keys not covered by id —
+        # such keys are part of the OCM identity and make it a distinct resource
+        # (e.g. SBOM resources share name/os/arch with binaries but add sbom-format)
+        return not (resource.extraIdentity.keys() - self.id.keys())
 
     def __post_init__(self):
         # basic validation: id.name must be present and non-empty
--- a/test/actions/release/test_release.py
+++ b/test/actions/release/test_release.py
@@ -16,6 +16,7 @@ sys.path.insert(
 
 import release
 import release_notes.ocm as rn_ocm
+import ocm
 
 
 def _make_component(version='1.0.0'):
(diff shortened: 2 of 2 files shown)


Michael5601 (approved):

.github/actions/release/release.py
```diff
@@ -35,7 +35,11 @@ def matches(self, resource: ocm.Resource):
             if not resource_value == v:
                 return False
 
-        return True
+        # reject if resource carries extraIdentity keys not covered by id —
+        # such keys are part of the OCM identity and make it a distinct resource
+        # (e.g. SBOM resources share name/os/arch with binaries but add sbom-format)
+        covered = {k for k in self.id if k not in ('name', 'version', 'type')}
```
8R0WNI3:
```suggestion
        covered = self.id.keys()
```
If the keys (e.g. `version`) would be removed from the `covered` list, the subsequent expression would be always `False` in case the `extraIdentity` holds one of those "standard" attribute names, wouldn't it?

ccwienk: good catch - updated accordingly

TuanAnh17N (approved):

TuanAnh17N (approved):

PR #1623, opened 2026-06-18
set delay to be min backoff_base and not 1

IndritFejza:
I saw that we are retrying on timeout issues after 1sec when retry is set to 1. 
Its because `backoff_base ** attempt = 1` on the zeroth attempt 

I think its best to wait a bit longer than 1sec. 
We can either set this to be backoff_base= 3 sec which is a tiny bit better but maybe increasing the wait-period when we retry so little would make sense (like 10 sec)

**Release note**:

```other operator
NONE
```

### DIFF
--- a/.github/actions/github-auth/get-token.py
+++ b/.github/actions/github-auth/get-token.py
@@ -68,7 +68,7 @@ def _open_with_retry(
                     file=sys.stderr,
                 )
                 sys.exit(1)
-            delay = min(backoff_base ** attempt, backoff_cap)
+            delay = min(backoff_base ** (attempt+1), backoff_cap)
             print(
                 f'WARNING: HTTP {e.code} from {url}, '
                 f'retrying in {delay:.0f}s ({attempt + 1}/{retries})...',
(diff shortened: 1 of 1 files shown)


ccwienk (commented):
generally-speaking, the diff does what it claims, and is reasonable assuming we _want_ longer back-off upon first error.

However, I wonder: is it actually helpful to do so? Originally, we had no retry-logic in place, at all, and (to my knowledge) never saw any issues w/ oidc-fed-svc.

Later-on I added retry-logic when we saw networking issues (probably you remember), which was, as it turned out, to no avail (as them networking-issues were persistent per gha-runner). So if at all, retries prolonged the amount of time for pipeline-runs to finally fail.

I kept retry-logic in, nonetheless (but decreased maximum amount of retry-duration / amount of retries), "just in case".

So I wonder: did you in act observe cases where existing retry-logic was actually triggered, and did not work well? Or asking more concretely: do you have any observation data that suggested to you the initial backoff should be increased?

ccwienk (approved):

PR #1627, opened 2026-06-19
drop ctx.py and dependent dead code

ccwienk:
**Release note**:
```other operator
Drop `ctx.py` (global execution-context module) and the code that depended exclusively on it:
`ccc/delivery.py`, `model/email.py`, `cli/gardener_ci/_ocm.py`,
`cli/gardener_ci/_release_notes.py`. The CLI commands removed from `_ocm.py` and
`_release_notes.py` are covered by `ocm/__main__.py` and
`.github/actions/release-notes/release_notes_cli.py` respectively.
```


model/__init__.py
```diff
@@ -130,59 +116,10 @@ def _parse_local_file(cfg_dir: str, cfg_src: LocalFileCfgSrc):
 
     @staticmethod
     def _parse_repo_file(
```
TuanAnh17N: this function is dead code, i think. its only caller (the elif isinstance(cfg_src, GithubRepoFileSrc) branch) is removed along with GithubRepoFileSrc itself, so this can be deleted aswell

ccwienk: done - good catch

Michael5601 (approved):

PR #1639, opened 2026-06-22
add CBOM generation alongside SBOM (cbomkit-theia)

ccwienk:
Add Cryptography Bill of Materials (CBOM) generation alongside the existing SBOM pipeline, using [cbomkit-theia](https://github.com/IBM/cbomkit-theia) (IBM).

CBOMs inventory cryptographic assets (certificates, keys, secrets, OpenSSL/JCA configuration) in OCI images.

**What changed:**

- `sbom/cbom.py` (new): `push_cbom_referrer`, `build_cbom_ocm_resources`; CBOM stored as OCI 1.1 referrer with `artifactType=application/vnd.cyclonedx+json;profile=cbom`
- `sbom/oci.py`: extracted generic `_build_referrer_manifest`/`_resolve_subject`/`_push_referrer` helpers; `push_sbom_referrer`/`push_sbom_referrers` now delegate to them
- `sbom/inject.py`: added `check_cbomkit_theia`, `_run_cbomkit_theia`; `scan_image` extended to produce CBOM (uses syft's CycloneDX output as `--bom` input — no second image pull); `build_sbom_ocm_resources` returns triple
- `ctt/process_dependencies.py`: CBOM generated and injected alongside SBOM during replication
- `oci-ocm.yaml`: cbomkit-theia install + CBOM generate/push/OCM-fragment steps for each platform build
- `merge-ocm-fragments/spdx_external.py`: generalised `_WANTED_DOCS` table + `_CacheResult.missing` replaces binary hit/miss; 3-layer cache (layer[2]=CBOM); CBOM-only partial hits run in parallel (no syft needed); `SCAN_CBOM_ONLY` status in step summary

**Release note**:
```feature operator
Add CBOM (Cryptography Bill of Materials) generation alongside SBOM for OCI images, using
cbomkit-theia. CBOMs are stored as OCI 1.1 referrer manifests and OCM resources alongside
the existing SPDX/CycloneDX SBOM documents. Affects oci-ocm.yaml (per-platform builds) and
merge-ocm-fragments (external images via release.yaml).
```


.github/actions/merge-ocm-fragments/spdx_external.py
```diff
@@ -621,102 +735,145 @@ def _is_oci_image(resource: dict) -> bool:
...
             spdx_blob_digest=spdx_digest,
             cdx_blob_digest=cdx_digest,
+            cbom_blob_digest=cbom_digest,
             component_version=component_version,
             tool_ver=tool_ver,
         )
-        new_resources.extend((spdx_resource, cdx_resource))
+        new_resources.extend(bom_resources)
+
+        was_full_hit = not cr.missing  # by construction: full hits have no missing
```
Michael5601: question: cr here is the returned `_CacheResult` from `_scan_image`, which always has a fully-populated cached dict (all three docs present), so in this line `was_full_hit `will always be `True`. This means scanned items are indistinguishable from original cache hits at this point.

The information of the original `_CacheResult` is gone as `_scan_image` returns a new instance.

.github/actions/merge-ocm-fragments/spdx_external.py
```diff
@@ -365,41 +414,83 @@ def _push_sboms_to_cache(
...
     resource_name = info.resource['name']
     source_ref = om.OciImageReference.to_image_ref(info.digest_ref)
     cache_repo = _cache_repo(cache_registry, cache_prefix, source_ref)
+    missing = miss.missing
+    cached = dict(miss.cached)  # copy; we'll populate it
+
+    need_syft   = 'spdx-2.3' in missing or 'cyclonedx-1.6' in missing
+    need_cbomkit = 'cbom-cyclonedx-1.6' in missing
 
     try:
```
Michael5601: **question (non-blocking):** The entire scan (syft + cbomkit-theia) is wrapped in a single try/except. If cbomkit-theia fails after syft already produced both SBOM formats, the exception propagates to the outer except Exception at line 494 and the function returns None. The successfully-produced SPDX and CycloneDX blobs are discarded and the cache is not updated. 

Is this wanted?

.github/actions/merge-ocm-fragments/spdx_external.py
```diff
@@ -621,102 +735,145 @@ def _is_oci_image(resource: dict) -> bool:
...
+        ]
+
+        if cbom_only_list:
+            logger.info(
+                f'phase 2: {len(cbom_only_list)} CBOM-only scan(s) (SBOM already cached)'
+            )
+            with concurrent.futures.ThreadPoolExecutor() as executor:
+                cbom_results = list(executor.map(
+                    lambda cr: _scan_image(
+                        cr, oci_client, cache_registry, cache_repo_prefix, syft_ver, tmpdir,
```
Michael5601: **question (non-blocking)**: cbom_only_list scans are dispatched via an unconstrained ThreadPoolExecutor.map(). Unlike the need_syft_list path, which goes through _run_scans_resource_aware() with disk/memory admission control. Can this lead to a bottleneck?

ccwienk: Acknowledged — and agreed the framing was too optimistic. cbomkit-theia still pulls the OCI image to scan for crypto assets, so the I/O pressure is not negligible. An unconstrained executor over a large list could easily saturate disk or memory the same way full scans would.

Will route cbom_only scans through `_run_scans_resource_aware` as well (or at minimum cap `max_workers`).

.github/actions/merge-ocm-fragments/spdx_external.py
ccwienk: You're right that the logic is broken — after  returns,  is always empty (the returned  is freshly constructed with all docs present), so  will always be  for scanned items too, and the CBOM-only / SCANNED status branches are never reached.

Will fix by preserving the original  tuple from the pre-scan  and using that for status classification, rather than inspecting the post-scan result.

.github/actions/merge-ocm-fragments/spdx_external.py
ccwienk: Partially intentional: the try/except is meant to be atomic — if any tool fails, we do not write a partial cache entry. However, the downside is that a cbomkit-theia failure causes syft to re-run on the next attempt, which is wasteful.

Will separate the two try/except blocks so a successful syft run is cached independently, and a cbomkit-theia failure only triggers a CBOM-only retry next time.

Michael5601 (approved):
lgtm

PR #1645, opened 2026-06-24
prepare: rename output component-descriptor -> component-descriptor-artefact

ccwienk:
Renames the `component-descriptor` workflow output of `prepare.yaml` to
`component-descriptor-artefact`, making it unambiguous that the value is an
artifact name, not inline YAML (incompatible change).

### DIFF
--- a/.github/workflows/prepare.yaml
+++ b/.github/workflows/prepare.yaml
@@ -187,15 +187,10 @@ on:
           It is output using the `capture-commit` action, and can be imported using the
           `import-commit` action.
         value: release-commit-objects
-      component-descriptor:
-        description: |
-          The name of the GHA artifact holding the base OCM Component-Descriptor.
-          Deprecated: use component-descriptor-artefact instead.
-        value: ${{ jobs.version-and-ocm.outputs.component-descriptor }}
       component-descriptor-artefact:
         description: |
           The name of the GHA artifact holding the base OCM Component-Descriptor.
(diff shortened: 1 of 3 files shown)


.github/workflows/prepare.yaml
```diff
@@ -213,7 +208,7 @@ jobs:
       can-push: ${{ steps.params.outputs.can-push }}
       version: ${{ steps.version.outputs.version }}
       commit-digest: ${{ steps.version.outputs.commit-digest }}
-      component-descriptor: ${{ steps.component-descriptor.outputs.component-descriptor-artifact }}
+      component-descriptor-artefact: ${{ steps.component-descriptor.outputs.component-descriptor-artifact }}
```
IndritFejza, line 211: I have noticed a mix of  usage of british/american `artefact` and `artifect` in the wf/actions in this repo.

Would it make sense to unify all usages and use only 1 term? 
As to no cause confusion.

That could potentially cause breaking changes ofc.

ccwienk: @IndritFejza : by default, this repository uses BE (so artEfact). however, GitHub-Actions-ArtIfacts are written according to AE, so in this specific case, AE seems better to avoid confusion w/ upstream API.

you are right, however, about incompatibility (so I would prefer to accept AE/BE-mix to avoid another round of downstream-updates..)

PR #1640, opened 2026-06-22
sbom: S3-bucket scanning support (gardenlinux)

ccwienk:
## Summary

- `sbom/s3.py`: public S3 download helper (no boto3)
- `sbom/oci.py`: standalone OCI manifest push for non-referrer SBOM storage
- `sbom/inject.py`: `scan_s3_resource`, `build_s3_sbom_ocm_resources`
- `ctt`: `inject_s3_sboms` flag + S3 injection loop in `process_replication_plan_step`
- `upload.py`: `_fmt_id`/`_filename` handle `cbom-format` extraIdentity key
- `sbom/iter.py`: exclude `cbom-format` from `source_extra` filter
- task specs in `sbom-tasks.d/`, unit tests in `test/sbom/`

Depends on: #(add-cbom-support PR)

PR #1646, opened 2026-06-24
sbom: S3-bucket scanning support (gardenlinux)

ccwienk:
## Summary

- `sbom/s3.py`: public S3 download helper (no boto3)
- `sbom/oci.py`: standalone OCI manifest push for non-referrer SBOM storage
- `sbom/inject.py`: `scan_s3_resource`, `build_s3_sbom_ocm_resources`
- `ctt`: `inject_s3_sboms` flag + S3 injection loop in `process_replication_plan_step`
- `upload.py`: `_fmt_id`/`_filename` handle `cbom-format` extraIdentity key
- `sbom/iter.py`: exclude `cbom-format` from `source_extra` filter
- task specs in `sbom-tasks.d/`, unit tests in `test/sbom/`

### DIFF
--- a/sbom/s3.py
+++ b/sbom/s3.py
@@ -1,0 +1,71 @@
+# SPDX-FileCopyrightText: 2024 SAP SE or an SAP affiliate company and Gardener contributors
+#
+# SPDX-License-Identifier: Apache-2.0
+'''
+Minimal S3 download helper for public buckets (no auth required).
+
+Only uses stdlib — no boto3/botocore dependency.  Virtual-hosted-style URLs are used when a
+region is known; path-style (us-east-1 endpoint) is used as fallback and lets AWS redirect.
+'''
+import urllib.error
+import urllib.request
+
+
+def s3_url(bucket: str, key: str, region: str | None = None) -> str:
+    '''Return the HTTPS URL for a public S3 object.'''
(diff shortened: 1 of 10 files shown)


sbom/s3.py
```diff
@@ -0,0 +1,73 @@
...
+    '''
+    Download a public S3 object and return its bytes.
+
+    Follows up to one redirect (region-correction redirect from us-east-1 endpoint).
+    Raises urllib.error.HTTPError on non-200 responses.
+    '''
+    url = s3_url(bucket, key, region)
+    req = urllib.request.Request(url, method='GET')
+    with urllib.request.urlopen(req) as resp:  # nosec B310
+        return resp.read()
```
Michael5601: **question (non-blocking):** this line reads the bytes into memory. Can this lead to a bottleneck on the runners if handling gardener images.

Michael5601:
also see my comments in https://github.com/gardener/cc-utils/pull/1640

Michael5601:
the rest looks good to me.

Michael5601 (approved):

sbom/s3.py
ccwienk: good catch. will switch to streaming

PR #1648, opened 2026-06-25
pin: update stale internal refs when dep gets a new pinned SHA

ccwienk:
## Problem

When an internal action/workflow changes on master and gets a new pinned SHA
on v1, any callers that already hold a pinned SHA for that dep are silently
left pointing at the old (stale) SHA.

Root cause: `_rewrite_parsed` skipped **all** already-pinned refs
unconditionally (`if _is_pinned(ref): continue`), and `_extract_internal_deps`
excluded them from the dependency graph entirely.  So the topological walk
never revisited a caller once its dep ref appeared to be "already pinned".

This is latent whenever the source passed to the pinner (`refs/capture-commit`
for releases) contains workflow files that were bootstrapped from a prior v1 —
which is exactly the normal case.  It manifested when `merge-ocm-fragments`
had its cbomkit-theia install fixed mid-release-cycle: the action itself got a
new pin commit on v1, but `release.yaml` / `post-build.yaml` etc. were left
pointing at the pre-fix SHA.

## Fix

Two targeted changes in `pin.py`:

- **`_extract_internal_deps`**: always yield internal deps regardless of
  whether the current ref is already pinned, so the topological sort is
  correct even when callers carry pinned SHAs.
- **`_rewrite_parsed`**: when a ref is already pinned, check whether
  `prefix_to_digest` holds a *different* (newer) digest for that dep and
  update if so; external deps (not in `prefix_to_digest`) are left untouched.

## Tests

Two regression tests added:
- `test_stale_pinned_internal_ref_is_updated` — verifies the rewrite happens
- `test_extract_internal_deps_finds_pinned_dep` — verifies graph inclusion

Plus a corrected existing test (`test_already_pinned_ref_unchanged` →
`test_already_pinned_ref_at_current_sha_unchanged`) that now correctly asserts
no change only when the SHA is already current, not when it differs.

### DIFF
--- a/.github/actions/pin-actions-and-workflows/pin.py
+++ b/.github/actions/pin-actions-and-workflows/pin.py
@@ -189,7 +189,7 @@ def _extract_internal_deps(
         if '@' not in after_repo:
             continue
         dep_prefix, ref = after_repo.rsplit('@', 1)
-        if not _is_pinned(ref, repo) and dep_prefix in known_prefixes:
+        if dep_prefix in known_prefixes:
             yield dep_prefix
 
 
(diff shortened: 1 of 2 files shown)


ccwienk:
superseded by new regression-test (#1649 )

PR #1656, opened 2026-07-03
Add OCM release notes when updating gardener images

hebelsan:
**Release note**:

```other operator
Add OCM release notes for Gardener image updates on the update-extension-provider-images workflow
```

### DIFF
--- a/.github/actions/update-extension-provider-images/update-images.py
+++ b/.github/actions/update-extension-provider-images/update-images.py
@@ -1207,10 +1207,10 @@ def create_release_notes(
 def write_ocm_release_notes(
     updates: list[ImageUpdate],
     images_data: ImagesData,
+    oci_client_: oci_client.Client,
     repo_dir: str = '.',
     ocm_repository: str = GARDENER_OCM_REPOSITORY,
 ) -> None:
-    oci_client_ = oci_client.client_with_dockerauth()
     ocm_repository_lookup = cnudie.retrieve.ocm_repository_lookup(ocm_repository)
     component_descriptor_lookup = cnudie.retrieve.create_default_component_descriptor_lookup(
         ocm_repository_lookup=ocm_repository_lookup,
@@ -1338,10 +1338,9 @@ def main() -> None:
             images_data.sort_images()
             write_yaml_file(images_data, args.images_yaml_path, updated_directives)
             print(f"\nUpdated {args.images_yaml_path} with {len(updates)} changes.", file=sys.stderr)
-            write_ocm_release_notes(updates, images_data)
-
-        create_release_notes(updates, images_data, all_available_tags, args.release_notes_path)
-        print(f"Created {args.release_notes_path}", file=sys.stderr)
+            write_ocm_release_notes(updates, images_data, oci_api)
+            create_release_notes(updates, images_data, all_available_tags, args.release_notes_path)
+            print(f"Created {args.release_notes_path}", file=sys.stderr)
 
         print("The following container images have been updated:")
         for update in updates:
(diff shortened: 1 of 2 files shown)


.github/actions/update-extension-provider-images/update-images.py
```diff
@@ -1197,6 +1204,58 @@ def create_release_notes(
             f.write("\n")
 
 
+def write_ocm_release_notes(
```
ccwienk, line 1209: `oci_client_` duplicates the `oci_api` client already created in `main()` — consider passing it in as a parameter to avoid a second connection pool.

.github/actions/update-extension-provider-images/update-images.py
```diff
@@ -1279,6 +1338,7 @@ def main() -> None:
             images_data.sort_images()
             write_yaml_file(images_data, args.images_yaml_path, updated_directives)
             print(f"\nUpdated {args.images_yaml_path} with {len(updates)} changes.", file=sys.stderr)
+            write_ocm_release_notes(updates, images_data)
```
ccwienk: `write_ocm_release_notes` is called inside `if updates:` but `create_release_notes` is called unconditionally within the outer `if all_new_versions:` block — the guards are inconsistent.

hebelsan: `create_release_notes` does nothing when updates is empty (it returns early).
But you are right, the inconsistency is confusing.
I added `create_release_notes` and its print into the same `if updates` path.

hebelsan:
Thanks for the review @ccwienk! 🚀
I've addressed your comments.

.github/actions/update-extension-provider-images/update-images.py
```diff
@@ -66,9 +66,16 @@
 from ruamel.yaml import YAML
 from ruamel.yaml import YAMLError
 
+import cnudie.retrieve
+import ocm
+import oci.auth
```
ccwienk: Unused import — nothing in this file references `oci.auth` directly.

.github/actions/update-extension-provider-images/update-images.py
```diff
@@ -1197,6 +1204,58 @@ def create_release_notes(
...
+    source_repo_by_name = {img.name: img.sourceRepository for img in images_data.images}
+
+    for update in updates:
+        source_repo = source_repo_by_name.get(update.image_name)
+        if not source_repo or not source_repo.startswith('github.com/gardener/'):
+            continue
+        if not update.old_tag:
+            continue
+
+        component_name = source_repo
```
ccwienk: Needless alias — `source_repo` already is the component name; just use it directly.

.github/actions/update-extension-provider-images/update-images.py
```diff
@@ -1197,6 +1204,58 @@ def create_release_notes(
...
+        print(f"Fetching OCM release notes for {component_name} {update.old_tag} -> {update.new_tag}", file=sys.stderr)
+        try:
+            release_notes_docs = list(rno.release_notes_for_vector(
+                upgrade_vector=upgrade_vector,
+                component_descriptor_lookup=component_descriptor_lookup,
+                version_lookup=version_lookup,
+                oci_client=oci_client_,
+                version_filter=is_final,
+            ))
+        except Exception as e:
```
ccwienk: Too broad — catches everything including bugs in this code (wrong API shape, `AttributeError`, etc.), silently turning them into a skipped update. Consider narrowing to the specific exceptions expected from the OCM/network calls.

hebelsan: I switched to:
```python
except (
            om.OciImageNotFoundException,
            requests.exceptions.HTTPError,
            requests.exceptions.ConnectionError,
            requests.exceptions.Timeout,
            ValueError,
) as e
```

ccwienk (approved):
All comments addressed, looks good.

PR #1658, opened 2026-07-06
ocm-upgrade: keep-going on callback failure

ccwienk:
## Summary

- Wrap each `create_upgrade_pullrequest` call in `create_upgrade_pullrequests` with `try/except subprocess.CalledProcessError`
- On callback failure: reset git worktree, yield an `UpgradeError` and continue to the next component
- Caller (action.yaml) collects errors separately, appends a warnings section to the job summary, and exits with code 1 if any failures occurred — so the job still fails but processes all components first

## Test plan

- [ ] Trigger the upgrade-dependencies workflow against a repo whose `.ci/set_dependency_version` callback fails for one component — confirm other upgrade-PRs are still created
- [ ] Confirm the job summary lists the failed component(s) under "Warnings"
- [ ] Confirm the job exits with a non-zero status when at least one callback failed

### DIFF
--- a/.github/actions/ocm-upgrade/action.yaml
+++ b/.github/actions/ocm-upgrade/action.yaml
@@ -234,8 +234,9 @@ runs:
           '${{ inputs.upstream-update-policy }}'.strip()
         )
 
-        created_upgrade_pullrequests = tuple(
-          ocm_upgrade.create_upgrade_pullrequests(
+        created_upgrade_pullrequests = []
+        upgrade_errors = []
+        for result in ocm_upgrade.create_upgrade_pullrequests(
             component=component,
             component_descriptor_lookup=component_descriptor_lookup,
             version_lookup=version_lookup,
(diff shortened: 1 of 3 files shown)


.github/actions/ocm-upgrade/action.yaml
```diff
@@ -272,5 +276,19 @@ runs:
...
         '''
 
+        if upgrade_errors:
+          summary += '\n        ## Warnings: Failed Upgrade-Attempts\n\n'
+          for ue in upgrade_errors:
+            uv = ue.upgrade_vector
+            summary += (
+              f'        - **{uv.component_name}**: '
+              f'`{uv.whence.version}` → `{uv.whither.version}` '
+              f'— callback exited with {ue.error.returncode}\n'
```
TuanAnh17N, line 286: formatted_traceback is captured in UpgradeError but never surfaced. maybe we could include it here for easier debugging

TuanAnh17N (approved):
rest lgtm

.github/actions/ocm-upgrade/action.yaml
ccwienk: sry, did not see initially. pushed to master in separate commit

PR #1663, opened 2026-07-07
ocm-validate: support transitive recursion, default to full closure

ccwienk:
## Summary

- `ocm-validate` action and `validate.py` hardcoded `recursion_depth=0`/`lookup=None`, so
  validation only covered the top-level component descriptor's own resources — transitive
  `componentReference` chains were never checked
- adds `ocm-repositories` (comma-separated) and `recursion-depth` inputs to `ocm-validate`
  action; wires them through `release` action and `release.yaml` workflow
- default is `-1` (full transitive closure); set explicitly to `0` to restore old behaviour

## Test plan

- [ ] trigger a release pipeline for a component with transitive references containing a
  deliberately stale OCI image — confirm validation now fails
- [ ] verify existing pipelines that do not pass `ocm-repositories` get a clear error rather
  than a silent no-op (ValueError raised in `validate.py`)

### DIFF
--- a/.github/actions/ocm-validate/validate.py
+++ b/.github/actions/ocm-validate/validate.py
@@ -1,5 +1,6 @@
 import yaml
 
+import cnudie.retrieve
 import ocm
 import ocm.iter
 import ocm.validate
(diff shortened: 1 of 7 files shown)


.github/actions/ocm-validate/validate.py
```diff
@@ -10,16 +11,27 @@
...
     component_descriptor = ocm.ComponentDescriptor.from_dict(component_descriptor)
 
+    if recursion_depth != 0 and ocm_repositories:
+        repos = [r.strip() for r in ocm_repositories.split(',') if r.strip()]
+        ocm_repo_lookup = cnudie.retrieve.ocm_repository_lookup(*repos)
+        lookup = cnudie.retrieve.create_default_component_descriptor_lookup(
+            ocm_repository_lookup=ocm_repo_lookup,
+        )
+    else:
+        lookup = None
```
TuanAnh17N, line 33: this else branch fires when recursion_depth != 0 but ocm_repositories is empty (e.g. the caller omits it), silently setting lookup = None. ocm.iter.iter will then skip transitive resolution with no warning, making the new default (-1) a silent no-op for most existing callers. i think we should rather raise in this case

ccwienk: Good catch — fixed in the latest commit (raises `ValueError` now).

TuanAnh17N (approved):

PR #1669, opened 2026-07-10
capture-commit/import-commit: use git bundle transport

ccwienk:
## Summary

- Replaces the tar-of-loose-objects transport in `capture-commit`/`import-commit` with `git bundle`, eliminating a race with git's background auto-gc/auto-maintenance under `set -euo pipefail` (surfaced in release runs as `find: '.git/objects/xx': No such file or directory`).
- `capture-commit` now writes a bundle (thin against remote-tracking refs where possible, full otherwise); `import-commit` consumes it via `git fetch <bundle> +refs/capture-commit:refs/capture-commit`.
- Action-level API preserved: input/output names, base64-encoding of the `commit-objects` output, the well-known `refs/capture-commit` side-effect, and the 64 KiB size threshold. `timestamp-reference` input remains required for backwards-compat but is now silently ignored.
- Adds broader act-based test coverage for the pair (multi-commit, multi-file, merge-commit, explicit-commit-input, cherry-pick, noop, two-captures) in a separate commit that locks in the pre-existing behaviour.

## Test plan

- [x] `test/actions/run-act-tests.sh capture-commit` — 8/8 pass
- [x] Full `build-and-test / test-actions` job on CI — green
- [x] All other CI jobs on the branch — green

**Release note**:
```bugfix operator
capture-commit/import-commit: switch to git-bundle transport, fixing a race with git auto-gc that could cause release pipelines to fail with 'find: .git/objects/xx: No such file or directory'.
```

### DIFF
--- a/.github/actions/capture-commit/capture-commit
+++ b/.github/actions/capture-commit/capture-commit
@@ -2,58 +2,55 @@
 
 set -euo pipefail
 
+# capture <commit> into a git-bundle file for consumption by import-commit.
+#
 # usage:
-# export GIT_DIR
-# $1: path to target-archive
-# $2: path to ref-file (must be older than commit)
-# $3: path to output-file for commit-digest (defaults to commit-digest)
-# $4: commit to capture (defaults to HEAD)
-
-# darwin/macOS is not supported. differences that would need addressing:
-# - `find -newermt` (GNU) -> `-newer` (BSD find; compares mtime, but no XY-syntax)
-if [ "$(uname -s)" = 'Darwin' ]; then
-  echo 'error: darwin/macOS is not supported'
-  exit 1
-fi
+# export GIT_DIR (usually $PWD/.git)
+# $1: path to output bundle
+# $2: (unused — kept for backwards-compat with the wrapper action; was the
+#      timestamp-reference file in the old find-based implementation)
+# $3: path to output-file for commit-digest (default 'commit-digest')
+# $4: commit to capture (default HEAD)
 
 if [ -z "${GIT_DIR:-}" ]; then
-  echo "must set GIT_DIR"
-  exit 1
+    echo 'must set GIT_DIR'
+    exit 1
 fi
 
 if [ -z "${1:-}" ]; then
-  echo "must pass target-file as ARGV[1]"
-  exit 1
-else
-  archive=${1}
+    echo 'must pass target-file as ARGV[1]'
+    exit 1
 fi
+archive=${1}
 
-if [ -z "${2:-}" ]; then
-  echo "must pass ref-file as ARGV[2]"
-  exit 1
-else
-  ref=$(readlink -f "${2}")
-fi
-
-if [ -z "${3:-}" ]; then
-  commit_digest_out='commit-digest'
-else
-  commit_digest_out=${3}
-fi
+commit_digest_out=${3:-commit-digest}
 
 if [ -z "${4:-}" ]; then
-  commit_digest="$(git rev-parse HEAD)"
+    commit_digest="$(git rev-parse HEAD)"
 else
-  commit_digest="$(git rev-parse ${4})"
+    commit_digest="$(git rev-parse "${4}")"
 fi
-echo "${commit_digest}" > "${GIT_DIR}/refs/capture-commit"
 
-object_paths=$(
-  cd "$(dirname "$GIT_DIR")"
-  find .git/objects -type f -newer "${ref}" | grep -v pack
-)
-object_paths+=('.git/refs/capture-commit')
+# contract: capture-commit is only ever intended for _local_, unpushed commits.
+# fail fast (with an expressive error) if the caller violates this — otherwise
+# `git bundle create --not --remotes` below would produce an empty bundle.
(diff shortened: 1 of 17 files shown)


.github/actions/capture-commit/capture-commit
```diff
@@ -2,58 +2,48 @@
...
+# a well-known ref so import-commit can look it up without needing to pass
+# commit-digest explicitly
+git update-ref refs/capture-commit "${commit_digest}"
+
+# prefer a thin bundle relative to any remote-tracking ref (receiver is assumed
+# to be at the same base). Fall back to a full-history bundle when there are
+# no remotes (typical for `git init`-based test fixtures) or when the
+# thin-bundle would be empty (commit already reachable from a remote).
+if git for-each-ref refs/remotes/ 2>/dev/null | grep -q .; then
+    git bundle create "${archive}" refs/capture-commit --not --remotes \
```
TuanAnh17N: If the captured commit is already fully reachable from the remote-tracking refs, `git bundle create --not --remotes` exits 128 (empty bundle) and the || fallback creates a full-history bundle instead. i think on a real repo that could leave commit-objects unset and break the import 🤔

test/actions/capture-commit/cherry-pick/workflow.yaml
```diff
@@ -0,0 +1,68 @@
...
+          git reset --hard HEAD~1
+          echo 'divergent' > other.txt
+          git add other.txt
+          git commit -m 'divergent commit'
+      - uses: gardener/cc-utils/.github/actions/import-commit@master
+        with:
+          commit-objects: ${{ steps.capture.outputs.commit-objects }}
+          commit-digest: ${{ steps.capture.outputs.commit-digest }}
+          after-import: cherry-pick
+      - name: assert-cherry-pick
```
TuanAnh17N, line 47: the existing round-trip test checks that commit-objects is non-empty right after capture, before even attempting the import. none of the 7 new tests do this

test/actions/capture-commit/merge-commit/workflow.yaml
```diff
@@ -0,0 +1,79 @@
...
+        with:
+          timestamp-reference: /tmp/ts-ref
+      - name: reset-to-base
+        shell: bash
+        run: git reset --hard '${{ steps.build.outputs.base }}'
+      - uses: gardener/cc-utils/.github/actions/import-commit@master
+        with:
+          commit-objects: ${{ steps.capture.outputs.commit-objects }}
+          commit-digest: ${{ steps.capture.outputs.commit-digest }}
+          after-import: rebase
```
TuanAnh17N: doesn't git rebase linearize history by default and drop the merge commit? if that's the case, the 
`parents -ne 3` assertion would always fail after the rebase. should this use after-import: noop instead, since the bundle import already guarantees the exact commit digest?

test/actions/capture-commit/two-captures/workflow.yaml
```diff
@@ -0,0 +1,98 @@
...
+      - id: capture-a
+        uses: gardener/cc-utils/.github/actions/capture-commit@master
+        with:
+          timestamp-reference: /tmp/ts-ref-1
+      - name: ts-ref-2-and-commit-B
+        id: b
+        shell: bash
+        run: |
+          # ensure ts-ref-2 is strictly newer than commit-A's objects
+          sleep 1
```
TuanAnh17N: the two sleep 1 calls in two-captures seem to be left over from the old timestamp-based approach. since timestamp-reference is now ignored by the bundle implementation, are these still needed?

.github/actions/capture-commit/capture-commit
ccwienk: Good catch, but on closer inspection this only triggers if the caller violates the design contract (capture-commit is only ever intended for _local, unpushed_ commits — the captured sha shouldn't be reachable from any remote-tracking ref). Under contract-compliant use, `--not --remotes` always yields a non-empty thin bundle.

Rather than silently falling back and risking a too-large full bundle (which would drop `commit-objects` via the 64 KiB gate and confuse the import step later), I've added an explicit contract check via `git branch -r --contains` that fails fast with an expressive error when the caller misuses the action. The `||` fallback is dropped.

test/actions/capture-commit/cherry-pick/workflow.yaml
ccwienk: Fixed — all 7 new tests now assert `commit-objects` is non-empty right after capture.

test/actions/capture-commit/merge-commit/workflow.yaml
ccwienk: Fixed — switched to `after-import: noop` and the assertion now runs against `refs/capture-commit` directly (via explicit `git reset --hard`). Cleaner and doesn't rely on the rebase-degenerates-to-fast-forward equivalence.

test/actions/capture-commit/two-captures/workflow.yaml
ccwienk: Fixed — both `sleep 1` calls (and the stale "strictly newer than commit-A's objects" comment) dropped.

TuanAnh17N (approved):

PR #1674, opened 2026-07-15
merge-ocm-fragments: configurable artifact-name; skip upload if empty

ccwienk:
Adds an `upload` boolean input (default: `true`) to `merge-ocm-fragments`.
When set to `false`, the artifact upload is skipped and the descriptor is
only available via the new `component-descriptor-path` output (local
filesystem path).

Fixes 409 conflicts when `helmchart-ocm` is used in a matrix: each
concurrent helmchart job calls `merge-ocm-fragments` to resolve OCM
resource references, but only needs the descriptor locally — not as a
shared run artifact. `helmchart-ocm` now passes `upload: false` and
uses the path output instead.

Also adds a `component-descriptor-path` input to the `helmchart` action
as a path-based alternative to `component-descriptor-artifact`.

**Release note**:
```bugfix operator
merge-ocm-fragments: add `upload` input (default: true) to skip artifact
upload when descriptor is only needed within the same job. Fixes 409
conflicts in matrix helmchart builds.
```

### DIFF
--- a/.github/actions/merge-ocm-fragments/action.yaml
+++ b/.github/actions/merge-ocm-fragments/action.yaml
@@ -98,13 +98,27 @@ inputs:
       When set to `true`, the SBOM cache is ignored and all external images are re-scanned.
       Useful when a newer syft version is desired. Defaults to `false`.
 
+  upload-artifact-name:
+    required: false
+    type: string
+    default: merged-component-descriptor
+    description: |
+      Name under which the merged component-descriptor is uploaded as a GHA artifact.
+      Set to an empty string to skip the upload (e.g. when the descriptor is only needed
+      within the same job, to avoid artifact-name collisions in matrix jobs).
+
 outputs:
   component-descriptor-artifact:
     description: |
       name of the GHA artifact containing the merged component-descriptor.
       The artifact holds a single file `component-descriptor.tar.gz` which
       in turn contains `component-descriptor.yaml`.
-    value: merged-component-descriptor
+      Empty if `upload-artifact-name` input is empty.
+    value: ${{ inputs.upload-artifact-name }}
+  component-descriptor-path:
+    description: |
+      Path to the merged component-descriptor.yaml on the runner filesystem.
+    value: ${{ inputs.outdir }}/component-descriptor.yaml
   name:
     description: |
       the component-name (for convenience)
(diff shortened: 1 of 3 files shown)


.github/actions/merge-ocm-fragments/action.yaml
```diff
@@ -359,13 +374,24 @@ runs:
             force_rescan='${{ inputs.spdx-force-rescan }}' == 'true',
         )
 
+    - name: artifact-name
+      id: artifact-name
+      shell: bash
+      run: |
+        if ${{ inputs.upload != 'false' }}; then
+          echo 'name=merged-component-descriptor' >> "${GITHUB_OUTPUT}"
```
DockToFuture: Wouldn't any other caller with a matrix and upload true run into the same problem again?

.github/actions/merge-ocm-fragments/action.yaml
```diff
@@ -359,13 +374,24 @@ runs:
...
+        else
+          echo 'name=' >> "${GITHUB_OUTPUT}"
+        fi
     - name: upload-component-descriptor-artifact
+      if: ${{ inputs.upload != 'false' }}
```
DockToFuture:
```suggestion
      if: ${{ inputs.upload }}
```

Same for other occurences

.github/actions/merge-ocm-fragments/action.yaml
ccwienk: yes. However, the merge-ocm-fragments action is _intended_ to only be called once (if uploading). it is used in helm-workflow as kind of a hack (well maybe not _as_ bad) to be able to get an early (not fully finalised) component-descriptor to have means to do early validation of ocm-resource-references.

that being said: we can also add an input by which the artefact-name can be controlled, to make the action more flexible for such cases.

ccwienk: Addressed by replacing the boolean `upload` input with a string `upload-artifact-name` (default: `merged-component-descriptor`). Empty string skips the upload. This lets any caller in a matrix avoid collisions by supplying a unique name, or pass `''` to skip entirely.

.github/actions/merge-ocm-fragments/action.yaml
ccwienk: You're right that `if: ${{ inputs.upload }}` is cleaner — though for a `type: boolean` GHA input, the value is actually passed as the string `'true'`/`'false'`, so `inputs.upload` would always be truthy (non-empty string). The `\!= 'false'` guard was therefore correct for the old design. Moot now anyway: both steps are gone, replaced by `inputs.artifact-name \!= ''`.

PR #1675, opened 2026-07-16
fix(kubernetes-auth): tolerate non-zero openssl exit code when retrieving server CA

Michael5601:
openssl s_client returns exit code 1 when it cannot fully verify the certificate chain against the system trust store, which is expected for clusters using a private/internal CA (e.g. SAP-internal clusters). With set -o pipefail this caused the prepare-server-ca step to fail silently even though the certificate was retrieved successfully.

  Fix:
  - Added || true to tolerate the non-zero openssl exit code
  - Added an explicit empty-check on server_ca afterwards so real failures (e.g. connection refused, no cert returned) still surface with a clear error message instead of failing silently in a later step
  
**Release note**:

```other operator

```

### DIFF
--- a/.github/actions/kubernetes-auth/action.yaml
+++ b/.github/actions/kubernetes-auth/action.yaml
@@ -76,10 +76,16 @@ runs:
             normalised_api_url="${normalised_api_url}:443"
           fi
 
-          server_ca=$(echo | openssl s_client -showcerts -connect "${normalised_api_url}" 2>/dev/null \
+          raw=$(echo | openssl s_client -showcerts -connect "${normalised_api_url}" 2>/dev/null || true)
+          server_ca=$(printf '%s' "${raw}" \
             | sed -n '/-----BEGIN CERTIFICATE-----/,/-----END CERTIFICATE-----/p' \
             | base64 -w 0 \
             | tr -d '\n')
+
+          if [ -z "${server_ca}" ]; then
+            echo "Error: failed to retrieve server CA from ${normalised_api_url}"
+            exit 1
+          fi
         fi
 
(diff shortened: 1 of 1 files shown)


TuanAnh17N (approved):

ccwienk (commented):
The `|| true` on the full pipeline is broader than needed — `sed`, `base64`, and `tr` failures would be silently swallowed too. The non-zero exit is specific to `openssl s_client` (it exits 1 when the server closes the connection, even after successfully writing certs to stdout).

Suggest splitting into two steps so the tolerance is scoped to `openssl` only:

```sh
raw=$(echo | openssl s_client -showcerts -connect "${normalised_api_url}" 2>/dev/null || true)
server_ca=$(printf '%s' "${raw}" \
  | sed -n '/-----BEGIN CERTIFICATE-----/,/-----END CERTIFICATE-----/p' \
  | base64 -w 0 \
  | tr -d '\n')
```

The empty-string check below stays as-is.

ccwienk:
The `|| true` on the full pipeline is broader than needed — `sed`, `base64`, and `tr` failures would be silently swallowed too. The non-zero exit is specific to `openssl s_client` (it exits 1 when the server closes the connection, even after successfully writing certs to stdout).

Suggest splitting into two steps so the tolerance is scoped to `openssl` only:

```sh
raw=$(echo | openssl s_client -showcerts -connect "${normalised_api_url}" 2>/dev/null || true)
server_ca=$(printf '%s' "${raw}" \
  | sed -n '/-----BEGIN CERTIFICATE-----/,/-----END CERTIFICATE-----/p' \
  | base64 -w 0 \
  | tr -d '\n')
```

The empty-string check below stays as-is.

TuanAnh17N (approved):
/lgtm

PR #1677, opened 2026-07-20
feat(oci): support docker credential helpers

ccwienk:
`docker_credentials_lookup` now honours `credHelpers` (per-registry) and `credsStore` (global fallback) fields from docker-cfg. Resolution order depends on `credential_helper_policy` (see below).

A new `CredentialHelperPolicy` parameter controls both resolution order and behaviour on missing/broken helpers:

- `STATIC_FIRST` *(default)* — static `auths` first; helpers used only as fallback. Safe default for consumers that receive updates uncontrolled — no surprises if a helper is present but broken.
- `DISABLED` — skip helpers entirely; only static `auths` are used. Recommended when docker-cfg is mounted from a host into a container where helpers are unavailable.
- `WARN` — helpers first (Docker-native order); missing/broken helper emits a log warning and falls through to static credentials.
- `FAIL` — helpers first; missing/broken helper raises `RuntimeError`.

Additional notes:
- `credHelpers` key matching is port-aware (prefers exact `host:port` match, falls back to host-only), consistent with `auths` lookup.
- The configured `credHelpers` key is passed as `server_url` to the helper binary, so helpers receive exactly what the user configured (relevant for local registries, e.g. `localhost:5000`).
- A `credential_helper_timeout_seconds` parameter (default: 60s) guards against hung helpers; pass `None` to disable for interactive/browser-based OAuth flows.
- PATH availability (`shutil.which`) is checked before invoking any helper.

**Release note**:

```feature developer
`docker_credentials_lookup` now supports docker credential helpers
(`credHelpers` / `credsStore` in docker-cfg). Resolution order and
behaviour on missing/broken helpers is controlled by the new
`credential_helper_policy` parameter (`STATIC_FIRST` / `DISABLED` /
`WARN` / `FAIL`). Default is `STATIC_FIRST` (static auths take
precedence; helpers used as fallback only).
```

### DIFF
--- a/oci/auth.py
+++ b/oci/auth.py
@@ -3,14 +3,26 @@ import collections.abc
 import dataclasses
 import enum
 import json
+import logging
 import operator
 import os
+import shutil
+import subprocess
 
 import oci.util
 
+logger = logging.getLogger(__name__)
 
-class AuthType(enum.Enum):
+
+class AuthType(str, enum.Enum):
     BASIC_AUTH = 'basic_auth'
+
+
+class CredentialHelperPolicy(str, enum.Enum):
+    DISABLED     = 'disabled'      # skip helpers entirely — static auths only
+    STATIC_FIRST = 'static_first'  # static auths → helpers (safe default)
+    WARN         = 'warn'          # helpers → static; missing/broken helper → log warning
+    FAIL         = 'fail'          # helpers → static; missing/broken helper → raise
 
 
 class Privileges(enum.Enum):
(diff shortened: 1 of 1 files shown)


oci/auth.py
```diff
@@ -145,20 +156,93 @@ def lookup_credentials(
...
+            f'credential helper {binary!r} configured in docker-cfg but not found on PATH'
+        )
+        return None
+
+    try:
+        result = subprocess.run(
+            [binary, 'get'],
+            input=server_url,
+            capture_output=True,
+            text=True,
```
TuanAnh17N, line 195: would adding a timeout make sense here? 🤔

oci/auth.py
```diff
@@ -206,6 +282,47 @@ def docker_auth_lookup(
         image_netloc = image_reference.split('/')[0]
         image_host = image_netloc.split(':')[0]
 
+        # 1. per-registry credential helper (credHelpers)
+        cred_helpers = docker_auth.get('credHelpers', {})
+        helpers_enabled = credential_helper_policy is not CredentialHelperPolicy.DISABLED
+        if helpers_enabled and image_host in cred_helpers:
+            creds = _invoke_credential_helper(
+                helper_name=cred_helpers[image_host],
+                server_url=image_host,
```
TuanAnh17N: is the image_host sufficient? or should this include the scheme

oci/auth.py
```diff
@@ -206,6 +282,47 @@ def docker_auth_lookup(
...
+            if creds is not None:
+                return creds
+            # helper was unavailable (warning already emitted) - fall through
+
+        # 2. global credential store (credsStore)
+        creds_store = docker_auth.get('credsStore')
+        if (
+            helpers_enabled
+            and creds_store
+            and image_host not in cred_helpers
```
TuanAnh17N: shouldn't we still try credsStore here if the per-registry helper was configured but failed? if broken-helper is set for gcr.io but unavailable, this condition skips osxkeychain entirely. but docker would still fall back to it

oci/auth.py
```diff
@@ -206,6 +282,47 @@ def docker_auth_lookup(
...
+        ):
+            # only try credsStore when there was no per-registry helper configured
+            # (if there was one but it was unavailable we already warned and fall through to auths)
+            creds = _invoke_credential_helper(
+                helper_name=creds_store,
+                server_url=image_host,
```
TuanAnh17N: same like above. missing scheme? 🤔 https:/

oci/auth.py
ccwienk: Added `credential_helper_timeout_seconds` (default 60s) to `docker_credentials_lookup`, threaded through to the subprocess call. Defaults to 60s as a reasonable guard against hangs in fully automated contexts; pass `None` to disable for interactive/browser-based OAuth flows.

oci/auth.py
ccwienk: Rather than prepending a scheme (which we cannot reliably infer from an OCI image reference, and would be wrong for http:// local registries), we now pass the configured key from `credHelpers` directly as `server_url`. This is what the user explicitly set and what Docker itself uses as the server identifier. Matching is now also port-aware (iterating and stripping ports, consistent with the `auths` lookup), so entries like `"localhost:5000"` are matched and passed through correctly.

oci/auth.py
ccwienk: Same reasoning as above — passing the configured `credHelpers` key directly rather than guessing a scheme. For `credsStore` (this call site) there is no configured key, so `image_host` is the right input there.

oci/auth.py
ccwienk: This intentionally does not match Docker's behaviour here. Docker itself does not fall back from a per-registry `credHelpers` entry to `credsStore` — if a `credHelpers` key exists for a host, it is definitive for that host regardless of outcome. Falling back to `credsStore` on failure would mask misconfiguration. The existing fall-through to `auths` (static credentials) is already the safety net for the mounted-config case.

TuanAnh17N (approved):

PR #1680, opened 2026-07-23
ocm-upgrade: live-check before upgrade-PR creation

ccwienk:
**Release note**:
```bugfix operator
Live-check added before upgrade-PR creation to guard against duplicate PRs
when concurrent runs both pass a stale open-PR snapshot check.
```

### DIFF
--- a/.github/actions/ocm-upgrade/ocm_upgrade.py
+++ b/.github/actions/ocm-upgrade/ocm_upgrade.py
@@ -505,6 +505,19 @@ def create_upgrade_pullrequests(
                 logger.info(f'upgrade-pullrequest for {uv=} already exists (skipping)')
                 continue
 
+            # live-check guards against duplicate creation when concurrent runs both passed the
+            # static check above before either had created a PR
+            if upgrade_pullrequest_exists(
+                upgrade_vector=uv,
+                upgrade_pullrequests=github.pullrequest.iter_upgrade_pullrequests(
+                    repository=repository,
+                    state='open',
+                ),
+                component_reference_name=component_reference_name,
+            ):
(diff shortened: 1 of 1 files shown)


TuanAnh17N (approved):
/lgtm

Michael5601:
I didn't think this through completely yet, but can't we use a concurrency group on the `create-upgrade-pullrequests` job in `upgrade-dependencies.yaml` to serialize concurrent runs? Something like:

```
concurrency:
  group: ${{ github.repository }}-upgrade-pullrequests
  cancel-in-progress: false
```

Since this is a `workflow_call` reusable workflow, `${{ github.repository }} `resolves to the calling repo (e.g. landscape-canary-garden), not cc-utils, so runs triggered by the same caller's schedule would queue rather than overlap. The live-check in the Python code is still good defense-in-depth, but this would close the race at the scheduler level.

I approved nonetheless as you can merge if you see any problems to my take.

Michael5601 (approved):

ccwienk:
Good point — technically feasible: `concurrency` is supported in reusable workflows, and `github.repository` resolves to the caller's context, so the group would be correctly scoped per caller.

That said, I'd rather leave this to the caller. Concurrency policy is a scheduling concern — different callers may have different schedules, different tolerance for queuing, or no schedule at all. Baking a concurrency group into the reusable workflow imposes a constraint callers can't opt out of without forking it.

The live-check added here is the right defence-in-depth at this level: it's a correctness guard ("don't create a duplicate") rather than a scheduling policy, so it's robust and caller-agnostic regardless of how the workflow is triggered.

The concurrency fix belongs in each caller — which is also where it was applied as the primary fix.

PR #1682, opened 2026-07-28
fix: re-fetch OIDC token on 401 to handle token expiry during retries

Michael5601:
The github-auth action fetched the OIDC token once upfront before the token exchange retry loop. With high retry counts and long backoff, the loop can run longer than the 5-minute OIDC token TTL, causing the exchange to fail with a 401 even though the infrastructure issue has self-healed.

The fix fetches the OIDC token once upfront and only re-fetches it when the token exchange returns a 401, keeping the number of OIDC token requests minimal. All other 4xx errors fail immediately since retrying the same request won't help. Transient errors (5xx, timeouts, 408, 429) continue to retry with the same token.

As part of this change, sys.exit(1) calls inside the HTTP layer were replaced with typed HttpError and TransientError exceptions, allowing exchange_token to distinguish between error types and react accordingly.

**Release note**:

```other operator
Re-fetches the OIDC token on 401 during token exchange retries, fixing authentication failures when retries span longer than the 5-minute OIDC token lifetime.
```

### DIFF
--- a/.github/actions/github-auth/get-token.py
+++ b/.github/actions/github-auth/get-token.py
@@ -23,6 +23,10 @@ def _make_pool() -> urllib3.PoolManager:
 _pool = _make_pool()
 
 
+def _is_transient_status(status: int) -> bool:
+    return status >= 500 or status in (408, 429)
+
+
 def _request_with_retry(
     method: str,
     url: str,
@@ -33,12 +37,12 @@ def _request_with_retry(
     retries: int = 6,
     backoff_base: float = 3.0,
     backoff_cap: float = 60.0,
-) -> dict:
+) -> urllib3.HTTPResponse:
     '''Send an HTTP request, retrying on transient errors with exponential backoff.
 
-    TCP connect timeouts are not retried: if the connection cannot be established,
-    the runner is likely blocked by a firewall and retrying on the same runner is futile.
-    4xx responses (except 408/429) are also not retried.
+    Transient errors (5xx, 408, 429, network failures) are retried up to `retries` times.
+    Non-transient 4xx responses (except 401) cause an immediate exit.
+    2xx and 401 responses are returned to the caller.
     '''
     timeout = urllib3.Timeout(connect=connect_timeout, read=read_timeout)
     for attempt in range(retries + 1):
@@ -51,12 +55,18 @@ def _request_with_retry(
                 timeout=timeout,
             )
         except urllib3.exceptions.ConnectTimeoutError:
+            reason = f'tcp connect timeout after {connect_timeout}s'
+            if attempt == retries:
+                print(f'ERROR: Request to {url} failed: {reason}', file=sys.stderr)
+                sys.exit(1)
+            delay = min(backoff_base ** (attempt + 1), backoff_cap)
             print(
-                f'ERROR: Request to {url} failed: tcp connect timeout after {connect_timeout}s'
-                f' - likely a firewall/network issue on this runner, retrying is futile',
+                f'WARNING: Request to {url} failed: {reason}, '
+                f'retrying in {delay:.0f}s ({attempt + 1}/{retries})...',
                 file=sys.stderr,
             )
-            sys.exit(1)
+            time.sleep(delay)
+            continue
         except urllib3.exceptions.ReadTimeoutError:
             reason = f'http read timeout after {read_timeout}s'
             if attempt == retries:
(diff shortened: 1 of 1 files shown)


ccwienk:
check/fix linter-error

.github/actions/github-auth/get-token.py
```diff
@@ -23,6 +23,19 @@ def _make_pool() -> urllib3.PoolManager:
...
+class HttpError(Exception):
+    def __init__(self, status: int, body: str):
+        self.status = status
+        self.body = body
+
+    def __str__(self) -> str:
+        return f'HTTP {self.status}: {self.body}'
+
+
+class TransientError(Exception):
```
ccwienk: urllib3 already features exception-types. I do not think it is necessary or adequate to define custom exceptions.

Instead, if needed, I suggest to define a util-function that determines whether an error is considered to be transient, or not, and use it (check whether inlining this code might not be "good enough", though).

---

More broadly, the two concerns handled here have different natures and should live in separate codepaths:

- **Transient errors** (timeouts, 5xx, 408, 429): retry with the same OIDC token and exponential backoff — already lives in `_request_with_retry`.
- **Token expiry (401)**: re-fetch the OIDC token, then retry immediately — no backoff needed since the new token is valid right away.

Merging both into a single retry loop with exception-type dispatch makes the logic harder to follow. Suggested restructuring:

**1. Add a small classifier (no new classes needed):**

```python
def _is_transient_status(status: int) -> bool:
    return status >= 500 or status in (408, 429)
```

This replaces the inline `resp.status < 500 and resp.status not in (408, 429)` predicate.

**2. Have `_request_with_retry` return the raw urllib3 response** rather than parsed data, so callers can inspect the status. `http_post` becomes the thin wrapper that asserts 2xx and parses JSON for the common case.

**3. `exchange_token` uses two clearly separate loops:**

```python
def exchange_token(..., oidc_refresh_retries=3, retries=6, ...):
    for oidc_attempt in range(oidc_refresh_retries + 1):
        id_token = get_oidc_token(request_url, request_token, audience)
        resp = _request_with_retry('POST', f'{token_server}/token-exchange', ..., retries=retries)
        if resp.status == 200:
            return json.loads(resp.data)['token']
        if resp.status == 401:
            if oidc_attempt == oidc_refresh_retries:
                print(
                    f'ERROR: token exchange returned 401 after re-fetching OIDC token'
                    f' {oidc_refresh_retries} times',
                    file=sys.stderr,
                )
                sys.exit(1)
            print(
                'WARNING: 401 on token exchange — OIDC token likely expired, re-fetching...',
                file=sys.stderr,
            )
            continue
        sys.exit(1)  # non-transient 4xx: _request_with_retry already exited; unreachable
```

The transient-error retry logic stays entirely in `_request_with_retry` (untouched), and the 401/OIDC-expiry logic lives entirely in `exchange_token` with no exception catching. The two retry budgets are independent and semantically clear.

ccwienk (changes requested):

.github/actions/github-auth/get-token.py
Michael5601: I think I incorporated all your suggestions to the code. Can you have a look again please.

.github/actions/github-auth/get-token.py
```diff
@@ -127,28 +136,7 @@ def http_get(
...
-        url,
-        headers={'Content-Type': 'application/json'},
-        body=json.dumps(payload).encode(),
-        connect_timeout=connect_timeout,
-        read_timeout=read_timeout,
-        retries=retries,
-        backoff_base=backoff_base,
-        backoff_cap=backoff_cap,
-    )
+    return json.loads(resp.data.decode())
```
ccwienk, line 142: `_request_with_retry` now returns 401 responses to the caller, but `http_get` calls `json.loads(resp.data.decode())` unconditionally. If the OIDC token request itself ever returns a 401, this will crash with `JSONDecodeError` instead of a clean error message. Add a status check before parsing:

```python
if resp.status != 200:
    print(f'ERROR: HTTP {resp.status} from {url}:\n{resp.data.decode()}', file=sys.stderr)
    sys.exit(1)
return json.loads(resp.data.decode())
```

Michael5601: good catch :) I added it.

ccwienk (approved):
All comments addressed, looks good.

PR #1683, opened 2026-07-30
feat(ocm): Add support for non-tar json component descriptor blobs

### LINKED ISSUE #204: Basic support for OCM v2
**Description**
Currently, OCM component descriptors created by the new OCM CLI are not compatible with ODG. As outlined https://github.com/open-component-model/open-delivery-gear/issues/162#issuecomment-4990201311, two adjustments are necessary as a baseline to provide support for components without the index manifest change:

- tar compression; cc-utils [assumes](https://github.com/gardener/cc-utils/blob/d672ccd576fa4426fff24d3275b44ae73c4d23d8/ocm/oci.py#L87) component-descriptor blob is compressed
- null value normalisation; cc-utils ocm model expects list values for repositoryContexts and componentReferences, whereas OCMv2 tooling may produce null values instead.

8R0WNI3:
Fixes https://github.com/open-component-model/open-delivery-gear/issues/204

**Release note**:

```feature developer
The Python OCM lib now features component descriptor blobs of non-tar type as well as json type too
```

### DIFF
--- a/cnudie/retrieve.py
+++ b/cnudie/retrieve.py
@@ -412,7 +412,7 @@ def component_descriptor_from_oci(
         )
 
         # by contract, there must be exactly one layer (tar w/ component-descriptor)
-        if not (layers_count := len(manifest.layers) == 1):
+        if (layers_count := len(manifest.layers)) != 1:
             logger.warning(f'XXX unexpected amount of {layers_count=}')
 
         layer_digest = manifest.layers[0].digest
--- a/cnudie/retrieve.py
+++ b/cnudie/retrieve.py
@@ -1,12 +1,10 @@
 import collections.abc
 import dataclasses
-import io
 import itertools
 import json
 import logging
 import os
 import shutil
-import tarfile
 import tempfile
 
 import cachetools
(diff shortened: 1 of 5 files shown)


ccwienk (commented):
LGTM overall — clean refactor, and the `layers_count` walrus bug fix is a nice bonus.

Two small notes:

- `component_descriptor_from_tarfileobj` (and the async `raw_component_descriptor_from_oci`) were public — did you check for external consumers? Happy to approve either way, just want to make sure it's on your radar.

- Minor readability nit in `normalize_nulls`:
  ```python
  k: normalize_nulls(v) if v is not None else ([] if k in null_allowed_for else v)
  ```
  The `else v` arm silently keeps `None` for non-listed keys, which is correct but not immediately obvious. A short comment there would help a future reader.

8R0WNI3:
> component_descriptor_from_tarfileobj (and the async raw_component_descriptor_from_oci) were public — did you check for external consumers?

I did check but I could not find any external consumers.

> The else v arm silently keeps None for non-listed keys, which is correct but not immediately obvious. A short comment there would help a future reader.

I have added a short comment on this :-)

ccwienk (approved):

PR #1685, opened 2026-08-03
fix(ocm-upgrade): pass reference_component to live-check

TuanAnh17N:
the live-check introduced in 179a29e was missing reference_component, causing a crash when parsing existing [ci:name:...] style pr titles.

**Release note**:

\`\`\`bugfix operator
Fix crash when existing upgrade-PRs use reference-name titles during live-check
\`\`\`


ccwienk (approved):
Fix is correct — passes the in-scope `component` to the live-check's `iter_upgrade_pullrequests` call, matching what's already passed to `create_upgrade_pullrequest`. Release note updated.

Michael5601 (approved):

PR #1686, opened 2026-08-04
ctt: add local_blobs target option for copy-by-reference replication

ccwienk:
Add a `local_blobs` option to ctt target configuration, controlling how local blobs inlined in component descriptor OCI artifacts are handled during replication.

**Background:** Registries such as AWS ECR impose a layer limit on OCI image manifests. Component descriptors with many inlined local blobs (e.g. SBOM documents in large component trees) can exceed this limit, causing replication to fail with `Limit on number of layers per image exceeded`.

**Change:** New enum option `local_blobs` on target `kwargs` (values: `copy_by_value` / `copy_by_reference`, default: `copy_by_value` — fully backwards-compatible).

When set to `copy_by_reference`, local blob layers are omitted from the replicated descriptor manifest. Instead, each `LocalBlobAccess` resource receives a `globalAccess` entry pointing back to the source OCI artifact and blob digest, so consumers can still resolve the content via the source registry.

**Release note**:
```feature operator
ctt targets now support a `local_blobs: copy_by_reference` option. When set, local blobs are not replicated as OCI layers but instead referenced back to the source registry via `globalAccess`. Use this for target registries with OCI layer limits (e.g. AWS ECR).
```

### DIFF
--- a/ctt/replicate.py
+++ b/ctt/replicate.py
@@ -24,6 +24,7 @@ def replicate_oci_artifact_with_patched_component_descriptor(
     src_ocm_repo: ocm.OciOcmRepository,
     oci_client: oci.client.Client,
     overwrite: bool=False,
+    rewrite_local_blobs: bool=False,
 ):
     if isinstance(src_ocm_repo, str):
         src_ocm_repo = ocm.OciOcmRepository(baseUrl=src_ocm_repo)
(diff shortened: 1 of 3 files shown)


ctt/replicate.py
```diff
@@ -48,6 +49,28 @@ def replicate_oci_artifact_with_patched_component_descriptor(
...
+
+    if rewrite_local_blobs:
+        skip = set()
+        for artefact in component.iter_artefacts():
+            access = artefact.access
+            if not isinstance(access, ocm.LocalBlobAccess):
+                continue
+            digest = access.localReference
+            if digest not in layer_sizes:
+                continue
```
Michael5601, line 66: **suggestion, non-blocking:** This silently leaves globalAccess unset and the blob not skipped. A `logger.warning` here would help diagnose unexpected states.

Michael5601 (approved):

Michael5601:
rest lgtm :)

PR #1687, opened 2026-08-04
trusted-checkout: fix HTTPS submodules being silently skipped

matthias-horne:
## Summary

- HTTPS submodules were silently skipped in `checkout-submodules.sh`: the `http*` branch hit `continue` without performing any checkout
- The anonymous fallback check is now moved before the URL-type branch, since it applies equally to both SSH and HTTPS submodules
- HTTPS submodules now extract `org/repo` by stripping scheme and host from the URL, then fall through to the same scoped-token fetch logic used for SSH submodules — ensuring a repository-scoped token is obtained rather than relying on the top-level checkout token

## Test plan

- [ ] Verify a repo with an HTTPS submodule checks out the submodule correctly when `submodules: recursive` is set
- [ ] Verify SSH submodule checkout behaviour is unchanged
- [ ] Verify anonymous fallback (no `token-server` or app credentials) still works for public submodules

### DIFF
--- a/.github/actions/trusted-checkout/checkout-submodules.sh
+++ b/.github/actions/trusted-checkout/checkout-submodules.sh
@@ -1,16 +1,16 @@
 #!/usr/bin/env bash
 
 # a (slightly hacky) script to be used as a workaround for allowing to checkout submodules
-# referenced with an SSH-URL with GitHub-Action's checkout-Action (using https / token-auth).
+# referenced with an SSH-URL or HTTPS-URL with GitHub-Action's checkout-Action (using https / token-auth).
 #
 # Will only work for submodules on same GitHub-Instance. Only needed on GHE (checkout-action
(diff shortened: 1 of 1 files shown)


ccwienk (approved):
One nit: the file header comment (line 4) still says "referenced with an SSH-URL" — worth updating to cover HTTPS as well.

Otherwise the fix is correct: moving the anonymous-fallback check before the URL-type branch is the right structural change, and the HTTPS stripping logic feeds cleanly into the same scoped-token fetch subshell. LGTM.

matthias-horne:
> One nit: the file header comment (line 4) still says "referenced with an SSH-URL" — worth updating to cover HTTPS as well.

Thanks for catching that. The header comment had a few more things that did not reflect the new state of the script. Should be more accurate now.

PR #1690, opened 2026-08-06
feat(ocm): Support more access/artefact types

### LINKED ISSUE #205: Full support of OCM v2
**Description**
Add support for OCI artifacts stored "locally", so being stored as regular OCI artifacts without tags, being referenced by a index manifest, mainly talking about this: https://github.com/open-component-model/open-delivery-gear/issues/162#issuecomment-4978717136

- Assure that nested OCI artifacts are support
- Check that in general now v2 artifacts work

8R0WNI3:
Fixes https://github.com/open-component-model/open-delivery-gear/issues/205

**Release note**:

```feature developer
The Python OCM library now supports more access/artefact types for (de-)serialisation
```

### DIFF
--- a/ocm/__init__.py
+++ b/ocm/__init__.py
@@ -68,6 +68,7 @@ class AccessType(enum.StrEnum):
 # accepted for deserialisation
 # note: the `/v1` suffix is _always_ optional (if absent, /v1 is implied)
 AccessType._value2member_map_ |= {
+    'github': AccessType.GITHUB,
     'github/v1': AccessType.GITHUB,
     'localBlob': AccessType.LOCAL_BLOB,
     'localFilesystemBlob': AccessType.LOCAL_BLOB,
@@ -77,6 +78,7 @@ AccessType._value2member_map_ |= {
     'ociArtefact': AccessType.OCI_REGISTRY,
     'ociArtifact': AccessType.OCI_REGISTRY,
     'ociArtifact/v1': AccessType.OCI_REGISTRY,
+    's3': AccessType.S3,
     's3/v1': AccessType.S3,
 }
 
@@ -207,9 +209,10 @@ class ArtefactType(enum.StrEnum):
 # note: the `/v1` suffix is _always_ optional (if absent, /v1 is implied)
 ArtefactType._value2member_map_ |= {
     'blob': ArtefactType.BLOB,
+    'git': ArtefactType.GIT,
     'git/v1': ArtefactType.GIT,
-    'ociImage/v1': ArtefactType.OCI_IMAGE,
     'ociImage': ArtefactType.OCI_IMAGE,
+    'ociImage/v1': ArtefactType.OCI_IMAGE,
     'helmChart': ArtefactType.HELM_CHART,
 }
 
@@ -657,9 +660,18 @@ def _read_schema_file(schema_file_path: str):
         return yaml.safe_load(f)
 
 
-def enum_or_string(v, enum_type: enum.Enum):
+def enum_or_string(
+    v,
+    enum_type: enum.Enum,
+    omit_v1_version: bool=False,
+):
+    if omit_v1_version and str(v).endswith('/v1'):
+        stripped_value = str(v).removesuffix('/v1')
+    else:
+        stripped_value = None
+
     try:
-        return enum_type(v)
+        return enum_type(stripped_value if stripped_value is not None else v)
     except ValueError:
         return str(v)
 
@@ -744,16 +756,16 @@ class ComponentDescriptor:
                 ],
                 type_hooks={
                     AccessType | str: functools.partial(
-                        enum_or_string, enum_type=AccessType
+                        enum_or_string, enum_type=AccessType, omit_v1_version=True
                     ),
                     ArtefactType | str: functools.partial(
-                        enum_or_string, enum_type=ArtefactType
+                        enum_or_string, enum_type=ArtefactType, omit_v1_version=True
                     ),
                     ArtifactIdentity | str: functools.partial(
-                        enum_or_string, enum_type=ArtefactType
+                        enum_or_string, enum_type=ArtefactType, omit_v1_version=True
                     ),
                     AccessType: functools.partial(
-                        enum_or_string, enum_type=AccessType
+                        enum_or_string, enum_type=AccessType, omit_v1_version=True
                     ),
                     datetime.datetime: dateparse,
                 },
--- a/ocm/__init__.py
+++ b/ocm/__init__.py
@@ -55,8 +55,10 @@ class SchemaVersion(enum.StrEnum):
 
 class AccessType(enum.StrEnum):
     GITHUB = 'github' # XXX: new: gitHub/v1
+    HELM = 'Helm/v1'
     LOCAL_BLOB = 'localBlob/v1'
     NONE = 'None'  # the resource is only declared informally (e.g. generic)
+    NPM = 'NPM/v1'
     OCI_BLOB = 'ociBlob/v1'
     OCI_REGISTRY = 'ociRegistry' # XXX: new: ociArtifact/v1
     RELATIVE_OCI_REFERENCE = 'relativeOciReference'
@@ -72,10 +74,14 @@ AccessType._value2member_map_ |= {
     'gitHub': AccessType.GITHUB, # deprecated
     'github': AccessType.GITHUB, # deprecated
     'github/v1': AccessType.GITHUB, # deprecated
+    'Helm': AccessType.HELM,
+    'helm': AccessType.HELM, # deprecated
     'LocalBlob': AccessType.LOCAL_BLOB,
     'localBlob': AccessType.LOCAL_BLOB, # deprecated
     'localFilesystemBlob': AccessType.LOCAL_BLOB,
     'none': AccessType.NONE,
+    'NPM': AccessType.NPM,
+    'npm': AccessType.NPM, # deprecated
     'OCIImage': AccessType.OCI_REGISTRY,
     'ociArtifact': AccessType.OCI_REGISTRY, # deprecated
     'ociArtifact/v1': AccessType.OCI_REGISTRY, # deprecated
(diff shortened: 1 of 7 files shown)


ocm/oci.py
```diff
@@ -102,6 +105,9 @@ def component_descriptor_from_blob(
             tre.add_note(f'{component_id=}')
             raise tre
 
+    if skip_serialisation:
+        return component_descriptor_blob.decode()
```
zkdev: why do we want to have this option at all?

8R0WNI3: In cases where one wants to inspect the raw blob instead of the serialised version of it (e.g. using the `raw` switch in ODG UI).

ocm/oci.py
```diff
@@ -115,3 +115,35 @@ def component_descriptor_from_blob(
...
+def find_component_descriptor_manifest_digest(
+    index_manifest: oci.model.OciImageManifestList,
+) -> str:
+    if not index_manifest.manifests:
+        raise ValueError(f'Index manifest appears to be empty: {index_manifest}')
+
+    component_descriptor_manifest_digest = None
+    for manifest in index_manifest.manifests:
+        if (
+            not manifest.annotations
```
ccwienk: split into separate if-continues

ccwienk (changes requested):
see inline comment. please split into two pullrequests:

first one:
```
 - 94bdd94 — feat(ocm): Support OCI index manifests for component descriptors
  - cf9df4f — feat(ocm): Support more component descriptor mimetypes
```

remaining commits into second one

ocm/__init__.py
```diff
@@ -68,16 +68,14 @@ class AccessType(enum.StrEnum):
 # accepted for deserialisation
 # note: the `/v1` suffix is _always_ optional (if absent, /v1 is implied)
 AccessType._value2member_map_ |= {
-    'github/v1': AccessType.GITHUB,
+    'github': AccessType.GITHUB,
```
ccwienk: I think fully ignoring versions is not a safe semantics. currently, no version-suffix is handled to be equal to version-suffix equal to `v1`, which (IIRC) was an acceptable approach considering the canonical form should start with `<type>/v1`, to be followed by greater versions, if needed.

if we simply strip and ignore the version, we would end up deserialising artefacts of actually incompatible versions (!== v1), which might either result in silently dropping extra attrs (which may lead to hard-to-debug follow-up errors), or deserialisation errors (as opposed to graciously falling back to `dict`).

ocm/__init__.py
```diff
@@ -85,6 +92,8 @@ class AccessType(enum.StrEnum):
     'ociBlob': AccessType.OCI_BLOB, # deprecated
     'S3': AccessType.S3,
     's3': AccessType.S3, # deprecated
+    'Wget': AccessType.WGET,
+    'wget': AccessType.WGET, # deprecated
```
ccwienk: is there anything special to wget that is different from plain http?

ocm/__init__.py
```diff
@@ -55,12 +55,15 @@ class SchemaVersion(enum.StrEnum):
 
 class AccessType(enum.StrEnum):
     GITHUB = 'github' # XXX: new: gitHub/v1
+    HELM = 'Helm/v1'
```
ccwienk, line 58: I suggest it might make sense to define a hook by which additional types can be plugged in by callers to avoid making this module a bottleneck for maintaining such types.

8R0WNI3:
> please split into two pullrequests

The new PR only containing the support for OCI index manifests can be found here: https://github.com/gardener/cc-utils/pull/1703

ocm/__init__.py
8R0WNI3: I am not aware of any deeper differences than convenience related changes. However, I'm just trying to mirror the supported access types of the [OCM spec](https://github.com/open-component-model/ocm-spec/tree/main/doc/04-extensions/02-access-types) to allow proper deserialisation of known types.

ocm/__init__.py
8R0WNI3: I think there are two cases:
1.  access / artefact types which are centrally defined and
2. access / artefact types which are defined and used by individual adopters of OCM

While I agree that (2) should not have to be maintained here, I do think that (1) should (this PR only focuses on (1)). Wdyt?

ocm/__init__.py
8R0WNI3: I have adjusted the function so that it will only remove the `v1` suffix for the ease of mapping to the suitable enum member, and I have added a new `_v2` type where available (currently only the S3 access has multiple versions).

ocm/__init__.py
ccwienk: agreed - if those are the core of "wellknown/wellsupported" types

ocm/__init__.py
```diff
@@ -55,29 +55,48 @@ class SchemaVersion(enum.StrEnum):
...
     S3 = 's3' # XXX: new: s3/v1
+    S3_v2 = 's3/v2'
+    WGET = 'Wget/v1'
 
 
 # hack: patch enum to accept "aliases"
 # -> the values defined in enum above will be  used for serialisation; the aliases are also
 # accepted for deserialisation
 # note: the `/v1` suffix is _always_ optional (if absent, /v1 is implied)
 AccessType._value2member_map_ |= {
```
ccwienk, line 73: No reason to remove these. Keeping `'github/v1'`, `'OCIRegistry/v1'`, `'ociArtifact/v1'`, `'s3/v1'`, `'git/v1'`, `'ociImage/v1'` in the map costs nothing and their removal is a silent compat break for any caller doing `AccessType('github/v1')` directly (i.e. not going through `enum_or_string`/`from_dict`). Just leave them alongside the new entries.

ocm/__init__.py
```diff
@@ -55,29 +55,48 @@ class SchemaVersion(enum.StrEnum):
...
     GITHUB = 'github' # XXX: new: gitHub/v1
+    HELM = 'Helm/v1'
     LOCAL_BLOB = 'localBlob/v1'
     NONE = 'None'  # the resource is only declared informally (e.g. generic)
+    NPM = 'NPM/v1'
     OCI_BLOB = 'ociBlob/v1'
     OCI_REGISTRY = 'ociRegistry' # XXX: new: ociArtifact/v1
     RELATIVE_OCI_REFERENCE = 'relativeOciReference'
     S3 = 's3' # XXX: new: s3/v1
+    S3_v2 = 's3/v2'
```
ccwienk: Nit: `S3_V2` — consistent with SCREAMING_SNAKE_CASE of all other members.

ocm/__init__.py
```diff
@@ -179,26 +198,61 @@ def hostname(self):
...
+@dc(kw_only=True)
+class NPMAccess(Access):
+    type: AccessTypeOrStr = AccessType.NPM
+    registry: str
+    package: str
+    version: str
+
+
+@dc(kw_only=True)
+class WgetAccess(Access):
```
ccwienk: I'd lean toward dropping `WgetAccess`, `AccessType.WGET`, and the related aliases from this PR entirely. `Wget` is a confusing name for what is essentially HTTP-with-options, and it feels out of place next to OCI/S3/Helm. Per the earlier discussion about an extensibility hook for non-core types — this looks like exactly the kind of peripheral type that should go through that mechanism rather than being baked in here.

ocm/__init__.py
```diff
@@ -179,26 +198,61 @@ def hostname(self):
...
+    type: AccessTypeOrStr = AccessType.WGET
+    url: str
+    mediaType: str | None = None
+    header: dict[str, str] | None = None
+    verb: str | None = None
+    body: bytes | None = None
```
ccwienk: What is the expected serialisation form? `bytes` will not round-trip cleanly through YAML/dacite. If the spec encodes this field as base64, the type should be `str`.

ocm/__init__.py
```diff
@@ -657,9 +719,18 @@ def _read_schema_file(schema_file_path: str):
...
+    v,
+    enum_type: enum.Enum,
+    omit_v1_version: bool=False,
+):
+    if omit_v1_version and str(v).endswith('/v1'):
+        stripped_value = str(v).removesuffix('/v1')
+    else:
+        stripped_value = None
+
     try:
```
ccwienk, line 723: `or` is the wrong idiom here: if `v == '/v1'`, `removesuffix` yields `''` which is falsy, so the fallback quietly reverts to `'/v1'`. Use `stripped_value if stripped_value is not None else v`.

test/ocm/init_test.py
```diff
@@ -0,0 +1,94 @@
+import ocm
+
+
+def test_access_type_aliases():
```
ccwienk, line 4: Should also cover the aliases that existed before this PR: `'github/v1'`, `'OCIRegistry/v1'`, `'ociArtifact/v1'`, `'s3/v1'`, `'git/v1'`, `'ociImage/v1'`. Had these been present they would have caught the compat break in the alias map.

test/ocm/oci_test.py
```diff
@@ -0,0 +1,58 @@
+import pytest
```
ccwienk: This file tests `ocm.oci.find_component_descriptor_manifest_digest`, which is not introduced in this PR — it belongs in #1703. Please remove it from here.

ccwienk (changes requested):
Additional review pass.

ocm/__init__.py
8R0WNI3: I don't insist on keeping the `Wget` parts here, but it is part of the core [OCM-spec](https://github.com/open-component-model/ocm-spec/blob/main/doc/04-extensions/02-access-types/wget.md).

test/ocm/oci_test.py
8R0WNI3: Do you want me to remove it entirely or should I create a dedicated PR to add this test file?

ocm/__init__.py
8R0WNI3: I have removed `Wget` from this PR.

test/ocm/oci_test.py
ccwienk: Since #1703 is already merged, the function under test exists — keep the file here. No point opening a third PR just for a test.

TuanAnh17N (approved):

PR #1703, opened 2026-08-11
feat(ocm): Support OCI index manifests

### LINKED ISSUE #205: Full support of OCM v2

8R0WNI3:
Fixes https://github.com/open-component-model/open-delivery-gear/issues/205

**Release note**:

```feature developer
The Python OCM library now supports OCI index manifests for component descriptors
```

### DIFF
--- a/cnudie/retrieve.py
+++ b/cnudie/retrieve.py
@@ -380,10 +380,21 @@ def component_descriptor_from_oci(
         manifest = oci_client.manifest(
             image_reference=target_ref,
             absent_ok=True,
+            accept=f'{om.OCI_MANIFEST_SCHEMA_V2_MIME}, {om.OCI_IMAGE_INDEX_MIME}',
         )
 
-        if manifest:
-            break
+        if not manifest:
+            continue
+
+        if manifest.mediaType == om.OCI_IMAGE_INDEX_MIME:
+            digest = ocm.oci.find_component_descriptor_manifest_digest(manifest)
+
+            manifest = oci_client.manifest(
+                image_reference=om.OciImageReference(target_ref).with_tag(digest),
+                accept=om.OCI_MANIFEST_SCHEMA_V2_MIME,
+            )
+
+        break
     else:
         manifest = None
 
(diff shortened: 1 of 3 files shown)


ccwienk:
**`skip_serialisation`** — a few concerns:

**Wrong name.** The blob is already serialised YAML/JSON on disk. What the flag skips is *de*serialisation (parsing into `ComponentDescriptor`). Minor but misleading.

**Polymorphic return type.** `component_descriptor_from_blob` and `component_descriptor_from_oci` now return `ComponentDescriptor | str`. That forces every caller to handle a union representing two fundamentally different use-cases — and threads a UI-layer concern (`raw` toggle in ODG) all the way down through the retrieval primitives.

**Is it actually needed here?** The use-case (ODG raw view) needs the raw YAML/JSON bytes of the OCI layer. Those bytes are available without any change to this library: the caller already has the OCI client and, after resolving the manifest, has the layer digest. A direct `oci_client.blob(ref, layer.digest)` call from ODG side gives exactly that without any API change here.

If a library-level helper is genuinely wanted, a separate function is cleaner:

```python
def raw_component_descriptor_blob_from_oci(
    name: str,
    version: str,
    ocm_repos: ...,
    oci_client: oc.Client,
    absent_ok: bool = False,
) -> bytes | None:
    ...
```

That keeps `component_descriptor_from_oci` single-purpose, avoids the union return type, and doesn't pollute the signature of the common path.

ccwienk:
To be more direct: I don't think this flag belongs in cc-utils at all. The need for raw bytes is an ODG-specific concern and should be solved there — e.g. by calling `oci_client.blob(ref, layer.digest)` directly. cc-utils should not accumulate application-layer requirements from its consumers.

8R0WNI3:
I have removed the commit in question from this PR.

ccwienk (approved):

PR #1725, opened 2026-08-19
fix(prune): skip out-of-scope OciAccess refs in _iter_resource_refs

ccwienk:
## Problem

`_iter_resource_refs` yielded all `OciAccess.imageReference` values unconditionally,
including refs that point to external registries (e.g. GAR, ECR). When
`delete_candidates` traverses a component descriptor to delete its resources,
it was submitting those external refs for deletion, resulting in 403/401 errors
against registries the pruning run has no business touching.

## Fix

Filter absolute `OciAccess` refs: only yield when the normalised ref starts with
`ocm_repo.oci_ref`. `RelativeOciAccess` refs are always within the target repo by
construction and are unaffected.

## Tests

Added two regression tests for `_iter_resource_refs`:
- out-of-scope `OciAccess` ref → not yielded
- in-scope `OciAccess` ref → yielded

### DIFF
--- a/ocm/prune.py
+++ b/ocm/prune.py
@@ -88,7 +88,11 @@ def _iter_resource_refs(
 ) -> collections.abc.Iterator[str]:
     access = node.resource.access
     if isinstance(access, ocm.OciAccess):
-        yield access.imageReference
+        # only yield refs that live within the target OCM repository; absolute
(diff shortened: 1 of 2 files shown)


ccwienk:
Superseded by #1727

PR #1744, opened 2026-09-07
oci: cleanup and fixes

ccwienk:
Three commits cleaning up the `oci` package after a code-review pass.

**Commit 1 — refactor + bug fix (`oci/__init__.py`)**

Extracted three helpers from the ~360-line `replicate_artifact` monolith:
`_apply_annotations`, `_replicate_manifest_list`, `_wrap_single_as_manifest_list`.

Bug fixed: the `NORMALISE_TO_MULTIARCH` branch called `oci_client.*` (the
function parameter, which is `None` when only `credentials_lookup` is provided)
instead of the resolved local `client` — causing `AttributeError` at runtime
for any caller that passes `credentials_lookup` and uses that mode.

Secondary fix: v1 cfg-blob synthesis now updates the manifest dataclass via
`dataclasses.replace` so a subsequent annotation patch re-serialises the
correct config digest rather than the stale pre-synthesis one.

**Commit 2 — `OciImageReference` hash/eq + `MimeTypes` (`oci/model.py`)**

`__hash__` previously hashed the original (un-normalised) string while `__eq__`
can match on normalised forms, violating the hash contract and breaking sets and
dicts. Fixed to use `oci.util.normalise_image_reference` directly (calling the
`@functools.cache`-decorated property from `__hash__` causes infinite recursion
because `functools.cache` hashes `self` as its own key).

`MimeTypes` promoted to `enum.StrEnum`. Values are `str` subclasses and
compare/interpolate identically — API-compatible.

**Commit 3 — miscellaneous cleanups (`oci/auth.py`, `oci/client.py`, `oci/workarounds.py`)**

- `docker_credentials_lookup`: stat-based mtime check before re-reading docker
  config on each auth lookup; re-reads only when the file changes.
- `client`: remove dead `_put_blob_chunked` (unreachable since the single-POST
  tempfile path was added); simplify `put_blob` branching accordingly.
- `workarounds`: `isinstance(x, bytes) or isinstance(x, str)` →
  `isinstance(x, (bytes, str))`.

**Release note**:
```bugfix developer
oci: fix replicate_artifact NORMALISE_TO_MULTIARCH crash when called with credentials_lookup; fix OciImageReference hash/eq contract
```

### DIFF
--- a/oci/__init__.py
+++ b/oci/__init__.py
@@ -61,6 +61,142 @@ class ReplicationMode(enum.Enum):
             raise NotImplementedError(self)
 
 
+def _apply_annotations(
+    manifest: om.OciImageManifest | om.OciImageManifestList,
+    annotations: dict[str, str],
+) -> bool:
+    '''Patches annotations into manifest in-place. Returns True if any value changed.'''
+    dirty = False
+    for k, v in annotations.items():
+        if manifest.annotations.get(k) != v:
+            manifest.annotations[k] = v
+            dirty = True
+    return dirty
+
+
+def _replicate_manifest_list(
+    raw_manifest: str,
+    manifest: om.OciImageManifestList,
+    src_image_reference: om.OciImageReference,
+    tgt_image_reference: om.OciImageReference,
+    client: oc.Client,
+    mode: ReplicationMode,
+    platform_filter: collections.abc.Callable[[om.OciPlatform], bool] | None,
+    annotations: dict[str, str] | None,
+) -> tuple[requests.Response, om.OciImageReference, bytes]:
+    src_name = src_image_reference.ref_without_tag
+    tgt_name = tgt_image_reference.ref_without_tag
+    manifest_dirty = False
+
+    for idx, sub_manifest in enumerate(tuple(manifest.manifests)):
+        src_reference = f'{src_name}@{sub_manifest.digest}'
+        tgt_reference = f'{tgt_name}'
+
+        if platform_filter:
+            platform = op.from_single_image(
+                image_reference=src_reference,
+                oci_client=client,
+                base_platform=sub_manifest.platform,
+            )
+            if not platform_filter(platform):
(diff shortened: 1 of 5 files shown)


oci/__init__.py
```diff
@@ -61,6 +61,142 @@ def accept_header(self):
...
+
+        submanifest_digest = f'sha256:{hashlib.sha256(submanifest_bytes).hexdigest()}'
+        if submanifest_digest != sub_manifest.digest:
+            patched = dataclasses.replace(
+                sub_manifest,
+                digest=submanifest_digest,
+                size=len(submanifest_bytes),
+            )
+            manifest.manifests.remove(sub_manifest)
+            manifest.manifests.insert(idx, patched)
```
Michael5601: `idx` comes from `enumerate(tuple(...))`, which is a snapshot of the original positions. When `platform_filter` removes an earlier entry (line 104), the live list shrinks, so `insert(idx, patched)` lands one slot too late for every preceding removal. Consider building a new list instead of mutating in-place to avoid the indexing drift.

Michael5601 (changes requested):

oci/__init__.py
ccwienk: good catch!

ccwienk:
Applied the suggestion in b6c64f493: `_replicate_manifest_list` now builds `replicated_manifests` and assigns `manifest.manifests` once at the end, instead of removing/inserting against a stale snapshot index. This fixes both the `platform_filter`-removal index drift you flagged and a second, subtler one: even without any removal, `remove` + `insert(idx, ...)` on the live list shifted the entry to position `idx-1`.

oci/__init__.py
ccwienk: Fixed in b6c64f493. Rebuilt the loop to collect into a new replicated_manifests list and assign manifest.manifests once after the loop, eliminating both the removal-driven index drift and the off-by-one caused by remove + insert(idx, ...) on the live list.

PR #1745, opened 2026-09-07
Add SAST linting to OCM image build workflow

Michael5601:
Adds a lint-pybandit job to the `build-ocm.yaml`workflow to connect the existing bandit SAST scan with the ODG delivery service.

The job runs `.ci/lint` (which already executes pylint, flake8, and bandit) and exports the output as an OCM resource fragment with sast/pybandit purposes, so the delivery service recognises the scan and resolves the no-linter finding for the github.com/gardener/cc-utils/ocm component.

**Release note**:

```other operator
Add SAST linting job to OCM image build workflow
```

### DIFF
--- a/.github/workflows/build-ocm.yaml
+++ b/.github/workflows/build-ocm.yaml
@@ -21,6 +21,54 @@ jobs:
       version-prerelease: ${{ (inputs.mode == 'snapshot' || inputs.mode == '') && 'dev0' || '' }}
       base-component-file: .ocm/ocm-component.yaml
 
+  lint-pybandit:
+    runs-on: ${{ vars.DEFAULT_RUNNER || 'ubuntu-latest' }}
+    permissions:
+      contents: read
+    steps:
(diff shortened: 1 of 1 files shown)


ccwienk (approved):

.github/workflows/build-ocm.yaml
```diff
@@ -21,6 +21,54 @@ jobs:
...
+          mkdir -p "${blobs_dir}"
+
+          bandit_logfile=bandit.log
+          bandit_evidence="${blobs_dir}/bandit.tar.gz"
+
+          bandit \
+            --configfile pyproject.toml \
+            --recursive \
+            -f txt -o "${bandit_logfile}" \
+            ocm/
```
TuanAnh17N, line 48: since bandit is called directly under `set -euo pipefail`, won't any finding abort the step before tar runs - meaning no evidence gets uploaded exctly when there's a real finding? 🤔

TuanAnh17N (approved):

PR #1757, opened 2026-09-18
ctt: add --paranoid to validate manifest-uploads via GET

ccwienk:
Adds a `--paranoid` flag to `ctt` that patches `oci.workarounds.patch_put_manifest_to_validate_via_get` onto the shared `oci_client`. After each successful manifest-PUT, a GET validates the manifest is actually served from the target (always digest-addressed to avoid false-positive passes through symbolic-tag indirection). On non-2xx, the `HTTPError` is re-raised with a diagnostic note, so replication fails early rather than being marked done.

**Release note**:
```feature operator
ctt: add `--paranoid` flag to validate OCI manifest-uploads via GET (works around registries that accept PUTs with HTTP 20x while silently dropping them)
```


ccwienk:
recreating from gardener-org branch (avoid fork)

